PR #3, opened 2018-02-08
SWS-81: Add React Storybook

mtho11:
This is a really nice UI tech to experiment with UI components in isolation. Especially stateless components.

To run:  `npm run storybook` and then goto `http://localhost:6006`

A great example of this is [Patternfly React](https://rawgit.com/patternfly/patternfly-react/gh-pages/index.html?knob-Label=Well%20done%21%20You%20successfully%20read%20this%20important%20alert%20message.&selectedKind=About%20Modal&selectedStory=AboutModal&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs)


.gitignore
```diff
@@ -22,3 +22,5 @@
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+
+yarn.lock
```
jotak: So, on the backend we decided to let the dependencies lock file in GIT, which is a common pattern in order to make sure that a given version is tight to given dependencies.

Should we do the same here or not? Or is this role devoted to just package.lock and not yarn.lock?

stories/index.stories.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
```
jotak: Can it be tested right now? Or do we need an additional html file to test it?

.gitignore
pilhuhn: Shouldn't package.json already have the exact versions? In a different context, I found checking in the .lock files to be more confusing than helping.
Of course that requires exact versions in package.json and not a "catch all" version > 0.0

stories/index.stories.js
pilhuhn: Was going to ask the same.

pilhuhn:
This needs a html page to get started with and also a bit of documentation.

.gitignore
jotak: package.json may or may not have the exact version. Having the .lock file in git is like a proof of any dependency update.

abonas:
the relevant [JIRA](https://issues.jboss.org/projects/SWS/issues/SWS-81) is not part of sprint 1, why is there work on this?
also the JIRA has some questions, I think it's best to answer them first and clarify the design/location of this.
@cc mwringe

.gitignore
jotak: See also "Tilde ranges" and "Caret ranges" that are frequently used : https://github.com/npm/node-semver#tilde-ranges-123-12-1

pilhuhn: Well with a tilde range we say "update at will" and then encode the result in the lock file.
Why not pick a specific version on purpose and encode this in the package.json?

jotak: I imagine they are good reasons (and others not so good) for that most javascripters use this instead of exact versions. For instance I believe any security related patches are automatically picked up that way.

Also, I wonder if transitive dependencies that are selected with tilde ranges upstream can be updated here even if the upstream project didn't change. I'm not sure if I'm clear, here is an example:
sws depends on foo "=1.0.0"
foo depends on bar "~1.0.0"
We build sws, we get foo 1.0.0 and bar 1.0.0

Then bar is updated to 1.0.1
We rebuild sws without changing any dependency, but even if we point strictly to foo 1.0.0, we get bar 1.0.1
This change could only be seen in the lock file I guess.

mtho11: @jotak @pilhuhn  I'm fine with locking the versions down. It should be a different jira though

mtho11:
This is basically a prerequisite to creating react components. The workflow is that you work on your components/tests and exercise them in the storybook. It is not something that you add later on in the project (nobody would ever go back and do it).

mtho11:
The storybook src is not in `src` but in its own directory at the same level called `stories`so they don't get compiled into the product.

mtho11:
@pilhuhn yes, I left out the documentation but an html page is not needed as it just assembles the js into its own dynamic page. I'll add the docs but the command is `npm run storybook` and then goto `http://localhost:6006`

stories/index.stories.js
mtho11: It can be tested right now see my instructions in another comment

.gitignore
mtho11: It is best practice to check-in the package-lock.json file to ensure consistent package dependencies.
https://stackoverflow.com/questions/44206782/do-i-commit-the-package-lock-json-file-created-by-npm-5

PR #4, opened 2018-02-08
SWS-91: Npm publishing

mtho11:
Publish the proper assets to npm

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "swsui",
-  "version": "0.1.0",
+  "version": "0.2.1",
   "bugs": {
     "url": "https://issues.jboss.org/projects/SWS/issues/"
   },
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "swsui",
-  "version": "0.1.0",
+  "version": "0.2.1",
```
jotak: already 0.2.1? Because you ran some test-purpose releases?

jotak:
I guess you should rebase so that the storybook related commit is removed from this PR

mtho11:
@jotak yes, I'll rebase

package.json
mtho11: I had to get the proper files published to npm for mazz so it required several versions.
And also `@swift-sunshine/swsui` didn't work (paid feature only now) so I had to rollback to just `swsui`

mtho11:
@jotak this should be good to go now

PR #10, opened 2018-02-14
Add Patternfly images to assets folder.

mtho11:
Not sure which assets we will need, so adding more than we need. We can trim down later.

Or we can delete certain ones right now.

[Create UI Scaffolding](https://issues.jboss.org/browse/SWS-56)


jotak:
Did you copy them from patternfly node_modules?
I wonder if we can get what we need copied at build time automatically?

aljesusg:
Like patternfly-react-demo?
cp -r node_modules/patternfly/dist/fonts src/
cp -r node_modules/patternfly/dist/img src/
and build the css from sass file

abonas:
what I am concerned about here is that assets from another repo are copy pasted here and we don't get any sync indication. what if openshift logo changes tomorrow? how would that be reflected in our app?  Also, why do we need a patternfly and openshift logo in the first place? where should it be displayed in our app?

mtho11:
I do see them in patternfly/dist/img so it is not needed.

PR #11, opened 2018-02-14
Modify typescript rules for development

mtho11:
Just try to do development without disabling some of these rules. I wasted hours. It is especially bad with the auto-compile of webpack.

I agree these aren't what we want for production builds but like @jotak suggested yesterday, it looks like we need a dev and prod separation. Someone can create a task for that.

[Create UI Scaffolding](https://issues.jboss.org/browse/SWS-56)

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -12,10 +12,10 @@
     "forceConsistentCasingInFileNames": true,
     "noImplicitReturns": true,
     "noImplicitThis": true,
-    "noImplicitAny": true,
+    "noImplicitAny": false,
     "strictNullChecks": true,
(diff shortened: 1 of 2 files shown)


mtho11:
Merge this if you want to get any work done ;)

jotak:
@mtho11 in this PR [ https://github.com/swift-sunshine/swsui/pull/15/files ] I only had to change no-console and no-any, and it's quite fine (it builds & run quite well)

mtho11:
Ok so https://github.com/swift-sunshine/swsui/pull/14 make this one old so I will close

PR #14, opened 2018-02-15
Configure prettier

jotak:
- 120 characters line length. This way it matches the tslint config. And the default was really small.
- Do not change semicolons. Rules between JS and TS differ
- Single quote = true

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,10 +19,7 @@
     "react-scripts-ts": "2.13.0"
   },
   "lint-staged": {
-    "src/**/*.{js,jsx,ts,tsx,json,css}": [
-      "prettier --single-quote --write",
-      "git add"
-    ]
+    "src/**/*.{js,jsx,ts,tsx,json,css}": ["prettier --single-quote --parser typescript --print-width 120 --write", "git add"]
   },
   "scripts": {
     "start": "react-scripts-ts start",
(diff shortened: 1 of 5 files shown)


jotak:
Side-comment: I'm starting to think that prettier causes more harm than it fixes.
For instance, configuring line width is a pain. Whatever length we choose, we'll end up with some bad-looking stuff (sometimes it's split where it shouldn't, or the contrary). This is the kind of thing I'd prefer to let up to the developer.

jotak:
Ok, I've committed some works and prettier just removed ending semi-colons. Now tslint complains. Either we put "--no-semi" and there's missing semi, or we don't put it and there's too many semi. It just doesn't work with Typescript.

jotak:
It's a bit better with the "typescript" parser :-)

But unfortunately some constructs seem impossible, like dealing with semi-colons on stuff like `export default interface XXX {...}`. I have to split it that way:

```
interface XXX { ... }

export default XXX;
```

jotak:
@aljesusg Yes... and I'm even wondering if we should keep it as a pre-commit hook. Anything as a pre-commit hook must be 100% reliable (because you commit something that may be different from what you tested), and I fear it is not the case here.

What if we just set it up on build / start scripts, but not as a pre-commit hook?

mtho11:
Yeah, I guess this is kind of an experiment

mtho11:
We can try removing the precommit hook until we are comfortable?

package.json
```diff
@@ -18,17 +18,11 @@
...
-      "git add"
-    ]
-  },
   "scripts": {
-    "start": "react-scripts-ts start",
-    "precommit": "lint-staged",
-    "build": "react-scripts-ts build",
-    "test": "react-scripts-ts test --env=jsdom",
+    "prettier": "prettier --single-quote --parser typescript --print-width 120 --write \"src/**/*.{js,jsx,ts,tsx,json}\"",
+    "start": "npm run prettier && react-scripts-ts start",
```
jotak: @mtho11 @aljesusg is it a correct way to embed script invocation or not? (ie. "start" invokes "prettier" script)

mtho11: lets roll with it. we can always change later

aljesusg: For me is ok have this task in start

PR #18, opened 2018-02-16
Set noImplicitAny flag in tsconfig.json to false 

mtho11:
Set noImplicitAny flag in tsconfig.json to false so we can use 3rd pa…rty libaries without @type libs.

I could not use the cytoscape.js library without this because it has no type libraries (*.d.ts)


mtho11:
IMO: this is little too tight anyway. Sometimes we may not care about the type and **any** is appropriate.
For reference: https://www.typescriptlang.org/docs/handbook/compiler-options.html

PR #20, opened 2018-02-16
Add a ServiceListPage

lucasponce:
Following the skeletons for main pages.
Added a ServiceList page.


lucasponce:
P.S.: With this small PR I want also to get familiar with the UI code, so probably I will add more incremental changes, to get earlier feedback instead to group everything in a single one.
Once everyone is familiar, we can change granularity of changes.

PR #22, opened 2018-02-16
Add lodash @types.

mtho11:
I forgot to add the lodash types

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
   },
   "devDependencies": {
     "@types/jest": "~22.1.1",
+    "@types/lodash": "~4.14.103",
     "@types/node": "~9.4.0",
     "@types/react": "~16.0.36",
     "@types/react-dom": "~16.0.3",


package.json
```diff
@@ -29,6 +29,7 @@
   },
   "devDependencies": {
     "@types/jest": "~22.1.1",
+    "@types/lodash": "~4.14.103",
```
abonas: Why is there a need to specify minor versions? (the 103 part here) 
the question of course stands for all the versions in the file, not only in lodash and should not block this PR

lucasponce: Isn't minor version in package.json a common approach ?
i.e.
https://github.com/patternfly/patternfly-react-demo-app/blob/master/package.json
https://github.com/patternfly/patternfly-react/blob/master/package.json
https://github.com/npm/npm/blob/latest/package.json
https://github.com/angular/angular/blob/master/package.json

abonas: @lucasponce but all those examples use caret, not tilde. and that's a slightly different meaning
https://github.com/npm/node-semver#tilde-ranges-123-12-1
https://github.com/npm/node-semver#caret-ranges-123-025-004

lucasponce: I see both uses in the package.json I attached as example
https://github.com/npm/npm/blob/latest/package.json#L36
https://github.com/npm/npm/blob/latest/package.json#L43

Or even in the lodash package.json itself
https://github.com/lodash/lodash/blob/master/package.json#L35
https://github.com/lodash/lodash/blob/master/package.json#L36

So, I guess @mtho11 has just used same depdency used from a template/example.

I don't think further justification should be necessary unless you are aware about some potential problem for this.

PR #24, opened 2018-02-19
SWS-104 Adds travis testing

josejulio:
Following React docs [1]

[1] https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#travis-ci

PR #25, opened 2018-02-19
SWS-112 Configure SCSS Patternfly resources

lucasponce:
- Adjust package.json to compile scss resources
- Copy assets on css/ relative folders
- Adjust App.tsx to point into compiled css
- Add a skeleton for patternfly react component

@mtho11 instead to hold this for the UI work I splitted and sent it separately as I think it will be shared by others.

So, with this, patternfly react seems to be working and resources configured as expected:

image

No magic here, just followed how patternfly recommends to configure this https://github.com/patternfly/patternfly-react-demo-app/blob/master/package-lock.json

Potential tweaks can be added later.


lucasponce:
Travis throws errors that are not related to this PR.

```
0.16s$ rake
rake aborted!
No Rakefile found (looking for: rakefile, Rakefile, rakefile.rb, Rakefile.rb)
/home/travis/.rvm/gems/ruby-2.4.1@global/gems/rake-12.3.0/exe/rake:27:in `<top (required)>'
(See full trace by running task with --trace)
The command "rake" exited with 1.
```

This creates some noise, so, unless is fixed in a different PR, I suggest to not count on travis for evaluate this one (or fix travis first and then I rebase it).

mtho11:
The rakefile thingy is weird, we are not using any ruby.
@josejulio any ideas?

josejulio:
Travis is just guessing, there is no `.travis.yml` on this code.
`Could not find .travis.yml, using standard configuration.`
We should ignore travis, at least until we review/merge #24

mtho11:
@josejulio thanks will merge #24

PR #26, opened 2018-02-19
Change homepage url inside package.json 

mtho11:
Change homepage url in package.json to be swift-sunshine instead of mtho11.


josejulio:
I'm not sure if that's the right way to use `homepage` on a React app.
It appears as if `homepage` is used to define the homepage for the actual running application.

See [here](https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#building-for-relative-paths) and [here](https://github.com/facebook/create-react-app/issues/1487)

jotak:
This is weird. See https://docs.npmjs.com/files/package.json#homepage
It looks like a misappropriation of this config flag by create-react-app

Btw, I still haven't understood, are we using create-react-app or not?

josejulio:
> This is weird. See https://docs.npmjs.com/files/package.json#homepage
> It looks like a misappropriation of this config flag by create-react-app
> 

Agree

> Btw, I still haven't understood, are we using create-react-app or not?

It seems so, [see this old README](https://github.com/swift-sunshine/swsui/blob/371243a0f8e17064944c90bce32d54ce9da5dd6b/README.md)

mtho11:
@jotak yes this project was created using create-react-app-ts
Hopefully, we won't have to eject the webpack config.

mtho11:
Changed to relative urls.

mtho11:
@mwringe so version 0.2.4 has the relative url changes

PR #27, opened 2018-02-19
SWS-104 Fails travis build if prettier needs to format a file

josejulio:


josejulio:
I noticed that there was a file that wasn't "pretty", but `travis` didn't complained about it. 
It's because prettier doesn't complain by default, but formats the file.

josejulio:
@mtho11 please review

PR #30, opened 2018-02-20
SWS-106 display metrics on service details page

jotak:
- Textual representation only (fancy charts are for later)
- Relaxed some tslint rules

### DIFF
--- a/src/pages/ServiceDetails/ServiceMetrics.tsx
+++ b/src/pages/ServiceDetails/ServiceMetrics.tsx
@@ -1,17 +1,133 @@
 import * as React from 'react';
+import axios from 'axios';
 import ServiceId from '../../types/ServiceId';
+import { MetricHistogram, MetricValue } from '../../types/Metrics';
 
-class ServiceMetrics extends React.Component<ServiceId> {
+type ServiceMetricsState = {
+  range: string;
+  requestCountIn?: MetricValue;
+  requestCountOut?: MetricValue;
+  requestSizeIn?: MetricHistogram;
+  requestSizeOut?: MetricHistogram;
+  requestDurationIn?: MetricHistogram;
+  requestDurationOut?: MetricHistogram;
+  responseSizeIn?: MetricHistogram;
+  responseSizeOut?: MetricHistogram;
+  healthyReplicas?: MetricValue;
+  totalReplicas?: MetricValue;
+};
+
+class ServiceMetrics extends React.Component<ServiceId, ServiceMetricsState> {
   constructor(props: ServiceId) {
     super(props);
+    this.state = {
+      range: '5m'
+    };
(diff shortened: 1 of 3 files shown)


tslint.json
```diff
@@ -22,7 +21,7 @@
         "jsx-no-lambda": false,
         "jsx-no-multiline-js": false,
         "label-position": true,
-        "max-line-length": [ true, 120 ],
+        "max-line-length": [ false, 120 ],
```
mtho11: These should probably be in a different PR. But I understand. Believe me ;)

mtho11: Right now, I have my own set of tslint.json for development.

src/pages/ServiceDetails/ServiceMetrics.tsx
```diff
@@ -1,18 +1,134 @@
...
   render() {
     return (
       <div>
-        == SERVICE METRICS ({this.props.namespace} / {this.props.service}) ==
+        <h1>
+          == SERVICE METRICS ({this.props.namespace} / {this.props.service}) ==
+        </h1>
+        Range:
+        <select value={this.state.range} onChange={this.onRangeChanged}>
+          <option value="1m">1 minute</option>
```
mtho11: At some point we should probably make this data driven and not hard coded.

jotak:
@abonas here's a screenshot: img

pilhuhn:
Great @jotak (I personally would more like the bullets be left-aligned, but that is not a blocker or anything).
Perhaps in the next round wrap the sections in some pattern fly box with INPUT/OUTPUT/.. being a header and the bullets being the box content

jotak:
@pilhuhn this is very temporary as the enxt step I think is to replace it with charts, so I didn't even bother in aligning / grouping stuff

pilhuhn:
On 21 Feb 2018, at 16:33, Joel Takvorian wrote:

> @pilhuhn this is very temporary as the enxt step I think is to replace 
> it with charts, so I didn't even bother in aligning / grouping stuff

I think this is already great as a table.

How do I access that page inside the UI ? What is the url?

jotak:
`/namespaces/tutorial/services/preference` but you'd need to fetch that PR server-side, which is not merged: https://github.com/swift-sunshine/swscore/pull/30

src/types/Metrics.ts
```diff
@@ -0,0 +1,10 @@
+export interface MetricValue {
+  Value: number;
+}
+
+export interface MetricHistogram {
+  Average: number;
```
mtho11: These should be lower cased No?

src/pages/ServiceDetails/ServiceMetrics.tsx
jotak: @mtho11 I've read your comment too late. You're right, I must bring changes on bother core&ui

PR #31, opened 2018-02-20
Api Service and configuration

aljesusg:
I am working in [SWS-69](https://issues.jboss.org/browse/SWS-69) about display information of service and I wanna make a proposal about how do the calls to the API and how load the configuration. 

@jotak. I saw your PR[#30](https://github.com/swift-sunshine/swsui/pull/30). What do you think about have the connector and configuration like this?

After that you can control then and catch with promise returned in the class
```js
  const api = new Api();
    api.route()
        .then(response => {
            console.log(response);
        })
        .catch(error => {
            console.log(error);

        });
```

JIRA [SWS-116](https://issues.jboss.org/browse/SWS-116)


lucasponce:
Let's split these things in JIRAs.
I found a similar one working on the service list, so, I stopped working on that and coding this
https://issues.jboss.org/browse/SWS-114
(It's as simple as this one, but I guess it will help to coordinate and give visibility to others as well).
I mean, this PR can go in a similar JIRA as I prepare for the navigation one, it's quite similar and probably I will reuse as well.

aljesusg:
Great @lucasponce I didn't see any Jira related with this. Should I open a Jira for this?

lucasponce:
@aljesusg sure, granularity is not the problem.
Now it's a little bit overwork but the idea is at least to track that when @mtho11 or others are planning something similar, it shows we are working on the same areas.

jotak:
@aljesusg : I don't see where the resource path gets appended to the url. I guess we should have something like: `api.route("/namespaces/etc.")` and the argument is appended to the base url, right?

aljesusg:
Yes, this is the idea, how have a methods for this paths, for example. Mi idea is have all related with the API in the same file.
```js
getNamespaceService(namespace_id: string, service_id:string) {

}
```

mtho11:
Do we want to have builder pattern to fluently build the request.  Easily readable and configurable, also the parameters are typesafe.
Or at least the `ApiOpions`, `ApiRequest` or `ApiConnection` Object could be configured by the Builder and passed to the Api.
`Request.withDefaultOptions()`

aljesusg:
ok @mtho11, @jotak and @lucasponce . What do you think about this new approach?

Now we call to the method function like this
```js
const api = new Api();
    api
      .call('GetServices', { namespace_id: '2', service_id: 2 }, {}, {})
      .then(response => {
        console.log(response);
      })
      .catch(error => {
        console.log(error);
      });
```

where our configuration json is

```json
{
  "version":  "0.1",
  "backend": {
    "url": "http://sws-istio-system.127.0.0.1.nip.io/api",
    "user": "jdoe",
    "password": "password",
    "endpoints":{
      "GetServices": {
        "path": "/namespaces/<namespace_id>/services",
        "method": "get"
      }
    }
  }
}

```

So the services/api is looking for the key "GetServices" in the endpoints in the configuration file and replace the "path" with regular expressions with the values  that you pass in the method { namespace_id: '2', service_id: 2 } , you can also add in 3º and 4º params of the method, the params for query and the data if you wanna make a post.

With this we can configure all endpoints in the configuration json. it could be more cleaner but it is a fast approach.

jmazzitelli:
BTW: this assume BASIC auth (user/pass). I've seen OpenShift things require Bearer Tokens. I don't know what we are going to require or need, so I suggest the credentials are wrapped in a "credentials" section that can support basic auth or bearer token auth.

For example, see for our [config/security/Credentials](https://github.com/swift-sunshine/swscore/blob/master/config/security/config_security.go#L16) struct, the [function to extract the http header based on auth being used](https://github.com/swift-sunshine/swscore/blob/master/config/security/config_security.go#L47), and a [validation function](https://github.com/swift-sunshine/swscore/blob/master/config/security/config_security.go#L31) to check if Credentials is valid.

src/config.json
```diff
@@ -1,8 +1,14 @@
...
-  backend: {
-    url: "http://sws-istio-system.127.0.0.1.nip.io/api",
-    user: "jdoe",
-    password: "password"
+  "version":  "0.1",
+  "backend": {
+    "url": "http://sws-istio-system.127.0.0.1.nip.io/api",
+    "user": "jdoe",
+    "password": "password",
+    "endpoints":{
```
jotak: I'm not for having everything in config. Url really makes sense because we may want to switch it depending on the environment. But IMO endpoints are not configuration. When reading the code it could be sometimes annoying to have to read several different files just in order to understand which endpoint is being used. Plus, it adds extra regexp processing that doesn't seem necessary to me. For that I'd prefer a more straight approach.

I don't know what the others think about it.

mtho11:
Personally, I really like where this is going, but would like to see more Type safety in arguments and responses. I should know what I'm putting in and getting out. Maps of strings are little bit loose and easy to make mistakes. Generics, interfaces or types would be my suggestion.
Others feel free to chime in...

I'm also pragmatic about getting this stuff in and used.

mtho11:
I don't think it should create a `new Api()` each time. I think it should come from a factory like `const api = Api.newRequest().withCredentials(user, password)` Instantiating a new instance each time seems overkill especially considering a certificate or bearer stuff may already be known.

src/config.json
lucasponce: It's odd to see this here, as well, but I'm not sure what are the best practices about how to setup this.
Final architecture will be a pod serving the api and sharing the static content, so I guess we should address that first, and perhaps second is to work on helpers setup for development mode.

src/types/Config.ts
```diff
@@ -0,0 +1,18 @@
+interface Config {
+  version: string;
+  backend: SWSCore;
+}
+
+interface SWSCore {
+  url: string;
+  user: string;
+  password: string;
+  endpoints: Map<string, Endpoint>;
```
jotak: Oh, I didn't know Typescript now have `Map`. Nice!

src/config.json
jotak: What about having the routes defined in Api.tsx (btw: should be .ts) instead of config.json? we would just call:
`api.getServices(namespace).then(...)`

Simple and efficient, no?

mtho11: works for me

lucasponce:
I am also kinda blocked by this for my next task. As most comments are related to the use, I suggest that @aljesusg fixes the travis issue and we can merge it.
So, we can start using it and address the enhancements suggested after we can give some tries putting together.

aljesusg:
If all is agree with this I am gonna clean this with all comments and push a new version today.

aljesusg:
Here is the new version.

- I changed the implementation to functions to avoid the new instance each time.
- I moved the config to a tsx.
```tsx
export function config() {
  return {
    version: '0.1',
    backend: {
      user: 'jdoe',
      password: 'password',
      endpoints: {
        Root: {
          path: '/api',
          method: 'get'
        },
        GetServices: {
          path: '/api/namespaces/<namespace_id>/services',
          method: 'get'
        },
        GetServiceMetrics: {
          path: '/api/namespaces/<namespace_id>/services/<service_id>/metrics',
          method: 'get'
        }
      }
    }
  };
}
```
- Now See [ServiceMetrics](https://github.com/swift-sunshine/swsui/pull/31/files#diff-54dd2350fed61642856e4025e72f734a) 
```js
API.newRequest(
      'GetServiceMetrics',
      { namespace_id: this.props.namespace, service_id: this.props.service },
      { range: this.state.range },
      {}
    )
```
I tested that the url is ok with queryparams and ids.
- If and endpoint is not defined this return an error with 'Endpoint not defined in config file'

src/services/Api.tsx
```diff
@@ -0,0 +1,57 @@
+import axios from 'axios';
+import { config } from '../config';
+
+export function newRequest(method: string, values: any, queryParams: any, data: any) {
```
mtho11: Why not use ES6 arrow syntax? `function` is so old school ;)
Plus with typescript, the arrow function fixes the `this` issues too -- bonus.

aljesusg:
I changed the syntax @mtho11

lucasponce:
LGTM, I see @mtho11 are addressed.
As primary user of this code, I'm for merge it and adjust once we link with the pages.
It's a start and the overall approach is shared by the team.

src/services/Api.tsx
```diff
@@ -0,0 +1,57 @@
+import axios from 'axios';
```
jotak: It should be .ts file as long as there's no JSX syntax inside. I think there's not here.

src/pages/ServiceDetails/ServiceMetrics.tsx
```diff
@@ -79,12 +79,15 @@ class ServiceMetrics extends React.Component<ServiceId, ServiceMetricsState> {
 
   fetchMetrics() {
     console.log('Fetching metrics...');
-    axios
-      .get(`/api/namespaces/${this.props.namespace}/services/${this.props.service}/metrics?range=${this.state.range}`)
-      // .get(`http://127.0.0.1:8000/api/namespaces/${this.props.namespace}/services/${this.props.service}/metrics?range=${this.state.range}`)
+    API.newRequest(
```
jotak: I still prefer `API.GetServiceMetrics(this.props.namespace, this.props.service, this.state.range)`
Config-based endpoints is more error prone. The compiler will never tell you if you're trying to reach an endpoint that doesn't exist

aljesusg: So...We'll need a method for each endpoint. What do you think about pass the query params like range inside a json to let axios work with them?

like  API.GetServiceMetrics(this.props.namespace, this.props.service, {range :this.state.range})
with this we can add more queryparams in future without change the method.

jotak: +1 :-)

aljesusg: Changed @jotak, review please

jotak:
LGTM, thanks @aljesusg

PR #32, opened 2018-02-21
Add editorconfig, the universal editor config settings

mtho11:
So folks can just get the right settings for the editor and have the IDE 'know' the right settings for indent spacing etc...

Most IDEs and editors support this.

http://editorconfig.org/


jotak:
Never used it but heard a lot about it... That may be the occasion for me to catch up. +1

PR #33, opened 2018-02-21
SWS-114 Add a minimal navigation

lucasponce:
- Define a vertical navigation
- Link with routes
- Fixes some style affecting proper components layout

Note, this is just to introduce a proper navigation into the application.
It doesn't enforce to choose vertical/horizontal but in the steps we are moving, it is good to have something.
Some rationale: I started working with the PatternFly React component but it has some bug with the Routes, so I prepared a simpler component that works (most work is just to investigate what people are doing and we are mimic the best practices defined in the patternfly react demo app).

With this, it will be easier to continue working on proper components and put things together.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,6 @@ cache:
     - node_modules
 script:
   - npm run prettier -- --list-different &&
+    npm run build-css &&
     npm run build &&
     npm test
-
--- a/package-lock.json
+++ b/package-lock.json
@@ -5188,7 +5188,8 @@
         },
         "jsbn": {
           "version": "0.1.1",
-          "bundled": true
+          "bundled": true,
+          "optional": true
         },
         "json-schema": {
           "version": "0.2.3",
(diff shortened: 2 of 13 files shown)


lucasponce:
image

image

abonas:
@mwringe @mtho11 fyi

lucasponce:
Locally I hit a different error, but related to the env.
Fixed in Linux like
```
echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p
```
It seems npm test needs some param in travis (probably headless related).
Still looking.

lucasponce:
Ok, there were a couple of things here:
- prettier detect something not well formated but the error was unclear.
- Also there was a missing a build-css in travis to make tests pass.

Now it's ok.

abonas:
nice!! please check the icons to:
service should be `pficon-service`
graph should be `pficon-topology`

lucasponce:
> nice!! please check the icons to:
> service should be pficon-service
> graph should be pficon-topology

Those will be adjusted in a following PR, this is just for the technology used in the navigation, all labels, sections, etc are just examples.

PR #34, opened 2018-02-21
Adds information on how to configure proxy to swscore server

josejulio:
This PR has the configuration required to point to swscore inside openshift without having to do any tricky stuff (build swscore docker again, rsync, etc) while developing the UI.

We only have to add the `proxy` configuration [1] to our `package.json` and point it to our swscore endpoint.

# SwsCore inside openshift
## UI
screenshot from 2018-02-21 12-39-33

## API
screenshot from 2018-02-21 12-39-41

# SwsUI in development
screenshot from 2018-02-21 12-40-47

## API
screenshot from 2018-02-21 12-46-15

Notes: 
- We can't access the API by writing its URL on the browser, is because the proxy will only serve requests without the `text/html` accept header [2].
- This only works when running `npm start` so a production build is not affected.
- This can be further configured [3] if we need to access many endpoints from different URLs.

[1] https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#proxying-api-requests-in-development
[2] https://github.com/facebook/create-react-app/issues/1378#issuecomment-272182020
[3] https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#configuring-the-proxy-manually

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -25,6 +25,41 @@ But, if it doesn't then use:
 This launches a development environment that instantly
 reloads any changes to the browser for rapid development.
 
+== Developing
+
+When developing, is usual to run this application outside of https://github.com/swift-sunshine/swscore[swscore].
+
+Is possible to proxy the API requests without having to push the changes to swscore.
+Add the https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#proxying-api-requests-in-development[proxy]
+property to `package.json` with the url of the swscore.
+[source, json]
+----
+{
+  "name": "swsui",
+  "version": "0.2.1",
+  "proxy": "http://sws-istio-system.127.0.0.1.nip.io",
+  "bugs": {
+...
+----
+
+Run `npm start` and try it!
+[source, bash]
+----
+$ curl -u jdoe:password http://localhost:3000/api
+Namespaces: [default istio-system kube-public kube-system myproject openshift openshift-infra openshift-node]
+
+Namespace: default Services [docker-registry kubernetes router]
+
+Service Name: docker-registry
+Service Labels:
+docker-registry = default
+Type: ClusterIP
+...
+----
+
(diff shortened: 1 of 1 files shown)


josejulio:
@jotak Let me know if this helps you

jmazzitelli:
1) we should put this in the swscore README so people that are using that can know how to get the UI to talk to it via this proxy

2) do we need to worry about CORS? There is a new server/cores_allow_all setting in the swscore yaml config file - do we need to set that to true?

josejulio:
> we should put this in the swscore README so people that are using that can know how to get the UI to talk to it via this proxy
 
Agree, will put it there.

> do we need to worry about CORS? There is a new server/cores_allow_all setting in the swscore yaml config file - do we need to set that to true?

I don't think we need to worry about that, they are on the same domain with this proxy.

josejulio:
> we should put this in the swscore README so people that are using that can know how to get the UI to talk to it via this proxy

I added the information to swsui Readme, I think it belongs here. People who uses this repo would use that information, not (i think) the other way.
Anyway i'm willing to put that info on swscore/Readme if you really think should belong there (or maybe refer to this docs from there?)

pilhuhn:
I think this is a good idea.
What we also should do if we go this route is to start adding (i)tests, that are run on travis or other CI, testing the final scenario. Otherwise we may run into too many "works on my laptop (only)" situations.

jmazzitelli:
@jotak I would keep the cors option, only because "you never know" :) The default is false, so its not like we have it on by default.

jmazzitelli:
@josejulio I'm ok leaving the info solely in the swsui README. Perhaps we could add some NOTE or other quick blurb on swscore README about this. I don't know where it would go. I'll leave that up to you if you want to make a PR for that or not. Doesn't matter to me.

PR #35, opened 2018-02-22
Add tslint rule for only ES6 arrow functions (no function syntax)

mtho11:
There really should be no need for 'function'. Especially when Traditional functions don’t bind lexical scope (and Typescript does), which can lead to unexpected behavior when accessing ‘this’.

https://palantir.github.io/tslint/rules/only-arrow-functions/

### DIFF
--- a/tslint.json
+++ b/tslint.json
@@ -60,6 +60,7 @@
             "check-open-brace",
             "check-whitespace"
         ],
+        "only-arrow-functions": true,
         "quotemark": [true, "single", "jsx-double"],
         "radix": true,
         "semicolon": [true, "always"],


jotak:
Actually, it seems that with that rules now it refuses writing methods such as:
`export function config() { ... }`

Locally I get:

```
/swsui/src/config.ts
(1,8): non-arrow functions are forbidden
```

And travis shows similar error

josejulio:
> Actually, it seems that with that rules now it refuses writing methods such as:
> export function config() { ... }

I think you have to write them like:
`export config = () => { ... }`

Or something like that :)

lucasponce:
This PR would need to rebase the pending 9 ones to adapt it.
I guess that we consolidate first for Sprint 1 and we can then apply code style changes.
Murphy has no mercy for demo effects.

cfcosta:
The earlier we do change the lints the better, more code will make them hard to change later. I'm +1 for this change.

mwringe:
This issue has been open for a while. Do we still want this or should we close the PR?

josejulio:
Superseded by  #119

PR #36, opened 2018-02-22
Move App out of src root and into components/App.

mtho11:
All the App.* files don't belong in the root. Let's put them in their own directory since it will only grow.


jotak:
actually... travis failed. Css needs to be regenerated I guess

jotak:
Superseded by #39 
@mtho11 ok to close?

PR #37, opened 2018-02-22
Use low camel case in json + adding pf spinner while loading metrics

jotak:


jotak:
related to https://github.com/swift-sunshine/swscore/pull/60

PR #39, opened 2018-02-22
SWS-130 Fix css generation issue - move App to dedicated folder

jotak:


jotak:
@lucasponce , @mtho11 this would supersede https://github.com/swift-sunshine/swsui/pull/36

Coincidentally because of patternfly, App.css must land in a 1-depth length directory from src root (previously, /src/css was ok, now it's in /src/app)

I'm using the rule of css generated next to their tsx counterpart, which is common practice.

The previous script was building only App.scss and no other scss file, so it's fixed with this patch.

jotak:
@lucasponce this was not part of a task actually, it's more a bug fix to allow other components scss to be built. I just created a jira ticket.

lucasponce:
> @lucasponce this was not part of a task actually, it's more a bug fix to allow other components scss to be built. I just created a jira ticket.

Thanks with this we will contribute to run a git log and check the JIRAs (Tasks or Bugs).

PR #40, opened 2018-02-22
[WIP][SWS-69]Display the service details information

aljesusg:
Working in Service Details Information UI

JIRA [SWS-69](https://issues.jboss.org/browse/SWS-69)

@jotak I added tabs for Information and Metrics, I think there is a lot of information about the service for show in only one page.(See GIF)

GIF with the last updated (Mock data):
peek 2018-02-22 18-29

### DIFF
--- a/src/pages/ServiceDetails/ServiceDetailsPage.tsx
+++ b/src/pages/ServiceDetails/ServiceDetailsPage.tsx
@@ -3,6 +3,7 @@ import { RouteComponentProps } from 'react-router-dom';
 import ServiceInfo from './ServiceInfo';
 import ServiceMetrics from './ServiceMetrics';
 import ServiceId from '../../types/ServiceId';
+import { Nav, NavItem, TabContainer, TabContent, TabPane } from 'patternfly-react';
 
 export default function ServiceDetails(routeProps: RouteComponentProps<ServiceId>) {
   return (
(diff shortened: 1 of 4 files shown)


lucasponce:
It's a nice progress.

Just some comments from "user's hat" perspective.
After playing with Istio some weeks, I guess that main info (for our user) could be resumed in:
- Labels per service
- Labels per pod
- Istio rules (as they are defined from previous labels).

So, I guess that if we bring three UXD engineers they would provide three different views (all of them correct) but in terms of what info is important, I guess my notes make sense.

With that in mind, any proposal that could help to identify that info in a first view, I guess it would be most welcome.

Other info can be showed with less priority (i.e. perhaps as secondary info), I don't know, but just leaving these thoughts here if it could help.

abonas:
GREAT
as we talked, some UX proposals:
1. make the service itself clickable instead of a "details" button (many services will result with a lot of details buttons - it won't be looking so nice, and those kind of buttons are less popular today in web)
2. First row to have service name, type, IP as one row rather than how it looks now (plus larger font please :))
3. no need to display twice the same information about pods/endpoints
4. it looks like either font size or also font type is different through the page - is it easy to fix?
5. from the name istio-system/productpage is not really clear that istio-system is the project/namespace name, I would separate them and make it clear "what is what".

lucasponce:
> make the service itself clickable instead of a "details" button (many services will result with a lot of details buttons - it won't be looking so nice, and those kind of buttons are less popular today in web)

Note that these changes will be addressed in a separate PR where we are working the Service List page itself, as list / detail pages have been splitted in different JIRAs.

pilhuhn:
Yes, please get this one merged and then address the suggestions from @abonas afterwards. This way we make visible progress sooner.

pilhuhn:
https://travis-ci.org/swift-sunshine/swsui/builds/344897548#L482

aljesusg:
Let me finish the view ^^. Can you add label "wip" or "not merge"? Thanks.

jotak:
The travis error may be fixed if you rebase (not sure)

PR #41, opened 2018-02-22
SWS-128 Fix tsconfig compiler error messages

mtho11:
These messages currently show up:
```
Failed to load tsconfig.json: Missing baseUrl in compilerOptions
Creating an optimized production build...
Found no baseUrl in tsconfig.json, not applying tsconfig-paths-webpack-plugin
Found no baseUrl in tsconfig.json, not applying tsconfig-paths-webpack-plugin
```

https://issues.jboss.org/browse/SWS-128

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,5 +1,6 @@
 {
   "compilerOptions": {
+    "baseUrl": "./",
     "outDir": "build/dist",
     "module": "esnext",
     "target": "es5",


lucasponce:
Travis is failing.
I saw that before, perhaps the travis command used for prettier is giving something different ?
I faced that yesterday and travis doesn't warn with specific error when prettier fails, but it throwed a similar error.

mtho11:
@lucasponce all checks have now passed

PR #43, opened 2018-02-22
Use nodejs 8 for CI

josejulio:
Also applies `prettier` to `src/pages/ServiceDetails/ServiceMetrics.tsx` as it was failing our CI


josejulio:
@mtho11 I think this file was failing our CI, please review

PR #44, opened 2018-02-23
SWS-63: Preliminar mock for ServiceList page

lucasponce:
- Changed proper icons on navigation and service
- Introduce basic features proposed for service list: filtering and paging
- This is pure mock with demo data, no connectivity with backend

I wanted to split SWS-63 in two parts, this is just a mock data showing the visual elements we are going to add, following PR will complete the connectivity with real backend calls, but to not send a big PR, I guess it's better to propose this way (note. I will ruse same JIRA code for next commit).

So, this page shows the list of services and will provide basic filtering by namespace or service name and also a basic pagination (as we don't know how many services we will have).

Also, in the list we can anticipate some info from the details, i.e. summary of pods, versions, istio rules, or even some mini graph. These is just food to thing for following sprints on planning phase.

The goal for Sprint 1 should be to make this works end-to-end.

This is a snapshot

image

React components are not definitive, just to put the elements in place, in the following PR once backend is connected those will be re-organized to share state correctly.

### DIFF
--- a/src/pages/ServiceList/ServiceListComponent.tsx
+++ b/src/pages/ServiceList/ServiceListComponent.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { ListView, ListViewItem, ListViewIcon, Button } from 'patternfly-react';
+import { ListView, ListViewItem, ListViewIcon } from 'patternfly-react';
 import { Link } from 'react-router-dom';
 
 class ServiceListComponent extends React.Component {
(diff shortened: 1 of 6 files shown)


src/pages/ServiceList/ServiceListComponent.tsx
```diff
@@ -1,30 +1,64 @@
...
-            actions={
-              <div>
-                <Link to={'/namespaces/istio-system/services/ProductPage'}>
-                  <Button>Details</Button>
-                </Link>
-              </div>
-            }
-            description={<span />}
-          />
+          <Link to={'/namespaces/istio-system/services/ProductPage'} style={{ color: 'black' }}>
```
pilhuhn: please use lowercase 'productpage' here as this is what the backend knows.
Same for ratings and reviews further down.

lucasponce: By covention K8S world works with lowercase names
https://kubernetes.io/docs/concepts/overview/working-with-objects/names/
So I agree we should move these even in a mock.
Also, UI speaking Kubernetes console also follows this
https://kubernetes.io/docs/tasks/access-application-cluster/web-ui-dashboard/

lucasponce:
Looking why travis is unhappy...

jotak:
LGTM (when travis will be happy)

lucasponce:
Okis, there were a strange issue, I used ES6 notation in some function and I moved back as prettier didn't like it.
Anyway, functionality is intact.
This should be ready to get in.

jotak:
@lucasponce what's the ES6 features not working with prettier? I also had some conflicts between tslint and prettier... If we have too many of these maybe we should remove prettier.

PR #45, opened 2018-02-23
SWS-131 change homepage (= base url) to /

jotak:
This is to enable client-side routing. Else HTML contains links such as "/services/static/js/main.f06a050d.js" instead of "/static/js/main.f06a050d.js" when hitting url "http://sws.../services"

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,5 +68,5 @@
   ],
   "author": "Red Hat",
   "license": "Apache-2.0",
-  "homepage": "."
+  "homepage": "/"
 }


pilhuhn:
When I deploy this together with https://github.com/swift-sunshine/swscore/pull/64, then the start page does not show at all.

lucasponce:
LGTM
Just one question, this has no effect when using npm start ? isn't it ?

jotak:
@lucasponce I tried with npm start as well and it was working. I hope I'm not missing any use case. Don't know why "." was set in a first place.

lucasponce:
Okis, just to confirm :-). Merging.

PR #46, opened 2018-02-23
SWS-132: Fix fluid navigation

lucasponce:
- Bring back PatternFly React Vertical Navigation Component
- Workaround issue with react-router-dom Link
- Simplify Routes declaration and management

Following patternfly-react-demo apps we didn't use the VerticalNav because combined with the Link from react-router-dom created some issues with the routes.

https://github.com/patternfly/patternfly-react/issues/239

So, in a first attempt we used similar navigation with plain divs and raw patternfly style.
That was fine until we realized that fluid navigation was broken as some missing pieces, so instead to include raw javascript to manage toggled and fluid style we have revisited the VerticalNav component from Patternfly React and introduced a `navigateTo` workaround.

As we need to use the route path on the workaround, I think is better to locate together routes and navigation.

Then we can simplify the navigation and fix the fluid issues using this component.

Check that fluid is fixed:
image

image


lucasponce:
Travis is complaining about a failing test. Looking.

lucasponce:
Locally the fix works fine but there is something odd in a test.
I will continue investigating it.

lucasponce:
Ok, I found the issue, quite strange one and described here
https://github.com/Microsoft/TypeScript-React-Starter/issues/8
So, I enabled the 
```
    "allowSyntheticDefaultImports": true
```
And code and tests are correct and Travis passes.

This is ready for merge.

lucasponce:
Please @mtho11 take a look :-) and we can win a weekend here.

mtho11:
Wow, there is weird little error. Good find!

PR #47, opened 2018-02-24
Add Cytoscape layout for the Service Graph page

mtho11:
Preliminary Service Graph with Cytoscape

image

https://issues.jboss.org/browse/SWS-73

Next steps to hook up to Jay's live data (we have already agreed on a cytoscape data format to send).
Still need to tweak look-n-feel more, add namespace selection and linking to other screens.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3720,6 +3720,31 @@
       "integrity": "sha1-GzN5LhHpFKL9bW7WRHRkRE5fpkA=",
       "dev": true
     },
+    "cytoscape": {
+      "version": "3.2.9",
+      "resolved": "https://registry.npmjs.org/cytoscape/-/cytoscape-3.2.9.tgz",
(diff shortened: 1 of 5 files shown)


mtho11:
Hopefully, we can merge this quickly so everyone can start playing (since we don't have much time). And then address all concerns.

mtho11:
Should a click on node take you to the detail page or render information about that node (say in a popup info block)?

mtho11:
Also, in general, I guess we are going with a white background in our main page section when the patternfly sections are dark?
Seems like we should be lite or dark profile?

mtho11:
Because we have no real wireframe mockups everyone is just making stuff up at this point. Every developer with their own style.

theute:
White background is what we (mostly) see in OpenShift console, so +1 on white background only.
And yes, let's merge this one when reviewed that it works and improve the LnF in other iterations

pilhuhn:
Checked out the current status locally and this shows nicely. So also +1 on merging now and refining afterwards.

PR #48, opened 2018-02-24
[SWS-69]Display the service details information

aljesusg:
Working in Service Details Information UI

JIRA [SWS-69](https://issues.jboss.org/browse/SWS-69)

@jotak I added tabs for Information and Metrics, I think there is a lot of information about the service for show in only one page.(See GIF)

GIF with the last updated:
peek 2018-02-28 18-32

Error message
screenshot from 2018-02-28 18-31-46


lucasponce:
Nice progress, a couple of suggestions from my side.
Let's think in our main example
image
So, when I click in reviews, main info should be 
- Versions of the service (in practice each version is mapped with a pod but the "relation" between pod/version is defined in the pod labels, that's why I commented labels should be interesting to show).
- Dependencies from sources (just to focus in a service and understand the incoming connnection, outcoming is less interesting as from Istio perspective we don't touch that).
- Istio rules (if they exist), so, they will explain the relation, weights between the connections before.

I guess rest of info is less interesting, well, that's relative, everything is important but as our focus is not kubernetes in general but to give a Istio view, I guess we should try to mark this info as main one (in terms of what I should put my attention when I enter in the details page).

Also, thinking that as the labels, and rules can be tiny, I wonder if we could use a card view with these concepts to avoid table data with all of this tiny info, like
http://www.patternfly.org/pattern-library/content-views/card-view/

So, as summary, I guess our goal should be to explain what dependencies I have and why a traffic rule is defined, and that should be driven the UX in the details.

Note I didn't forget the metrics, just focus on the service details and rules, I think the tabs for details/metrics makes sense for this iteration.

aljesusg:
Mmm ok I don't know what information will provide the endpoint yet. Where can I find some example about this information?

About the card-view, patternfly-react not have it yet, I can add it with the core implementation while they finish the component.

lucasponce:
> Mmm ok I don't know what information will provide the endpoint yet. Where can I find an example?

In the SWS-68 we discussed about it
https://issues.jboss.org/browse/SWS-68?focusedCommentId=13530601&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13530601

And in the root handler I prepared a demo focused on this info as well:
https://github.com/swift-sunshine/swscore/blob/master/handlers/root.go#L64

Istio rules has not a well predefined format, so we will need to play with it in future iterations, but basically they define traffic conditons based on labels.

> About the card-view, patternfly-react not have it yet, I can add it with the core implementation while they finish the component.

I don't think we should be limited by the patternfly-react components, some of them are just pure css and we can woraround them in our component. Indeed the patternfly-react demo apps combines both.

But I guess that using card-style view instead of tables would help to group better the info (inside the cards we can use lists/tables, of course, but the idea is to reduce the scrolling effect of the user).

aljesusg:
But in the root handler there isn't information about versions or istio rules format :'(

About components I was thinking in collapse lists to avoid scrolls. Thanks @lucasponce.

lucasponce:
> But in the root handler there isn't information about versions or istio rules format :'(

Versions are defined as labels, there is not a "version" type as it is, meaning that istio uses labels to select pods and control traffic, that's why I put the labels as first class element on the visualization.

About Istio rules, you are right, are quite open, and for a first iteration we didn't enter to map all the details, just an example to clarify things:

This is a rule defined for the `reviews` service, and basically adds a filter to move traffic to a pod selected by a labels expression.
```
apiVersion: config.istio.io/v1alpha2
kind: RouteRule
metadata:
  name: reviews-test-v2
spec:
  destination:
    name: reviews
  precedence: 2
  match:
    request:
      headers:
        cookie:
          regex: "^(.*?;)?(user=jason)(;.*)?$"
  route:
  - labels:
      version: v2
```
So, for this reason, from Istio perspective I would like to get in a first sight without entering in too many click this info.

About UX, it's quite open, and I guess that will iterate a lot, I imagine that having something like cards grouping info a first sight, and inside the cards list that can collapse to show details could be a nice start.

aljesusg:
Ok I understand now, Tomorrow will be a study day !!! ^^

abonas:
what is the Dependencies about? I'm not sure I understand what entity/relationship is that

lucasponce:
> what is the Dependencies about? I'm not sure I understand what entity/relationship is that

I refer about the sources dependencies for a service.
Using the example, let's focus on `reviews` service, this one is invoked by `productpage` service and also it invokes `ratings` service, too. 
Istio rules will cover the traffic for incoming/sources services (but not for services invoked from inside, so `ratings` won't be interesting).

So, `reviews` have three versions and it might have some Istio rules defining logic that control traffic to one to another service.
So, as a user I want to know which are the sources dependencies to understand why I defined the rules for this specific service.

But from the details perspective, perhaps it's less interesting to focus that `reviews` service invokes `details` (this would be showed when click on `details` service details page).

abonas:
@lucasponce perhaps the term dependencies confuses me here. I reread your comment above several times, I am still not sure. is it who depends on  (and depends means who is called by) the current selected service? or who is the current dependent service called by/from?

lucasponce:
@abonas not sure if there is a better term here, but I mean "dependencies" for a service, in the context of the discussion, the overall sources (who calls the service) and destinations (others services called by this service).

So, in the example attached, "reviews" service has "productpage" as a source, and also "ratings" is being called from "reviews" service.

The point is that, focus on the istio rules we should display the "sources" services, as those are affected by the control traffic.

I used "Dependencies" to refer the inter-relations of a service with other services.

pilhuhn:
Can we please get this merged and then refine afterwards?

jotak:
@aljesusg the details/metrics split with tabs in fine for me.
Currently this is only mocked data.
Do you want to continue to work on this PR, or open another one to finish the task?

aljesusg:
I don't when we can finish this, the endpoint is not ready so we can merge and work in another PR I think. What do you think @lucasponce @abonas ?

lucasponce:
If it's a question,  I would prefer to see the mock more structured, just to progress in the visual elements.
If we want to merge and refine later, is fine with me as well, but then we should adopt a different approach of review.
For example, letting merge minor pieces and give a full review when JIRA owner asks for it.
If not the meaning of these discussions is lost.

aljesusg:
Updated version, now is connected with backend, See the gif with the new view.
@lucasponce What do you think?

lucasponce:
@aljesusg it looks very nice, it's getting shape. Two things:
- "Dependencies" -> "Source Services" ( I guess that is more clear )
- "Source Services" -> "Route Rules" or "Istio Route Rules" as it is specific for Istio I guess.

Perhaps minos UX adjustments can be done, but now it represents better the concepts we have discussed, I click on a service and I see:
- Main service information.
- Pods with it's labels representing the versions of a service.
- Source services (dependencies with this service).
- Istio route rules that affect to the source services.

lucasponce:
I am not sure if @aljesusg is available today.
I can take this commit, rebase locally and merge from my repo, so we can make it for Sprint 1.
Any objections @jotak, @abonas, @pilhuhn ?

There are UX things to improve but from visualization concepts discussion comments are addressed.

pilhuhn:
On 1 Mar 2018, at 9:22, Lucas Ponce wrote:

> I am not sure if @aljesusg is available today.
> I can take this commit, rebase locally and merge from my repo, so we 
> can make it for Sprint 1.

Yes please get it in.

lucasponce:
Conflicts addressed on https://github.com/swift-sunshine/swsui/pull/67

PR #49, opened 2018-02-26
SWS-131: prefix all client routes with /console

jotak:
Needs https://github.com/swift-sunshine/swscore/pull/64 on server side
PS: for motivations, see discussion on server-side PR

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -6,7 +6,7 @@ import { BrowserRouter } from 'react-router-dom';
 class App extends React.Component {
   render() {
     return (
-      <BrowserRouter>
+      <BrowserRouter basename="/console">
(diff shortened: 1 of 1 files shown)


lucasponce:
mmm is this going to work with the local npm start development mode ?

jotak:
@lucasponce yes it does. React-router does all the work (and apparently does it well)

lucasponce:
@jotak if this is tested with the development trick to use proxy on package.json
```
  "proxy": "http://sws-istio-system.127.0.0.1.nip.io",
```
then it's fine :-). Just wanted to double check to not break the common flow of development UI + backend.

jotak:
Just tested with proxy, I confirm it works

PR #50, opened 2018-02-26
No var keyword tslint

mtho11:
ES6 best practice is to not use `var` but instead use `const` or `let` which are more explicit and have lexical scope boundaries that are more concrete (no hoisting errors). With `var` we also don't know the intention: is it mutable or immutable?


josejulio:
Some [info](https://eslint.org/docs/rules/no-var) on why `var` is discouraged.

PR #51, opened 2018-02-27
Remove Home Page until we actually have a real one. 

mtho11:
Until then, we can use service-graph page as home page. Also remove react.js logo.svg since it was only used on homepage.


jotak:
@mtho11 I get the same error than travis when running test locally. It looks like in-test rendering is suffering with cytoscape?

lucasponce:
@mtho11 can you rebase, or take a look into the travis error ?
Also I promise to invite you to your preferred beer if you also add a JIRA into the commit.

I'm going to start to make people pay $5 per commit without JIRA code :-).

lucasponce:
Ok I will merge.

Note:
nBeersMikeWillPay = nBeersMikeWillPay + 1;

lucasponce:
Note, we will fix test in a following PR to not stall these PRs as jest tests are not yet formelly defined.

PR #54, opened 2018-02-27
SWS-63: Services list UI page

lucasponce:
- Support for filtering by namespace and/or service name.
- Support for pagination.
- Sorted alphabetically namespaces/servicename.
- Selected filtered are maintained across navigation.

image
image
image
image


lucasponce:
Probably I need to rebase and check that travis is ok but I guess it is almost ready.
@aljesusg take a look as well as we need to sync them both pages.
In theory, details shouldn't take anything special as filtered are propagated.

aljesusg:
If you are saving the status of the filters, I don't think it will harm the navigation from service details, anyway I am gonna check it. This is a nice improve in the view @lucasponce ^^ Have you seen the sort addon https://rawgit.com/patternfly/patternfly-react/gh-pages/index.html?selectedKind=Sort&selectedStory=Sort&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs?

lucasponce:
@aljesusg yes, selected filters are stored in a namespace, so if the filter is reused by the graph, same filtering will be propagated.

I think the sort addon is a good idea and it is cheap to add it, I'm going to introduce it in the PR.

lucasponce:
Also I realized I didn't show the errors, I guess it's also interesting to add an alert at the top of the component to cover when API throws any error (I put it in the console but I forgot to add it on this one).

lucasponce:
Marked as DNM to complete errors handling and sorted addon.

aljesusg:
In addition @lucasponce  maybe you can add another select for sort with namespace or service option. What do you think?

lucasponce:
> In addition @lucasponce maybe you can add another select for sort with namespace or service option. What do you think?

Yeah, I'm going to play with the sort addon and add the easy ones.
By default is ordered by namespace first and servicename second, but without options to choose.
Storing the state sounds feasable, I'll update the PR with that.

lucasponce:
```coveralls``` is showing info not really useful here and creating some noise.
For example, on this PR, I've just fixed a failing navigation tests and the coverage has been reduced.
So, that's useful in information terms but not to mark a check as unsuccessful.

**Comment by lucasponce** [comment]
Errors will be propagated with an Alert after the h2 title as it is shown
image

**Comment by abonas** [comment]
Is the above error example just to show how it looks or this is also the real error's content? (I have comments on its content but I want to make sure if it is relevant or not)

**Comment by lucasponce** [comment]
> Is the above error example just to show how it looks or this is also the real error's content? (I have comments on its content but I want to make sure if it is relevant or not)

It is the real error content.

**Comment by abonas** [comment]
> It is the real error content.

I don't entirely understand what is the actual reason for the error, as it says that it failed to fetch namespaces, but it also says "404" which is typically "not found" - not found what?
In any case, the word "error" appears 3 times in the sentence and that's not very readable.
Also I would not put the "404" in UI. The users might not know what that is or how to fix it.

**Comment by lucasponce** [comment]
Final implementation for this PR: filtering, sorting, pagination and errors handling.
image
image
image

**Comment by lucasponce** [comment]
@aljesusg @mtho11 coveralls is not happy as I reduced the test coverage when fixing a navigation fix, but travis passes.
So, in order to not block other pending work I would suggest to consolidate.

**Comment by abonas** [comment]
minor comment - the service icon is misplaced inside its circle, it is leaning towards the right part. it should be repositioned/resized

File: src/components/ServiceFilter/ServiceFilter.tsx [hunk 1]
```diff
@@ -0,0 +1,235 @@
...
+} from '../../types/ServiceFilter';
+import * as API from '../../services/Api';
+import { Namespace } from '../../types/Namespace';
+
+const serviceNameFilter: FilterType = {
+  id: 'servicename',
+  title: 'Service Name',
+  placeholder: 'Filter by Service Name',
+  filterType: 'text',
+  filterValues: []
```
jotak: can we type filterValues or not? Array of strings?

src/pages/ServiceList/ServiceListComponent.tsx
```diff
@@ -1,67 +1,267 @@
...
+        .catch(namespacesError => {
+          console.error(JSON.stringify(namespacesError));
+          this.handleError('Error fetching namespace list.');
+        });
+    } else {
+      this.fetchServices(namespacesSelected, servicenameFilters);
+    }
+  }
+
+  fetchServices(namespaces: string[], servicenameFilters: string[]) {
```
jotak: Or just: `const promises = namespaces.map(ns => API.GetServices(ns))`
(no blocker)

By the way: I've seen that API functions are Upper camel case and not lower, I started to change it in another PR, it's weird to see a mix of up and low camel case.

src/pages/ServiceList/ServiceListComponent.tsx
```diff
@@ -1,67 +1,267 @@
...
+  constructor(props: ServiceListComponentProps) {
+    super(props);
+    this.filterChange = this.filterChange.bind(this);
+    this.handleError = this.handleError.bind(this);
+    this.pageSet = this.pageSet.bind(this);
+    this.pageSelect = this.pageSelect.bind(this);
+    this.updateSortField = this.updateSortField.bind(this);
+    this.updateSortDirection = this.updateSortDirection.bind(this);
+    this.state = {
+      loading: true,
```
jotak: In my experience it's better to set the `loading: true ` flag not in constructor, but at the beginning of the function that actually loads stuff. For instance here, loading is triggered both via `componentWillMount` and `filterChange`. I bet the latter won't show the spinner because of that.

src/components/ServiceFilter/ServiceFilter.tsx
lucasponce: yes, this is part of the FilterType class, which might get two types of filtering: values entered in the input text or a select with values.
The FilterType class doesn't store the values entered by the user, but potential values in case filterType is 'select'.

src/pages/ServiceList/ServiceListPage.tsx
```diff
@@ -14,24 +14,38 @@ type ServiceListProps = {
 class ServiceListPage extends React.Component<ServiceListProps, ServiceListState> {
   constructor(props: ServiceListProps) {
     super(props);
-
     console.log('Starting ServiceListPage');
+
+    this.dismissAlert = this.dismissAlert.bind(this);
```
jotak: +1, but thinking out loud, I wonder if this part could be generalized at higher level (like in App.tsx) so that we don't have to repeat alert management everywhere.

lucasponce: Probably, perhaps not for this iteration, but yes, a shared alert manager would be interesting.
This one is quite basic, just to not throw in the console the errors.

src/pages/ServiceList/ServiceListComponent.tsx
lucasponce: > By the way: I've seen that API functions are Upper camel case and not lower, I started to change it in another PR, it's weird to see a mix of up and low camel case.

I thought it was some convention, I'm not against that, but as we are time constraint I would focus in consolidate before to introduce refactoring that are just code style.

src/pages/ServiceList/ServiceListComponent.tsx
lucasponce: Mmm locally is working, but I will take a look.

lucasponce: Ok, true, fix it.

lucasponce: I updated in componentWillMount and filterChange method as well.
Just leave here a default value in the constructor.

lucasponce:
@jotak @aljesusg @mtho11 anything else for this PR ?
To start all the merge juggling and focus on the pending items for the Sprint 1 ?

PR #55, opened 2018-02-27
SWS-149: Visualize the direction of the flow/traffic

israel-hdez:
With these small changes, the graph will now show arrows:
image

### DIFF
--- a/src/pages/ServiceGraph/CytoscapeConfig.ts
+++ b/src/pages/ServiceGraph/CytoscapeConfig.ts
@@ -33,9 +33,10 @@ export class CytoscapeConfig {
         css: {
           width: 2,
           content: 'data(text)',
-          'target-arrow-shape': 'triangle',
+          'target-arrow-shape': 'vee',
           'line-color': 'data(color)',
-          'target-arrow-color': 'black'
+          'target-arrow-color': 'black',
+          'curve-style': 'bezier'
         }
       },
       {


src/pages/ServiceGraph/CytoscapeConfig.ts
```diff
@@ -33,9 +33,10 @@ export class CytoscapeConfig {
         css: {
           width: 2,
           content: 'data(text)',
-          'target-arrow-shape': 'triangle',
```
mtho11: Strange so they changed the symbol name for target-arrow-shape somewhere along the way ;

israel-hdez: Hmmm not really...
Check this image: https://issues.jboss.org/secure/attachment/12431890/screenshot-1.png
I used ´vee´ because I like it a little more than ´triangle´. But I can change it back to triangle if you think it's better.

mtho11: Nope. It looks good!

pilhuhn:
Just as a note: if we'd use 'breadthfirst' layout, it would have a 'directed' attribute that does exactly this.

abonas:
@pilhuhn isn't your PR adding the breadthfirst layout? in that case it would make this PR redundant? if so, can you please link those 2?

pilhuhn:
@abonas @israel-hdez wrote in the Jira ( https://issues.jboss.org/browse/SWS-149?focusedCommentId=13539172&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13539172 ) that the arrows are independent of the layout.

PR #58, opened 2018-02-28
[SWSQE-44] New tests for Services components

vnugent:
- Mock Api service to avoid calling axios
- Add tests for `ServiceMetrics` and `ServiceInfo`
- Move `ServiceMetrics.fetchData()` from constructor to `componentDidMount()`

### DIFF
--- a/src/pages/ServiceDetails/ServiceMetrics.tsx
+++ b/src/pages/ServiceDetails/ServiceMetrics.tsx
@@ -31,17 +31,30 @@ class ServiceMetrics extends React.Component<ServiceId, ServiceMetricsState> {
     this.state = {
       range: '5m',
       loading: false,
-      delayedLoading: false
+      delayedLoading: false,
+      requestCountIn: undefined,
+      requestCountOut: undefined,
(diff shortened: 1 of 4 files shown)


vnugent:
Please review

lucasponce:
I think these tests are ok, but perhaps we could define a "how to write jest test", I mean, we are adding UI changes very quickly, and we should be able to check if a test is making sense to be changed (i.e. if we drop a property, and test is looking for that property, I don't think test should fail, but just update to the new UI).

I mean, I don't mind to merge but I would prefer to be on the same page as @vnugent is doing a lot of job here and we want to avoid that we make changes that break tests on purpose.

I don't know, just leaving this note here to coordinate on how to sync with jest testing.

jotak:
Hey @vnugent , I'm sorry I didn't see this PR, and my last one added other tests for Metrics page.
I can add a new test inspired from yours if it makes sense.

vnugent:
@jotak this PR is quite old and some of it may have been superseded by recent updates to __mocks__/Api.ts.  I'm going to close this PR.

PR #56, opened 2018-02-28
SWS-165: Add Namespace and navigation to incorporate namespace

mtho11:

PR #59, opened 2018-02-28
SWS-164 Wire Graph to backend.

pilhuhn:
Wire the graph screen with the backend API.

Also uses breadthfirst layout so that the thingies overlap less

bildschirmfoto 2018-02-28 um 09 33 57
https://issues.jboss.org/browse/SWS-164


src/pages/ServiceGraph/ServiceGraphPage.tsx
```diff
@@ -25,7 +25,16 @@ class ServiceGraphPage extends React.Component<ServiceGraphProps, ServiceGraphSt
   }
 
   componentDidMount() {
-    this.setState({ elements: FakeData.getElements() });
+    API.GetGraphElements('istio-system', null) // TODO put real namespace
```
abonas: is this todo documented somewhere outside of code? JIRA?

pilhuhn:
On 28 Feb 2018, at 12:12, abonas wrote:

> is this todo documented somewhere outside of code? JIRA?

Yes inside SWS-164 Jira and I think I also added it to SWS-165,
but can make it more obvious.

src/services/Api.ts
```diff
@@ -31,3 +31,7 @@ const newRequest = (method: string, url: string, queryParams: any, data: any) =>
 export const GetServiceMetrics = (namespace: String, service: String, params: any) => {
   return newRequest('get', `/api/namespaces/${namespace}/services/${service}/metrics`, params, {});
 };
+
+export const GetGraphElements = (namespace: String, params: any) => {
+  return newRequest('get', `/api/namespaces/${namespace}/graphs`, params, {});
```
abonas: I have some concerns about this kind of REST api. What I mean - graphs is not a resource on a namespace. Users calling that kind of api do not receive any resource, they receive a specific representation of the entities/resources in that namespace. that should be an input parameter on a GET request on a namespace resource, not a separate api.

pilhuhn:
> calling that kind of api do not receive any resource, they receive a 
> specific representation of the entities/resources in that namespace. 
> that should be an input parameter on a GET request on a namespace 
> resource, not a separate api.

Right. With that argumentation, it should perhaps even a
media type and not a queryParam that tells the type of result.

On the other hand, a resource does not necessarily be a 'leaf'
in a tree, but can be a compound (which then probably has links
to more data). E.g. a bill, that then also links out to individual line
items (descriptions).

It is exposed like this on the server for now, so the UI has to follow 
it.

abonas:
> Right. With that argumentation, it should perhaps even a
> media type and not a queryParam that tells the type of result.
> 
> On the other hand, a resource does not necessarily be a 'leaf'
> in a tree, but can be a compound (which then probably has links
> to more data). E.g. a bill, that then also links out to individual line
> items (descriptions).
> 
> It is exposed like this on the server for now, so the UI has to follow
> it.

yes, I commented on that api on server side as well in the PRs that added it.
media type also makes sense.
having "graphs" in plural is even more confusing

mtho11:
@pilhuhn if you merge https://github.com/swift-sunshine/swsui/pull/56 you'll get namespace param

src/services/Api.ts
mwringe: Can you please open a JIRA for this so that we can keep track and discuss it there?
Its outside of the scope for this PR.

mwringe:
We are missing live updating of this data, right?

pilhuhn:
On 28 Feb 2018, at 16:52, Matthew Wringe wrote:

> We are missing live updating of this data, right?

Yes. And without this PR we miss any real data :-)

mwringe:
True, lets skip live data and get this PR in :)

jshaughn:
> yes, I commented on that api on server side as well in the PRs that added it. media type also makes sense. having "graphs" in plural is even more confusing

@abonas  Please add a jira for saturn team and we'll iterate on the graphs rest api in sprint 2.

src/pages/ServiceGraph/ServiceGraphPage.tsx
mwringe: Tracking in https://issues.jboss.org/browse/SWS-176

mwringe:
For linking the live data to a particular namespace, that is now being tracked in https://issues.jboss.org/browse/SWS-176

pilhuhn:
@mwringe I am waiting on #56 being merged and will re-do on top of this then.

abonas:
@jshaughn @mwringe https://issues.jboss.org/browse/SWS-181

PR #60, opened 2018-02-28
SWS165: Add namespace to service-graph

mtho11:


src/components/Nav/__tests__/Navigation.test.tsx
```diff
@@ -27,12 +27,12 @@ describe('Test suite Navigation', () => {
...
-  it('should render Services view', function() {
-    const wrapper = render(
+  it('should render Services view', () => {
+    const wrapper = mount(
       <MemoryRouter initialEntries={BROWSING_HISTORY} initialIndex={0}>
         <Navigation />
       </MemoryRouter>
     );
-    expect(wrapper.text()).toMatch('Services');
+    expect(wrapper.text()).toMatch('Namespace');
```
vnugent: This assertion is not very robust at the moment. `Namespace` is one of the sorting options.

lucasponce: This deserves a discussion.
Sorting options might not be available if there is a problem in the backend.
I've added an errors handlers and I guess now it should render, but this test gave some problems.

Perhaps it is just to sync on how to write the tests and check that works with connectivity and without connectivity as we have in place end-to-end flows with the REST endpoints.

I guess if we want to add coverage we will need to mock those calls, as well.

vnugent: For now we can just add back the matching string to 'Service'.   
```
expect(wrapper.text()).toMatch('Services');
```
IMO from UXD perspective we want to show filter/sorting options regardless of backend data to help users 'learn' about the UI.

PR #69, opened 2018-03-01
SWS-187: Fix unused navigation entry

lucasponce:
- Select Graph as default one

As Home page is removed, Overview entry in navigation should be removed as well, it was creating some errors.
Now default page is Graph.

image

### DIFF
--- a/src/components/Nav/Navigation.tsx
+++ b/src/components/Nav/Navigation.tsx
@@ -7,8 +7,6 @@ import ServiceDetailsPage from '../../pages/ServiceDetails/ServiceDetailsPage';
 import ServiceGraphPage from '../../pages/ServiceGraph/ServiceGraphPage';
 import ServiceListPage from '../../pages/ServiceList/ServiceListPage';
 
-const homePath = '/';
-const homeTitle = 'Overview';
 const serviceGraphPath = '/service-graph/istio-system';
 const serviceGraphTitle = 'Graph';
 const servicesPath = '/services';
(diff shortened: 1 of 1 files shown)


lucasponce:
Probably I need to adjust the jest tests.

lucasponce:
Jest tests are failing as there is some missing work for testing the Graph page when it doesn't have backend connectivity.

So, I would suggest to ignore jest tests for now as there are several things to discuss properly:
- How to mock the backend calls in jest environment.
- Adjust correctly the UI tests per component/page and how to test them.

I guess it's a task that can be defined separately with its proper JIRA tracking.

jotak:
@lucasponce do you know how did it work in a first place?

jotak:
Oh, maybe tests are rendering the homepage, which was a static thing before and that why it was not failing

pilhuhn:
Could we instead of removing the homepage just show the list of namespaces 
and the link to http://sws-istio-system.192.168.64.9.nip.io/console/service-graph/{namespace} ?

lucasponce:
I guess it was an agreement to remove the Homepage and @mtho11 prepared a PR for that, which was reviewd and approved.
This one is to fix a missing piece in that PR.
If we cant to review it and add a new page it's fine with me, but then the scope of this task is different.

lucasponce:
> Oh, maybe tests are rendering the homepage, which was a static thing before and that why it was not failing

Yes, it's related to the rendering page and make pages to work without backend, so, with mocked REST calls.
Which is good and fine, just commenting that has a litte bit more of work and coordination and I guess we should be defined this, I put a comment in a related PR about that.

jotak:
@pilhuhn I'm merging since it fixes a bug we currently have, but we all agree we'll have to setup a real homepage then.

PR #71, opened 2018-03-01
SWS-192 Show matching details in ServiceDetails page

lucasponce:
- Minor tweaks on style

image

Basically this is a ephemeral UI changes for demo:
- Matching is now show in the rules, as it might have a complex type inside, I decided to put it as a parsed json source, using list indented wasn't looking good. We can revisit this on Sprint 2.
- Also trying to reduce space, Endpoints was showing redundant ports, so we could unify it.
- Adding some labels to define what is what.

Goal is to show the info, UI is open to change quite soon, but at least BookInfo scenarios should be covered.

### DIFF
--- a/src/pages/ServiceDetails/ServiceInfo/ServiceInfo.tsx
+++ b/src/pages/ServiceDetails/ServiceInfo/ServiceInfo.tsx
@@ -95,7 +95,7 @@ class ServiceInfo extends React.Component<ServiceId, ServiceInfoState> {
                 title={this.state.name}
                 items={
                   <Row>
-                    <Col xs={12} sm={6} md={6} lg={6}>
+                    <Col xs={12} sm={6} md={4} lg={4}>
                       <div className="progress-description">
                         <strong>Labels</strong>
                       </div>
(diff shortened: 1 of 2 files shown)


lucasponce:
@xeviknal @aljesusg @jotak just to make it for Sprint 1, it's just pure style, I know Alberto is working in a refactoring for Sprint 2, but with this we will be able to show all elements in the BookInfo demo.

src/pages/ServiceDetails/ServiceInfo/ServiceInfo.tsx
```diff
@@ -109,12 +109,14 @@ class ServiceInfo extends React.Component<ServiceId, ServiceInfoState> {
                           rightText={this.state.labels ? this.state.labels[key] : ''}
                         />
                       ))}
-                      <div className="progress-description">
-                        <strong>Type</strong> {this.state ? this.state.type : ''} <strong> Ip</strong>{' '}
-                        {this.state ? this.state.ip : ''}
+                      <div>
+                        <strong>Type</strong> {this.state ? this.state.type : ''}
```
jotak: this.state is never supposed to be null or undefined afaik; what can be null or undefined, however, is `this.state.type`. So I would rather write `{this.state.type ? this.state.type : ''}`

src/pages/ServiceDetails/ServiceInfo/ServiceInfo.tsx
```diff
@@ -109,12 +109,14 @@ class ServiceInfo extends React.Component<ServiceId, ServiceInfoState> {
...
+                      </div>
+                      <div>
+                        <strong> Ip</strong> {this.state ? this.state.ip : ''}
```
jotak: same here

src/pages/ServiceDetails/ServiceInfo/ServiceInfo.tsx
jotak: Actually `{undefined}` or `{null}` would just write nothing, so, I guess the null check is just useless

lucasponce:
Addressed, thx @xeviknal and @jotak

PR #75, opened 2018-03-01
Skip full App render smoke test

vnugent:


mwringe:
If the tasks for fixing this and getting the tests enabled again could be linked here, then I think we should merge this PR.

lucasponce:
I agree to merge this one and fix if there is anything pending to have a travis clean.
Then (shortly) we can define how to plan these tests better at same time components are being developed.

PR #76, opened 2018-03-02
Automatically refit the service graph on resize.

cfcosta:
Automatically refits the cytoscape graph on resize, trying to keep everything on screen.

Bad video (too fast, but it should show it changing automatically when the window size changes.

### DIFF
--- a/src/components/CytoscapeLayout/CytoscapeLayout.tsx
+++ b/src/components/CytoscapeLayout/CytoscapeLayout.tsx
@@ -14,6 +14,7 @@ type CytoscapeLayoutProps = {
 
 export default class CytoscapeLayout extends React.Component<CytoscapeLayoutProps, CytoscapeLayoutState> {
   cy: any;
+  dimensions: [number, number];
 
   constructor(props: CytoscapeLayoutProps) {
     super(props);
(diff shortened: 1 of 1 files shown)


src/components/CytoscapeLayout/CytoscapeLayout.tsx
```diff
@@ -50,10 +52,15 @@ export default class CytoscapeLayout extends React.Component<CytoscapeLayoutProp
...
   cyRef(cy: any) {
     this.cy = cy;
+
     cy.on('tap', 'node', (evt: any) => {
       let node = evt.target;
       console.log('clicked on: ' + node.id());
     });
+
+    cy.on('resize', (evt: any) => {
+            this.cy.fit();
```
mtho11: spacing is off here. Prettier says confirms this.

src/components/CytoscapeLayout/CytoscapeLayout.tsx
```diff
@@ -14,6 +14,7 @@ type CytoscapeLayoutProps = {
 
 export default class CytoscapeLayout extends React.Component<CytoscapeLayoutProps, CytoscapeLayoutState> {
   cy: any;
+  dimensions: [number, number];
```
mtho11: dimensions is unused as it is declared later in a function with `let`

josejulio:
> trying to keep everything on screen

Not sure if this PR does that, but I think that as an user i would be annoyed if i lose my zoom level when resizing the window.

mwringe:
The service graph resize seems to be working for me. Do we still need this?

mwringe:
Trying out this PR, I don't think it works really as expected, there are some weird behaviours if you are zoomed in and go to resize this screen.
This needs more investigation.

rhqci:

israel-hdez:
Closing, because the graph sizing seems to be already fixed.

PR #84, opened 2018-03-02
fix output metrics link to grafana

jotak:


pilhuhn:
Sorry I f'd up in the original commit.

PR #85, opened 2018-03-02
Hotfix - grafana output link fixed

xeviknal:
There was a wrong url in grafana output section, within service details metrics page.


pilhuhn:
I think #84 fixed that already ?

PR #86, opened 2018-03-02
Some fun with Patternfly

pilhuhn:
* Health now has an icon for its state (ok, warn, unknown)
* Sparklines -- those need a real backend. At the moment I am only using the data from the in/out histogram to populate 3 data points per chart.

bildschirmfoto 2018-03-02 um 21 09 14

More Info about Sparklines (especially for React) : https://rawgit.com/patternfly/patternfly-react/gh-pages/index.html?selectedKind=Chart&selectedStory=Line%20Charts&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs

Sparklines on Patternfly in General: http://www.patternfly.org/pattern-library/data-visualization/sparkline/

### DIFF
--- a/src/pages/ServiceDetails/ServiceMetrics.tsx
+++ b/src/pages/ServiceDetails/ServiceMetrics.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import ServiceId from '../../types/ServiceId';
 import { MetricHistogram, MetricValue } from '../../types/Metrics';
 import * as API from '../../services/Api';
+import { SparklineChart } from 'patternfly-react';
 
 interface GrafanaInfo {
   url: string;
@@ -84,10 +85,13 @@ class ServiceMetrics extends React.Component<ServiceId, ServiceMetricsState> {
                 <div className="card-pf-title">
                   <span className="fa fa-heart"/>
                   Health</div>
-                <div className="card-pf-body">{this.health()}</div>
+                <div className="card-pf-body">
+                  <span className={this.healthIcon()}/><span className="">{this.health()}</span>
+                </div>
+
               </div>
             </div>
-            <div className="col-xs-4">
+            <div className="col-md-4">
               <div className="card-pf-accented card-pf-aggregate-status">
                 <h3 className="card-pf-title">
                   <span className="fa fa-bar-chart"/>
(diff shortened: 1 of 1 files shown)


src/pages/ServiceDetails/ServiceMetrics.tsx
```diff
@@ -98,10 +102,20 @@ class ServiceMetrics extends React.Component<ServiceId, ServiceMetricsState> {
                   <li>Request duration: {this.histogram(this.state.requestDurationIn)}</li>
                   <li>Response size: {this.histogram(this.state.responseSizeIn)}</li>
                 </ul>
+
+                <SparklineChart
+                  id="line-chart-2"
```
mtho11: Since this is using an 'id' I assume this will only be used in one place in the application. There can only be one id of a given name in the application. Really cuts down on reusability

src/pages/ServiceDetails/ServiceMetrics.tsx
```diff
@@ -121,6 +144,14 @@ class ServiceMetrics extends React.Component<ServiceId, ServiceMetricsState> {
     }
   }
 
+  _createRow(what: MetricHistogram|undefined) {
+    let column: any = [];
+    if (what) {
+      column.push('data1', what.average, what.median, what.percentile99);
```
mtho11: Minor nit. Better to use `map` instead of array.push for immutability issues.

src/pages/ServiceDetails/ServiceMetrics.tsx
pilhuhn: This whole PR was not meant "for production", but rather to research what could be done and how.

jotak:
Nice. Sparklines are a nice complement if we choose to display data as textual values / arrays as currently.
I will also do some tests with full line charts, and a red/green donut for the health. We can start discussions then with some various proposals.

pilhuhn:
On 5 Mar 2018, at 7:59, Joel Takvorian wrote:

> Nice. Sparklines are a nice complement if we choose to display data as 
> textual values / arrays as currently.

Those mini-sparklines can be a good way to convey trend data
as hover on the edges of the graph.

Also they are good to give a quick overview in such a details
page, where embedding Grafana would take much more screen
estate.

pilhuhn:
For clarification: this PR was never meant to go into production "as is", but rather to present what I found out about Sparklines - espcially as the main patternfly site is lacking the sparkline + react example.
It is meant for others to look at it and take it from there.

jotak:
@pilhuhn , more fun here: https://screenshots.firefox.com/6k4a2tSNV8VPmNQW/localhost

Btw didn't you have troubles with patternfly charts and react? For me, the charts won't display when something change in the page that triggers a new render - like changing the rate interval.

jotak:
Also I wonder if Health should rather go to service Info instead of Metrics.

pilhuhn:
@jotak I think I saw something that on the reload the id was already "used" - attributed to that `on*()` method I used.

Nice graphs.

Also I agree, that the health should be on the 1st page and/or even in the graph (and the overarching traffic light page)

pilhuhn:
I think this is no longer needed as sparkline charts are showing up in the UI

PR #87, opened 2018-03-03
 SWS-210 Add about box with the UI version

josejulio:
screenshot from 2018-03-02 23-58-44

screenshot from 2018-03-02 23-58-52

Source (when built with `npm run build`):
screenshot from 2018-03-03 22-32-27


tslint.json
```diff
@@ -63,7 +63,7 @@
         ],
         "quotemark": [true, "single", "jsx-double"],
         "radix": true,
-        "semicolon": [true, "always"],
+        "semicolon": [true, "always", "ignore-bound-class-methods"],
```
mtho11: Technically, this should go in a different PR. Someone may take issue with the lint rule and not your code

josejulio: I had to add that rule because the way I used some methods: https://github.com/swift-sunshine/swsui/pull/87/files#diff-7db5c52cc4566caf6c568f8b8c05773bR16
I would have to create a JIRA if a different PR, but next time i'm going to use a separate commit

PR #89, opened 2018-03-04
SWS-243 Adds back the precommit hook

josejulio:
Prettier runs on the precommit. It will fail the commit if any file is updated by it, a message is show to warn the developer.

Moved husky and prettier to devDependencies and removed lint-staged as is no longer used.

@jotak I think you removed the precommit because it staged the files automatically, this does something similar to swscore, please review.

screenshot from 2018-03-03 22-36-33

Note: There is one file that is failing the linter, fixed in PR #88

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -986,11 +986,6 @@
         "color-convert": "1.9.1"
       }
     },
-    "any-observable": {
-      "version": "0.2.0",
-      "resolved": "https://registry.npmjs.org/any-observable/-/any-observable-0.2.0.tgz",
-      "integrity": "sha1-xnhwBYADV5AJCD9UrAq6+1wz0kI="
-    },
     "anymatch": {
       "version": "1.3.2",
       "resolved": "https://registry.npmjs.org/anymatch/-/anymatch-1.3.2.tgz",
@@ -999,11 +994,6 @@
         "micromatch": "2.3.11",
         "normalize-path": "2.1.1"
       }
-    },
-    "app-root-path": {
-      "version": "2.0.1",
-      "resolved": "https://registry.npmjs.org/app-root-path/-/app-root-path-2.0.1.tgz",
-      "integrity": "sha1-zWLc+OT9WkF+/GZNLlsQZTxlG0Y="
     },
     "append-transform": {
       "version": "0.4.0",
@@ -3166,20 +3156,6 @@
         "restore-cursor": "2.0.0"
       }
     },
-    "cli-spinners": {
-      "version": "0.1.2",
-      "resolved": "https://registry.npmjs.org/cli-spinners/-/cli-spinners-0.1.2.tgz",
-      "integrity": "sha1-u3ZNiOGF+54eaiofGXcjGPYF4xw="
-    },
-    "cli-truncate": {
-      "version": "0.2.1",
-      "resolved": "https://registry.npmjs.org/cli-truncate/-/cli-truncate-0.2.1.tgz",
-      "integrity": "sha1-nxXPuwcFAFNpIWxiasfQWrkN1XQ=",
-      "requires": {
-        "slice-ansi": "0.0.4",
-        "string-width": "1.0.2"
-      }
-    },
     "cli-width": {
       "version": "2.2.0",
       "resolved": "https://registry.npmjs.org/cli-width/-/cli-width-2.2.0.tgz",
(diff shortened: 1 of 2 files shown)


mtho11:
I think we should try adding it back for this sprint and judge from there. I have seen people (including myself) forget to format even as easy as it is now.

josejulio:
> I think we should try adding it back for this sprint and judge from there. I have seen people (including myself) forget to format even as easy as it is now.

Yes, is failing most of time because someone forgot (and that happens easily) to prettier their files.
Having this precommit hook is middle ground from autocommiting files and not doing anything, plus is similar to what swscore is doing.

lucasponce:
https://docs.jboss.org/display/SWS/Project+Rules

"Commit messages must begin the JIRA name associated with the PR. If the PR is to address SWS-99 then the commit message for the PR must start with "SWS-99: " and then followed by the detailed commit message."

jotak:
@lucasponce what value would bring a JIRA ticket here? Nothing to test for QE. Nothing to track for sprints. No feature. Isn't it over-processing?

lucasponce:
> @lucasponce what value would bring a JIRA ticket here? Nothing to test for QE. Nothing to track for sprints. No feature. Isn't it over-processing?

I think that could be subjective.

I'm not arguing that, but then it should be clarified when to use JIRA codes in commits and when not, and everyone in the same page.

If not project rules are useless.

jotak:
@lucasponce I know and that's why I already brought this topic on ml, without effect unfortunately.
I we can avoid flooding our jira notifs even more than they are with trivial stuffs it would be great :)

lucasponce:
> @lucasponce I know and that's why I already brought this topic on ml, without effect unfortunately.
I we can avoid flooding our jira notifs even more than they are with trivial stuffs it would be great :)

Probably I don't disagree, but until we are in the same page, and everyone accepts same concepts as "trivial" I guess we should follow the rules (if not, again, why project rules are defined ? also, if project rules are wrong, they can be modified).

I don't think we should increase the number of JIRAs, i.e. if I miss something trivial for a related JIRA I reuse the same code, as it will help to understand the context.

vnugent:
My 2 cents on the pre-commit hook. May I propose a more proactive and preventative measure as part of the code-commit process?

0. Ctrl-Shift-I (assuming you're using Visual Studio Code on Fedora) to prettify your code
1. Squash your commits
2. `npm run lint`

mtho11:
I'm quite happy with prettier doing all the heavy duty formatting. It's just one less decision that Developers need to think about and it ensures consistency.

But I can go either way on the pre-commit hook.

josejulio:
> Travis failed because of prettier... how ironic :-)

Actually, that's the point of this PR. Is getting frequent that a developer forgets to run prettier on their code. The error shows but reviewers merge anyway.
The intention of this precommit hook is to make aware the developers that they forgot to do this (in case of an error).
This always run prettier before commiting and if there is an error, warn the developer (and fail the commit). This behavior of course can be skipped. Instructions are print on the screen on how do to Skip the check.

The specifics of the fails are related to #88, will rebase to have travis green. Unless new unpretty code is in master :)

vnugent:
I think prettier and tslint are 2 different tools, no? 
 `prettier`: formats code to meet *most* of tslint rules. 

Ultimately it's `tslint`, the gatekeeper that fails the build.  IMO throwing everything at travis is not the way to go.

josejulio:
> Ultimately it's tslint, the gatekeeper that fails the build. IMO throwing everything at travis is not the way to go.

Actually, I don't think (i might be wrong) that we are even running `tslint` on travis. I didn't even noticed it was a dependency.

I would say that using either one is OK, unless they both complement each other. e.g. some rules not being tested by tslint but by prettier.

At any point, i truly believe that having `prettier` run on the developer side is good, one thing less to worry about (running prettier before commiting) and ensuring that we only commit formatted code. 
On the current PR, it won't auto-commit for you, we could test this way and if there is no issues, we could move to auto-commit. 
Thoughts?

vnugent:
@josejulio `tslint` should be automatically run as a part of `npm build` or `npm start`

josejulio:
@lucasponce @jotak 
Created SWS-243 and added the number to the commit and PR title.

josejulio:
> @josejulio tslint should be automatically run as a part of npm run build or npm start. Check out the last 2 lines.

You are right on that, though we had to do an explicit call.

mtho11:
Ok, merging as _I think_ we have agreement on this. Although, we can also change this at any time if we feel this is not the **right solution**.

mtho11:
This one is under _trial assessment_ and subject to change.

PR #90, opened 2018-03-05
SWS-171 More configurable grafana link

jotak:
Note also that the http call /api/grafana may return 204 if the link should not be displayed


jotak:
Needs https://github.com/swift-sunshine/swscore/pull/85 on server-side

PR #91, opened 2018-03-05
SWS-161 Let prettier infer the format by the path

josejulio:
This allows to parse json files
 Removes options on package.json, they were already defined in .prettierrc.json

Before:
screenshot from 2018-03-05 11-31-29

After:
screenshot from 2018-03-05 11-31-37

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,5 +1,4 @@
 {
   "printWidth": 120,
-  "parser": "typescript",
   "singleQuote": true
 }
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "copy-fonts": "cp -r node_modules/patternfly/dist/fonts src/",
     "copy-img": "cp -r node_modules/patternfly/dist/img src/",
     "build-css": "node-sass-chokidar src/ --output-style compressed $npm_package_sassIncludes_src $npm_package_sassIncludes_patternfly $npm_package_sassIncludes_patternflyReact $npm_package_sassIncludes_bootstrap $npm_package_sassIncludes_fontAwesome -o src/",
-    "prettier": "prettier --single-quote --parser typescript --print-width 120 --write \"src/**/*.{js,jsx,ts,tsx,json}\"",
+    "prettier": "prettier --write \"src/**/*.{js,jsx,ts,tsx,json}\"",
     "lint": "tslint --project ./tsconfig.json",
     "start": "npm run copy-fonts && npm run copy-img && npm run build-css && npm run prettier && react-scripts-ts start",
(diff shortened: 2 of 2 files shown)


.prettierrc.json
```diff
@@ -1,5 +1,4 @@
 {
   "printWidth": 120,
-  "parser": "typescript",
```
mtho11: I'm concerned about removing this!

josejulio: According to `prettier` (emphasis is mine):
> Prettier automatically infers the parser from the input file path, so **you shouldn't** have to change this setting [1]

[1] https://prettier.io/docs/en/options.html#parser

jotak: Hmm... Did you try it? It's simple: if it works when you run both tslint and prettier, I'm fine with that.
I introduced this change because I had conflicts between prettier and tslint about semi-colons: the former was adding semi-colon, the latter was removing them.
Maybe is it because of .tsx files and not .ts ?

package.json
```diff
@@ -24,7 +24,7 @@
     "copy-fonts": "cp -r node_modules/patternfly/dist/fonts src/",
     "copy-img": "cp -r node_modules/patternfly/dist/img src/",
     "build-css": "node-sass-chokidar src/ --output-style compressed $npm_package_sassIncludes_src $npm_package_sassIncludes_patternfly $npm_package_sassIncludes_patternflyReact $npm_package_sassIncludes_bootstrap $npm_package_sassIncludes_fontAwesome -o src/",
-    "prettier": "prettier --single-quote --parser typescript --print-width 120 --write \"src/**/*.{js,jsx,ts,tsx,json}\"",
```
jotak: did you test it with our codebase?

josejulio: yes

.prettierrc.json
josejulio: > I introduced this change because I had conflicts between prettier and tslint about semi-colons: the former was adding semi-colon, the latter was removing them.
Maybe is it because of .tsx files and not .ts ?

Not sure about that, `ts-lint` is configured [1] to add semicolons:
```json
"semicolon": [true, "always"],
```
and `prettier` [2] default is `true`

`npm run prettier` and `npm run build` runs fine, no file is modified and there is no trace of errors. 

[1] https://github.com/swift-sunshine/swsui/blob/master/tslint.json#L66
[2] https://prettier.io/docs/en/options.html#semicolons

josejulio: Also travis is running it too without fails (no file needs to be modified) https://travis-ci.org/swift-sunshine/swsui/builds/349408486#L3781

PR #93, opened 2018-03-06
SWS 210 - Loads swscore's version

josejulio:
When is loading:
screenshot from 2018-03-06 17-24-26

When is loaded:
screenshot from 2018-03-06 17-24-47

### DIFF
--- a/src/components/About/AboutUIModal.tsx
+++ b/src/components/About/AboutUIModal.tsx
@@ -1,7 +1,6 @@
 import * as React from 'react';
 import { AboutModal, Spinner } from 'patternfly-react';
 import * as API from '../../services/Api';
-import update from 'immutability-helper';
 
 const pfLogo = require('../../img/logo-alt.svg');
 const SWS_CORE_COMMIT_HASH = 'SWS core commit hash';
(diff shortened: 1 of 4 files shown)


src/components/About/AboutUIModal.tsx
```diff
@@ -1,24 +1,71 @@
...
-    this.setState({ showModal: true });
+    this.setState(state => {
+      if (!state.loadingVersions) {
+        getStatus().then(
+          status => {
+            this.setState(innerState => {
+              return update(innerState, {
+                loadingVersions: { $set: false },
+                versions: { $set: status }
+              });
```
jotak: what's the difference between this and:

```
           this.setState({
               loadingVersions: false,
               versions: status
             });
```
?

In other words I have the impression that react already uses an immutable state, no?

josejulio: I'm telling:
Set loadingVersions to false and versions to status, but don't touch anything else.
e.g. if showModal = true, it will stay as true, if is false, it will stay as false.

I could of course just use 
```javascript
 this.setState({
               loadingVersions: false,
               versions: status,
               showModal: innerState.showModal
             });
```
But this will start getting complicated if i have to keep track of everything in the state for a single update. 
e.g. If i had a new attribute, I would have to track every `setState` and update accordingly. Type safety helps, but if an attribute is optional, it won't warn if i accidentally remove it.

jotak: I don't think react's `setState` is a dumb setter, I think it already merges the input object with the existing one so you don't have to forward all the unmodified properties.

... or else the code I wrote til now is completely bugged and just work by chance? :-)

josejulio: I though of it as a dumb setter :). 

I'll check more about it and report back, i got this usage from https://github.com/kolodny/immutability-helper

josejulio: `This performs a shallow merge of stateChange into the new state`

I'll update my code.

josejulio:
@jotak removed the `update` calls and tested.

PR #94, opened 2018-03-06
DivlessWrapper allows us to remove the unneeded 'div' wrapper

mtho11:
These High Order Components are very powerful for templates and such, as they can wrap (or decorate) their children. This is a simple one but good for starters and highly useful.
I didn't globally replace all the top level divs; I figured the individual authors can do this. More of these coming...

I also didn't create a jira for this since it is just a refactoring.

### DIFF
--- a/src/components/Nav/Navigation.tsx
+++ b/src/components/Nav/Navigation.tsx
@@ -7,6 +7,7 @@ import HelpDropdown from './HelpDropdown';
 import ServiceDetailsPage from '../../pages/ServiceDetails/ServiceDetailsPage';
 import ServiceGraphPage from '../../pages/ServiceGraph/ServiceGraphPage';
 import ServiceListPage from '../../pages/ServiceList/ServiceListPage';
+import DivlessWrapper from '../DivlessWrapper/DivlessWrapper';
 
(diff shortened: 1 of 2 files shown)


src/components/Nav/Navigation.tsx
```diff
@@ -36,7 +37,7 @@ class Navigation extends React.Component {
 
   render() {
     return (
-      <div>
+      <DivlessWrapper>
```
josejulio: Won’t <> and </> do the same?

josejulio: https://reactjs.org/blog/2017/11/28/react-v16.2.0-fragment-support.html

mtho11: Hmm. I didn't know about <> and </>

jshaughn: fwiw, <DivlessWrapper> looks less like a mistake than </>

josejulio: Seems that `<> stuff</> ` is just a shorter version of  `<React.Fragment>stuff<React.Fragment/>`

https://reactjs.org/blog/2017/11/28/react-v16.2.0-fragment-support.html#jsx-fragment-syntax

PR #95, opened 2018-03-07
SWS-260 Add tests and mock for service API

aljesusg:
Jira: [SWS-260](https://issues.jboss.org/browse/SWS-260)

Add tests and mocks for Service API. 

With this we can mock the API calls in the tests.
Example test of ServiceDescription, thanks for @vnugent, great masterclass 

```js
import { shallow } from 'enzyme';
import { ServiceInfoDescription, ServiceInfoCard } from '../index';

jest.mock('../../../../services/Api');

const API = require('../../../../services/Api');

describe('#ServiceInfoDescription render correctly with data', () => {
  it('should load service detail data', () => {
    return API.GetServiceDetail('istio-system', 'reviews').then(data => {
      const wrapper = shallow(
        <ServiceInfoDescription
          name={data.name}
          labels={data.labels}
          type={data.type}
          ip={data.ip}
          endpoints={data.endpoints}
        />
      );
      expect(wrapper).toBeDefined();
      expect(wrapper.find(ServiceInfoCard).props().iconType).toEqual('pf');
      expect(wrapper.find(ServiceInfoCard).props().iconName).toEqual('service');
      expect(wrapper.find(ServiceInfoCard).props().title).toEqual(data.name);
    });
  });
});

```

### DIFF
--- a/src/services/__mockData__/getGrafanaInfo.json
+++ b/src/services/__mockData__/getGrafanaInfo.json
@@ -1,0 +1,4 @@
+{
+  "url": "http://172.30.139.113:3000",
+  "variablesSuffix": "svc.cluster.local"
+}
--- a/src/services/__mockData__/getGrafanaInfo.json
+++ b/src/services/__mockData__/getGrafanaInfo.json
@@ -1,4 +1,7 @@
 {
   "url": "http://172.30.139.113:3000",
-  "variablesSuffix": "svc.cluster.local"
+  "variablesSuffix": "svc.cluster.local",
+  "dashboard": "istio-dashboard",
+  "varServiceSource": "var-source",
+  "varServiceDest": "var-http_destination"
 }
--- a/src/services/__mockData__/getGraphElements.json
+++ b/src/services/__mockData__/getGraphElements.json
@@ -1,0 +1,75 @@
+{
+  "elements": {
+    "nodes": [
+      {
+        "data": {
+          "id": "n2",
+          "text": "details (v1)",
+          "service": "details.istio-system.svc.cluster.local",
+          "version": "v1",
+          "link_prom_graph":
+            "http://prometheus:9090/graph?g0.range_input=1h\u0026g0.tab=0\u0026g0.expr=istio_request_count%7Bdestination_service%3D%22details.istio-system.svc.cluster.local%22%2Cdestination_version%3D%22v1%22%7D"
+        }
+      },
+      {
+        "data": {
+          "id": "n1",
+          "text": "productpage (v1)",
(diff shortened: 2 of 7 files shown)


aljesusg:
@jotak , @lucasponce can you review this? There is only tests and mocks for service/API, the code not change and this don't impact.

lucasponce:
Please @vnugent, can you also give us your opinion here ?
Main code LGTM and if everyone agreed we can follow @aljesusg format for upcoming tests.

src/services/__mockData__/getGrafanaInfo.json
```diff
@@ -0,0 +1,4 @@
+{
+  "url": "http://172.30.139.113:3000",
+  "variablesSuffix": "svc.cluster.local"
```
jotak: this has changed via a recent PR. You should have now:
```
    "url": "http://172.30.139.113:3000",
    "variablesSuffix": "svc.cluster.local",
    "dashboard": "istio-dashboard",
    "varServiceSource": "var-source",
    "varServiceDest": "var-http_destination"
```

vnugent:
The mock API part and unit tests look good to me.  We can refactor them later. Some suggestions to streamline the test:
- we may want to define expected data as first class Javascript object instead of plain json file and having to load them via file system call.  

```
import {ServiceInfoDescription} from './ServiceInfoDescription';
import {service_detail_data} from './ServiceDetailData;

const expected_data = service_detail_data;

jest.mock('../../../../services/Api');

it('verifies ServiceInfo lifecycle', () => {
     // API.GetServiceDetail(...) will be invoked behind the scene
    // in componentDidMount()
      const wrapper = shallow(
        <ServiceInfo namespace='isitosystem' service='review/>
     );

     const svcInfoDesc = wrapper.find(ServiceInfoDescription);
     expect(svcInfoDesc.props().port).toMatch(expected_data.ports.port);
});
```

- since ServiceInfoDescription is a pure component and renders a lot of html, we can either spot check its output against expected data (as plain old javascript objectd), or use the [snapshot feature](https://facebook.github.io/jest/docs/en/snapshot-testing.html) to diff the output.  There's no API call involved in the component so IMO I wouldn't want to introduce the mock API in the test.

aljesusg:
@vnugent Should we use snapshoot for this? We need to make some changes or add some library to it I think. Thanks

vnugent:
We can do snapshot test later

aljesusg:
@jotak I updated the mockData. @vnugent ok I'll wait until this to continue working in the serviceInfo tests, anyway I'll need this for the serviceInfo component that make a call to the API. Thanks

PR #96, opened 2018-03-07
SWS-261 match backend changes for metrics model

jotak:
Keep the existing layout and display

PR related to https://github.com/swift-sunshine/swscore/pull/94


jotak:
This PR doesn't change anything yet on visible features but prefigures future changes to display charts.

PR #98, opened 2018-03-07
SWS-265: add in loading page for the service graph

mwringe:
Adds in a loading page for when we are fetching data from the backend and we don't have a response yet.

An easy way to test is to shutdown your p8s instance as it will take a while for the backend to respond

[there are other issues created to handle the error page and empty page]

### DIFF
--- a/src/components/CytoscapeLayout/CytoscapeLayout.tsx
+++ b/src/components/CytoscapeLayout/CytoscapeLayout.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import { ReactCytoscape } from 'react-cytoscape';
+import { Spinner } from 'patternfly-react';
 import PropTypes from 'prop-types';
 
 import * as API from '../../services/Api';
@@ -8,6 +9,7 @@ import { refreshSettings } from '../../model/RefreshSettings';
 
 type CytoscapeLayoutState = {
   elements?: any;
+  isLoading: boolean;
 };
 
 type CytoscapeLayoutProps = {
@@ -30,7 +32,8 @@ export default class CytoscapeLayout extends React.Component<CytoscapeLayoutProp
     console.log('Starting ServiceGraphPage for namespace ' + this.props.namespace);
 
     this.state = {
-      elements: {}
+      elements: [],
+      isLoading: false
     };
     this.updateGraphElements = this.updateGraphElements.bind(this);
   }
@@ -129,30 +132,36 @@ export default class CytoscapeLayout extends React.Component<CytoscapeLayoutProp
   render() {
     return (
       <div id="cytoscape-container" style={{ marginRight: '25em' }}>
-        <ReactCytoscape
-          containerID="cy"
-          cyRef={cy => {
-            this.cyRef(cy);
-          }}
-          elements={this.state.elements}
-          style={GraphStyles.styles()}
-          cytoscapeOptions={GraphStyles.options()}
-          layout={this.props.layout}
-        />
+        <Spinner loading={this.state.isLoading}>
+          <ReactCytoscape
+            containerID="cy"
+            cyRef={cy => {
+              this.cyRef(cy);
+            }}
+            elements={this.state.elements}
+            style={GraphStyles.styles()}
+            cytoscapeOptions={GraphStyles.options()}
+            layout={this.props.layout}
+          />
+        </Spinner>
       </div>
     );
   }
 
   updateGraphElements(props: any) {
     let params = { interval: props.interval };
+    this.setState({ isLoading: true });
     API.GetGraphElements(props.namespace, params)
       .then(response => {
         const elements =
           response['data'] && response['data'].elements ? response['data'].elements : { nodes: [], edges: [] };
-        this.setState({ elements: elements });
+        this.setState({ elements: elements, isLoading: false });
       })
       .catch(error => {
-        this.setState({});
+        this.setState({
+          elements: [],
+          isLoading: false
+        });
         console.error(error);
       });
   }


src/components/CytoscapeLayout/CytoscapeLayout.tsx
```diff
@@ -67,25 +74,33 @@ export default class CytoscapeLayout extends React.Component<CytoscapeLayoutProp
...
-            maximalAdjustments: 2,
-            spacingFactor: 1
-          }}
-        />
-      </div>
-    );
+    if (this.state.isLoading === true) {
+      return (
+        <div id="cytoscape-container" style={{ marginRight: '25em' }}>
+          <Spinner loading={true} />
```
josejulio: IIRC you can put the ReactCytoscape inside the Spinner and it will hide it when loading=true. With that you can remove the if/else

mwringe: Oh, I didn't know I could do that. That makes things easier, I will update

jotak: Hmm ... I just ran some tests with this `<Spinner loading={this.state.isLoading}>` and embedded jsx, but it is actually less efficient since all the child JSX is still interpreted, functions calls executed etc. Whereas it doesn't if you check with `if (this.state.isLoading)`.

josejulio: Im curious on how you ran your tests. I did some quick testing and not everything seems to be executed.

e.g.
This doesn't print "hello world" in the console, nor is the DOM element <div>hello world</div> in the tree (not even with hidden or something like that).

```json
<Spinner style={{ marginTop: '15px' }} loading={true} size="lg" inverse={true}>
           <div ref={am => { console.log('hello world'); }}>hello world</div>
          </Spinner>
```

Looking at the code [1] if loading == true, it won't return the children (although i'm not an expert in react, so just guessing).

[1] https://github.com/patternfly/patternfly-react/blob/master/src/components/Spinner/Spinner.js#L20

jotak: @josejulio I did write something like `{console.log(...)}` in children, not as an anonymous function like you do.
So you're right that it's not hidden, the children dom is really not interpreted. But embedded js code is executed.
Of course since js code must be as minimal as possible in a render function anyway, there shouldn't be much impact - But still it's good to keep that in mind

To summarize: @mwringe you can do as you want :)

mwringe:
Ok, I updated this PR with the suggestion to put ReactCytoscape inside the Spinner

mwringe:
What needs to be done to get this merged?

mtho11:
@mwringe This needs rebasing

src/components/CytoscapeLayout/CytoscapeLayout.tsx
```diff
@@ -1,11 +1,13 @@
 import * as React from 'react';
 import { ReactCytoscape } from 'react-cytoscape';
+import { Spinner } from 'patternfly-react';
 import { CytoscapeConfig } from './CytoscapeConfig';
 import * as API from '../../services/Api';
 import PropTypes from 'prop-types';
 
 type CytoscapeLayoutState = {
   elements?: any;
+  isLoading?: boolean;
```
josejulio: Wondering, if we always have `isLoading` (and is being used) why we have it as optional? (same could be said from elements, but not touched in this PR).

mtho11: I agree with @josejulio probably not optional and defaults to false;

src/components/CytoscapeLayout/CytoscapeLayout.tsx
```diff
@@ -26,7 +28,8 @@ export default class CytoscapeLayout extends React.Component<CytoscapeLayoutProp
     console.log('Starting ServiceGraphPage for namespace ' + this.props.namespace);
 
     this.state = {
-      elements: []
+      elements: [],
+      isLoading: true
```
mtho11: Defaults to true? Shouldn't it only be set to true when we make the request?
Otherwise, the app first starts up it can have the empty state pattern.

mwringe: I don't understand. The first thing the app is going to do it to fetch the information from the server, so the default state when the page is first loaded is that its fetching data from the backend.

We could set it to be 'isLoading' when the server goes to do the fetch the first time. But then what should be displayed in this area for the split second before we try and fetch something from the server?

> Otherwise, the app first starts up it can have the empty state pattern.
I don't understand what you mean here. We first need to fetch the data from the backend, and then if its empty we show the empty state pattern. We shouldn't show the empty state pattern until we know if the data is actually empty or not.

src/components/CytoscapeLayout/CytoscapeLayout.tsx
mwringe: Why false?

src/components/CytoscapeLayout/CytoscapeLayout.tsx
josejulio: I think we should default `isLoading` to `false` and have `updateGraphElements` set `isLoading` to `true`.

Because `updateGraphElements` is also called when a property (the namespace) is changed, and I suppose it would be good to show the loading while loading more data.

Having it this way, we won't see an empty state because we are using `ComponentDidMount` to do the first `updateGraphElements` call (emphasis mine).

> Calling setState() in this method will trigger an extra rendering, but it will happen before the browser updates the screen. This guarantees that even though the render() will be called twice in this case, the **user won’t see the intermediate state**.



[1] https://reactjs.org/docs/react-component.html#componentdidmount

mtho11: This is the lifecycle:
1) It draws the component and creates the DOM first in `render`
2) Renders *all* child components
3) Only then does it do a `componentDidMount` and fetch

So there can be a longer than expected time between an initial state in the constructor and the actual fetch completing.

mtho11: I don't think we want to do the isLoading until we start the fetch.

mwringe:
@mtho11 Ok, its been updated to include the recommended changes. Thanks.

josejulio:
LGTM.
Loading spinner is too close to the horizontal top line, but i suppose we can give some styling as a follow up.

loading

PR #99, opened 2018-03-08
[SWS-256] Restore main app test

vnugent:
- Re-enable full App render smoke test
- Mock Axios module to return http 200 for all calls
- Move `axios-mock-adapter` module to `devDependencies`. Not sure why it was added in regular deps.
- Add a null/undefined guard for `ServiceGraphPage.availableNamespaces`
- Temporarily mock Cytoscape graph component due to DOM/div hack in browser resize logic

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,6 @@
   },
   "dependencies": {
     "axios": "~0.17.1",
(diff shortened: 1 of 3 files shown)


vnugent:
Please review

PR #100, opened 2018-03-08
SWS-259 add tests to ServiceDetailInfo

aljesusg:
Jira [SWS-259](https://issues.jboss.org/browse/SWS-259)

reviews

@vnugent Here is the WIP of tests about ServiceDetail Info, It's working with snaphots !!! Thanks

_Changed dependencies model by Backend now the json is a Map<string,string[]>_

### DIFF
--- a/src/pages/ServiceDetails/ServiceInfo.tsx
+++ b/src/pages/ServiceDetails/ServiceInfo.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import ServiceId from '../../types/ServiceId';
 import { ServiceInfoPods, ServiceInfoRules, ServiceInfoRoutes, ServiceInfoDescription } from './ServiceInfo/';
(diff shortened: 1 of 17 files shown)


jotak:
Hey, a personal question about using snapshots. I understand it's an easy way to test. But since it doesn't enforce any specific criteria for matching that is written in test, I'm wondering how confident we can be, when the snapshot doesn't match because we changed something intentionally, in that the new snapshot if correct? (if you see what I mean)

jotak:
ok, self-answering, the snapshot is of course commited so author and reviewers can check if changes are suspicious.

xeviknal:
Perhaps it will be good to have one reviewer from QE. Perhaps @vnugent can have a look into it?

vnugent:
Tests look good. Since `ServiceInfoBadge` renders a lot of text so it's a good candidate for snapshot test. Generally we may want to wait a bit on writing snapshot tests until the UI is more hashed out.

PR #101, opened 2018-03-09
SWS-215: Side panel mock-up with fake data

israel-hdez:
It looks like this:
image

The panel is always there (even if the graph is blank).

### DIFF
--- a/src/app/__tests__/App.test.tsx
+++ b/src/app/__tests__/App.test.tsx
@@ -27,6 +27,10 @@ process.env.REACT_APP_NAME = 'swsui-test';
 process.env.REACT_APP_VERSION = '1.0.1';
 process.env.REACT_APP_GIT_HASH = '89323';
 
+// TODO: properly handle SVG and D3 in the following 2 components
+jest.mock('../../pages/ServiceDetails/ServiceInfo/ServiceInfoBadge');
+jest.mock('../../pages/ServiceGraph/SummaryPanel');
+
 it('renders full App without crashing', () => {
   const div = document.createElement('div');
   ReactDOM.render(<App />, div);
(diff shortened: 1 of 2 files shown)


mtho11:
I guess we are just waiting for CI to pass. @mwringe

mtho11:
@vnugent looks like we have the same issue with canvas as we do now with svg

vnugent:
Adding jest.mock(path to SummaryPanel) to App.test.tsx to temporarily exclude the component. Run ‘npm test’ to verify locally.

israel-hdez:
@vnugent It works with SummaryPanel.
But, then CI fails with Patternfly because AreaChart and PieChart are also using D3 internally (which, I guess, use SVG).

I tried doing:
```ts
const patternfly = require.requireActual('patternfly-react');
patternfly.PieChart = jest.fn();
patternfly.AreaChart = jest.fn();
```

but it doesn't work. It fails with the error:

`TypeError: Cannot set property PieChart of [object Object] which has only a getter`

I don't know what to do in this case.

vnugent:
I tried to push a fix to `israel-hdez` but it didn't go through: `remote: Permission to israel-hdez/swsui.git denied to vnugent.`

Isn't there a repo setting that allows other to push commits to your PR?

I'd add the following to App.test.tsx
```
// TODO: properly handle SVG and D3 in the following 2 components
jest.mock('../../pages/ServiceDetails/ServiceInfo/ServiceInfoBadge');
jest.mock('../../pages/ServiceGraph/SummaryPanel');
```

josejulio:
> Isn't there a repo setting that allows others to push commits to your PR?

You can send a PR to his branch :)

israel-hdez:
@mtho11 @mwringe Travis build: https://travis-ci.org/swift-sunshine/swsui/builds/352411052
Looks like it is green :)

PR #102, opened 2018-03-09
Restore cytoscape smoke test

josejulio:
Related to SWS-256

Cytoscape smoke test was failing because of two reasons:

1) Cytoscape uses `canvas`, but jsdom doesn't implement it.
2) Cytoscape relies on the values `paddingXXX` provided by `window.getComputedStyle`,  but `jsdom` implementation is not complete, not all values are returned with sane defaults.

Fixed 1) by adding `jest-canvas-mock`.
Fixed 2) by mocking getComputedStyle and adding default for `paddingXXX`.

@vnugent please take a look

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1226,9 +1226,10 @@
       }
     },
     "axios-mock-adapter": {
-      "version": "1.12.0",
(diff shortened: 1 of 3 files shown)


vnugent:
@josejulio Good finding!  LGTM.

PR #105, opened 2018-03-09
Slight tweak to how cola is rendered

mwringe:


mwringe:
Before:
screenshot from 2018-03-09 17-37-52

After:
screenshot from 2018-03-09 17-38-07

mwringe:
Do we want the animation? I purposely got rid of it :)

PR #107, opened 2018-03-10
[SWS-248] add auto-refresh for fetching graph data

vnugent:
- add a simple timer (setInterval) to refresh graph data at 30s interval
- introduce a global refresh settings  `src/model/RefreshSettings`
- unit test


src/model/RefreshSettings.ts
```diff
@@ -0,0 +1,7 @@
+class RefreshSettings {
+  autoRefresh: boolean = true;
+  interval: number = 30 * 1000; // hardcoded value for now
```
mwringe: Is there any reason why we don't just hardcode to a much smaller interval? I think a few seconds is all we should really wait for at this point.

[Yes, it will put more strain on the backend and p8s, but I think 30s is too long for now in our testing purposes].

vnugent: It was a coin toss.  How about 5 seconds?

PR #108, opened 2018-03-11
For service graph - cytoscape graph, use Hexagon instead of circles 

mtho11:
To improve a more sophisticated look, use Hexagon instead of just circles. Also, move node labels down to bottom right for easier readability. And move labels offset to lines so they are move easily seen.

sws_ui_and_https___react-bootstrap_github_io_components_button-group_

https://issues.jboss.org/browse/SWS-289


mwringe:
-1 on hexagons :)

josejulio:
Why a different shape?

mtho11:
IMO it looks much more professional.

mtho11:
Closing this as it looks like the interests is in circles and not hexagons.

PR #109, opened 2018-03-12
SWS-262 Show metrics as charts:

jotak:
- integrate patternfly c3-based line charts
- use patternfly styles for metrics options
- separate metrics options bar as a new component

### DIFF
--- a/src/components/MetricsOptions/MetricsOptionsBar.tsx
+++ b/src/components/MetricsOptions/MetricsOptionsBar.tsx
@@ -1,0 +1,105 @@
+import * as React from 'react';
+import { Toolbar, DropdownButton, MenuItem } from 'patternfly-react';
+import { MetricsOptions } from '../../types/MetricsOptions';
+
+interface Props {
+  onOptionsChanged: (opts: MetricsOptions) => void;
+}
+
+interface MetricsOptionsState extends MetricsOptions {}
+
+export class MetricsOptionsBar extends React.Component<Props, MetricsOptionsState> {
+  rateIntervals = [['1m', '1 minute'], ['5m', '5 minutes'], ['10m', '10 minutes'], ['30m', '30 minutes']];
+
+  steps = [
+    ['1', '1 second'],
+    ['5', '5 seconds'],
+    ['15', '15 seconds'],
+    ['30', '30 seconds'],
+    ['60', '1 minute'],
+    ['1800', '30 minutes'],
+    ['3600', '1 hour']
+  ];
+
+  durations = [
+    ['300', 'Last 5 minutes'],
+    ['600', 'Last 10 minutes'],
+    ['1800', 'Last 30 minutes'],
+    ['3600', 'Last hour'],
+    ['10800', 'Last 3 hours'],
+    ['21600', 'Last 6 hours'],
+    ['43200', 'Last 12 hours'],
+    ['86400', 'Last day']
+  ];
+
+  constructor(props: Props) {
+    super(props);
+
+    this.onRateIntervalChanged = this.onRateIntervalChanged.bind(this);
+    this.onDurationChanged = this.onDurationChanged.bind(this);
+    this.onStepChanged = this.onStepChanged.bind(this);
+
+    this.state = {
+      rateInterval: this.rateIntervals[0][0],
+      duration: this.durations[1][0],
+      step: this.steps[2][0],
+      filterLabels: new Map(),
+      byLabels: []
+    };
+  }
+
+  componentWillMount() {
+    // Init state upstream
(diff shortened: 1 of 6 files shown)


jotak:
screenshot: https://screenshots.firefox.com/M50S3ItEg0dk5ZuP/localhost

jotak:
Do not merge yet: adding tests...

lucasponce:
@jotak could you add a quick screenshot of the expected UI result ?
Just to check that locally works well.
Thx

jotak:
@lucasponce there's already one above :)

lucasponce:
> @lucasponce there's already one above :)
Ah, true, sorry, I missed as normally I expected embedded in the comments.

src/components/MetricsOptions/MetricsOptionsBar.tsx
```diff
@@ -0,0 +1,105 @@
...
+    this.setState({ step: key }, () => {
+      this.props.onOptionsChanged(this.state);
+    });
+  }
+
+  render() {
+    return (
+      <Toolbar>
+        <div className="form-group">
+          <DropdownButton id="duration" title="Duration" onSelect={this.onDurationChanged}>
```
lucasponce: No change requested, but Filter from patternfly didn't suit for this scenario ?

jotak: Filter is more complex, here I only want a simple dropdown, with single selection... as simple as it can be :)

lucasponce: I guess it gives you for free the MenuItem as well, from a similar json as you created at the beginning of the page.
No change intended, just sharing info :-).

src/components/MetricsOptions/__tests__/MetricsOptionsBar.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+    const wrapper = shallow(<MetricsOptionsBar onOptionsChanged={jest.fn()} />);
+    const eltDuration = wrapper.find('#duration');
+    expect(eltDuration.length).toBe(1);
+    const eltStep = wrapper.find('#step');
+    expect(eltStep.length).toBe(1);
+    const eltRateInterval = wrapper.find('#rateInterval');
+    expect(eltRateInterval.length).toBe(1);
+  });
+
+  it('changes trigger parent callback', () => {
```
lucasponce: This test is interesting.
There are not enough tests in any project.
My only comment is about the  tradeof of this tests in the UI.
Current UI is ephemeral as is going to be changed as soon as we review basic UX flows, so perhaps these tests have a very low granilarity.
If these can be reused, it's fine, but I think with a 'renders initial layout' would be enough for the stage we are.

src/pages/ServiceDetails/ServiceMetrics.tsx
```diff
@@ -160,10 +71,11 @@ class ServiceMetrics extends React.Component<ServiceId, ServiceMetricsState> {
           requestDurationOut: metrics.histograms['request_duration_out'],
           responseSizeIn: metrics.histograms['response_size_in'],
           responseSizeOut: metrics.histograms['response_size_out'],
-          health: health
+          health: metrics.health
         });
       })
       .catch(error => {
+        // TODO: show error alert
```
lucasponce: I guess alert handling won't be part of the current PR ? am I right ?
(I am fine with it, just checking before to merge).

src/pages/ServiceDetails/ServiceMetrics.tsx
```diff
@@ -211,6 +130,155 @@ class ServiceMetrics extends React.Component<ServiceId, ServiceMetricsState> {
...
+        </div>
+        <div className="row row-cards-pf">
+          <div className="col-xs-6">
+            <div className="card-pf-accented card-pf-aggregate-status">
+              <h3 className="card-pf-title">
+                <span className="fa fa-bar-chart" />
+                Input
+              </h3>
+              <div className="card-pf-body">
+                {this.renderMetric('requestCountIn', 'Request count rate', this.state.requestCountIn)}
```
lucasponce: My first thought is that renderMetric and renderHistogram perhaps might have their own component, just to reduce the complexity of this render.
But it is fine, just a comment for future refactorings (when it makes sense ;-).

pilhuhn:
The screenshot looks great. 
I wonder though if the charts are way to big, as the page requires a considerable amount of scrolling. I would also put the duration on the 1st or 2nd row. The sizes are probably less interesting.
Good job!

jotak:
Thanks @pilhuhn 
I can do several screenshots with various arrangements and bring them to public debate

PR #110, opened 2018-03-12
[WIP] Sws 221

jshaughn:
@mtho11 , let me know if you see why the graph layout isn't changing in the browser despite what seems to be a valid update of the layout.  Also, any other feedback about the PR is welcome.


vnugent:
what will the incoming data look like?  an array of graphs?

jshaughn:
> what will the incoming data look like? an array of graphs

@vnugent , the incoming data (from the rest api) is the cytoscape config for a single graph as defined by the namespace and interval variables.  The layout choice displays the same config in different ways.

jshaughn:
@mtho11 OK, the PR is updated, rebased and squashed.  It seems to be working as expected for me.  I migrated to fat arrows in places, and extracted out a couple of buttongroup components (in the fashion of your original cytoscapetoolbar).  Hopefully it's good to merge.  A follow on PR will work to parse the query params (layout and interval) and seed the initial render so we can have bookmarks.

src/components/CytoscapeLayout/CytoscapeLayout.tsx
```diff
@@ -121,12 +122,12 @@ export default class CytoscapeLayout extends React.Component<CytoscapeLayoutProp
     );
   }
 
-  updateGraphElements(newNamespace: string) {
-    API.GetGraphElements(newNamespace, null)
+  updateGraphElements(props: any) {
+    let params = { interval: props.interval };
```
mtho11: `const` instead of `let`

src/pages/ServiceGraph/ServiceGraphPage.tsx
```diff
@@ -91,32 +61,21 @@ export default class ServiceGraphPage extends React.Component<RouteComponentProp
...
-              {this.state.availableNamespaces.map(ns => (
-                <MenuItem key={ns.name} active={ns.name === this.state.graphNamespace} eventKey={ns.name}>
-                  {ns.name}
-                </MenuItem>
-              ))}
-            </DropdownButton>
-          </h2>
-          {alertsDiv}
-        </PfHeader>
+      <div className="container-fluid container-pf-nav-pf-vertical">
```
mtho11: Please use `PfContainerNavVertical` instead of `<div className="container-fluid container-pf-nav-pf-vertical">`. Trying to componentize as much of this as possible. If need to change a style we can just change it in the component and it will be reflected everywhere in the app. Also add `PfHeader`.

jshaughn:
@mtho11 Thanks, I'll work the comments into the next PR.

PR #113, opened 2018-03-13
SWS-313 Handle graph query params

jshaughn:
- add bookmarkability by honoring graph url query params
- also, fix refresh bug introduced with original time filter PR

### DIFF
--- a/src/components/GraphFilter/GraphFilter.tsx
+++ b/src/components/GraphFilter/GraphFilter.tsx
@@ -10,9 +10,9 @@ import { IntervalButtonGroup } from './IntervalButtonGroup';
 import { LayoutButtonGroup } from './LayoutButtonGroup';
 
 export namespace GraphFilters {
-  let graphInterval: string;
-  let graphLayout: any;
-  let graphNamespace: string;
+  let graphInterval: string = '30s';
+  let graphLayout: any = DagreGraph.getLayout();
+  let graphNamespace: string = 'istio-system';
 
   export const getGraphInterval = () => {
     return graphInterval;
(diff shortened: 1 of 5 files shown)


src/pages/ServiceGraph/ServiceGraphPage.tsx
```diff
@@ -27,10 +29,14 @@ export default class ServiceGraphPage extends React.Component<RouteComponentProp
     this.filterChange = this.filterChange.bind(this);
     this.handleError = this.handleError.bind(this);
 
+    const search = routeProps.location.search;
+    const params = new URLSearchParams(search);
+    let graphInterval = params.get('interval');
```
mtho11: `const` instead of `let`

mtho11:
@vnugent that is counter to what I have learned my react classes. You should not cause any side-effects in `componentWillMount` and it mainly exists for historical reasons.

vnugent:
we're saying the same thing :) In the PR data is fetched in componentWillMount()
https://github.com/jshaughn/swsui/blob/c73c2a53489f1f19bf3c1642aab07f48043d342e/src/components/GraphFilter/GraphFilter.tsx#L69

src/components/GraphFilter/GraphFilter.tsx
```diff
@@ -66,7 +66,7 @@ export class GraphFilter extends React.Component<GraphFilterProps, GraphFilterSt
     };
   }
 
-  componentWillMount() {
+  componentDidMount() {
```
mtho11: In the future should also do a `.catch()` for failures

PR #114, opened 2018-03-14
SWS 299 - Highlight connection and nodes for services

josejulio:
- Highlight related nodes when clicking a service.
 - Mouseover will only highlight related nodes of the service (if one
   is selected).
 - Moved all the highlight code out of the CytoscapeLayout component.

sws-299-2

Issues: When the graph is refreshed, we lose the current selected service.

### DIFF
--- a/src/components/CytoscapeLayout/graphs/GraphHighlighter.ts
+++ b/src/components/CytoscapeLayout/graphs/GraphHighlighter.ts
@@ -1,0 +1,126 @@
+const DIM_CLASS: string = 'mousedim';
+
+export class GraphHighlighter {
+  cy: any;
+  service: any;
+  nodeOrEdge: any;
+
+  constructor(cy: any) {
+    this.cy = cy;
+    this.cy.on('tap', this.onTapEvent);
+    this.cy.on('mouseover', 'node,edge', this.onMouseOverNodeOrEdge);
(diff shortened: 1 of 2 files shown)


josejulio:
@mtho11 I moved the highlighting logic to a separate class. I'm wondering if there is a more React-way of doing that.

src/components/CytoscapeLayout/graphs/GraphHighlighter.ts
```diff
@@ -0,0 +1,126 @@
+const DIM_CLASS: string = 'mousedim';
+
+export class GraphHighlighter {
```
mtho11: Could this not just be a function with some props passed into it?

mtho11:
I like what you did with moving the highlighting logic out of the class. It makes it less likely for merge conflicts.

mtho11:
Ideally, I guess it would be a High Order Component that decorates the CytoscapeLayout with Highlighting capabilities.
But we can take the functionality now and refactor later.

src/components/CytoscapeLayout/graphs/GraphHighlighter.ts
josejulio: By `props` you mean regular arguments to the function? or a React Component as defined [here](https://reactjs.org/docs/components-and-props.html#functional-and-class-components)?

PR #115, opened 2018-03-14
trivial - add catch

jshaughn:
@mtho11 just adding that missing catch you noticed.


mtho11:
Can you hook this into showing the error on screen with `setState` and `alertVisible`

jshaughn:
@mtho11 , updated with [I think] what you want.  By the way, it would be nicer if we added more to the errors returned from API issues.  I tested this to force a 404 and the message was not much more than, "got a 404", appending the request URL would be useful.  Currently above my pay scale but something to look at doing.

PR #116, opened 2018-03-14
SWS-277 allow grouping metrics by label

jotak:
- allow grouping by label
- add unit to charts
- some refactoring in MetricService to compute as few as possible during rendering
- replace option "step" with "ticks", to prevent having too many datapoints freezing the browser when we have low step + high duration. Step = duration / ticks

Screenshot: https://screenshots.firefox.com/vO3BZbotsIKtJA5I/localhost

### DIFF
--- a/src/components/MetricsOptions/__tests__/MetricsOptionsBar.test.tsx
+++ b/src/components/MetricsOptions/__tests__/MetricsOptionsBar.test.tsx
@@ -11,21 +11,17 @@ const lastOptionsChanged = () => {
 describe('MetricsOptionsBar', () => {
   it('renders initial layout', () => {
     const wrapper = shallow(<MetricsOptionsBar onOptionsChanged={jest.fn()} />);
-    const eltDuration = wrapper.find('#duration');
-    expect(eltDuration.length).toBe(1);
-    const eltStep = wrapper.find('#step');
-    expect(eltStep.length).toBe(1);
-    const eltRateInterval = wrapper.find('#rateInterval');
-    expect(eltRateInterval.length).toBe(1);
+    expect(wrapper).toMatchSnapshot();
   });
 
   it('changes trigger parent callback', () => {
     const wrapper = mount(<MetricsOptionsBar onOptionsChanged={optionsChanged} />);
     expect(optionsChanged).toHaveBeenCalledTimes(1);
     const opts = lastOptionsChanged();
-    expect(opts).toHaveProperty('duration', '600');
-    expect(opts).toHaveProperty('step', '15');
-    expect(opts).toHaveProperty('rateInterval', '1m');
+    // Step = duration / ticks
+    expect(opts).toHaveProperty('duration', MetricsOptionsBar.DefaultDuration);
+    expect(opts).toHaveProperty('step', MetricsOptionsBar.DefaultDuration / MetricsOptionsBar.DefaultTicks);
+    expect(opts).toHaveProperty('rateInterval', MetricsOptionsBar.DefaultRateInterval);
 
     let elt = wrapper
       .find('#duration')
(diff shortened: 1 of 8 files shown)


abonas:
Is the group by option a standard control "as is" ?
I would expect that the already selected criterias on that dropdown "group by" would be highlighted/have some mark next to them. But if that is that control's behavior, then it's not an ask from this PR.

abonas:
can you please explain a bit further on the ticks part? an example would help (I feel I'm lacking some knowledge here or we differ on terminology)

jotak:
@abonas , about group by: I think it's possible to make them appear as checkboxes with patternfly. I'm going to try that. Or another possibility could be to remove from the list the ones that have been selected.

About the ticks and step: in the prometheus API we must define the _time step_ between two datapoints. So in my previous PR I introduced that _step_ option. But it turned out that when setting a low step (like 1 second) and high duration (like 6 hours) we end up with too many datapoints (3600*6=21600 for a single timeseries), which impacts badly the UX (not to say: freezing the browser) when it was rendered on charts.
So by replacing _steps_ by _ticks_, and thus keeping ticks under control, we avoid this problem. Ticks are the number of datapoints we want to fetch per timeseries. Maybe there's a better terminology for that, I don't know.

jotak:
@abonas and to be honest, these kind of settings may not be kept in the future. Maybe we can consider that we want something simple enough to not bother the user with this kind of settings, and choose for her/him what is the right _step_ to make the chart useful looking pretty.

But for the time being I wanted to show them so that we all know it's an option that _can_ be used in the Prometheus API. Grafana shows it, for instance.

jotak:
About the group-by filter: patternfly has the concept of "checkbox filters"
But it is not (yet?) implemented in patternfly-react. The current behaviour with the group-by filter is what is demoed on the storybook: [2]
So as said above another option could be to remove active filters from the dropdown list.

[1] http://www.patternfly.org/pattern-library/forms-and-controls/filter/
[2] https://rawgit.com/patternfly/patternfly-react/gh-pages/index.html?selectedKind=Filter&selectedStory=Filter&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs

lucasponce:
UI style comment:
image

It looks that the align is broken when some filter results are selected (at least from the first screenshot).
I hit that as well in the services filter and I could solve it re-organizing the tags, perhaps worth to take a look.

src/components/MetricsOptions/__tests__/MetricsOptionsBar.test.tsx
```diff
@@ -11,44 +11,45 @@ const lastOptionsChanged = () => {
 describe('MetricsOptionsBar', () => {
   it('renders initial layout', () => {
     const wrapper = shallow(<MetricsOptionsBar onOptionsChanged={jest.fn()} />);
-    const eltDuration = wrapper.find('#duration');
-    expect(eltDuration.length).toBe(1);
-    const eltStep = wrapper.find('#step');
-    expect(eltStep.length).toBe(1);
-    const eltRateInterval = wrapper.find('#rateInterval');
-    expect(eltRateInterval.length).toBe(1);
+    expect(wrapper).toMatchSnapshot();
```
lucasponce: When do you record the snapshot ?
If this is to ensure that only intended changes in the UI are detected fine, but it is an open discussion the granularity of UI tests.

jotak: Enzyme creates a snapshot when the test is run and store it to a file. If the file exists, snapshots are compared and the test fails if it doesn't match. If the file doesn't exist (which is the case for the first run obviously), the test passes.
So, in this PR it doesn't test anything, but in future PRs it will.

It looks like a convenient way to test UI without giving headaches.

When we want to change the UI, we just delete the snapshot, run the test and it will pass. We must just make sure the diff with the old snapshot is consistent with the intended result. And reviewers will check that, too.

I think we should give it a try. It seems to be a very common pattern with testing with enzyme. @aljesusg uses it as well in his PR https://github.com/kiali/swsui/pull/100

lucasponce:
Most LGTM, a couple of comments, but no real changes requested for this PR.
The only request is to validate that the style is not broken when you have results in the filter.
That one I think deserves to fix it (or confirm is fixed).

jotak:
@lucasponce regarding style issue, do you mean this: https://screenshots.firefox.com/OlDchBlmzFd3NPt3/screenshots.firefox.com ?

I saw it but really don't know how I can fix it, it's really inside patternfly, isn't it?

lucasponce:
> I saw it but really don't know how I can fix it, it's really inside patternfly, isn't it?

I hit exactly same issue when adding the sort addons, so I added children capability:

https://github.com/kiali/swsui/blob/master/src/components/ServiceFilter/ServiceFilter.tsx#L204
https://github.com/kiali/swsui/blob/master/src/pages/ServiceList/ServiceListComponent.tsx#L248

Perhaps you can give it a try.

jotak:
Screenshot update:
screenshot

@lucasponce I took the opportunity to move out some logic about adding/removing items to a helper class.
At first I wanted to create a new react component for that, but it turned out to be difficult (it needs 2 "mount" point in the DOM for display, one for filter, one for results; which means it would have been 2 components, and I would have to share some state between those so... in the end a small helper class is simpler)

PR #117, opened 2018-03-14
SWS-322 Make service graph text more readable

mtho11:
Make service graph labels text more readable via:

- give more space between edge/node and label text
- make edge label text a bit smaller font
- make edge label text gray instead of black

cola

dagre

breadthfirst

https://issues.jboss.org/browse/SWS-322

### DIFF
--- a/src/components/CytoscapeLayout/graphs/GraphStyles.ts
+++ b/src/components/CytoscapeLayout/graphs/GraphStyles.ts
@@ -16,8 +16,9 @@ export class GraphStyles {
           'border-width': '1px',
           'border-color': '#000',
           'font-size': '10px',
-          'text-valign': 'center',
-          'text-halign': 'right'
+          'text-valign': 'bottom',
+          'text-halign': 'right',
+          'text-outline-color': '#77828C'
         }
       },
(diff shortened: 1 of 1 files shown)


jmazzitelli:
productpage(v1) still looks like it could go in a better spot. plenty of whitespace around that node, yet the graph puts it over connectors anyway.

But, I'm sure that's the graph impl and nothing we can do.

LGTM otherwise

jmazzitelli:
Thought - how hard is it to make the nodes rectangles and put the names in the rectangles?

mwringe:
> Thought - how hard is it to make the nodes rectangles and put the names in the rectangles?

I think if we can put the name inside the node it might make it more clear what label is applied to what node. Is this something we can easily test out with cytoscape?

> I'd rather not see the rates by default, but assuming I'm the minority

I don't think your are necessarily the minority. I also suspect they will go away at some point (or not be displayed by default). We should be able to get the information from the side panel. But I think we are missing something to visualise how the traffic is split up across the node (this was created a while back https://issues.jboss.org/browse/SWS-264)

PR #118, opened 2018-03-14
SWS-181: change rest endpoint to graph from graphs

mwringe:


mwringe:
@jshaughn the swsui pr for https://github.com/kiali/swscore/pull/88

Lets see if we can get these merged.

PR #121, opened 2018-03-15
[SWS-323] Refactor SummaryPanel component

vnugent:
- break up `SummaryPanel` into sub-components and move them to `components/SummaryPanel` dir
- add url hot link to service detail page
---
summary-panel


src/components/SummaryPanel/ErrorRatePieChart.tsx
```diff
@@ -0,0 +1,30 @@
+import * as React from 'react';
+
+import { PieChart } from 'patternfly-react';
+
+type ErrorRatePieChartPropType = {
+  percentError: number;
```
jmazzitelli: The data in the graph is not going to be percentage, its going to be a fraction "error rate" So the data we put in this component won't be a percentage value. Right @jshaughn ?

src/components/SummaryPanel/ErrorRatePieChart.tsx
```diff
@@ -0,0 +1,30 @@
...
+};
+
+export class ErrorRatePieChart extends React.Component<ErrorRatePieChartPropType, {}> {
+  render() {
+    const successRate: number = 100 - this.props.percentError;
```
jmazzitelli: The data in the graph is not going to be percentage, its going to be a fraction "error rate". So the data we put in this component won't be a percentage value. Right @jshaughn ?

src/components/SummaryPanel/SummaryPanel.tsx
```diff
@@ -0,0 +1,64 @@
...
+          <p>
+            <strong>Labels:</strong>
+            <br />
+            {this.renderLabels()}
+          </p>
+          <hr />
+          <div style={{ fontSize: '1.2em' }}>
+            {this.renderIncomingRpsChart()}
+            {this.renderOutgoingRpsChart()}
+            <ErrorRatePieChart percentError={10} />
```
jmazzitelli: Here's where the data won't be a percentage, it will be a rate (error per second or something like that).

src/components/SummaryPanel/ErrorRatePieChart.tsx
jshaughn: I think percentError (a percentage) is more useful to show than RateError (requestsPerSec returning failure status codes (> 2XX)).  The % can be calculated either by the caller or here given total rate and failure rate, which are supplied in the graph nodes/edges.

src/components/SummaryPanel/SummaryPanel.tsx
jmazzitelli: OK, so either we require the caller to calculate the percentError here or we add a "failureRate" and "totalRate" attributes here. I suspect we want to just leave this the way it is and expect the caller to pass in the properly calculated percentError. Yes? In which case we can accept this PR as is.??

src/components/SummaryPanel/ErrorRatePieChart.tsx
jmazzitelli: see my comment below where the percentError is set for another question.

src/components/SummaryPanel/ErrorRatePieChart.tsx
mtho11: We should probably just merge this so people aren't stepping on each other.

PR #122, opened 2018-03-15
sws-300 summary panel for graph edge

jshaughn:
set up infrastructure for changing summary panels based on what is
clicked in the main graph.

### DIFF
--- a/src/components/CytoscapeLayout/__tests__/CytoscapeLayout.test.tsx
+++ b/src/components/CytoscapeLayout/__tests__/CytoscapeLayout.test.tsx
@@ -11,11 +11,14 @@ jest.mock('../../../services/Api');
 jest.useFakeTimers();
 
 const testNamespace = 'ISTIO_SYSTEM';
+const testHandler = () => {
+  console.log('click');
+};
 
 describe('CytographLayout component test', () => {
   it('should set correct elements data', async () => {
     const wrapper = await shallow(
-      <CytoscapeLayout namespace={testNamespace} layout={ColaGraph.getLayout()} interval="30s" />
+      <CytoscapeLayout namespace={testNamespace} layout={ColaGraph.getLayout()} interval="30s" onClick={testHandler} />
     );
     wrapper.update();
     expect(wrapper.instance().state.elements.nodes).toEqual(GRAPH_DATA[testNamespace].elements.nodes);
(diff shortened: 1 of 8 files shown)


jmazzitelli:
of course, I would like travis to pass the tests, too. :-p

src/pages/ServiceGraph/SummaryPanel.tsx
```diff
@@ -0,0 +1,23 @@
...
+type SummaryPanelState = {
+  // stateless
+};
+
+type SummaryPanelProps = {
+  display: string;
+  // data: any;
+};
+
+export default class SummaryPanel extends React.Component<SummaryPanelProps, SummaryPanelState> {
```
vnugent: Instead of re-introducing older version of `SummaryPanel`  as `SummaryPanelGraph` and `SummaryPanelEdge` we could rename `SummaryPanel` to `SummaryPanelHOC` (higher-order component) and in its `render()` produce the appropriate SummaryPanel.

vnugent: if we want the test to pass

```
// App.test.tsx

// comment the following two lines
jest.mock('../../components/SummaryPanel/RpsChart');
jest.mock('../../components/SummaryPanel/ErrorRatePieChart');

// add the following
jest.mock('../../pages/ServiceGraph/SummaryPanelGraph');
jest.mock('../../pages/ServiceGraph/SummaryPanelEdge);
```

jmazzitelli:
OK, @jshaughn  and myself pair-programmed this and it looks good to me.

We now have separate side-panel graph components depending on the graph element selected. This allows us to populate the side panel with any data relevant for the selected element.

src/components/CytoscapeLayout/__tests__/CytoscapeLayout.test.tsx
```diff
@@ -11,11 +11,14 @@ jest.mock('../../../services/Api');
 jest.useFakeTimers();
 
 const testNamespace = 'ISTIO_SYSTEM';
+const testHandler = () => {
```
israel-hdez: What will be the goal of this handler? :)

israel-hdez: This is a test, so I don't understand well...

src/pages/ServiceGraph/ServiceGraphPage.tsx
```diff
@@ -23,10 +24,15 @@ export default class ServiceGraphPage extends React.Component<RouteComponentProp
     super(routeProps);
     this.state = {
       alertVisible: false,
-      alertDetails: ''
+      alertDetails: '',
+      // This will have two fields:
+      //   summaryTarget: The cytoscape element
+      //   summaryType  : one of 'graph', 'node', 'edge', 'group'
```
israel-hdez: Perhaps, this comment should be "coded" in the type.

jshaughn:
One note: we need to maybe merge the former click handling in the graphHighlighter into the new click handling.  This PR also removes the click link to the service detail, that should be in the summary panel only.

jmazzitelli:
Yes, the onTap handler in GraphHighlighter is broke anyway - if I click an edge (connector) or a root node, and then click the group box, both get hightlighted (so it looks like I have both selected at the same time).

src/components/CytoscapeLayout/__tests__/CytoscapeLayout.test.tsx
vnugent: @israel-hdez you can simulate a mouse click in the test.   `testHandler()` function allows you to assert that on-click logic gets call (as it would in the real code)

PR #128, opened 2018-03-16
Patch a DDOS Vulnerbility in react-scripts-ts 

mtho11:
DDOS patch for react-script-ts from dependency in braces package: https://snyk.io/vuln/npm:braces:20180219


jotak:
I'm not opposed to this change but the link you provides mentions vuln is for package <2.3.1, we are currently 2.13.0 so not affected?

PR #129, opened 2018-03-16
Bump all versions in package.json up (and corresponding package-lock)

mtho11:
Bump the node package versions.

I tested `npm start`, `npm run build` and `npm test`.


vnugent:
@jotak do you know why the snapshot assertion in `MetricsOptionsBar.test` are failing?

mtho11:
I'm repushing adding the ts-jest library it was complaining about.

jotak:
I ran `npm install` and `npm test` from this PR and it didn't fail for me.
Then I removed my `node_modules` and repeat the operation and it failed.

So it must be a change in how snapshots are generated with the new dependencies. I'm not sure which is the 'guilty' library, since `enzyme-adapter-react-16` and `enzyme` didn't change, maybe the change comes from caret or tilde resolution? Or maybe it's a side-effect from another dependency.

To avoid that in the future maybe we should prefer tilde over caret most of the time, or even strict version?

Anyway @mtho11 can you delete the failing snapshot file to fix it? It's `src/components/MetricsOptions/__tests__/__snapshots__/MetricsOptionsBar.test.tsx.snap`

jotak:
By the way, when running `npm install` locally it complained about the version of `node-sass-chokidar`, apparently `1.1.1` isn't valid anymore, should be bumped to `1.1.2`

mtho11:
I guess this is waiting until after sprint 2. First thing sprint 3.

mtho11:
I updated the libs to latest versions.

mtho11:
Closed in favor of https://github.com/kiali/kiali-ui/pull/190

PR #130, opened 2018-03-16
sws-160: version group side panel with live data

jmazzitelli:
partial SWS-160... shows live data except spark graph. we need to unmock that


jmazzitelli:
screenshot from 2018-03-16 16-46-25

jmazzitelli:
do not merge yet. still have to do the spark graphs

jmazzitelli:
Working with @jshaughn on this - this is going to be merged and then we are going to work on getting it into his PR so avoid wasting time on conflicts and integration. The group side panel is done except for sparklines. We'll work on that next.

PR #132, opened 2018-03-19
SWS-285 Update UI to fetch health from new health endpoint

jotak:
Depends on https://github.com/kiali/swscore/pull/118


jotak:
PS: there is no visible change expected

pilhuhn:
I like the idea of separating health from metrics

PR #134, opened 2018-03-19
[SWS-342] Remove auto-refresh logic. Add a manual refresh button

vnugent:
- removed auto-refresh logic. 
- added a manual refresh button
- adjusted test

Do not merge until I can confirm Refresh button location in relation to the graph


abonas:
since you are changing the UI and adding a new button, please add a screenshot to the PR.
I also recommend adding a message "last refresh at X" so users looking at the screen understand when it was last refreshed

vnugent:
For now Refresh button goes in the graph component, below the toolbar.  I wanted to put it on the far right of the toolbar but the way the code is structured I'll have to make changes to Filter and the parent page. 

image

vnugent:
I don't have permission to remove the label.

vnugent:
FYI enhancement is tracked in SWS-341

PR #136, opened 2018-03-19
get the sparkline graphs to display with live data

jmazzitelli:


jmazzitelli:
sparkline data is now graphing properly.

still need to get the "options" correctly passed in.

PR #137, opened 2018-03-20
SWS-302 Spark Chart

jshaughn:
Roughly working, needs some work still.

### DIFF
--- a/src/pages/ServiceGraph/SummaryPanelEdge.tsx
+++ b/src/pages/ServiceGraph/SummaryPanelEdge.tsx
@@ -3,8 +3,24 @@ import ServiceInfoBadge from '../../pages/ServiceDetails/ServiceInfo/ServiceInfo
 import { RateTable } from '../../components/SummaryPanel/RateTable';
 import { RpsChart } from '../../components/SummaryPanel/RpsChart';
 import { SummaryPanelPropType } from '../../types/Graph';
+import * as API from '../../services/Api';
+import * as M from '../../types/Metrics';
 
-export default class SummaryPanelEdge extends React.Component<SummaryPanelPropType> {
+type SummaryPanelEdgeState = {
+  loading: boolean;
+  source: string;
+  sourceService: string;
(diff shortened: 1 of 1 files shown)


src/pages/ServiceGraph/SummaryPanelEdge.tsx
```diff
@@ -13,34 +29,97 @@ export default class SummaryPanelEdge extends React.Component<SummaryPanelPropTy
...
+    };
+    API.getServiceMetrics(this.state.destNamespace, this.state.destService, options)
+      .then(response => {
+        const data: M.Metrics = response['data'];
+        const metrics: Map<String, M.MetricGroup> = data.metrics;
+        const mg: M.MetricGroup = metrics['request_count_in'];
+        const tsa: M.TimeSeries[] = mg.matrix;
+        let reqRates: number[] = [];
+        let errRates: number[] = [];
+        for (let i = 0; i < tsa.length; ++i) {
```
jmazzitelli: this is what I didn't understand when I was doing this  on the group page - why are you getting more than one timeseries into a single sparkline? That doesn't make sense - it is as if you are combining multiple timeseries into one - basically appending one timeseries to the end of the other. That doesn't seem right to me.

jshaughn: It's a complex structure coming back.  There are several queries being performed, each returning a bunch of potential time series.  First, we narrow to the query we're interested in; for the edge summary, it's the 'request_count_in'  for the specified destination service.  The returned matrix is an array of TimeSeries.  Each TS is described by the discerning labels (see 'byLabelsIn[]': 'source_service,source_version' above).  So, one TS for each unique service feeding the destination service.  For the edge, we're only interested in one specific source service,  so if you look down a little further at lines 72:73 you can see I'm tossing all but the TS I want.  Once I have that I want to run through the array of datapoint for that TS.  Note that line 78 is currently just for testing, I just wanted to generate a line of error rates at 1/2 the actual.

PR #138, opened 2018-03-20
SWS-349 SWSUI deploy job fails after GitHub organization renaming

theute:
Attempt to fix the Travis deploy job


theute:
I don't know if it fixes the problem, but I guess it won't hurt...

PR #139, opened 2018-03-20
SWS-286 Make Service Health donut as a separate component

jotak:
- Remove from Service Metrics
- Add to Service Details

Health donut

### DIFF
--- a/src/components/ServiceHealth/ServiceHealth.tsx
+++ b/src/components/ServiceHealth/ServiceHealth.tsx
@@ -1,0 +1,73 @@
+import * as React from 'react';
+import { DonutChart } from 'patternfly-react';
+import Health from '../../types/Health';
+
+interface Props {
+  health?: Health;
+  size: number;
+  thickness?: number;
+}
+
+class ServiceHealth extends React.Component<Props, {}> {
+  static DonutColors = {
+    // From Patternfly status palette
+    Healthy: '#3f9c35',
+    Failure: '#cc0000',
+    'N/A': '#707070'
+  };
+
+  donutData: any;
+  donutSize: any;
+  donutTitle: string;
+
+  constructor(props: Props) {
+    super(props);
+    this.onGetProps(props);
+  }
+
+  componentWillReceiveProps(nextProps: Props) {
+    this.onGetProps(nextProps);
+  }
+
+  onGetProps(props: Props) {
+    this.donutSize = {
+      width: props.size,
+      height: props.size
+    };
+    let columns: [string, number][] = [];
+    if (props.health && props.health.totalReplicas > 0) {
+      columns = [
+        ['N/A', 0],
+        ['Healthy', props.health.healthyReplicas],
+        ['Failure', props.health.totalReplicas - props.health.healthyReplicas]
+      ];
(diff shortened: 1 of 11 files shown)


jotak:
@lucasponce / @xeviknal / @abonas what do you think about the layout?

lucasponce:
> what do you think about the layout?

I think in terms of space, health can be smaller and be shown in the service description card.
I personally don't think it deserves a specific card for it.

Potential suggestions, would be to use a smaller graph and use the initial title, like:
image

Or similar, UX discussions can go anywhere, but I guess it would make sense to optimize space there.

jotak:
With some tweaking I can have a result like this:

health

To be honest I'm not really convinced. We can make it appear smaller, or even text-based, without putting it at an unusual place.

Would it make sense to have it in the "deployments" card? Since it's actually an information about the deployments.
We already have `Pod status: 1/1` there. We could change that to `Deployed pods: 1/1` and below `Pods health: 1/1` ?

lucasponce:
> Would it make sense to have it in the "deployments" card? Since it's actually an information about the deployments.

That sounds good to me as well.
Or just to put in in the description, right to Endpoints, as it is info for the whole service, I guess.
Also, it has space there and probably visally it's similar.

pilhuhn:
The small one in the top right corner looks odd to me.
Patternfly already has this type of card where you have a donut and some details below
http://www.patternfly.org/pattern-library/cards/utilization-trend-card/#card-with-a-single-metric
(we would not show the sparkline).
If I understand correctly, the health and the pod-status 1/1 are the same thing, so that would match.

jotak:
@pilhuhn they are not the same thing: if i understand correctly pods status comes from k8s deployment status. Health come from envoy metrics. They may not be always identical, for instance when a pod is up but badly performing it might be excluded from envoy cluster, so it's still deployed but will be marked as unhealthy.

xeviknal:
@jotak good point, I was about to ask you for that.
Then, I am up to add this donut next to endpoints column, not as a different card/box, but as a normal column.

pilhuhn:
On 21 Mar 2018, at 11:43, Joel Takvorian wrote:

> performing it might be excluded from envoy cluster, so it's still 
> deployed but will be marked as unhealthy.

Right. So in this case (if istio is not deciding to kill the entire 
pod),
we need to re-think those two classes of "health", and make the
final health dependent on both.

And then I have the impression that we need to revisit the entire notion
of "health" in a later stage when it is possible to also derive it from
different sources. E.g. if there are 10 pods for a service and 1 is slow
and there is low traffic, I'd not say the app is unhealthy or degraded

-- 
hwr@pilhuhn.de  m:0179/207 4919  b:http://pilhuhn.blogspot.com

jotak:
Proposal with smaller donut and included in main card:

donut

@pilhuhn I agree that we must define what exactly is health. For now I'm just taking the istio definition which itself is based on envoy's.

lucasponce:
> Proposal with smaller donut and included in main card:

I like this alternative, it looks more in line with the elements of the service details page.

jotak:
I rebased and push the last proposed alternative

PR #140, opened 2018-03-20
SWS-236 Support Istio Rules in the UI

lucasponce:
- Add API service for rule list and rule details.
- Add a rules list page.
- Add a rule details page.
- Add test coverage for API services.
- Refactor shared components between services/rules pages.

### DIFF
--- a/src/components/NamespaceFilter/NamespaceFilter.tsx
+++ b/src/components/NamespaceFilter/NamespaceFilter.tsx
@@ -4,21 +4,13 @@ import {
   ActiveFilter,
   FilterType,
   FilterValue,
-  ServiceFilterProps,
-  ServiceFilterState
-} from '../../types/ServiceFilter';
+  NamespaceFilterProps,
+  NamespaceFilterState
+} from '../../types/NamespaceFilter';
 import * as API from '../../services/Api';
 import { Namespace } from '../../types/Namespace';
 
-const serviceNameFilter: FilterType = {
-  id: 'servicename',
-  title: 'Service Name',
-  placeholder: 'Filter by Service Name',
-  filterType: 'text',
-  filterValues: []
-};
-
-export namespace ServiceFilterSelected {
+export namespace NamespaceFilterSelected {
   let selectedFilters: ActiveFilter[] = [];
 
   export const setSelected = (activeFilters: ActiveFilter[]) => {
@@ -30,8 +22,8 @@ export namespace ServiceFilterSelected {
   };
 }
 
-export class ServiceFilter extends React.Component<ServiceFilterProps, ServiceFilterState> {
-  constructor(props: ServiceFilterProps) {
+export class NamespaceFilter extends React.Component<NamespaceFilterProps, NamespaceFilterState> {
+  constructor(props: NamespaceFilterProps) {
     super(props);
 
     this.updateCurrentValue = this.updateCurrentValue.bind(this);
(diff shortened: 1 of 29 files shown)


lucasponce:
This PR creates a new page to list Istio Rules in the system:

image

The list page shows a preliminar handler and instances info of the rule, meanwhile the details page shows match and handler/instance details:

image

Handler and instances objects have a pretty open format, so not easy to decide how to show them in the UI, as a preliminar step, these complex types are shown as a raw json, to postpone decision about what is the best way.

Also, there are Istio Rules created by default in the system:

image

And others created by user as part of telemetry or policy rules:

image

So, the goal for this PR is to bring this functionality into the UI for Sprint 2, meanwhile we can refine in details for how to show any specific handler/instance in Sprint 3.

lucasponce:
Please @aljesusg and @xeviknal could you review this one ?

I had some conflicts as I perform some refactoring to reuse same components between services/rules pages, but I think all are solved in the PR.

Thanks.

jotak:
Based on the screenshots I see that several textareas are defined quite small (in terms of rows), needing to scroll even for small pieces of text , like here https://user-images.githubusercontent.com/1662329/37660479-d14ebef4-2c52-11e8-9255-1f2d9579a8ad.png
They could be made a little bigger?

lucasponce:
> Based on the screenshots I see that several textareas are defined quite small (in terms of rows), needing to scroll even for small pieces of text , like here https://user-images.githubusercontent.com/1662329/37660479-d14ebef4-2c52-11e8-9255-1f2d9579a8ad.png
They could be made a little bigger?

Yes, but manually, as text is completely arbitrary in terms of size, I used the default one and let user to zoom in

image

src/components/NamespaceFilter/NamespaceFilter.tsx
```diff
@@ -49,14 +41,25 @@ export class ServiceFilter extends React.Component<ServiceFilterProps, ServiceFi
...
-      filterTypeList: [serviceNameFilter, namespaceFilter],
-      activeFilters: ServiceFilterSelected.getSelected(),
+      currentFilterType: initialFilters[0],
+      filterTypeList: initialFilters,
+      activeFilters: NamespaceFilterSelected.getSelected(),
       currentValue: ''
     };
   }
 
+  initialFilterList(namespaceFilter: FilterType) {
```
jotak: nitpicking, but good to know, you can copy an array more easily with `splice`:
`return this.props.initialFilters.splice().concat(namespaceFilter)`

jotak: sorry, not splice, slice. I always mix them up :/
`return this.props.initialFilters.slice().concat(namespaceFilter)`

src/pages/IstioRuleDetails/IstioRuleInfo/IstioRuleInfoDescription.tsx
```diff
@@ -0,0 +1,36 @@
+import * as React from 'react';
+import PfInfoCard from '../../../components/Pf/PfInfoCard';
+import { Col, Row } from 'patternfly-react';
+
+interface IstioInfoDescriptionProps {
```
jotak: Looking at `RuleItem` and parent component usage, `name` and `match` should not be optional, or did I miss something?

lucasponce: Not sure if I understand the question.
This one is a mapping that comes from the istio rule details
```
      let data = response['data'];
      this.setState({
        name: data.name,
        match: data.match,
        actions: data.actions
      });
```
Meanwhile RuleItem is used in the rules list page to map the list of rules.

jotak: Yeah what I mean is why are they declared optional `name?: string` instead of just `name: string` ?

lucasponce: Ah, true, name is always there, but match can be empty, I'll fix the name one.
Thx !

lucasponce:
@jotak comments addressed.

Yeah, it has several files, but I think is just a clone of the list/details pages used for services but adapted to the rules objects, with a couple of refactoring to put in commons the shared components.

pilhuhn:
I just switched to this locally and don't see the new istio-rules menu entry

jotak:
merged, thanks @lucasponce

PR #143, opened 2018-03-20
SWS-158: provide in-out rate table 

jmazzitelli:
this adds customization settings for the pie chart component


jmazzitelli:
screenshot from 2018-03-20 16-32-45

PR #144, opened 2018-03-20
SWS-217: Display data in side panel about a selected versioned service (node) in graph

israel-hdez:
Using the group panel as a template. The RpsChart can now show the
timestamps of the data points if data is received with the rigth format.

Pulling out "toC3Columns" function to an utils object to reuse the code.

### DIFF
--- a/src/components/SummaryPanel/RpsChart.tsx
+++ b/src/components/SummaryPanel/RpsChart.tsx
@@ -4,8 +4,8 @@ import { AreaChart } from 'patternfly-react';
 
 type RpsChartTypeProp = {
   label: string;
-  dataRps: number[];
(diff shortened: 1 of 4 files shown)


rhqci:

jmazzitelli:
I see it - I like it! In fact, we should make all the users of the RpsChart do this. @israel-hdez is there a way you can retrofit this RpsChart into the other side panels and make it a requirement the data passed in matches the map of string,number[] (don't make it optional)?

screenshot from 2018-03-21 09-56-52

israel-hdez:
Following up with https://github.com/kiali/swsui/pull/155 to display timestamps in all side panels.

PR #145, opened 2018-03-21
title "Ip" looks better when it is in upper case "IP"

jkandasa:


abonas:
@jkandasa please check the test failures.
other than that +1 on the change from me

rhqci:

jotak:
The snapshot test fails which is expected for this change, @jkandasa you just just have to delete the snapshot (I think it's this one: `src/pages/ServiceDetails/ServiceInfo/__tests__/ServiceInfoDescription.test.tsxsrc/pages/ServiceDetails/ServiceInfo/__tests__/ServiceInfoDescription.test.tsx`) then run `npm test` and commit with the updated snapshot.

jkandasa:
@abonas @jotak Thank you! Updated the snapshot file.

PR #146, opened 2018-03-21
shouldComponentUpdate to help be able to select graph items.

jmazzitelli:
this really doesn't fix the core problem, but it makes it a little better. At least now you can see the blue selection (and the selections don't disappear) if you select an item twice.

We may or may not want to merge this. As I say, it doesn't really fix the problem, but at least it allows you to select an item and see the blue outline some times.

### DIFF
--- a/src/pages/ServiceGraph/ServiceGraphPage.tsx
+++ b/src/pages/ServiceGraph/ServiceGraphPage.tsx
@@ -69,6 +69,13 @@ export default class ServiceGraphPage extends React.Component<RouteComponentProp
     this.setState({ summaryData: data });
   };
 
+  shouldComponentUpdate(nextProps: any, nextState: any) {
+    if (this.state.summaryData.summaryType !== nextState.summaryData.summaryType) {
+      return true;
+    }
+    return false;
+  }
+
   render() {
     let alertsDiv = <div />;
     if (this.state.alertVisible) {


jmazzitelli:
screenshot from 2018-03-20 23-27-11

src/pages/ServiceGraph/ServiceGraphPage.tsx
```diff
@@ -69,6 +69,13 @@ export default class ServiceGraphPage extends React.Component<RouteComponentProp
     this.setState({ summaryData: data });
   };
 
+  shouldComponentUpdate(nextProps: any, nextState: any) {
+    if (this.state.summaryData.summaryType !== nextState.summaryData.summaryType) {
```
josejulio: I think that it might be better to return:
```typescript
return this.state.isLoading || nextState.isLoading;
```

We only want to update the component when we are leaving an `isLoading` or entering into a `isLoading`.
Briefly tested, but looks OK.

Maybe further refining to something like
```typescript
return this.state.isLoading !== nextState.isLoading;
```
"When leaving or entering into an isLoading state"

rhqci:

src/pages/ServiceGraph/ServiceGraphPage.tsx
josejulio: Sorry, I used this into `CytoscapeLayout.tsx`

PR #147, opened 2018-03-21
KIALI-299 Fix GraphHighlighter using ClickEvents from CytoscapeLayout

josejulio:
- Removed mouse hover events and only use clicks.

~~This PR considers we can click, tested using the suggested change in: https://github.com/kiali/swsui/pull/146#discussion_r175989257~~

~~I was trying to record a gif, but my box crashed while trying, i will try again later.~~
@jmazzitelli recorded a nice gif already, thanks!
out ogv

### DIFF
--- a/src/components/CytoscapeLayout/CytoscapeLayout.tsx
+++ b/src/components/CytoscapeLayout/CytoscapeLayout.tsx
@@ -11,12 +11,17 @@ type CytoscapeLayoutProps = {
   namespace: string;
   layout: any;
   interval: string;
-  onClick: PropTypes.func;
+  onClick: (event: CytoscapeClickEvent) => void;
 };
 
 type CytoscapeLayoutState = {
   elements?: any;
-  isLoading: boolean;
+  loading: boolean;
+};
+
+export type CytoscapeClickEvent = {
+  summaryType: string;
+  summaryTarget: any;
 };
 
 export default class CytoscapeLayout extends React.Component<CytoscapeLayoutProps, CytoscapeLayoutState> {
@@ -34,9 +39,10 @@ export default class CytoscapeLayout extends React.Component<CytoscapeLayoutProp
 
     this.state = {
       elements: [],
-      isLoading: false
+      loading: false
     };
     this.updateGraphElements = this.updateGraphElements.bind(this);
+    this.handleTap = this.handleTap.bind(this);
   }
 
   resizeWindow() {
@@ -63,6 +69,10 @@ export default class CytoscapeLayout extends React.Component<CytoscapeLayoutProp
     }
   }
 
+  shouldComponentUpdate(nextProps: any, nextState: any) {
+    return this.state.loading !== nextState.loading;
+  }
+
   cyRef(cy: any) {
     this.cy = cy;
     this.graphHighlighter = new GraphHighlighter(cy);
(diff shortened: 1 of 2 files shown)


rhqci:

jmazzitelli:
So, correct me if I'm wrong, but this now removes the feature of dimming the graph as we move the mouse over the graph. If so, I'm not sure if this is what we want, as it completely reverts these JIRAs that we implemented last sprint:

https://issues.jboss.org/browse/KIALI-218
https://issues.jboss.org/browse/KIALI-303

jmazzitelli:
adding some stakeholders to the "reviewers" list - what do you folks think? Do we keep the feature of "mouse hover dimming unrelated parts of the graph" or do we remove that and just have the feature "click an element in the graph in order to dim unrelated parts of the graph"??

I am not sure, myself. I even mentioned last week that I thought the mouse-hover dimming the graph got annoying as it seemed to make the graph "flash" as you moved the mouse around. So I am considering this PR as potentially an enhancement :)

pilhuhn:
On 21 Mar 2018, at 14:13, John Mazzitelli wrote:

> you folks think? Do we keep the feature of "mouse hover dimming 
> unrelated parts of the graph" or do we remove that and just have the 
> feature "click an element in the graph in order to dim unrelated parts 
> of the graph"??

I like the dim-on-mouseover.

The flashing on reload is a much bigger issue, as this happens
without user interaction.
When the user interacts with the UI, then the "flashing" is intended.

A question is though going forward how this works on mobile devices,
where a finger-over event does not really exist.

jmazzitelli:
OK, I just ran this locally - and it also removes the "hightlighting" feature as well (i.e. when I click an element, I get no indication what is currently selected other than looking at the dimmed elements and trying to infer that information. Before we would get a blue outline over the selected item so it was clear what was selected (and thus, it tells you the element that owns the live side panel data).

So I'm putting a DNM label on this until we figure out how we want this to work

josejulio:
It’s right, it removes the dimming when hovering and instead uses a click to focus approach. It seemed more intuitive to me with the way the group box was asked to work, also sws-303 reads:

> when I *click*/hover over a connection 

So it looks like the original intention was, “what is better between click and hover”, but we had already something on the click 

I’ll upload a gif later showing how it works

josejulio:
> OK, I just ran this locally - and it also removes the "hightlighting" feature as well

"This PR considers we can click, tested using the suggested change in: #146 (comment)"
We will need to work out that first, but i tested adding to `CytoscapeLayout.tsx`
```typescript
shouldComponentUpdate(nextProps: any, nextState: any) {
  return this.state.isLoading !== nextState.isLoading;
}
```

jmazzitelli:
Yes, that works better. @josejulio - put that in this PR. As it makes it work nicer - no mouse-over, but it does everything else (highlights on selection, dims the graph, populates side panel with live data of selected element).

jmazzitelli:
out ogv

josejulio:
> put that in this PR. As it makes it work nicer

Done!

I'll put your gif on the top, thanks!

josejulio:
> The flashing on reload is a much bigger issue, as this happens
> without user interaction.

It doesn't happen anymore after #134 (not without user interaction)

PR #148, opened 2018-03-21
Add Destination Policies to the Service Details page

lucasponce:
This requires https://github.com/kiali/swsui/pull/140 be merged first as it has dependencies introduced there.

This is a preliminar way to present DestinationPolicies objects (load balancers, circuit breakers) attached/grouped per service, in the Service Details page.

It is just a list, not open the discussion about better UX, as further discussion about how to re-org hasn't started yet, but it makes sense to bring it into the details page as a preliminar step.

image

### DIFF
--- a/src/pages/ServiceDetails/ServiceInfo/ServiceInfoDestinationPolicies.tsx
+++ b/src/pages/ServiceDetails/ServiceInfo/ServiceInfoDestinationPolicies.tsx
@@ -1,0 +1,122 @@
+import * as React from 'react';
+import { DestinationPolicy } from '../../../types/ServiceInfo';
+import PfInfoCard from '../../../components/Pf/PfInfoCard';
+import ServiceInfoBadge from './ServiceInfoBadge';
+
+interface ServiceInfoDestinationPoliciesProps {
+  destinationPolicies?: DestinationPolicy[];
+}
+
+class ServiceInfoDestinationPolicies extends React.Component<ServiceInfoDestinationPoliciesProps> {
+  constructor(props: ServiceInfoDestinationPoliciesProps) {
+    super(props);
+  }
+
+  render() {
+    return (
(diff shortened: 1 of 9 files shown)


aljesusg:
I wanna move ServiceInfoBadge to components too, There are another views with this.

src/pages/ServiceDetails/ServiceInfo/ServiceInfoDestinationPolicies.tsx
```diff
@@ -0,0 +1,122 @@
...
+            <div key={'rule' + i}>
+              <div>
+                <strong>Name</strong>
+                {': '}
+                {dPolicy.name}
+              </div>
+              {dPolicySource}
+              {destinationLabels}
+              {loadBalancing}
+              {circuitBreaker}
```
xeviknal: Would it be useful to split this render in different components? Having one per each last 2 lines above (LoadBalancing, CircuitBreaker). I think this way we let those 2 parts evolve independently of this component and making modifications easier.
I am not very familiar with React already, so maybe I am over modularizing this.

lucasponce: Probably, but I wouldn't introduce that refactor on this PR as it would depend on other factor.
Also, LoadBalancing and CircuitBreaker have not enough entity yet to split them in different components IMO.

lucasponce:
Rebased and solved some conflicts with master.

src/pages/ServiceDetails/ServiceInfo/ServiceInfoDestinationPolicies.tsx
xeviknal: I think in a not far future we should try to make those conceptual entities containers, for a better maintainability.

PR #154, opened 2018-03-21
fix behavior of loading graphs

jmazzitelli:
- add loading state where needed
- show loading message when downloading the data
- don't use larger fonts - use standard font size

### DIFF
--- a/src/pages/ServiceGraph/SummaryPanelEdge.tsx
+++ b/src/pages/ServiceGraph/SummaryPanelEdge.tsx
@@ -117,7 +117,7 @@ export default class SummaryPanelEdge extends React.Component<SummaryPanelPropTy
             rate4xx={rate4xx}
             rate5xx={rate5xx}
           />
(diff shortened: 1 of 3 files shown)


pilhuhn:
Some smaller issues: for groups, the 2nd sparkline is out of the container.
bildschirmfoto 2018-03-21 um 16 46 32

For single nodes, there is an additional ; 
bildschirmfoto 2018-03-21 um 16 46 46

And I think the 'Err' is too close to the % (?)

jmazzitelli:
I fixed the semicolon - that was just a typo. That's gone now.

As for the graphs outside the container, I do not see that. I assume it has to do with screen size/resolution. We do plan on working on better look-n-feel issues like this later.

pilhuhn:
Yes it has to do with the resolution. But it is also that the containers behave differently

israel-hdez:
@pilhuhn I think I have fixed the issue about the graphs out of the container. I'll push the fix in another PR.

PR #155, opened 2018-03-21
Homologate all side panels to show timestamps for RpsCharts

israel-hdez:
- Change RpsChart to only allow data with timestamps.
- Fix styles of the side panel to avoid overflow of the content.


israel-hdez:
@pilhuhn This should fix the overflow issue on the side panels.

PR #156, opened 2018-03-21
KIALI-362 Add pod status for each service listed

xeviknal:
Adding pod status to every service in the service list.

pod-status-service-list

There are no test for this Service List, so I couldn't update them. I am planning to add a subsequent PR adding them ( I think tests are out of KIALI-362 scope).


jotak:
@lucasponce no it's fine, I was just experimenting but I'll consolidate over this work

PR #157, opened 2018-03-21
KIALI-331 graph summary for namespace

jshaughn:
This aggregates info about all services reflected in the namespace graph
also:
- remove horizontal break in RateTable to conserve space/improve locality
- add updateTime to graph state (for future use)

NOTE: This depends on new backend support, setting to do not merge until core PR is merged.

### DIFF
--- a/src/pages/ServiceGraph/ServiceGraphPage.tsx
+++ b/src/pages/ServiceGraph/ServiceGraphPage.tsx
@@ -18,6 +18,7 @@ type ServiceGraphPageProps = {
   alertVisible: boolean;
   alertDetails: string;
   summaryData: any;
+  updateTime: string;
 };
 
 export default class ServiceGraphPage extends React.Component<RouteComponentProps<NamespaceId>, ServiceGraphPageProps> {
@@ -30,7 +31,8 @@ export default class ServiceGraphPage extends React.Component<RouteComponentProp
     this.state = {
       alertVisible: false,
       alertDetails: '',
-      summaryData: { summaryType: 'graph' }
+      summaryData: { summaryType: 'graph' },
+      updateTime: new Date().toLocaleString()
     };
 
     this.filterChange = this.filterChange.bind(this);
@@ -66,7 +68,9 @@ export default class ServiceGraphPage extends React.Component<RouteComponentProp
   }
 
   handleGraphClick = (data: any) => {
-    this.setState({ summaryData: data });
+    if (data !== undefined) {
+      this.setState({ summaryData: data, updateTime: new Date().toLocaleString() });
+    }
   };
 
   render() {
@@ -86,7 +90,11 @@ export default class ServiceGraphPage extends React.Component<RouteComponentProp
           <GraphFilter onFilterChange={this.filterChange} onError={this.handleError} />
         </PfHeader>
         <div style={{ position: 'relative' }}>
-          <SummaryPanel data={this.state.summaryData} rateInterval={GraphFilters.getGraphInterval()} />
+          <SummaryPanel
+            data={this.state.summaryData}
+            namespace={GraphFilters.getGraphNamespace()}
+            rateInterval={GraphFilters.getGraphInterval()}
+          />
           <CytoscapeLayout
             namespace={GraphFilters.getGraphNamespace()}
             layout={GraphFilters.getGraphLayout()}
--- a/src/pages/ServiceGraph/SummaryPanelGraph.tsx
+++ b/src/pages/ServiceGraph/SummaryPanelGraph.tsx
@@ -1,11 +1,26 @@
 import * as React from 'react';
-
 import ServiceInfoBadge from '../../pages/ServiceDetails/ServiceInfo/ServiceInfoBadge';
-import { ErrorRatePieChart } from '../../components/SummaryPanel/ErrorRatePieChart';
+import { RateTable } from '../../components/SummaryPanel/RateTable';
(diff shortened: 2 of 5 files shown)


jshaughn:
This can be reviewed but is blocked by backend https://github.com/kiali/swscore/pull/126

mwringe:
"Click graph background to see summary information..."

Can we not just run this on the page load so that people don't need to click on the background to see something?

mwringe:
switching views from a namespace that has a service mesh to a namespace that doesnt also doesn't properly update the graph (it updates the name, but not the other details). This might go with my comment above with reloading the information when the page is loaded, or the namespace changes

jmazzitelli:
@mwringe Yes, I think this is related to something else I noticed today. Click any edge, see the side panel load with data, click another edge... most of the side panel data is static (doesn't change). We need to fix something about the click handler.

jshaughn:
> Can we not just run this on the page load so that people don't need to click on the background to see something?

No, we can't :)   This is what I was whining about yesterday.  At least I haven't figured out a way to do it yet, I'll look more today but let's get this in as is first, so that we at least have something to show.

src/pages/ServiceGraph/SummaryPanelGraph.tsx
```diff
@@ -14,46 +29,147 @@ export default class SummaryPanelGraph extends React.Component<SummaryPanelPropT
...
-            {this.renderOutgoingRpsChart()}
-            <ErrorRatePieChart percentError={10} />
+          <RateTable
+            title="Traffic (requests per second):"
+            rate={this.state.rate}
+            rate3xx={this.state.rate3xx}
+            rate4xx={this.state.rate4xx}
+            rate5xx={this.state.rate5xx}
+          />
+          <div style={{ fontSize: '1.2em' }}>
```
jmazzitelli: remove this font size style - keep fonts consistently sized across all panels.

src/pages/ServiceGraph/ServiceGraphPage.tsx
```diff
@@ -66,7 +68,9 @@ export default class ServiceGraphPage extends React.Component<RouteComponentProp
   }
 
   handleGraphClick = (data: any) => {
-    this.setState({ summaryData: data });
+    if (data !== undefined) {
```
mtho11: `undefined` is falsey so you can just use `if(data)`

https://www.sitepoint.com/javascript-truthy-falsy/

src/pages/ServiceGraph/SummaryPanelGraph.tsx
```diff
@@ -14,46 +29,147 @@ export default class SummaryPanelGraph extends React.Component<SummaryPanelPropT
...
+      <ServiceInfoBadge scale={0.8} style="plastic" leftText="edges" rightText={numEdges} color="green" />
     </>
   );
 
-  renderIncomingRpsChart = () => {
-    const dataRps: any = [['x', 1500, 3500, 5500, 7500, 9500, 10500], ['RPS', 350, 400, 150, 850, 50, 220]];
-    const dataErrors: any = [['x', 1500, 3500, 5500, 7500, 9500, 10500], ['Error', 140, 100, 50, 700, 10, 110]];
+  private renderRpsChart = () => {
+    if (this.state.loading) {
+      return <h3>loading chart...</h3>;
```
jmazzitelli: use \<strong> here - keep font sizes consistently the same size across all panels.

src/pages/ServiceGraph/SummaryPanelGraph.tsx
```diff
@@ -14,46 +29,147 @@ export default class SummaryPanelGraph extends React.Component<SummaryPanelPropT
...
+        this.setState({
+          loading: false,
+          reqRates: reqRates,
+          errRates: errRates
+        });
+      })
+      .catch(error => {
+        this.setState({ loading: false });
+        console.error(error);
+        // this.props.onError(error);
```
mtho11: At some point we need to start handling these errors

src/pages/ServiceGraph/SummaryPanelGraph.tsx
```diff
@@ -14,46 +29,147 @@ export default class SummaryPanelGraph extends React.Component<SummaryPanelPropT
...
+    });
+  };
+
+  private renderLabels = (numNodes: string, numEdges: string) => (
     <>
-      <ServiceInfoBadge scale={0.8} style="plastic" leftText="app" rightText="bookinfo" color="green" />
-      <ServiceInfoBadge scale={0.8} style="plastic" leftText="app" rightText="product" color="green" />
-      <ServiceInfoBadge scale={0.8} style="plastic" leftText="version" rightText="v5" color="navy" />
+      <ServiceInfoBadge scale={0.8} style="plastic" leftText="services" rightText={numNodes} color="green" />
+      <ServiceInfoBadge scale={0.8} style="plastic" leftText="edges" rightText={numEdges} color="green" />
```
jmazzitelli: also use patternfly colors:   color="#2d7623" // pf-green-500

src/pages/ServiceGraph/SummaryPanelGraph.tsx
mtho11: If this is used in more than one place then it needs to be its own component so the style changes can be managed in one place.

mtho11: Ideally, we would have components all the way down the tree until we get to leaf nodes that would consist of html markup.

PR #158, opened 2018-03-21
KIALI-267: Add empty state page when there is no service graph available

mwringe:


mwringe:
Adding 'Do Not Merge', awaiting this PR to be merged first https://github.com/kiali/swsui/pull/147 which will require some minor changes to this PR.

vnugent:
Just a thought.  Maybe returning a simple `<div>This is an empty graph.</div>` instead of introducing more components?  For example,
```
render() {
   if (elements is empty) {
     return <div>This is an empty graph... explain more what the user can do.</div> 
  }
  return <normal cytoscape component>
}
```

mwringe:
I was trying to introduce the more components to do it the patternfly way (and it also looks nicer :)

mwringe:
screenshot from 2018-03-22 15-04-58

mwringe:
It doesn't resolve the side panel, but this might be a good start to getting something on the page instead of the blank page we see now when there is no service graph available

mtho11:
I'm +1 on more components and less markup. Components are more structured and reusable. React is about _Components_.

vnugent:
I'm in agreement with creating new components.  I made the comment hastily due to the amount of look-and-feel/presentation code added to CytoscapeLayout where a simple `<div>help message</div>` can later be replaced with a more polished component once we start addressing UXD.  

Upon running the PR locally the help message is actually quite nice. I'd suggest to extract the code to a separate file/component as we want to keep CytoscapeLayout as lean as possible.  

Also I'm proposing keeping conditional rendering check in `CytoscapeLayout.render()` as it would read like pseudocode for someone who is not familiar with the code.  Furthermore we're not leaking internal graph data structure to the Help message component.
```
render() {
   if (elements is empty) {
     return <HelpMessage refreshOnClick={ props.refreshOnClick} /> 
  }
  return <normal cytoscape component>
}
```

mwringe:
Lets see if we can get this closed

A couple of questions:

1) do we want to make this just for the service graph area(like it is now)? or also have cover the side panel? Eg if there is no istio service graph for a namespace, should we even display the side panel?

2) for moving this to a separate file/component, I assume this should just be a specific component for the service graph and not abstracted to be something more reusable in other areas.

3) do we want explicit checks (if {elements is empty} then {display emptyState} else {display ReactCytoscape}) or embedded like the spinner (<emptyState><ReactCytoscape></emptyState>)?

vnugent:
> do we want to make this just for the service graph area(like it is now)? or also have cover the side panel? Eg if there is no istio service graph for a namespace, should we even display the side panel?

Good point.  Since the URL drives the active namespace value, which could be invalid, non-existent. It doesn't make sense to show a help message with blank summary panel on the side.  I'd move the conditional rendering check up to the parent.  But IMO PR is fine as is. We can re-factor later.

josejulio:
> do we want explicit checks (if {elements is empty} then {display emptyState} else {display ReactCytoscape}) or embedded like the spinner ()?

I'm +1 on embedded for this simple usage, we are not doing anything else that requires a if/else

abonas:
in the above screenshot, there is a "plus" sign (+) . typically this is used when a "create new ...." is available to users so they can click that + sign and create something, which is not the case here.

mwringe:
This was updated a couple of days ago, can we please get this reviewed and finally closed?

Moving to get the whole page (including sidepanel) is a bit more tricky than expected, so I didn't do it at this point (we need to figure out how to properly size the cytoscape graph, currently its hacked up together and causes problems)

josejulio:
Current work doesn't have the plus sign.

screenshot from 2018-04-10 11-47-19

PR #163, opened 2018-03-22
[minor] ServiceHealth: move some class fields to constants, …

jotak:
…add showTitle param


jotak:
@lucasponce simple tiny change...

PR #167, opened 2018-03-22
KIALI-396: Ensure charts are updated when a versioned node of the graph is selected

israel-hdez:
The sparklines will now update correctly, which are the only objects not being updated correctly when selection from the graph is changed by the user.

This fix is only for versioned nodes.


jmazzitelli:
this looks good to me. Putting @jshaughn on the reviewer list for him to look.

PR #168, opened 2018-03-22
KIALI-396 Fix edge summary rendering, and simplify impl.

jshaughn:
- also, prevent link to "unknown" source service

### DIFF
--- a/src/pages/ServiceGraph/SummaryPanelEdge.tsx
+++ b/src/pages/ServiceGraph/SummaryPanelEdge.tsx
@@ -9,15 +9,6 @@ import graphUtils from '../../utils/graphing';
 
 type SummaryPanelEdgeState = {
   loading: boolean;
-  sourceService: string;
-  sourceServiceName: string;
-  sourceNamespace: string;
-  sourceVersion: string;
-  dest: string;
-  destService: string;
-  destServiceName: string;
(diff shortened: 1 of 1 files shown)


src/pages/ServiceGraph/SummaryPanelEdge.tsx
```diff
@@ -33,82 +24,58 @@ export default class SummaryPanelEdge extends React.Component<SummaryPanelPropTy
...
-        {this.state.destServiceName}
-      </a>
-    );
+    const sourceLink = <a href={`../namespaces/${sourceNamespace}/services/${sourceService}`}>{sourceServiceName}</a>;
+    const destLink = <a href={`../namespaces/${destNamespace}/services/${destService}`}>{destServiceName}</a>;
 
     return (
       <div className="panel panel-default" style={SummaryPanelEdge.panelStyle}>
-        <div className="panel-heading">Edge Source: {sourceLink}</div>
+        <div hidden={sourceServiceName === 'unknown'} className="panel-heading">
```
israel-hdez: Not a blocker, but a suggesetion: Perhaps this can take advantage of conditional rendering instead of injecting two blocks: https://reactjs.org/docs/conditional-rendering.html#inline-if-else-with-conditional-operator

PR #169, opened 2018-03-22
get the group spark graph to load properly

jmazzitelli:


src/pages/ServiceGraph/SummaryPanelGroup.tsx
```diff
@@ -155,7 +133,38 @@ export default class SummaryPanelGroup extends React.Component<SummaryPanelPropT
     );
   }
 
-  renderVersionBadges = () => {
+  private updateRpsCharts = (props: SummaryPanelPropType) => {
+    const namespace = this.props.data.summaryTarget.data('service').split('.')[1];
```
jshaughn: Should not be this.props, just props (true in a couple of places)

PR #171, opened 2018-03-22
Allow sparklines to "morph" when changing (versioned) node selection in graph

israel-hdez:


israel-hdez:
@jmazzitelli this is for you :)

israel-hdez:
Although I think we should show some "loading" indicator while the data is being fetched from the backend.

jmazzitelli:
Yes, I'm not convinced the morphing is the way to go. The reason is sometimes it might take a while to load the data. Seeing the "loading" message is helpful. Otherwise, there is no indicator that the graph I am looking at is "old" and will soon morph with the new data when it comes in.

jmazzitelli:
Perhaps have some kind of "loading" icon somewhere near/on the spark graph? I dunno.  But without some kind of loading indicator or message, the morphing alone (while cool :)) is not good enough.

israel-hdez:
Yes, I agree it's not enough.
We could show a small spinner. I think there is enough space to add it next to the "incoming/outgoing" texts. Anyway, lets push this to the next sprint. It's too late now to think about these things.

israel-hdez:
Closing because of very, very low priority.

PR #172, opened 2018-03-22
KIALI-400: Set step and rateInterval dependent upon duration.

jmazzitelli:
This ensures the graph will always have 30 data points, thus making the graphs look consistent but also allowing the server-side request to execute fast even when processing as large a data set as 30 days worth.
Note the current implementation uses a consistent rate interval of 1m. Not sure if it would be more efficient to increase the interval as the data size gets larger (e.g. is it better to query with 5m rate intervals over a 7d or 30d duration?). Leaving this as-is will allow us to change intervals per duration if we need to in the future.

### DIFF
--- a/src/components/GraphFilter/GraphFilter.tsx
+++ b/src/components/GraphFilter/GraphFilter.tsx
@@ -12,7 +12,21 @@ import { LayoutButtonGroup } from './LayoutButtonGroup';
 import { KlayGraph } from '../CytoscapeLayout/graphs/KlayGraph';
 
 export namespace GraphFilters {
+  const graphQueryOptionsPerDuration = {
+    '60': { step: 2, rateInterval: '1m' }, // 1m - 60/2=30 buckets which is 1 datapoint per 2s
+    '600': { step: 20, rateInterval: '1m' }, // 10m - 600/20=30 buckets which is 1 datapoint per 20s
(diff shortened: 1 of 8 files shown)


jmazzitelli:
As the comments indicate in the code, depending on what duration the user selects, an appropriate "step" option is used. (rateInterval is remaining constant at "1m" unless and until we know of a reason to increase it as duration increases).

The way the steps are defines are to ensure the sparkgraphs always have at most 30 buckets or data points (actually it will be 31 datapoints if you count the initial data point at the start of the graph).

If you lack data, you will have less datapoints, but if, for example, you have more than 1 days worth of data, if you select a duration of 1d or less you will always have 30 buckets in your graph (31 datapoints counting the initial data point).

Each bucket worth of data will represent a different amount of time duration depending on the total duration selected in the UI. This is the breakdown:

If you select a duration of ... each bucket represents one time duration of:
1m ... 2 seconds
10m ... 20 seconds
30m ... 1 minutes
1h ... 2 minutes
4h ... 8 minutes
8h ... 16 minutes
1d ... 48 minutes
7d ... 5.6 hours
30d ... 1 day

jmazzitelli:
Definitely can calculate this for sure. duration divided by 30 is the step to use. The thinking was I was going to have to have a custom rateInterval too (which probably isn't calculable) hence this map/hardcode. We can calculate later if we want.

PR #174, opened 2018-03-23
[KIALI-399] Select sidebar according to browser url

vnugent:
- The sidebar is working as expected as you click on it
- If you bookmark and revisit a page later, while correct view is shown 'Graph' always gets selected.
- Partially addressed KIALI-399.  The hotlink is still incorrect.  For a smooth transition between views it's better to use React Router to update browser url.
https://github.com/kiali/swsui/blob/master/src/pages/ServiceGraph/SummaryPanelGraph.tsx#L78
 ```
const servicesLink = <a href={`../services?namespace=${this.props.namespace}`}>{this.props.namespace}</a>;
```

### DIFF
--- a/src/components/Nav/Navigation.tsx
+++ b/src/components/Nav/Navigation.tsx
@@ -11,16 +11,24 @@ import ServiceGraphPage from '../../pages/ServiceGraph/ServiceGraphPage';
 import ServiceListPage from '../../pages/ServiceList/ServiceListPage';
 
 const istioRulesPath = '/rules';
-const istioRulesTitle = 'Istio Mixer';
+export const istioRulesTitle = 'Istio Mixer';
 const serviceGraphPath = '/service-graph/istio-system';
-const serviceGraphTitle = 'Graph';
+export const serviceGraphTitle = 'Graph';
 const servicesPath = '/services';
-const servicesTitle = 'Services';
+export const servicesTitle = 'Services';
 
 const pfLogo = require('../../img/logo-alt.svg');
 const pfBrand = require('../../assets/img/kiali-title.svg');
 
-class Navigation extends React.Component {
+type PropsType = {
+  location: any;
+};
+
+type StateType = {
+  selectedItem: string;
+};
+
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by lucasponce** [comment]
> const servicesLink = <a href={`../services?namespace=${this.props.namespace}`}>{this.props.namespace}</a>;

The hotlink is still a different topic, I guess.
The filter state is stored globally, so if you want to move to services list page with a particular filter state you can change that.

The service details page can show an example of that:
https://github.com/kiali/swsui/blob/master/src/pages/ServiceDetails/ServiceDetailsPage.tsx#L25

Another thing is if we want a bookmark maintaining the filtering state, but I guess that is a separate thing.

File: src/components/Nav/Navigation.tsx [hunk 1]
```diff
@@ -11,25 +11,60 @@ import ServiceGraphPage from '../../pages/ServiceGraph/ServiceGraphPage';
...
+    // handle initial path from the browser
+    const selected = this.parseInitialPath(props.location.pathname);
+    this.state = {
+      selectedItem: `/${selected}/`
+    };
   }
 
+  parseInitialPath = (pathname: string) => {
+    let selected = '';
+    if (pathname.startsWith('/namespaces') || pathname.startsWith('/services')) {
```
lucasponce: Minor. For consistency, I would suggest to use the constants for the path.
So, then, if some change, we have them located in a single point.
(And also to follow same style within same file).

src/components/Nav/Navigation.tsx
```diff
@@ -11,25 +11,60 @@ import ServiceGraphPage from '../../pages/ServiceGraph/ServiceGraphPage';
...
+      selected = serviceGraphTitle;
+    } else if (pathname.startsWith('/rules')) {
+      selected = istioRulesTitle;
+    } else {
+      selected = serviceGraphTitle;
+    }
+    return selected;
+  };
+
+  setControlledState = event => {
```
lucasponce: +1
I forgot this in the preliminar version, thanks !

PR #175, opened 2018-03-23
[WIP] Kiali 345  Node/Edge label toggle

mtho11:
A preliminary version of the Edge and Node toggle. Visually this works. However, I have no handle to the CytoscapeLayout component because of no commonality in the component tree (it's not a parent, just another node in the tree somewhere). 

label-toggle

- [ ] Use MobX to gain publish/subscribe capability to a common datastore.

https://issues.jboss.org/browse/KIALI-345

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -7,9 +7,7 @@ class App extends React.Component {
   render() {
     return (
       <BrowserRouter basename="/console">
-        <div>
-          <Navigation />
-        </div>
+        <Navigation />
       </BrowserRouter>
     );
   }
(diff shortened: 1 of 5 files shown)


mtho11:
FWIW: the toggle Switch component is actually in Patternfly React but there is no documentation of it -- only looking through the source code: 

[Patternfly Storybook](https://rawgit.com/patternfly/patternfly-react/gh-pages/index.html?selectedKind=About%20Modal&selectedStory=AboutModal&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs)

abonas:
Can you please clarify what is the expected functionality/use case here? What does toggling of nodes do? Why is it needed? thank you

mtho11:
@abonas the use case here is turning on/off the labels for nodes and edges. Sometimes the graph gets very crowded and unreadable with the labels in the picture. This gives the user the ability to control what is shown.

Perhaps the labels shown in the toggles should be renamed 'Edge Labels' and 'Node Labels' to avoid the confusion? I can understand the confusion.

mtho11:
Closed in favor of #245

PR #177, opened 2018-03-23
SWS-403 handle errors in ServiceMetrics

jotak:


jotak:
@lucasponce while I think this code change is safe, we can wait the demo before merging it (demo code freezing?). It doesn't bring much value to show anyway.

lucasponce:
> @lucasponce while I think this code change is safe, we can wait the demo before merging it (demo code freezing?). It doesn't bring much value to show anyway.

I'm fine merging it if you want. No problem from my side.

jotak:
then let's go :)

PR #179, opened 2018-03-23
[KIALI-406][Refactor] bind method to this with helper

aljesusg:
Refactor bind items with a helper.

- More cleaner and reusable.

### DIFF
--- a/src/components/CytoscapeLayout/CytoscapeLayout.tsx
+++ b/src/components/CytoscapeLayout/CytoscapeLayout.tsx
@@ -6,6 +6,7 @@ import * as API from '../../services/Api';
 import { GraphStyles } from './graphs/GraphStyles';
 import { GraphHighlighter } from './graphs/GraphHighlighter';
 import ReactCytoscape from './ReactCytoscape';
+import { bindMethods } from '../../utils/helpers';
 
 type CytoscapeLayoutProps = {
   namespace: string;
@@ -45,10 +46,7 @@ export default class CytoscapeLayout extends React.Component<CytoscapeLayoutProp
       elements: [],
       loading: false
     };
-    this.updateGraphElements = this.updateGraphElements.bind(this);
-    this.handleTap = this.handleTap.bind(this);
-    this.handleMouseIn = this.handleMouseIn.bind(this);
-    this.handleMouseOut = this.handleMouseOut.bind(this);
+    bindMethods(this, ['updateGraphElements', 'handleTap', 'handleMouseIn', 'handleMouseOut']);
   }
 
   resizeWindow() {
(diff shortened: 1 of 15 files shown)


mtho11:
This is nice. You should show some example syntax changes in the PR description (before/after).
The only downside is that strings are not typesafe (possibly handled with Generics with an improvement later), but much more compact in syntax.

The normal React binding syntax irritates me as well. As it needs some syntactic sugar.

josejulio:
I have been using the `public class fields syntax` [1]  to avoid (as React docs say) the annoying syntax
But i suppose we should agree on what syntax use and stick to that one :)

> This is not React-specific behavior; it is a part of how functions work in JavaScript. Generally, if you refer to a method without () after it, such as onClick={this.handleClick}, you should bind that method.
>
> If calling bind annoys you, there are two ways you can get around this. If you are using the experimental public class fields syntax, you can use class fields to correctly bind callbacks:
```typescript
handleClick = () => {
  console.log('this is:', this);
}
```
instead of 
``` typescript
handleClick() {
  console.log('this is:', this);
}
// ...
constructor() {
  this.handleClick = this.handleClick.bind(this)
}
```


[1] https://reactjs.org/docs/handling-events.html

src/components/CytoscapeLayout/CytoscapeLayout.tsx
```diff
@@ -45,10 +46,7 @@ export default class CytoscapeLayout extends React.Component<CytoscapeLayoutProp
       elements: [],
       loading: false
     };
-    this.updateGraphElements = this.updateGraphElements.bind(this);
-    this.handleTap = this.handleTap.bind(this);
-    this.handleMouseIn = this.handleMouseIn.bind(this);
-    this.handleMouseOut = this.handleMouseOut.bind(this);
+    bindMethods(this, ['updateGraphElements', 'handleTap', 'handleMouseIn', 'handleMouseOut']);
```
jotak: I'm -1 on this because you loose any compiler error in case of mistyping the function name.
Or could you put function reference in the array instead of strings?
`bindMethods(this, [this.updateGraphElements, this.handleTap, this.handleMouseIn, this.handleMouseOut]);`

jotak: PS: I realize it may not be possible to use function references since `bind` doesn't modify the function passed as parameter, but return a new bound function instead

jshaughn:
> The only downside is that strings are not typesafe (possibly handled with Generics with an improvement later), but much more compact in syntax.

For me this is big downside.  Compact syntax just makes things more and more dense and harder to understand.  I'd rather have something explicit and easier to understand, especially if it enhances type safety.

vnugent:
Why can't we use fat arrow => function?

jotak:
@vnugent yes, that was @josejulio suggestion too

mtho11:
Many of already do use fat arrow, but for those who don't...

mtho11:
@aljesusg I guess we should close in favor of fat arrow. I believe that is the consensus.

aljesusg:
Yes...Thanks @mtho11

PR #187, opened 2018-03-26
KIALI-427 bump version of ssri.

pilhuhn:
I locally see an error

```
events.js:137
      throw er; // Unhandled 'error' event
      ^
``` 
on `npm test`, but I see the same with the current version

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2903,7 +2903,7 @@
         "move-concurrently": "1.0.1",
         "promise-inflight": "1.0.1",
         "rimraf": "2.6.2",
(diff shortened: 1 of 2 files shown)


mtho11:
I actually don't know what the ssri package is.

mtho11:
This is not bumping but adding `ssri`

pilhuhn:
On 26 Mar 2018, at 16:39, Mike Thompson wrote:

> This is not bumping but adding `ssri`

It is already pulled in indirectly at the old version.
Adding it explicitly pulls the new / correct version.

PR #193, opened 2018-03-26
KIALI-180: Visually identify versioned nodes in service graph.

israel-hdez:
Changing borders of versioned microservices and making the group box to be more snug.
With these changes, the graph will look like this:

image

### DIFF
--- a/src/components/CytoscapeLayout/graphs/GraphStyles.ts
+++ b/src/components/CytoscapeLayout/graphs/GraphStyles.ts
@@ -36,13 +36,23 @@ export class GraphStyles {
         // version group boxes
         selector: '$node > node',
         css: {
-          'padding-top': '10px',
-          'padding-left': '20px',
-          'padding-bottom': '10px',
-          'padding-right': '20px',
+          'padding-top': '5px',
+          'padding-left': '5px',
+          'padding-bottom': '5px',
+          'padding-right': '5px',
           'text-valign': 'top',
           'text-halign': 'center',
-          'background-color': '#fbeabc' // pf-gold-100
+          'background-color': '#fbeabc', // pf-gold-100
+          'border-color': '#b58100', // pf-gold-500
+          'border-width': '2px'
+        }
+      },
+      {
+        // versioned node in a group
+        selector: 'node > $node',
+        css: {
+          'border-color': '#b58100', // pf-gold-500
+          'border-width': '2px'
         }
       },
       {


israel-hdez:
@gbaufake does this helps?

mtho11:
I think we need pictures of what it looks like with the various layouts (because we have not decided on one yet).

mwringe:
I don't know if I like the square boxes, it makes it seem like its a different component type

israel-hdez:
@mwringe Yes, you have a point. I can omit the change of the shape and change only the color of the borders. As I said, perhaps only for me is not enough the color of the border. But if you are ok... I think it's ok.

mtho11:
This needs to be rebased with master as the labels now are centered over the shapes.

israel-hdez:
@mtho11 Rebased. I'm keeping the original shape. Now that only border colors are changed, do you still need screenshots of the other layouts?

mtho11:
@israel-hdez I think we do because the layout shown above may not be the couple we go with.

israel-hdez:
@mtho11 I updated the main description with an image of all layouts.

pilhuhn:
I am not convinced, as the grouping-box should only contain services that are the same app in different versions. 
I.e. in the Dagre layout, only the reviews should be in the group 'box' and the ratings should be outside.
One can argue that users can move the rating node into the reviews box and the distinction is needed.
But what if the user also moves the details into the box - would that get a different border/shape from the ratings?
I think if we have a sensible default layout, that already takes care of this, then the different colors/shapes are not necessarily needed.

israel-hdez:
@pilhuhn In the dagre layout I moved the nodes to enclose the ratings to show how it looks. Initially, the box wasn't enclosing it. If the user moves a node into the box, it's appearance won't change. It's not becoming part of the group, the user is only changing the position of the node.
I've googled a little and I cannot find any layout or extension that "fixes" nodes/groups overlapping when the user changes positioning. I would need to implement that logic.

mwringe:
The main task for this is https://issues.jboss.org/browse/KIALI-180 it might be better to discuss more there (especially if we re-open another PR and this one gets closed).

We have a problem with the grouping of the boxes, but I don't think we yet have a good solution to ti

mtho11:
Is the first image Dagre or Cose?

israel-hdez:
@mtho11 I got that layout for both Dagre and Cose. Perhaps it was fixed in #196 .

jshaughn:
i fixed that problem with cose rendering as dagre...

mtho11:
Needs a rebase...

israel-hdez:
Solving KIALI-180 with #218.

PR #190, opened 2018-03-26
kiali-440: Bump all versions in package.json (and package-lock.json).

mtho11:
Bump all the versions to latest.

PR #195, opened 2018-03-27
Fix tests after upgrade KIALI-440

aljesusg:
Require merged PR [#190](https://github.com/kiali/kiali-ui/pull/190)

cc @jotak @mtho11


jotak:
As discussed on mailing list, we could keep all the snapshots (and just update them). I'd rather be for that solution. But let's wait for @mtho11 opinion :)
Apart from that, +1 for the fix on ServiceInfo

aljesusg:
I added the Badge tests too ^^

mtho11:
Nice work on the fix @aljesusg @jotak!

mtho11:
Let's give the snapshots one more chance. But not add any new snapshots. So far, these snapshots have been very man hours intensive. I'm not really sure they are worth it -- IMO.

I'd rather just test the functionality via assertions than test the actual markup, which is not that important and changes much more often.

src/components/Badge/__tests__/Badge.test.tsx
```diff
@@ -2,10 +2,157 @@ import * as React from 'react';
...
+    const wrapper = mockBadge('', '', scale);
+    let textSize = 11 * scale;
+    let gElement = wrapper.find('g').getElements()[0];
+    expect(gElement.props.fontSize).toEqual(textSize);
+    expect(gElement.props.fontFamily).toEqual('DejaVu Sans,Verdana,Geneva,sans-serif');
+    expect(gElement.props.fill).toEqual('#fff');
+  });
+
+  it('should render text elements with correct size and props', () => {
+    let scale = 1;
```
mtho11: Most of these `let` can be `const`

jotak:
@mtho11 I don't see these snapshots so hour intensive, here in the end the problem was more due to the deps update itself than to the snapshots - and the biggest issue we had with @aljesusg was about a traditional test, not a snapshot test.
We must just get used to run `npm test -- -u` locally.

src/components/Badge/__tests__/Badge.test.tsx
aljesusg: Changed to const ;)

aljesusg:
@mtho11 this is ready I updated the PR

aljesusg:
Thanks !!! @mtho11 and sorry for delay

PR #197, opened 2018-03-27
move group label to right side and rotate.  

jshaughn:
Move away from center because there is likely always an edge through the center.  Also, thin edges
for a generally cleaner look and less overlap when crossing over text.

screen

### DIFF
--- a/src/components/CytoscapeLayout/graphs/GraphStyles.ts
+++ b/src/components/CytoscapeLayout/graphs/GraphStyles.ts
@@ -36,19 +36,18 @@ export class GraphStyles {
         // version group boxes
         selector: '$node > node',
         css: {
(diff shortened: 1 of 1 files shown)


jshaughn:
mazz, I upped the edge thickness from 1 to 2 since you thought the edges were too thin after being reduced from 3 to 1.  You're wrong but what can I do, you are the reviewer...

mtho11:
Personally, I'm not that wild about the sideways (180 degree) stuff.

jshaughn:
@mtho11 I don't love it but I like it much, much better than the alternatives.  Anyway, it's a work in progress and UX is soon to intervene.  I still do like the vizceral approach of small nodes, small labels, and left-to-right flow.

PR #202, opened 2018-03-28
introduce Mobx model and 'side-way' injection of namespace list (Part1)

vnugent:
- introduce a simple Mobx model for storing namespace list
- refactor `GraphFilter` to use the model directly [*]
- unit test

At first glance there's no apparent benefit of keeping namespace list in a common/external model.  In the next PR I'll inject the model in `ServiceListComponent`.  `GraphFilter` and `ServiceListComponent` will then 'react' (re-render) automatically when namespace list changes.  Without using Mobx  it'd be cumbersome to pass around a share namespace list as prop since the two components are under separate trees.

[*] 'side-way' injection:  top-down Provider/inject would be a better choice, but I couldn't figure out how to use inject directly due to complex typescript hacks.
https://github.com/mobxjs/mobx-react/issues/256


mtho11:
Let's use MobX strict mode so that people can't mutate the store everywhere, but only via `actions`.

mtho11:
> but I couldn't figure out how to use inject directly due to complex typescript hacks.

@vnugent I'm surprised there are any complex typescript hacks because MobX is written in typescript.

vnugent:
@mtho11 yes it's enabled in the model
https://github.com/kiali/kiali-ui/pull/202/files#diff-37239ea437b4c404e0e900f3528d5e82R6

Here's a quick test.  Changing the model directly resulting in a runtime error.
image

src/components/GraphFilter/GraphFilter.tsx
```diff
@@ -59,12 +78,13 @@ export class GraphFilter extends React.Component<GraphFilterProps, GraphFilterSt
...
               title={this.props.activeNamespace.name}
               onSelect={this.updateNamespace}
+              onClick={this.reloadData}
             >
-              {this.state.availableNamespaces.map(ns => (
-                <MenuItem key={ns.name} active={ns.name === this.props.activeNamespace.name} eventKey={ns.name}>
-                  {ns.name}
-                </MenuItem>
-              ))}
+              <NSList
```
mtho11: Let's spell out `Namespace` its only a few more characters and is meaningful to everyone reading the code (at least in the TagName)

vnugent: ack

src/components/GraphFilter/GraphFilter.tsx
```diff
@@ -1,30 +1,49 @@
...
 import { ButtonToolbar } from 'react-bootstrap';
 
 import { GraphFilterProps, GraphFilterState } from '../../types/GraphFilter';
-import * as API from '../../services/Api';
+import Namespace from '../../types/Namespace';
 import { DurationButtonGroup } from './DurationButtonGroup';
 import { LayoutButtonGroup } from './LayoutButtonGroup';
+import { namespaceStore } from '../../model/Namespace';
 
+type DropdownListType = {
```
jotak: Just for my understanding: when is it better to use type, when is it better to use interface? Is it just a matter of taste, or is there any commonly accepted guidelines on this?

src/components/GraphFilter/GraphFilter.tsx
```diff
@@ -1,30 +1,49 @@
...
+  list: Namespace[];
+  current: Namespace;
+  onSelect: Function;
+};
+
+export class DropdownList extends React.PureComponent<DropdownListType, {}> {
```
jotak: I'm just reading about that ( https://reactjs.org/docs/react-api.html#reactpurecomponent )
So IIRC we should use PureComponent as much as possible, as I think many of our components rendering functions are only state/props dependant.

vnugent:
@jotak
Error handling is now in the model.

jotak:
@vnugent yes but it still has to be displayed somehow on screen

vnugent:
@jotak before the refactoring GraphFilter dropdown didn't display any error messages in case of a backend error (also it didn't make sense to do so for a toolbar).

Since error flag and message are observables, they can be used when we introduce a global error handling component. 
 
Do you think the dropdown component itself should show the error?

src/components/GraphFilter/GraphFilter.tsx
josejulio: I guess we should use `interfaces` when creating types that will be used externally to give more context when an error happens.

> One difference is that interfaces create a new name that is used everywhere. Type aliases don’t create a new name — for instance, error messages won’t use the alias name.

https://www.typescriptlang.org/docs/handbook/advanced-types.html#interfaces-vs-type-aliases

src/components/GraphFilter/GraphFilter.tsx
```diff
@@ -1,30 +1,49 @@
...
 
   componentDidMount() {
-    // TODO: [KIALI-436] API.GetNamespaces() is also called in Services component.
-    // We should consolidate them into one.
-    API.GetNamespaces()
-      .then(this.setNamespaces)
-      .catch(error => {
-        this.props.onError(error);
-      });
+    this.reloadData();
```
josejulio: If different components implement this, isn't that going to cause an api call for each component (that uses the namespace)?

src/components/GraphFilter/GraphFilter.tsx
```diff
@@ -1,30 +1,49 @@
...
   }
 
+  reloadData = () => {
+    namespaceStore.fetchNamespacesFromBackend();
+  };
+
   setNamespaces = (response: any) => {
     this.setState({ availableNamespaces: response['data'] });
   };
```
josejulio: I suppose this can go away.

src/components/GraphFilter/GraphFilter.tsx
```diff
@@ -55,16 +74,12 @@ export class GraphFilter extends React.Component<GraphFilterProps, GraphFilterSt
...
-              onSelect={this.updateNamespace}
-            >
-              {this.state.availableNamespaces.map(ns => (
-                <MenuItem key={ns.name} active={ns.name === this.props.activeNamespace.name} eventKey={ns.name}>
-                  {ns.name}
-                </MenuItem>
-              ))}
+            <DropdownButton id="namespace-selector" title={this.props.activeNamespace.name} onClick={this.reloadData}>
+              <DropdownList
+                list={namespaceStore.namespaceList}
```
josejulio: Question, should we move this list to a props? `list={this.props.namespaces}`? That would make it stateless (desired), right?

src/components/GraphFilter/GraphFilter.tsx
vnugent: Yup, good catch! Thanks.

src/components/GraphFilter/GraphFilter.tsx
vnugent: Ideally yes. This direct “sideway” injection (as oppose to “top down” via props) is a temp solution due to difficulty I had with typescript as mentioned in the PR foot note.

jotak:
> @jotak before the refactoring GraphFilter dropdown didn't display any error messages in case of a backend error (also it didn't make sense to do so for a toolbar).

I see in diff that there was an `onError` handler, but I haven't followed how it was used from the parent component. I agree that it should be connected to some error handling component at a higher level.

src/components/GraphFilter/GraphFilter.tsx
mtho11: To sum up, if you plan on inheriting or extending then `interface` is better. Type is better if its just local and/or you need to use `type guards`

mtho11:
@vnugent I guess this one should be close now and any applicable changes moved over to redux.

vnugent:
@mtho11 I'm closing this PR.  New task: https://issues.jboss.org/browse/KIALI-516

PR #203, opened 2018-03-29
Improve l+f by moving summary labels into panel header

jshaughn:
increase size for readability.  This also saves space as we look to add
more summary info to the side panels.

Here is an idea of what it looks like using edge as an example:

screen


abonas:
@jshaughn it will be more easier to understand what is the change about if you add "before" and "after" screenshots and not only the "after" one :)

PR #205, opened 2018-03-29
kiali-464: make npm package versions static.

mtho11:
For stability reasons lock the versions in `package.json`
Note: package-lock.json did not change.

https://issues.jboss.org/browse/KIALI-464


mtho11:
This was discussed in the ML.

PR #207, opened 2018-04-02
KIALI-430 Fix about name UI

aljesusg:
about

[KIALI-430](https://issues.jboss.org/browse/KIALI-430)


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@kiali/kiali-ui",
```
josejulio: We need `@kiali/kiali-ui` to be able to publish it to [npm kiali's account](https://www.npmjs.com/package/@kiali/kiali-ui)

mtho11: I guess I don't understand the purpose of this because the package name is already correct.

mtho11: Ok, Maybe we also want to display `Npm Package Name` as well? Since that is the artifact that gets published. So a link on that item that would take you to https://www.npmjs.com/package/@kiali/kiali-ui would be useful to view the dependencies and version info.

aljesusg:
@josejulio @mtho11 I hardcode the value.

aljesusg:
Thanks @mtho11 & @josejulio

PR #208, opened 2018-04-02
KIALI-388: Ordering by error rate

israel-hdez:
Displaying error rates in the service list and allowing the user to sort by error rate:

image

This PR depends on PR kiali/kiali#138

(The error rates in the image are random numbers artificially set, to illustrate how it will look)


src/pages/ServiceList/ServiceListComponent.tsx
```diff
@@ -130,7 +138,7 @@ class ServiceListComponent extends React.Component<ServiceListComponentProps, Se
     });
   }
 
-  updateServices() {
+  updateServices(rateInterval?: string) {
```
mtho11: Maybe a default `rateInterval`?

src/pages/ServiceList/ServiceListComponent.tsx
```diff
@@ -43,6 +49,7 @@ type ServiceListComponentState = {
   pagination: Pagination;
   currentSortField: SortField;
   isSortAscending: boolean;
+  rateInterval: string;
```
mtho11: rateInterval should be a typesafe `enum` since it is bounded.

I have not worried about these kinds of things yet, but we need to start tightening up on this now.

abonas:
in the screenshot, error rate is (expectedly) percentage. but next to it there is an "a-z" sign which is more suitable for alphabetical order. is this something that is built in or there is another sign suitable for order of numeric values?

src/pages/ServiceList/ServiceListComponent.tsx
jotak: It may be bounded in the component, but not in the model (the rest api accepts unbounded rate interval as long as it has the right format), so I don't know if it's really better to define an arbitrary enum type.

josejulio:
~~Some notes [1] from patternfly~~
> This icon and text would display regardless of data type. The icon and text use “A to Z” as an example to communicate sort order, and are not intended to indicate type of data that displays in the selected column.


But then, there is this:

screenshot from 2018-04-03 12-02-45

see here: https://rawgit.com/patternfly/patternfly-react/gh-pages/index.html?selectedKind=Sort&selectedStory=Sort&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs
Select "Birth Month"

[1] http://www.patternfly.org/pattern-library/forms-and-controls/sort/#design

src/pages/ServiceList/ServiceListComponent.tsx
israel-hdez: Sure! Apparently, typescript allows to do `updateServices(rateInterval = MetricsOptionsBar.DefaultRateInterval)` in the definition of the function. I'll use that.

src/pages/ServiceList/ServiceListComponent.tsx
israel-hdez: I'm +1 with Joel's comment. Perhaps a better approach is to make an interface/class like this one:

```ts
interface TimeDuration {
    duration: number;
    timeUnit: TimeUnitEnum;
}
```

where `TimeUnitEnum` is a real enumeration with items for seconds, minutes, hours, etc. This way, we can query the rest API with any allowed rate interval rather than having a few predefined intervals.

mwringe:
Can we have the error rate and the pod status as separate columns on the page? Having them one on top of the other I find makes it hard to quickly glance over what is happening. We should have the room for it.

Can we also have the same kind of green checkbox or warning/error symbols besides the error rate? Like what we do with the pod status indicators.

israel-hdez:
@mwringe I moved the error rates to a new column as you requested.

@abonas @josejulio The "a-z" sign now changes to a "1-9" sign when error rate is selected and goes back to "a-z" when something else is selected.

I updated the image in the main description.

src/pages/ServiceList/ServiceListComponent.tsx
israel-hdez: I think I'll fix this in a follow-up PR, so that we can better discuss how to deal with this.

mwringe:
A couple of comments:

- It seems like this PR is missing tests, can we please get some tests added to this?

- It seems like if there isn't any request data, we just show "Error rate: 0.00%" which is a bit confusing. Could we change this so that it indicates that there is no available request data?

mwringe:
Another comment, the 'description' segment has the name of the service and the pod status. While the 'additional-info' segment has the error rate.

The error rate shouldn't be separate on its own. I think this is also why the columns look out of place with different spaces between them

israel-hdez:
@mwringe I updated the PR and the image for your comments. The tests I would like to do them in a separate PR since the service list page does not have any tests.

mwringe:
It looks good, but I can't see what the rate interval is after I select it. Shouldn't this be displayed somewhere?

PR #209, opened 2018-04-03
App test: setup mocks to avoid error while getting namespaces

jotak:


vnugent:
LGMT.  I tried to address the same thing in https://github.com/kiali/kiali-ui/pull/202

PR #210, opened 2018-04-03
KIALI-126 Refresh Metrics

aljesusg:
refresh_metrics
 
[KIALI-126](https://issues.jboss.org/browse/KIALI-126)

Refresh Metrics with a Polling Interval selector.

Interval options : Pause, 5s, 10s, 30s, 1m and 5m

screenshot from 2018-04-03 17-37-59

### DIFF
--- a/src/components/MetricsOptions/MetricsOptionsBar.tsx
+++ b/src/components/MetricsOptions/MetricsOptionsBar.tsx
@@ -1,15 +1,17 @@
 import * as React from 'react';
-import { Toolbar, DropdownButton, MenuItem } from 'patternfly-react';
+import { Toolbar, ToolbarRightContent, DropdownButton, MenuItem, Icon } from 'patternfly-react';
 
 import ValueSelectHelper from './ValueSelectHelper';
 import MetricsOptions from '../../types/MetricsOptions';
 
 interface Props {
-  onOptionsChanged: (opts: MetricsOptions) => void;
+  onOptionsChanged: (opts: MetricsOptions, interval: number) => void;
+  loading?: boolean;
 }
 
 interface MetricsOptionsState {
   rateInterval: string;
+  pollInterval: number;
   duration: number;
   ticks: number;
   groupByLabels: string[];
@@ -23,6 +25,16 @@ interface GroupByLabel {
 export class MetricsOptionsBar extends React.Component<Props, MetricsOptionsState> {
   static RateIntervals = [['1m', '1 minute'], ['5m', '5 minutes'], ['10m', '10 minutes'], ['30m', '30 minutes']];
   static DefaultRateInterval = MetricsOptionsBar.RateIntervals[0][0];
+
+  static PollIntervals = [
+    [0, 'Pause'],
+    [5000, '5 seconds'],
+    [10000, '10 seconds'],
+    [30000, '30 seconds'],
+    [60000, '1 minute'],
+    [300000, '5 minutes']
+  ];
+  static DefaultPollInterval = 5000;
 
   static Ticks = [10, 20, 30, 50, 100, 200];
   static DefaultTicks = MetricsOptionsBar.Ticks[2];
@@ -64,6 +76,7 @@ export class MetricsOptionsBar extends React.Component<Props, MetricsOptionsStat
     super(props);
 
     this.onRateIntervalChanged = this.onRateIntervalChanged.bind(this);
+    this.onPollIntervalChanged = this.onPollIntervalChanged.bind(this);
     this.onDurationChanged = this.onDurationChanged.bind(this);
     this.onTicksChanged = this.onTicksChanged.bind(this);
     this.changedGroupByLabel = this.changedGroupByLabel.bind(this);
@@ -77,6 +90,7 @@ export class MetricsOptionsBar extends React.Component<Props, MetricsOptionsStat
 
     this.state = {
       rateInterval: MetricsOptionsBar.DefaultRateInterval,
+      pollInterval: MetricsOptionsBar.DefaultPollInterval,
       duration: MetricsOptionsBar.DefaultDuration,
       ticks: MetricsOptionsBar.DefaultTicks,
       groupByLabels: []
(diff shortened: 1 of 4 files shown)


aljesusg:
@jotak @mtho11 This is ready.

src/pages/ServiceDetails/ServiceMetrics.tsx
```diff
@@ -37,66 +38,77 @@ class ServiceMetrics extends React.Component<ServiceId, ServiceMetricsState> {
...
-  onOptionsChanged = (options: MetricsOptions) => this.fetchMetrics(options);
+  onOptionsChanged = (options: MetricsOptions) => {
+    if (this.state.pollMetrics !== undefined) {
+      clearTimeout(this.state.pollMetrics);
+    }
+    this.fetchMetrics(options, this);
+  };
 
-  fetchMetrics(options: MetricsOptions) {
+  fetchMetrics(options: MetricsOptions, object: any) {
```
jotak: Was the switch from `this` to `object` necessary? We could bind fetchMetrics to `this` in constructor, if it was the problem. (Or maybe writing it as arrow function would do the trick)

src/pages/ServiceDetails/ServiceMetrics.tsx
```diff
@@ -37,66 +38,77 @@ class ServiceMetrics extends React.Component<ServiceId, ServiceMetricsState> {
...
-          responseSizeOut: this.nameHistogram(
+          responseSizeOut: object.nameHistogram(
             metrics.histograms['response_size_out'],
             'Response size (bytes)',
             options.byLabelsOut
           )
         });
+        if (options.pollInterval !== undefined) {
+          if (options.pollInterval > 0) {
+            object.setState({ pollMetrics: setTimeout(object.fetchMetrics, options.pollInterval, options, object) });
```
jotak: `setInterval` is usually better suited for repeated tasks.
Also, the state of `pollInterval` shouldn't necessarily changes on `fetchMetrics`. So maybe it's better to set `pollInterval` from `onOptionsChanged` and clear and/or set interval from there and not from `fetchMetrics`

src/pages/ServiceDetails/ServiceMetrics.tsx
```diff
@@ -259,7 +268,8 @@ class ServiceMetrics extends React.Component<ServiceId, ServiceMetricsState> {
         }
       }
     };
-    return <LineChart id={id} title={{ text: title }} data={data} axis={axis} point={{ show: false }} />;
+    let obj = <LineChart id={id} title={{ text: title }} data={data} axis={axis} point={{ show: false }} />;
```
jotak: Is it intentional change? (I'm asking, because I'd rather write the shorter 1-line version, but if there are reasons against it then ok)

src/components/MetricsOptions/MetricsOptionsBar.tsx
```diff
@@ -152,7 +173,21 @@ export class MetricsOptionsBar extends React.Component<Props, MetricsOptionsStat
...
+              <MenuItem key={r[0]} active={r[0] === this.state.pollInterval} eventKey={r[0]}>
+                {r[1]}
+              </MenuItem>
+            ))}
+          </DropdownButton>
         </div>
+        <ToolbarRightContent>
+          {this.props.loading && (
+            <span>
+              <Icon name="spinner" spin={true} size="lg" /> Loading
```
jotak: It looks different from the patternfly spinner, no? Is it intentional?

src/types/MetricsOptions.ts
```diff
@@ -1,5 +1,6 @@
 interface MetricsOptions {
   rateInterval?: string;
+  pollInterval?: number;
```
jotak: IMO it shouldn't go in MetricsOptions, since this is what is sent to the server (and the server API doesn't have a `pollInterval` query param).
So, it could be a second parameter in MetricsOptionsBar `onOptionsChanged` props.

src/components/MetricsOptions/MetricsOptionsBar.tsx
aljesusg: Yes, render the icon spinner consume less memory. But for me there is not problem if you prefer Spimmer

src/pages/ServiceDetails/ServiceMetrics.tsx
aljesusg: No sorry , I was playing with bind the c3 chart ^^

src/pages/ServiceDetails/ServiceMetrics.tsx
aljesusg: Ok I fix this.

src/types/MetricsOptions.ts
aljesusg: But onOptionsChanged: (opts: MetricsOptions) => void; Is the type MetricsOptions I can't define pollInterval if I not define it in MetricOptions type.

src/components/MetricsOptions/MetricsOptionsBar.tsx
jotak: It's just a matter of being consistent with the other places we have a spinner. I haven't seen if they have been changed in the other places, but I think either we change it everywhere, or nowhere.

jotak: PS: maybe if the PF spinner is an issue it should be addressed globally through another task

aljesusg: mmm I agree with you, Should I keep Spinner while we discuss this with others?

src/types/MetricsOptions.ts
jotak: I'm not sure to understand, you can change the method signature in `MetricsOptionsBar.tsx` no?
Or an alternative, to really decorrelate API options versus UI options: pass another function as argument to `MetricsOptionsBar`, such as `onPollIntervalChanged`. That would be even better because there's no reason to fetch metrics immediately after changing the poll interval.

jotak:
@vnugent +1, my bad I recommended to use either `bind` or fat arrows but our guidelines are now to use always fat arrows instead of bind.

src/pages/ServiceDetails/ServiceMetrics.tsx
jotak: thanks :)

src/types/MetricsOptions.ts
aljesusg: Ok I understand now ^^

aljesusg:
@jotak @vnugent I did the changes

aljesusg:
Thanks @jotak

PR #126, opened 2018-03-16
Adds storybook for some components

josejulio:
- CytoscapeLayout
  - MetricOptions
  - Pf
  - ServiceFilter

peek 2018-04-03 00-14

> Storybook is a development environment for UI components. It allows you to browse a component library, view the different states of each component, and interactively develop and test components.
> Storybook runs outside of your app. This allows you to develop UI components in isolation, which can improve component reuse, testability, and development speed. You can build quickly without having to worry about application-specific dependencies.

storybook demo

https://github.com/storybooks/storybook

PR #211, opened 2018-04-04
after running "npm install" i'm getting this diff in the package-lock.json file.

jmazzitelli:
Do we need to commit this?


cfcosta:
Does the same diff gets generated with the latest npm?

jmazzitelli:
This occurs with both npm 5.6.0 and 5.8.0.

josejulio:
It seems that my `npm` removed that in #126, not sure why or how, but is adding it back when i run npm install. So maybe i forgot to run `npm install` after finishing?

PR #214, opened 2018-04-05
Kiali-486: Upgrade to React 16.3.1 from 16.2.0 

mtho11:
Also, upgrade other packages.

See: https://reactjs.org/blog/2018/03/29/react-v-16-3.html

**Please note lifecycle changes:https://reactjs.org/blog/2018/03/29/react-v-16-3.html#component-lifecycle-changes**

So the following lifecycle methods have been **deprecated** (and will be removed in React 17):

-  componentWillMount
- componentWillReceiveProps
- componentWillUpdate

This PR just addresses the `componentWillMount` because its an easy fix by replacing with `componentDidMount`. The others require different args and some methods are replaced by static ones so we'll hit that in a different PR as time allows.


mtho11:
This is good to merge now.

PR #215, opened 2018-04-06
KIALI-480 Global health indicator

jotak:
- Replace the donut with the global indicator (colored icon) in service details
- Replace global pod status with health in services list
- Aggregate health sources into a global health
- Unit tests

Related to https://github.com/kiali/kiali/pull/145

### DIFF
--- a/src/components/ServiceHealth/ServiceHealth.tsx
+++ b/src/components/ServiceHealth/ServiceHealth.tsx
@@ -17,7 +17,7 @@ interface Status {
   name: string;
   color: string;
   priority: number;
-  jsx: (size?: string) => JSX.Element;
+  jsx: (size: number, title: string) => JSX.Element;
 }
 
 // Colors from Patternfly status palette https://www.patternfly.org/styles/color-palette/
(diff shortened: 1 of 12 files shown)


jotak:
Service list:
Service list

Service details:
Service details

abonas:
commenting purely the ux - in the service details page above, the Health saying "failure/no active deployment" doesn't sound very clear. Meaning that the terminology of failure/active deployment and health somehow does not seem to sound clear together imo.

jotak:
@abonas do you mean that you wouldn't include the deployment status in health calculation?
(PS if you don't mind we can continue discussion on kiali-dev ml there's an open topic for health)

jshaughn:
@jotak, I like this enhanced health determination.  Perhaps 'failure' is unnecessary?  Maybe just 'no active deployment' in red. (or whatever the explanation).  Also, I personally would rather see the health maybe be more prominent.  Either 1st column or brought up next to the service name, although maybe that would affect sorting.

lucasponce:
> Perhaps 'failure' is unnecessary? Maybe just 'no active deployment' in red. (or whatever the explanation).

+1

The explanation messages are a great idea, perhaps just the color identifies the "failure" and it sounds redundant.

Pure UX comment, in the service list, having the same icon in health and error rate makes me wonder what are the differences between them.

I understand that one belongs to the pod health, and the other of the envoy health ?
But it is confusing, so, if the goal is to show some minimal info that "hey, this service has something wrong, you need to enter in the details to get more details" perhaps just having a single icon is enough ?.

I mean, I understand that the service list shouldn't enter in the details, just to mark which services have problems (and perhaps to order/filter in the future).

jotak:
@lucasponce that's true, about error rates actually I started my task in parallel with @aljesusg (I think) working on error rates and I wasn't aware this was going to be added next to it. We may want to merge them into a single indicator, because error rates is also a relevant source for computing health.

jotak:
I created a ticket to merge error rates and health, as it may also involve the backend code change: https://issues.jboss.org/browse/KIALI-505
So if you agree, for this PR we just add the health regardless of error rate

jotak:
Also, a suggestion for change in pod status under "deployment", as you can see on the second screenshot, there is a status icon next to the "0 / 0" text. It may be redundant or even contradictory with the Health, I suggest that we remove the icon so that any visual representation of healthiness is only done through Health. We keep however the x/y text: only facts, no interpretation there. WDYT?

jotak:
FYI here's another screenshot that show degraded mode. I removed the redundant label "Healthy/Degraded/Failure" under the icon, given the icon and color speak for itself.

Degraded details

pilhuhn:
Probably not related to this PR: the reviews-v1 has 0/0 pods, but shows green. Shouldn't that be flagged as well? And then the overall result even alert instead of warn?

xeviknal:
@jotak what about paint in red the pod status when expected pods doesn't match the running ones? (and remove the icon). If you guys like, I could apply the change in a separated JIRA. Thoughts?

jotak:
@xeviknal for me, the issue is not the icon itself, but the fact that there is some "business rule" used here to determine if it's OK or KO, which is different from the business rule used for health and thus could result in a contradictory signal (like in the above screenshot, 0/0 = green for deployments status versus red for health)

Another option could be to share the same rule of course.

xeviknal:
got it! IMO, I would follow same rule but keep some quick visual trick in which is the deployment failing.
Imagine you have more than 10 deployments, I think it would be great to have it identified by coloring.
Wearing a customer hat, I imagine I would start on service list, detect that one service has a degraded behavior. Right after, I would get inside service page, I would read that the degradation is because of a 'pod deployment degradation'. And I would like to see quickly which deployments has that problem.
Just my opinion ^^

jotak:
I updated Deployment status in this PR (it's related to health anyway).
It know reuse health's status checker, see screenshot:

deployment status

xeviknal:
@jotak looks good to me how the pod statuses are displayed now :dancing_men:

PR #216, opened 2018-04-06
KIALI-266: Display graph load errors in a notificacion

israel-hdez:
Sample when the app/cluster is down:
image

Sample if only the graph fails to load:
image


cfcosta:
Maybe this should be a component by itself? I like the idea of giving it an error and not have worry on how it will be rendered.

israel-hdez:
@cfcosta The alerts block is now a component.

PR #217, opened 2018-04-06
KIALI-491: Display rate interval for the service page

israel-hdez:
image

### DIFF
--- a/src/pages/ServiceList/ServiceListComponent.tsx
+++ b/src/pages/ServiceList/ServiceListComponent.tsx
@@ -308,6 +308,10 @@ class ServiceListComponent extends React.Component<ServiceListComponentProps, Se
     if (this.state.loading) {
       serviceListComponent = <div className="spinner spinner-sm left-spinner" />;
     } else {
+      let rateIntervalSelected = MetricsOptionsBar.RateIntervals.find(el => {
+        return el[0] === this.state.rateInterval;
+      });
+
       serviceListComponent = (
         <div>
           <NamespaceFilter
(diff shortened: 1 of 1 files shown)


abonas:
Something confuses me here (it might be completely my own misunderstanding and also unrelated to this particular PR)
1. Displaying rate interval means to me read only.
The above screenshot suggests that users *control* the rate interval.

2. the rate interval term itself - does this mean "refresh frequency of the UI vs server"? "refresh interval in which data is fetched by Kiali from Istio"? another third option? perhaps the term is not 100% accurate here

israel-hdez:
@abonas "Rate interval" is for the error rate percentage. If you choose "Last 10 minutes" the error rate for the last 10 minutes will be calculated and shown in the list. 

So, for example, if a service receives lots of requests but there were a high number of errors in the last few minutes, selecting "Last 30 minutes" may show a lower error rate compared to the error rate of the "Last 5 minutes".

Perhaps instead of "Rate interval" I could change the label to "Error rate for".... Or perhaps I'm misusing the _rate_ word.

cfcosta:
I think "Rate Interval" is ok enough, maybe "Error rate interval"?

jotak:
Technically, the "rate interval" is what is passed to prometheus `rate()` / `irate()` function, like the `5m`  described here: https://prometheus.io/docs/prometheus/latest/querying/functions/#rate(), which if I'm correct stands for "rolling rate computed over interval [N-5 minutes, N]". Since in this case I guess we fetch only the latest datapoint, it's really the last 5 minutes from now - but in the metrics screen it has to be distinguished with the observed duration which is in general bigger than the rate interval.

PR #221, opened 2018-04-09
KIALI-358 Wraps Switch into an ErrorBoundary object

josejulio:
This make possible to catch unhandled render errors without breaking
 the whole application, the user is still able to change pages and use
 the rest of the application.

peek 2018-04-09 17-53

Note: The flashy error screen is show only in development mode, while the "Something wrong happened" is show always.

### DIFF
--- a/src/components/Nav/Navigation.tsx
+++ b/src/components/Nav/Navigation.tsx
@@ -1,7 +1,9 @@
 import * as React from 'react';
 import { VerticalNav } from 'patternfly-react';
 import PropTypes from 'prop-types';
-import { Route, Switch, Redirect } from 'react-router-dom';
+import { Route, Redirect } from 'react-router-dom';
+
+import SwitchErrorBoundary from '../SwitchErrorBoundary/SwitchErrorBoundary';
 
 import IstioRulesPage from '../../pages/IstioRulesList/IstioRuleListPage';
 import IstioRuleDetailsPage from '../../pages/IstioRuleDetails/IstioRuleDetailsPage';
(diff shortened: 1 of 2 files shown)


src/components/Nav/Navigation.tsx
```diff
@@ -105,15 +107,21 @@ class Navigation extends React.Component<PropsType, StateType> {
           <VerticalNav.Item title={istioRulesTitle} iconClass="fa pficon-migration" onClick={this.navigateTo} />
           <VerticalNav.Item title={servicesJaeger} iconClass="fa fa-paw" onClick={this.navigateTo} />
         </VerticalNav>
-        <Switch>
+        <SwitchErrorBoundary
+          fallBackComponent={() => (
+            <div className="container-fluid container-pf-nav-pf-vertical">
```
mtho11: Please use `PfContainerNavVertical` instead

src/components/SwitchErrorBoundary/SwitchErrorBoundary.tsx
```diff
@@ -0,0 +1,48 @@
...
+    super(props);
+    this.state = { hasError: false };
+    this.show = false;
+  }
+
+  componentDidCatch(error: any, info: any) {
+    if (this.props.onError) {
+      this.props.onError(error, info);
+    }
+    this.setState({ hasError: true });
```
mtho11: This is great for now, but will be cool to see this integrated into redux.

src/components/Nav/Navigation.tsx
josejulio: Right, I always forget about that :), thanks.

src/components/SwitchErrorBoundary/SwitchErrorBoundary.tsx
josejulio: Yes! that's the idea :)

PR #222, opened 2018-04-10
KIALI-196 Unify error messages on services and rules pages

lucasponce:
- Add explanation of the error when possible.

This fixes a bug reported by QE related the confusion with the messages in the error handling.

### DIFF
--- a/src/pages/IstioRuleDetails/IstioRuleInfo.tsx
+++ b/src/pages/IstioRuleDetails/IstioRuleInfo.tsx
@@ -46,13 +46,9 @@ class IstioRuleInfo extends React.Component<RuleId, RuleInfoState> {
         });
       })
       .catch(error => {
-        let errorMessage = 'Could not fetch IstioRule details.';
-        if (error['response']['data'] && error['response']['data']['error']) {
-          errorMessage = errorMessage + ' Error: [ ' + error['response']['data']['error'] + ' ]';
-        }
         this.setState({
           error: true,
-          errorMessage: errorMessage
+          errorMessage: API.GetErrorMsg('Could not fetch IstioRule details.', error)
         });
       });
   }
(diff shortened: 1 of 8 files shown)


src/pages/IstioRulesList/IstioRuleListComponent.tsx
```diff
@@ -147,7 +147,11 @@ class IstioRuleListComponent extends React.Component<IstioRuleListComponentProps
         })
         .catch(namespacesError => {
           console.error(JSON.stringify(namespacesError));
-          this.handleError('Error fetching namespace list.');
+          let errorMessage = 'Could not fetch namespace list.';
+          if (namespacesError['response']['data'] && namespacesError['response']['data']['error']) {
+            errorMessage = errorMessage + ' Error: [ ' + namespacesError['response']['data']['error'] + ' ]';
+          }
```
cfcosta: This seems copy-pasted everywhere, can it be extracted somewhere? Something like `errorMessageFor('Could not fetch namespace list.')`.

src/pages/IstioRuleDetails/IstioRuleInfo.tsx
```diff
@@ -36,14 +36,25 @@ class IstioRuleInfo extends React.Component<RuleId, RuleInfoState> {
...
+      .then(response => {
+        let data = response['data'];
+        this.setState({
+          name: data.name,
+          match: data.match,
+          actions: data.actions
+        });
+      })
+      .catch(error => {
+        let errorMessage = 'Could not fetch IstioRule details.';
```
jotak: According to this page [1] `error` may not have a `response` nested field. Maybe we should have a shared function in API that transforms the axios error object in something more directly usable.

[1] https://github.com/axios/axios#handling-errors

mtho11:
Can we at least create a Jira to add this to other pages as well right?

PR #223, opened 2018-04-10
Kiali-268 Show Istio version in about modal

aljesusg:
version


src/services/Api.ts
```diff
@@ -66,6 +66,10 @@ export const getJaegerInfo = () => {
   return newRequest('get', `/api/jaeger`, {}, {});
 };
 
+export const GetIstioVersion = () => {
+  return newRequest('get', `/api/istio`, {}, {});
```
josejulio: I was thinking that `/api/status` could return us all these values (istio, prometheus, jaeger?) instead of separate calls.

https://github.com/aljesusg/kiali-ui/blob/605dcfd35478a4d9bf7fbe07388062dfb23a2a86/src/services/Api.ts#L34

PR #228, opened 2018-04-11
KIALI-506 Add Istio Sidecar flag on Service pages

lucasponce:
- Add filtering and sorting by Istio services

Look and feel is temporal, pretty soon, it will change when UX teams suggest better design, but to keep things simple we have added an "Istio" icon on the service list and service details pages.

image

image

Service list page can now filter and sort by Istio based services:

image

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -9,3 +9,22 @@ body {
 .left-spinner {
   margin: 10px 10px;
 }
+
+.IstioLogo {
+  height: 25px;
+}
+
+.ServiceList-Heading {
+  vertical-align: top;
+}
+
+.ServiceList-Title {
+  display: inline-block;
+}
+
+.ServiceList-IstioLogo {
+  display: inline-block;
+  position:relative;
+  top:-10px;
+  padding-right: 10px;
+}
--- a/src/index.css
+++ b/src/index.css
@@ -10,21 +10,3 @@ body {
   margin: 10px 10px;
 }
 
-.IstioLogo {
-  height: 25px;
-}
-
-.ServiceList-Heading {
-  vertical-align: top;
-}
-
-.ServiceList-Title {
-  display: inline-block;
-}
-
-.ServiceList-IstioLogo {
-  display: inline-block;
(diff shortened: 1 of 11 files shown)


lucasponce:
Mmm it seems I haven't rebased yet, I will update this PR once conflicts has been resolved.

src/index.css
```diff
@@ -9,3 +9,22 @@ body {
 .left-spinner {
   margin: 10px 10px;
 }
+
```
jotak: Isn't it supposed to go in component scss?

lucasponce: The .left-spinner was there before :-) and that makes me think to add it here as first place.
Yes, it is better to move to a close scss file at same component/page level.

lucasponce:
> One thing that I don't like, though, is that this is a shotgun-surgery (https://en.wikipedia.org/wiki/Shotgun_surgery), so maybe that's a sign that this needs to be encapsulated so that all istio-sidecar related stuff is on one class/component. If we do want to have that on the service graph in the future, then this is a must.

That can lead in a more deep conversation, but yes, it's right, in the whole code there are areas that can be candidates for a refactoring and optimization.

But not for this PR :-), as the goal is to bring these features ready for the Sprint to receive feedback from user perspective. Also with the latest discussions with UX team, things can change as soon as we receive new feedback, so I guess it's wise to introduce refactoring in an iterative way as soon as we consolidate things.

src/index.css
jotak: yup sorry, my comment was for lines _below_

jotak:
@cfcosta a principle I like to follow is to make reusable code only when there's at least two candidates for reuse, else stick to YAGNI recipe. Here there's two parts:
- the istio icon (few lines of code, almost no business code, it could be easily turned into a small component of its own if it's reused)
- the istio filter: it's more complex and may not easily be turned into a reusable component, so better here to wait to have 2 use cases - it may turn out that a shared component would bring far more complexity than not shared.

PR #229, opened 2018-04-11
KIALI-497 Show health details

jotak:
- Show details on modal window when a health indicator is clicked
- Rename ServiceHealth component => HealthIndicator
- Move some common functions to HealthHelper


jotak:
Screenshot:

Health details

lucasponce:
Instead a modal, which is something we have introduced now, why now just a tab with the details ?

Well, UX will say, but I tend to see modals quite strong in terms of usability, and perhaps the health info is not worth to use it on this scenario.

A pair of tabs, "Health" / "Health Details" sounds less aggressive.

Or just simpler, just add a new card under deployments with Health Details, that also works and we eliminate the modal. 

(I think).

jotak:
It could be just a tooltip as well, maybe less aggressive:

https://rawgit.com/patternfly/patternfly-react/gh-pages/index.html?knob-Accent=false&knob-Trigger=hover%20focus&knob-Popover%20Title=Popover&knob-Popover%20Content=%3Cstrong%3EHoly%20guacamole%21%3C%2Fstrong%3E%20Check%20this%20info.&knob-Justified=false&knob-Right%20aligned=false&knob-Tab%203%20Disabled=false&knob-Tab%202%20Disabled=false&knob-Tooltip=%3Cstrong%3EHoly%20guacamole%21%3C%2Fstrong%3E%20Check%20this%20info.&knob-Tab%201%20Disabled=false&knob-Placement=right&knob-Root%20Close=false&knob-Match%20Height=true&knob-Tab%203%20Tittle=Tab%203&knob-Tab%202%20Tittle=Tab%202&knob-Tab%201%20Tittle=Tab%201&selectedKind=Tooltip&selectedStory=Tooltip&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs

lucasponce:
Tooltip also works to me.

My rationale was to not "stop" user with a model for this. In terms of UX, a modal is strong and I guess it means "hey, stop here, and don't move until you complete something", so apart of the "about" version or similar, modals work better for mandatory steps in forms or things like that, so that is why it sounded weird in a first sight.

The tooltip perhaps is less obvious (as you need to hover) but doesn't "stop" user.

Also, a card is not a bad decision, as on this stage I guess goal is to show UX all info we want to offer to the user, and they will propose the better approach. A card, even in a overloaded page, is more useful than a tooltip that needs to know that there is an "hover" attached to an icon (I think).

cfcosta:
+1 on the tooltip.

jotak:
Here's the tooltip version:

tooltip

lucasponce:
LGTM as well

PR #231, opened 2018-04-11
KIALI-502 make sure summary panelresets when graph is refreshed

jshaughn:
note - also includes a fix for %error calc in summary panel rps chart

### DIFF
--- a/src/components/SummaryPanel/RpsChart.tsx
+++ b/src/components/SummaryPanel/RpsChart.tsx
@@ -51,7 +51,7 @@ export class RpsChart extends React.Component<RpsChartTypeProp, {}> {
       maxRps = sample > maxRps ? sample : maxRps;
       sumRps += +sample;
     }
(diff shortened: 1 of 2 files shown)


src/components/SummaryPanel/RpsChart.tsx
```diff
@@ -51,7 +51,7 @@ export class RpsChart extends React.Component<RpsChartTypeProp, {}> {
       maxRps = sample > maxRps ? sample : maxRps;
       sumRps += +sample;
     }
-    const avgRps = len === 0 ? 0 : sumRps / (len - 1);
+    const avgRps = len < 2 ? 0 : sumRps / (len - 1);
```
cfcosta: Minor, but I think this fix should be on a different commit so it can be referenced/queried if necessary.

jshaughn: Guilty as charged, I have a [bad] habit of including tweaks.  I will try to be better.

PR #232, opened 2018-04-11
minor - apply default export convention to summary panel components

jshaughn:
I was doing a little react learning and saw that we were inconsistent on the use of export default.  I'll let Mike decide if this makes sense.

### DIFF
--- a/src/components/SummaryPanel/InOutRateTable.tsx
+++ b/src/components/SummaryPanel/InOutRateTable.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { ErrorRatePieChart } from './ErrorRatePieChart';
+import ErrorRatePieChart from './ErrorRatePieChart';
 
 type InOutRateTablePropType = {
   title: string;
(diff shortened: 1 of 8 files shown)


src/components/SummaryPanel/InOutRateTable.tsx
```diff
@@ -13,7 +13,7 @@ type InOutRateTablePropType = {
   outRate5xx: number;
 };
 
-export class InOutRateTable extends React.Component<InOutRateTablePropType, {}> {
+export default class InOutRateTable extends React.Component<InOutRateTablePropType, {}> {
```
mtho11: I **do** like this for most use cases because it's nice and concise. 

However, when you decorate a class with a high order function the other method is needed:

`const FilterLink = connect(
  mapStateToProps,
  mapDispatchToProps
)(Link)

export default FilterLink`

And with _Redux_ now we will be doing more High Order functions like _connect_.

jshaughn: Not sure what this means, I'll follow up with you tomorrow...

PR #234, opened 2018-04-11
KIALI-500 don't require graph click before showing full graph summary

jshaughn:
A little trickery needed to avoid infinite rendering loop, but I think
this does the job.

### DIFF
--- a/src/pages/ServiceGraph/ServiceGraphPage.tsx
+++ b/src/pages/ServiceGraph/ServiceGraphPage.tsx
@@ -32,6 +32,7 @@ type ServiceGraphPageState = {
   graphTimestamp: string;
   graphData: any;
   isLoading: boolean;
+  isReady: boolean;
   params: GraphParamsType;
 };
 
@@ -56,6 +57,7 @@ export default class ServiceGraphPage extends React.Component<
     const { graphDuration, graphLayout } = this.parseProps(routeProps.location.search);
     this.state = {
       isLoading: false,
+      isReady: false,
       alertVisible: false,
       alertDetails: {},
       summaryData: { summaryType: 'graph' },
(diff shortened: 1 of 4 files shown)


src/pages/ServiceGraph/ServiceGraphPage.tsx
```diff
@@ -124,6 +126,12 @@ export default class ServiceGraphPage extends React.Component<
     }
   };
 
+  handleReady = (cy: any) => {
+    if (cy !== undefined) {
```
mtho11: Remember our friend **falsey**?
Falsey gives us a cleaner syntax of `if(!cy) {`

src/pages/ServiceGraph/SummaryPanelGraph.tsx
```diff
@@ -49,28 +49,28 @@ export default class SummaryPanelGraph extends React.Component<SummaryPanelPropT
...
     };
-    if (cy) {
-      cy.edges().forEach(edge => {
-        rate += +safeRate(edge.data('rate'));
-        rate3xx += +safeRate(edge.data('rate3XX'));
-        rate4xx += +safeRate(edge.data('rate4XX'));
-        rate5xx += +safeRate(edge.data('rate5XX'));
-      });
-    }
+    cy.edges().forEach(edge => {
```
mtho11: Try not to use `forEach` in the future, it forces you to use mutable data structures. Most of the time `map` or `reduce` is a better option because its immutable. With React/Redux now, immutability is a big issue.

src/pages/ServiceGraph/ServiceGraphPage.tsx
cfcosta: @mtho11 falsey?

jshaughn: d'oh, right, should just be:   if (cy)

src/pages/ServiceGraph/SummaryPanelGraph.tsx
jshaughn: OK.  This was existing code, but I was the original author.  I'll look at the alternatives.

PR #238, opened 2018-04-12
Fix warnings and uncaught errors in tests

jotak:
- Add a unique 'id' in Tooltips
- Mock API.getNamespaceMetrics
- Had to rewrite a part of HealthDetails.tsx due to a prettier versus tslint conflict

### DIFF
--- a/src/components/ServiceHealth/HealthDetails.tsx
+++ b/src/components/ServiceHealth/HealthDetails.tsx
@@ -5,6 +5,7 @@ import { Health } from '../../types/Health';
 import * as H from './HealthHelper';
 
 interface Props {
+  id: string;
   health: Health;
   headline: string;
   placement?: string;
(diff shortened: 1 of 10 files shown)


jotak:
Note, the mentioned prettier versus tslint conflict is due to prettier formatting blocs of code with new lines after `{` whereas tslint asks (probably just in JSX) that there are no space after `{`.

It's not the first time prettier is bugging me. Ok to use formatting tools as long as they're not broken :-/

josejulio:
> It's not the first time prettier is bugging me. Ok to use formatting tools as long as they're not broken :-/

I don't think they are broken, but not compatible on all their rules. Can you configure either one to match the other (for this case) ?

https://prettier.io/docs/en/configuration.html
https://palantir.github.io/tslint/usage/configuration/

jotak:
@josejulio haven't gone through that painful process yet. Maybe next time.

josejulio:
> @josejulio haven't gone through that painful process yet. Maybe next time.
> Note that prettier config is quite basic and generally it's turning on or off a feature, not fine-tuning how it works.

Seems someone already did that: https://www.npmjs.com/package/tslint-config-prettier

PR #239, opened 2018-04-12
KIALI-472 use popper badges to mark nodes that have circuit breakers

jmazzitelli:
This is what the circuit breaker badging will look like and how it behaves:

badging

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4711,6 +4711,14 @@
         "klayjs": "0.4.1"
       }
     },
+    "cytoscape-popper": {
+      "version": "1.0.1",
+      "resolved": "https://registry.npmjs.org/cytoscape-popper/-/cytoscape-popper-1.0.1.tgz",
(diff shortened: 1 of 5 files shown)


jmazzitelli:
do not merge this yet. Once @cfcosta gets his server-side work done, I will add the necessary code to put the badges where they really need to do (see the TODO in the code where that new code will go).

jmazzitelli:
this PR now should be ready for whatever implementation is on the server side.  So long as the node in the JSON graph has an attribute "hasCircuitBreaker" whose value is "true" the badge will show up

lucasponce:
Minor comment. The CB and other potential elements, like load balancers and route rules I guess it is better added at the box in the groupes services instead per version as it shows in the UI.

But not a blocker, I think that is UXD arena and probably we need more formal feedback about how is the best way to present visually this information.

PR #243, opened 2018-04-12
[KIALI-516] Use Redux-Namespace dropdown, part 2 of 2

vnugent:
[x] Introduce Redux, reducers and actions for managing namespace list
[x] `AutoUpdateNamespaceList` container component reloads the list on every click
[x] Save graph params to session [localStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage) 
[x] extract URL-related props to a new component, `ServiceGraphRouteHandler` to make it easier to test.  `ServiceGraphPage` now requires namespace, duration, layout as props
[x] unit test change handlers

Todo:
[ ] USe stricter types for actions and reducers KIALI-588

@mtho11 I'm probably duplicating a bit of what you're doing.  I simply picked redux-thunk instead of saga because that's what the tutorial shows.

### DIFF
--- a/src/components/GraphFilter/GraphFilter.tsx
+++ b/src/components/GraphFilter/GraphFilter.tsx
@@ -1,48 +1,31 @@
 import * as React from 'react';
-import { Toolbar, Button, ButtonGroup, DropdownButton, MenuItem, Switch, Icon } from 'patternfly-react';
+import { Toolbar, Button, ButtonGroup, Switch, Icon } from 'patternfly-react';
 
 import { GraphFilterProps, GraphFilterState } from '../../types/GraphFilter';
-import * as API from '../../services/Api';
 import { ToolbarDropdown } from '../ToolbarDropdown/ToolbarDropdown';
+import AutoUpdateNamespaceList from '../../containers/AutoUpdateNamespaceList';
 
 export default class GraphFilter extends React.Component<GraphFilterProps, GraphFilterState> {
   constructor(props: GraphFilterProps) {
     super(props);
-    this.state = {
-      availableNamespaces: []
-    };
   }
-
-  componentDidMount() {
-    // TODO: [KIALI-436] API.getNamespaces() is also called in Services component.
-    // We should consolidate them into one.
-    API.getNamespaces()
-      .then(this.setNamespaces)
-      .catch(error => {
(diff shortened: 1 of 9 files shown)


vnugent:
What do you all think about a global namespace dropdown selector?  This structure is in-line with OpenShift's,  <root> -> namespace -> overview | deployment | pods ...

kiali-dropdown-global

vnugent:
@jotak we talked about the screen mock up at the call today and the decision was to keep the dropdown where it is.

jotak:
@vnugent ok, then I guess all namespace dropdowns in various screens will be "connected" via redux?

vnugent:
Redux doesn't prescribe where you should connect the store to components.  The recommended practice I saw from their docs is to have a presentation component that is pure (in this case, a dropdown taking a list of namespaces as prop) and a separate 'container' component that connects the redux store to the presentation component. 
`AutoUpdateNamespaceList` is the container one that can be reused throughout the app   https://github.com/vnugent/kiali-ui/blob/85898231f074fa4452fd97e33dcdb674814ad02a/src/containers/AutoUpdateNamespaceList.tsx

vnugent:
new screen cast
kiali-516

mtho11:
My concern is this PR is getting Big. 18 files and 400+ lines.  The rebasing alone has got to be annoying. Any way to break this up, and get some smaller chunks into the codebase? It's good stuff.
_I know the pain._

mtho11:
Yes! I have been struggling with Sagas(generators), Thunk, Promise, Observer!
All so very different.

vnugent:
Majority of the new additions are new files for Redux reducers, actions, etc. Main change to the current code base is pulling url params handling logic to its own component. Let me see if I can break it up

mtho11:
Needs rebasing @vnugent

vnugent:
kiali-servicegraph 1

vnugent:
Could someone please review and merge.  I've been rebasing this PR 3x :(

mtho11:
@vnugent needs another rebase.
Don't feel bad I'm having to do the same thing ;)

mtho11:
@vnugent can you change your commit message to have Jira number and descriiption instead of 'rebase 4'

PR #244, opened 2018-04-12
KIALI-520 replace 'Refresh' button

jshaughn:
- replace text button with standard refresh icon button
- move button to options bar and place to far right
- as an added bonus, protect from button clicks while graph is refreshing


mtho11:
Can you add a before/after picture since this is very visual

jshaughn:
@mtho11 , sure, sorry:

Before:
screen-before

After:
screen-after

PR #248, opened 2018-04-13
R&D: View source of route rules

pilhuhn:
This is the counterpart of https://github.com/kiali/kiali/pull/155 to display routing rules and then optionally fire the ACE editor component to get a source view of the routing rule.

This is a synthetic rule generated via the above PR. It only shows 2 destinations for the reviews destination, as the v1 has been detected, but did not yet receive traffic.

bildschirmfoto 2018-04-13 um 10 58 24

Fired the viewer (ACE editor) 
bildschirmfoto 2018-04-13 um 10 58 36

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3436,6 +3436,11 @@
         }
       }
     },
+    "brace": {
+      "version": "0.11.1",
+      "resolved": "https://registry.npmjs.org/brace/-/brace-0.11.1.tgz",
+      "integrity": "sha1-SJb8ydVE7vRfS7dmDbMg07N5/lg="
+    },
     "brace-expansion": {
       "version": "1.1.8",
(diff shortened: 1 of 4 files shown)


lucasponce:
This is interesting.
Basically this approach is taking the source services info from a service, and convert them into a "synthetic" route rule, so at the end, the user will have all routing info from a single area.

I imagine as long as the source services info changes, these synthetic (or "infered" rules) will change as well.

I have mixed feelings, 50/50 I think, I like the concept to infered some "runtime" rules, and then eliminate the soure services tab, as at end, there is some kind of duplication then. And in the other side, I like to see separated those "static vs runtime" concepts.

I don't know, but the ACE editor is a good component, I guess I was not aware of that earlier, it would improved a lot those boring textareas we have in other pages.

lucasponce:
@pilhuhn I'm going to close the PR I think is outdated and also the counterpart was closed.

PR #249, opened 2018-04-13
Use window.setInterval and not the node one.

josejulio:
We are running it on a browser, in the end it will use that.
We are wrongly type-hinting that we are using the other one and could cause
problems if someone tries to access the `NodeJS.Timer` object created.

I tested and it returns a `number`, not a `NodeJS.Timer`, calling `ref()` on it
causes an error to be throw.

### DIFF
--- a/src/pages/ServiceDetails/ServiceMetrics.tsx
+++ b/src/pages/ServiceDetails/ServiceMetrics.tsx
@@ -29,7 +29,7 @@ type ServiceMetricsState = {
   responseSizeOut?: M.Histogram;
   grafanaLinkIn?: string;
   grafanaLinkOut?: string;
-  pollMetrics?: NodeJS.Timer;
+  pollMetrics?: number;
 };
 
 class ServiceMetrics extends React.Component<ServiceId, ServiceMetricsState> {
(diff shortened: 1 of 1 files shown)


src/pages/ServiceDetails/ServiceMetrics.tsx
```diff
@@ -59,12 +59,12 @@ class ServiceMetrics extends React.Component<ServiceId, ServiceMetricsState> {
   };
 
   onPollIntervalChanged = (pollInterval: number) => {
-    let newRefInterval: NodeJS.Timer | undefined = undefined;
+    let newRefInterval: number | undefined = undefined;
     if (this.state.pollMetrics) {
       clearInterval(this.state.pollMetrics);
```
jotak: maybe here also `window.clearInterval` would avoid confusion, or is it not necessary?

jotak:
Good catch, thanks @josejulio 
I was misled by my IDE which told me it was a nodejs object, I took that for an internal typescript detail.
Is there a way to tell typescript what our runtime context is?

src/pages/ServiceDetails/ServiceMetrics.tsx
josejulio: There wasn't any warning or error with that, i'll double check and maybe it wouldn't hurt to call it like that.

josejulio:
> Is there a way to tell typescript what our runtime context is?

AFAIK it infers the context based on our dependencies. I'll try to research a bit more on this subject.

PR #251, opened 2018-04-13
KIALI-564 - Process all rates in one loop, calling cy.edges() only once.

josejulio:
This is an answer to  @jmazzitelli concern about looping 4x more than the original code.


mtho11:
@josejulio please rebase

josejulio:
> @josejulio this should have been tracked in a jira for sprint 4, we ended up making fixes in the same place. Your code is probably better, but we need to avoid this sort of overlap.

Sorry, just created the [jira](https://issues.jboss.org/browse/KIALI-564) and updated my changes.

PR #259, opened 2018-04-17
[KIALI-567] Hide Service Detail Information if not have a Istio enabled or no data

aljesusg:
[KIALI-567](https://issues.jboss.org/browse/KIALI-567)

Auto scale columns.

## Istio-System Ingress Before (No istio-sidecar deployed)
istio-system_ingress_before

## Istio-System Ingress After (No istio-sidecar deployed)
istio_ingress_after

## Router Default Before (No istio-sidecar deployed)
service_router_details_before

## Router Default After (No istio-sidecar deployed)
default_router_after

## Istio-System Details Before (istio-sidecar deployed)
istio-system_details_before

## Istio-System Details After (istio-sidecar deployed)
after

### DIFF
--- a/src/services/__mocks__/Api.ts
+++ b/src/services/__mocks__/Api.ts
@@ -1,4 +1,5 @@
 import * as GraphData from '../__mockData__/getGraphElements';
+import { AxiosError } from 'axios';
 
 const fs = require('fs');
 
@@ -123,4 +124,12 @@ export const getNamespaceMetrics = (namespace: String, params: any) => {
       }
     }
   });
+};
+
+export const GetErrorMsg = (msg: string, error: AxiosError) => {
+  let errorMessage = msg;
+  if (error && error.response && error.response.data && error.response.data['error']) {
+    errorMessage = `${msg} Error: [ ${error.response.data['error']} ]`;
+  }
+  return errorMessage;
 };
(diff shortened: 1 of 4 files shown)


pilhuhn:
I like that, but I think showing health as n/a is not correct.
We probably need to re-think the definition of health and perhaps
in this case fall back to what openshift has (e.g. enough pods deployed)

src/services/__mocks__/Api.ts
```diff
@@ -124,3 +125,11 @@ export const getNamespaceMetrics = (namespace: String, params: any) => {
     }
   });
 };
+
+export const GetErrorMsg = (msg: string, error: AxiosError) => {
```
xeviknal: Is that connected to the Pull Request? Perhaps commited by error? Otherwise, what is it for?

aljesusg: Because the ServiceInfo in some cases when you test the catch fail because this methods is not in the mock.

aljesusg:
I am gonna assign this to @abonas to discuss this @xeviknal

aljesusg:
Sorry I wanted to assign this to @serenamarie125

jotak:
@pilhuhn about health (btw, note that it's unrelated to this PR) :
> perhaps in this case fall back to what openshift has (e.g. enough pods deployed)

This is already what is done, but here as you can see there is no information about the deployments, it's not just health. So we are really missing the information. Sometimes we are unable to match a service with the related deployments. @xeviknal mentioned this in a mail last week (kiali-dev: "Issue grouping deployments for a service")

PR #262, opened 2018-04-17
[KIALI-474] Add button to toggle Circuit Breakers on Graph

cfcosta:
The button is shown on the right side of the filter bar, like this:

It can also be defined on the URL, for repeatable graph generation.

### DIFF
--- a/src/components/CytoscapeLayout/CytoscapeLayout.tsx
+++ b/src/components/CytoscapeLayout/CytoscapeLayout.tsx
@@ -100,6 +100,10 @@ export default class CytoscapeLayout extends React.Component<CytoscapeLayoutProp
 
     // when the graph is fully populated and ready, we need to add appropriate badges to the nodes
     this.cy.ready((evt: any) => {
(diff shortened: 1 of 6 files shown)


pilhuhn:
On 18 Apr 2018, at 1:19, Josejulio Martínez wrote:

> I think it should be "Hide Circuit Breakers" / "Show Circuit Breakers" 
> or something like that to show what is actually going to do 
> (Show/Hide).

+3 Especially as "Toggle" could also mean to (de)activate the the CBs in 
the real istio-RRs.

jshaughn:
My 2 cents, this seems like a checkbox thing to me, just a checkbox labeled with 'circuit breakers'.

mtho11:
Looks like it should be a toggle button. I think we are going to have a whole row of these _visibility layer toggles._
toggle

cfcosta:
Fixed failing tests, and added the filter on a new row:

mtho11:
Remember to squash the commits

PR #263, opened 2018-04-18
KIALI-582 remove presentation concerns from graph json

jshaughn:
Remove use of json-supplied text, style and color fields to determine
node and edge presention.
A couple if notes for a couple of tabgential changes:
- updated edge fontsize from 9px to 7px
- updated edge text-margin-x from 10px to 6px
- dropped service name from grouped service nodes
- alphabetized node and edge settings for clarity (well, not diff clarity...)

### DIFF
--- a/src/components/CytoscapeLayout/CytoscapeLayout.tsx
+++ b/src/components/CytoscapeLayout/CytoscapeLayout.tsx
@@ -106,7 +106,7 @@ export default class CytoscapeLayout extends React.Component<CytoscapeLayoutProp
           if (ele.isParent()) {
             return false; // we never need the group box elements, filter them out
           }
-          return ele.data('hasCircuitBreaker') === 'true';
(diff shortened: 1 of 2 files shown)


jshaughn:
No real changes to the look of things other than those mentioned in the description.  But this is not using any style/color/text from the graph json.

screen

jshaughn:
Oh, I just realized it also wraps the versions...  Anyway, let me know if you disagree with any of the tangential changes (other than the version-only nodes, because that is a UX request).

mtho11:
Can you paste the json in here?

jshaughn:
@mtho11 Here is json that roughly resembles that used with the graph image above:

[screen-json.txt](https://github.com/kiali/kiali-ui/files/1925774/screen-json.txt)

jshaughn:
@mtho11 Thanks for the review, I'll consider maybe changing the field names going forward but for now I liked the common prefix for the appended fields.

PR #92, opened 2018-03-06
SWS-258 Separate ServiceDetailInfo components

aljesusg:
Related with  : [SWS-69](https://issues.jboss.org/browse/SWS-69)
Jira: [SWS-258](https://issues.jboss.org/browse/SWS-258) Subtask of [SWS-186](https://issues.jboss.org/browse/SWS-186) 

reviews

PR #264, opened 2018-04-19
KIALI-591 remove duplicate fetching from ServiceInfo

jotak:
@aljesusg @xeviknal @lucasponce do you know if there was a good reason to have this "fetch" both from componentDidMount and componentWillReceiveProps ?

It makes 2 calls to the rest api where 1 would be enough.
I've tried and it seems to work properly, but maybe I missed some cases?

### DIFF
--- a/src/pages/ServiceDetails/ServiceInfo.tsx
+++ b/src/pages/ServiceDetails/ServiceInfo.tsx
@@ -47,10 +47,6 @@ class ServiceInfo extends React.Component<ServiceId, ServiceInfoState> {
     };
   }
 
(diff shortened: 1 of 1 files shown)


xeviknal:
No, I don't. It looks that this was introduced here: https://github.com/kiali/kiali-ui/pull/92 so @aljesusg probably can tell.

aljesusg:
Yes, we can remove the fetch, before the upgrade of libraries the service info didn't update the info but now it should work with only willreceiveprops.

Thanks @jotak

PR #265, opened 2018-04-19
KIALI-505 Add error rate to health

jotak:
- Need to add rateInterval information to health tooltip
- ServiceErrorRate now doesn't display an icon, but the the ratio errors/total
- Fix some CSS on service list health icon placement
- Sorting in service list: needed to change comparator functions so that it can compare other than just ServiceItem fields (here: health)
- Make ServiceItem derived from ServiceOverview (ServiceOverview renamed so that it matches server-side model)

PS for reviewers: I did a different commit for snapshots to make it easier to review

Core counterpart: https://github.com/kiali/kiali/pull/165


src/components/ServiceHealth/HealthDetails.tsx
```diff
@@ -36,28 +38,39 @@ export class HealthDetails extends React.PureComponent<Props, {}> {
...
                   </li>
                 );
               })}
             </ul>
             <strong>
-              {this.renderStatus(H.ratioCheck(health.envoy.healthy, health.envoy.total))}&nbsp; Envoy health
+              {this.renderStatus(H.ratioCheck(health.envoy.healthy, health.envoy.total))}
+              {' Envoy Health:'}
+            </strong>
+            {' ' + health.envoy.healthy + ' / ' + health.envoy.total}
```
lucasponce: Just a refresh to confirm, envoy healthy gets a "healthy / non healthy" status from the envoy proxy and this is a ratio of all proxies status for a given service.

jotak: yes, or in terms of pods , normally envoy.total == total number of running deployment pods in a service

src/pages/ServiceList/ServiceListComponent.tsx
```diff
@@ -7,39 +7,61 @@ import { ActiveFilter, FilterType } from '../../types/NamespaceFilter';
...
+import { getRequestErrorsRatio } from '../../components/ServiceHealth/HealthHelper';
 import { HealthIndicator, DisplayMode } from '../../components/ServiceHealth/HealthIndicator';
 import ServiceErrorRate from './ServiceErrorRate';
-import ServiceItemComparer from './ServiceItemComparer';
 import RateIntervalToolbarItem from './RateIntervalToolbarItem';
 
 import './ServiceListComponent.css';
 
-const sortFields: SortField[] = [
+export const sortFields: SortField[] = [
```
lucasponce: I don't like to have mixed criteria for export functions.
I think this object should be uppercase, Alberto started that and I don't see bad to follow that rule.
Having lowercase and uppercase exports in methods are confusing.
I would prefer to use a single criteria, and just to follow the preliminar, majority one.

jotak: sure .. actually this is only exported to be accessible for test, let me see if I can change the test and not export it at all

jotak: I wonder if we should start to use something to test non-exported stuff (like "rewire", see https://stackoverflow.com/questions/14874208/how-to-access-and-test-an-internal-non-exports-function-in-a-node-js-module )

It's a shame we have to export functions in order to test them. What do you think?

cc @mtho11 / @vnugent

jotak: @lucasponce yes I fully understand your point (+100 on having unified rules, I already asked for this in the past but it didn't got much attention) and I agree not to leave my change as is.

But I'm stepping to another point about testing non-exported stuff because it's also the subject here. I would actually prefer not to export things that are not meant to be accessed from other modules. I've just discovered rewire now, I also see here are [other suggestions](https://stackoverflow.com/a/32071275/3697695) for similar problem.

jotak: By the way, in API.ts I don't really understand why it's upper camel case. It's the only place where we have it. Just have a search of "export const" in the project: almost everything is low camel case, which is the most common in js.

lucasponce: That's the point I tried to spot.

I guess the rule should be:
- Let's follow what it is in the code, as someone introduced first.
- If a style should be changed, some discussion, and a global change.
- But to introduce a partial change is not good, as it grows two different styles, and it creates confusion.
- If there is no clear consensus, just the first style should prevail.

I don't have a strong opinion on style, personally I basically imitate what there is in the code, but we are in a point that code grows without a clear control.

jshaughn: If there is something done in non-test code that is there specifically as a test-hook I recommend adding a comment saying, "blah blah because it is a test hook".

jotak: Ok, I added a comment

jotak:
@serenamarie125 my concern with click is that people may not notice the icon is clickable, unless we change it completely to display it as a button (which IMO would not be very pretty). 
Is there other alternatives?

I'm playing with `tabindex` to make it triggered from keyboard (it works pretty well on service details page but for some reason it's breaking the tab iteration in service list page) - do you think it would be ok in terms of UX & accessibility?

jotak:
@serenamarie125 I've created a JIRA for that: https://issues.jboss.org/browse/KIALI-606
I'd prefer not to block this PR for it - it doesn't introduce any new tooltip anyway, they were already there. Moreover current kiali state is in regression until we merge it since the backend PR kiali/kiali#165 is already merged.

jotak:
Just noticed that `popover` could be used with `hover` too ... it's pretty much the same actually, except about style. So here's an update with popover instead of tooltip:

popover

abonas:
@serenamarie125 please see @jotak comment from above - is this ok to proceed merging this and take care of accessibility in the follow up PR or you request this one to be blocked and handle it here?

PR #266, opened 2018-04-19
KIALI-595 fix service count in graph summary

jshaughn:
- remove ingress
- fix regression wrt removal of composite nodes

### DIFF
--- a/src/components/CytoscapeLayout/graphs/GraphStyles.ts
+++ b/src/components/CytoscapeLayout/graphs/GraphStyles.ts
@@ -21,7 +21,7 @@ export class GraphStyles {
           'background-color': '#f9d67a', // pf-gold-200
           'border-color': '#030303', // pf-black
           'border-style': (ele: any) => {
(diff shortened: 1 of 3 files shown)


src/components/CytoscapeLayout/graphs/GraphStyles.ts
```diff
@@ -41,6 +41,12 @@ export class GraphStyles {
           'border-color': '#0088ce' // pf-blue
         }
       },
+      {
+        selector: 'node[isRoot]',
+        style: {
+          shape: 'diamond'
```
jmazzitelli: this introduces a different shape for a node? I'm not convinced I should be the one saying OK to this.

jshaughn:
@jmazzitelli Yes, this introduces diamonds for nodes outside the requested namespace (or namespaces in the future).  It looks like this:

screen-diamond

@pilhuhn is +1 on the change so I think it's OK.  Everything is subject to change...

PR #267, opened 2018-04-19
Introduce Redux and friends for namespace dropdown - Part 1

vnugent:
As discussed in https://github.com/kiali/kiali-ui/pull/243 this PR only contains Redux-related additions without any changes to existing features.

### DIFF
--- a/src/actions/NamespaceAction.ts
+++ b/src/actions/NamespaceAction.ts
@@ -1,0 +1,64 @@
+import * as Api from '../services/Api';
+
+export enum NamespaceActionType {
+  RELOAD = 'reload',
+  API_INITIATE_REQUEST = 'api_initiate_request',
+  API_RECEIVE_LIST = 'api_receive_list'
+}
+
+export const reload = () => {
+  return {
+    type: NamespaceActionType.RELOAD
+  };
+};
+
+export const apiInitiateRequest = () => {
+  return {
+    type: NamespaceActionType.API_INITIATE_REQUEST
+  };
+};
+
+export const apiReceiveList = newList => {
+  return {
+    type: NamespaceActionType.API_RECEIVE_LIST,
+    list: newList,
+    receivedAt: Date.now()
+  };
+};
+
+export const asyncFetchNamespaces = () => {
+  return dispatch => {
+    dispatch(apiInitiateRequest());
+    return Api.GetNamespaces()
+      .then(response => response['data'])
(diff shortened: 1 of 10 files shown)


src/actions/NamespaceAction.ts
```diff
@@ -0,0 +1,64 @@
...
+export const asyncFetchNamespaces = () => {
+  return dispatch => {
+    dispatch(apiInitiateRequest());
+    return Api.GetNamespaces()
+      .then(response => response['data'])
+      .then(data => dispatch(apiReceiveList(data)));
+  };
+};
+
+const shouldFetchNamespaces = state => {
```
jmazzitelli: what  happens if a new namespace is added in OpenShift/k8s after the initial namespace fetch is done?  Am I reading this right or do the namespaces only get fetched once and never more? Thus if a new namespace is added after, it will never be seen.

vnugent: whenever the user clicks on the dropdown the namespace will get reloaded from the backend.  In the future we can have a separate timer-driven refreshing logic.

src/actions/NamespaceAction.ts
```diff
@@ -0,0 +1,64 @@
...
+  // a cached value is already available.
+  return (dispatch, getState) => {
+    console.log('fetchNamespacesIfNeeded()');
+    if (shouldFetchNamespaces(getState())) {
+      console.log('begin async fetching namespaces');
+      // Dispatch a thunk from thunk!
+      return dispatch(asyncFetchNamespaces());
+    } else {
+      // Let the calling code know there's nothing to wait for.
+      return Promise.resolve();
```
josejulio: It will return a resolved promise if we are fetching?
I suppose this function won't be used to actually get the namespaces, right?

vnugent: Yes it will return the resolved promise.  `fetchNamespacesIfNeeded()` action is mapped to the `refresh` callback:

https://github.com/vnugent/kiali-ui/blob/aeaf8358e654398a67c43e87cfad6572a9f5789b/src/containers/AutoUpdateNamespaceList.tsx#L14

As you see in the NamespaceDropdown.componentDidMount() and onSelect(), when call this.props.refresh(), Redux dispatches `fetchNamespacesIfNeeded()`

https://github.com/vnugent/kiali-ui/blob/aeaf8358e654398a67c43e87cfad6572a9f5789b/src/components/NamespaceDropdown.tsx#L19

PR #268, opened 2018-04-20
KIALI-592: Show circuit breakers when graph node is selected

israel-hdez:
**Don't review yet.**

image

### DIFF
--- a/src/pages/ServiceGraph/SummaryPanelNode.tsx
+++ b/src/pages/ServiceGraph/SummaryPanelNode.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import * as API from '../../services/Api';
+import * as pf from 'patternfly-react';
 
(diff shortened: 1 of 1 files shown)


mwringe:
@lucasponce Ideally what we want to do is show the circuit breaker in the graph and if the circuit breaker has been activated during the requested time period then mark it as being red (like how we handle things with errors in the connections).

A user would then be able to click on the service and see in the side panel when the circuit breaker had been activated. With a link to the service detail page to get more information.

This works similar to how we handle things like requests with graphs on the side panel.

I don't know if we want to put the same information about the circuit breaker in the side panel as what we have in the detail page, but I think we would want at least the name, graph and link to the service detail page for more information.

israel-hdez:
@lucasponce @mwringe Yes, data is duplicated. So, what I can do is just show this:

image

Then, the user can use the header link to go to the details page and check the config.

PR #269, opened 2018-04-20
[Cleaning] unify camel-case naming

jotak:
Use low camel case for functions.
Also remove debugging console logs


aljesusg:
I saw late the chat...I prefer start with lowercase like you did. I added some methods with Uppercase because I saw the method with uppercase in some view but after I added getJaegerInfo in lowercase ^^. So For me it is ok after travis.

mtho11:
@jotak that is funny, I was actually going to do this today ;)

PR #274, opened 2018-04-23
KIALI-601 Simplify metrics options

jotak:
- Remove Ticks and Rate interval
- Compute them automatically
- In the same time, fix bug where charts were not correctly refreshed when grouping options are used


jotak:
@jshaughn @jmazzitelli can you review please, especially the second commit "KIALI-608 Use shared rate interval calculation for graph panels" which impacts the graph.

The function `computePrometheusQueryInterval` does the same as what was statically computed for the durations, ie. divide duration by desired number of datapoints.

src/components/GraphFilter/DurationDropdown.tsx
```diff
@@ -13,6 +13,7 @@ type DurationDropdownState = {
 };
 
 export class DurationDropdown extends React.Component<DurationDropdownProps, DurationDropdownState> {
+  // TODO: use enum for type-safety and to make sure we always initialize this component with a valid value
```
jshaughn: There is a jira for this already, can you update the comment to:
`// TODO: [KIALI-437] use enum for type-safety and to make sure we always initialize this component with a valid value`

src/components/GraphFilter/QueryOptions.ts
```diff
@@ -1,24 +0,0 @@
-// TODO [KIALI-437] Use enums instead of string keys
```
jshaughn: I'm glad to see this file go away.

jotak:
@vnugent but, unless the duration is a global setting that affects all the pages, I'm not sure the default has to be the same everywhere. So to me it makes sense to declare it wherever it's used.

PR #275, opened 2018-04-23
[KIALI-607] Add labels and state to dropdowns in Graph toolbar

aljesusg:
Added labels, state and removed react-bootstrap styles by patternfly-styles

## Before

before

## After

after


lucasponce:
And the dotted style for unused nodes and edges ?

aljesusg:
I'm confused with your question @lucasponce I didn't make any changes in nodes and edges styles.

lucasponce:
Nevermind, sorry, I saw the first two nodes and I thought there was a change in the global style.
So, you had traffic to ratings and kiali master node in your example, isn't it ?
Otherwise, it should show a dotted border, I thought perhaps the style was changed but I see there is not changed on that area.

src/components/GraphFilter/DurationDropdown.tsx
```diff
@@ -24,27 +25,43 @@ export class DurationDropdown extends React.Component<DurationDropdownProps, Dur
...
     return (
-      <DropdownButton id="duration" title="Duration" onSelect={this.onDurationChanged}>
-        {DurationDropdown.DurationIntervals.map(r => (
-          <MenuItem key={r[0]} active={r[0] === this.state.currentDuration} eventKey={r[0]}>
-            {r[1]}
-          </MenuItem>
-        ))}
-      </DropdownButton>
+      <div className="form-group">
+        <label style={{ paddingRight: '0.5em' }}>Duration:</label>
```
mtho11: If we are going to be doing this in multiple places, then we should create our own component for dropdowns that has an optional label property for it

aljesusg:
Yes ^^, I always have a watcher doing requests

src/components/GraphFilter/LayoutDropdown.tsx
```diff
@@ -20,33 +21,49 @@ export class LayoutDropdown extends React.Component<LayoutDropdownProps, LayoutD
...
     super(props);
     this.state = {
-      currentGraphType: props.initialLayout
+      currentGraphType: props.initialLayout,
+      currentGraphTypeLabel: this.getGraphTypeName(props.initialLayout)
     };
   }
 
+  getGraphTypeName = (graphType: string) => {
+    let label = LayoutDropdown.InitialLabel;
```
mtho11: `InitialLabel` should be camelCase.

src/components/GraphFilter/LayoutDropdown.tsx
```diff
@@ -20,33 +21,49 @@ export class LayoutDropdown extends React.Component<LayoutDropdownProps, LayoutD
...
+      if (subarray[0] === graphType) {
+        label = subarray[1];
+      }
+    });
+    return label;
+  };
+
   layout = (graphType: string) => {
-    this.setState({ currentGraphType: graphType });
+    this.setState({ currentGraphType: graphType, currentGraphTypeLabel: this.getGraphTypeName(graphType) });
```
mtho11: Does `currentGraphTypeLabel` need to be stored in the state since it is just derived from `graphType`?

src/components/GraphFilter/LayoutDropdown.tsx
```diff
@@ -20,33 +21,49 @@ export class LayoutDropdown extends React.Component<LayoutDropdownProps, LayoutD
...
-            active={r[0] === this.state.currentGraphType}
-            id={r[0]}
-            eventKey={r[0]}
-            disabled={this.props.disabled}
-          >
-            {r[1]}
-          </MenuItem>
-        ))}
-      </DropdownButton>
+      <div className="form-group">
```
mtho11: This could also use the new component here.

src/components/GraphFilter/LayoutDropdown.tsx
```diff
@@ -20,33 +21,49 @@ export class LayoutDropdown extends React.Component<LayoutDropdownProps, LayoutD
     ['dagre', 'Dagre'],
     ['klay', 'Klay']
   ];
+  static InitialLabel = 'Select Graph Type';
```
mtho11: This is more of a question on style: Do we want to use ALL_CAPS for static types or ??

aljesusg: I am working in a JIRA to create a style code guide, in some places there is car with camelcase and orhers with uppercase for example. We need to have the code more aligned

src/components/GraphFilter/LayoutDropdown.tsx
aljesusg: Mmm No ^^ We can call him in the element. You are right

src/components/GraphFilter/LayoutDropdown.tsx
```diff
@@ -20,33 +21,49 @@ export class LayoutDropdown extends React.Component<LayoutDropdownProps, LayoutD
...
-            id={r[0]}
-            eventKey={r[0]}
-            disabled={this.props.disabled}
-          >
-            {r[1]}
-          </MenuItem>
-        ))}
-      </DropdownButton>
+      <div className="form-group">
+        <label style={{ paddingRight: '0.5em' }}>Graph Type:</label>
```
jshaughn: Minor, but I think 'Graph Layout' or just 'Layout' may a better label.  We're not really changing the type of information presented in the graph, only the layout.  @mtho11 What do you think?

mtho11: @jshaughn yes, I agree `Layout` is more precise. And we are already on the Service Graph page so `Graph` is probably redundant.

src/components/GraphFilter/DurationDropdown.tsx
```diff
@@ -24,27 +25,43 @@ export class DurationDropdown extends React.Component<DurationDropdownProps, Dur
...
   constructor(props: DurationDropdownProps) {
     super(props);
     this.state = {
-      currentDuration: props.initialDuration
+      currentDuration: props.initialDuration,
+      currentDurationName: this.getDurationName(props.initialDuration)
     };
   }
 
+  getDurationName = (duration: number) => {
```
jotak: Instead of having a lookup function to retrieve the name from the duration, you could directly have the tuple [duration, name] as key of `MenuItem`, it would make it a bit simpler.

src/components/GraphFilter/DurationDropdown.tsx
```diff
@@ -24,27 +25,43 @@ export class DurationDropdown extends React.Component<DurationDropdownProps, Dur
...
-          <MenuItem key={r[0]} active={r[0] === this.state.currentDuration} eventKey={r[0]}>
-            {r[1]}
-          </MenuItem>
-        ))}
-      </DropdownButton>
+      <div className="form-group">
+        <label style={{ paddingRight: '0.5em' }}>Duration:</label>
+        <DropdownButton id="duration" title={this.state.currentDurationName} onSelect={this.onDurationChanged}>
+          {DurationDropdown.DurationIntervals.map(r => (
+            <MenuItem key={r[0]} active={r[0] === this.state.currentDuration} eventKey={r[0]}>
```
jotak: cf my comment above, you could have:
`<MenuItem key={r[0]} active={r[0] === this.state.currentDuration} eventKey={r}>`

src/components/GraphFilter/LayoutDropdown.tsx
jotak: Yeah I think in some cases I also wrote upper camel case for static fields. We must define our standard. I have no opinion about what is better.

src/components/GraphFilter/LayoutDropdown.tsx
```diff
@@ -20,33 +21,49 @@ export class LayoutDropdown extends React.Component<LayoutDropdownProps, LayoutD
...
   constructor(props: LayoutDropdownProps) {
     super(props);
     this.state = {
-      currentGraphType: props.initialLayout
+      currentGraphType: props.initialLayout,
+      currentGraphTypeLabel: this.getGraphTypeName(props.initialLayout)
     };
   }
 
+  getGraphTypeName = (graphType: string) => {
```
jotak: Same as above: lookup function can be avoided if you pass the label+value as event key.

aljesusg:
@mtho11 I did the changes. Thanks.

mtho11:
Thanks @aljesusg

PR #276, opened 2018-04-24
KIALI-584 Add VirtualServices and DestinationRules

lucasponce:
- Introduce a tab level for Deployments, Sources and Istio Objects
- Refactor add-hoc Istio Object components for a generic one

image
image
image
image
image
image


aljesusg:
@lucasponce you need to launch npm run test -- u to update snapshoots

src/pages/ServiceDetails/ServiceInfo.tsx
```diff
@@ -139,26 +165,52 @@ class ServiceInfo extends React.Component<ServiceId, ServiceInfoState> {
...
-                {(destinationPolicies.length > 0 || this.state.istio_sidecar) && (
-                  <ServiceInfoDestinationPolicies destinationPolicies={destinationPolicies} />
-                )}
-              </Col>
-            )}
+            <Col xs={12} sm={12} md={12} lg={12}>
+              <TabContainer id="service-tabs" defaultActiveKey={101}>
+                <div>
+                  <Nav bsClass="nav nav-tabs nav-tabs-pf">
+                    <NavItem eventKey={101}>{'Deployments (' + deployments.length + ')'}</NavItem>
```
jotak: Can you explain how these numbers are chosen? (101, 102, 103 etc.)
In what scope do they have to be unique: in the tabcontainer?

abonas:
what is the difference between DestinationPolicy and DestinationRule? is it the same thing in different api versions?

src/pages/ServiceDetails/ServiceInfo.tsx
lucasponce: Let me check, I guess I used these IDs for me, as I had a inner tab and those were more clear.
But I don't remember a warning o problem, let me check.
I can change them to 1,2,3,4 if that's clear for you.

abonas:
> Both APIs can be used and the is not a clear direct translation/conversion, so a user needs to deal with objects in both versions.

Is there anywhere documentation of their differences? any important data v1 shows that v3 doesn't? if not, I think it would be best to stick to one of them (latest sounds best) to avoid confusions. 
cc @pilhuhn

lucasponce:
> Is there anywhere documentation of their differences? any important data v1 shows that v3 doesn't? if not, I think it would be best to stick to one of them (latest sounds best) to avoid confusions.

That is not possible.

You can create objects in one or another API, and there is not visibility.
Transforming objects from one version to other is a experimental, not completed feature in Istio as well.

So I think we need to deal with both.

abonas:
> That is not possible.
> 
> You can create objects in one or another API, and there is not visibility.
> Transforming objects from one version to other is a experimental, not completed feature in Istio as well.
> 
> So I think we need to deal with both.

I think I have some gaps in understanding here.
If a user creates an object via v1, does this object show up in v3 when fetching objects, or not?
in other words, are those completely different objects that exist for v1 and v3  separately? so you could have a destination policy A on v1, but that A does not exist completely on v3 ? 
Or A exists on v1, and A' on v3, where A' representing the same info from A but in a different format?

lucasponce:
> If a user creates an object via v1, does this object show up in v3 when fetching objects, or not?

No, objects created in v1 and v3 are totally independent. It is close to a 1 to 1 translation in terms of functionality but as user you don't see v1 objects from v3 APi and viceversa.

> so you could have a destination policy A on v1, but that A does not exist completely on v3 ?

That's correct.

There is no visibility, and API has changed, but both can be effectively used.

So, we can't rely only on the latest v3 API on this case.

https://istio.io/docs/reference/commands/istioctl.html#istioctl%20experimental%20convert-networking-config

This is pure AlphaX API that I think it will change, but to avoid confusion, we need to support both.

abonas:
@lucasponce that is a very strange approach that istio chose :)
I think we can still decide and declare we work only with latest. 
All apis are alpha, and anyway it will most likely change, but then at least we won't have to maintain logic and UI for both. 
As there are no users yet, I think it is a valid direction to consider. I'd like to hear @pilhuhn @mwringe thoughts on this.

lucasponce:
> I think we can still decide and declare we work only with latest.

Sorry, I am not agree.

If you have defined an object in v1, you can modified them from v3, so, we need to support both.
I don't think Kiali can transparently convert them, that will confuse users.

abonas:
> I think we can still decide and declare we work only with latest.
> 
> Sorry, I am not agree.
> 
> If you have defined an object in v1, you can modified them from v3, so, we need to support both.
> I don't think Kiali can transparently convert them, that will confuse users.


I'm not expecting introducing any conversion here.
But perhaps I'm confused again. you [wrote above
](https://github.com/kiali/kiali-ui/pull/276#issuecomment-383915643) that an object created via v1 does not exist at all on v3, but here you wrote that if an object was defined on v1, it can be modified via v3, how does those approaches coexist?

lucasponce:
> that an object created via v1 does not exist at all on v3, but here you wrote that if an object was defined on v1, it can be modified via v3, how does those approaches coexist?

Sorry, fast typing, I've modified my comment.

You can't modify an object from a different api.

All objects from v1 and v3 are independent and user has visibility of them.
Istio Pilot send them to envoy at it's dealt at that level but there is no conversion.

The roadmap for this is asked in the Istio forums and due the experimental nature of this, the best approach in my opinion is to support both.

mwringe:
Since Istio 0.7.1 supports both, I think we need to as well.

pilhuhn:
I think the point is that the alpha 3 api is going to superseed the alpha1 api at some point in time. And perhaps even be replaced by a beta1 api, that is again looking differently. 

Going forward we should probably (only) support latest. There is in fact an Istio PR that can convert alpha1 to alpha3 rules: https://github.com/istio/istio/pull/5178, which I can imaging to end up in Istio 0.8

lucasponce:
> Going forward we should probably (only) support latest. There is in fact an Istio PR that can convert alpha1 to alpha3 rules: istio/istio#5178, which I can imaging to end up in Istio 0.8

I am not agree with that solution. Objects are independent, and users works with both.
So Kiali should support both.

If you dictate that change, I'll do it, but I am againt it.

I think it is better to support both now and modify kiali at same pace Istio consolidates its API.

lucasponce:
Istio 0.8.0 is not going to change this API, at least that is what we can see on istio master branch.

I suggest to merge this, as this UI PR is blocking other areas that are in development.

As always, best methodology for these discussions is to present a feature to real users and get the feedback.

abonas:
@lucasponce @jotak @pilhuhn @mwringe @aljesusg fyi this comment:
https://github.com/istio/istio/pull/5178#issuecomment-384315062
it means we'll have to cleanup relatively soon...

PR #277, opened 2018-04-24
KIALI-379 fix yaxis precision on charts

jotak:


lucasponce:
It seems fixed:

image

PR #278, opened 2018-04-24
KIALI-605: Fix sidebar focus when navigating from graph to service list

israel-hdez:
Sidebar kept "Graph" hightlighted when user clicked on the namespace
link in the side panel to go to the serivce list page.

### DIFF
--- a/src/components/Nav/Navigation.tsx
+++ b/src/components/Nav/Navigation.tsx
@@ -40,19 +40,18 @@ class Navigation extends React.Component<PropsType, StateType> {
     router: PropTypes.object
   };
 
-  constructor(props: any) {
-    super(props);
-    this.navigateTo = this.navigateTo.bind(this);
+  static getDerivedStateFromProps(nextProps: PropsType, prevState: StateType): Partial<StateType> | null {
+    const selected = Navigation.parsePath(nextProps.location.pathname);
+    if (selected === prevState.selectedItem) {
+      return null;
+    }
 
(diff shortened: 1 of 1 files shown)


src/components/Nav/Navigation.tsx
```diff
@@ -77,16 +76,11 @@ class Navigation extends React.Component<PropsType, StateType> {
...
-    if (e.title === servicesTitle) {
-      this.context.router.history.push(servicesPath);
-    } else if (e.title === istioRulesTitle) {
-      this.context.router.history.push(istioRulesPath);
-    } else if (e.title === servicesJaeger) {
-      this.context.router.history.push(servicesJaegerPath);
-    } else {
-      this.context.router.history.push(serviceGraphPath);
-    }
+  componentWillReceiveProps(nextProps: PropsType) {
```
josejulio: This method seems deprecated/unsafe [1], please use ´getDerivedStateFromProps´ [2]


[1] https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops
[2] https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops

israel-hdez: Oh... I'll fix that...
So... I need to re-read the docs.

josejulio: I suppose it will suffice to read from here https://github.com/kiali/kiali-ui/pull/214#issue-179554533 (the comments and the links)

PR #280, opened 2018-04-25
Typestyle introduction

mtho11:
A simple example of libs and how to use typestyle  to achieve typesafe css styling and more...


mtho11:
One of the major issues with CSS is that the failures are totally silent. So you got the wrong style... a one character mistake will display as a huge visual failure for the user.
TypeStyle is an added protection against CSS style errors.

jotak:
I understand the advantages of typestyle over raw CSS.
What I don't get is the advantage over inline style, since it achieves (I think) pretty much the same without needing a lib. Styles within JSX also get type-safety, IDE auto-completion and so on with the type React.CSSProperties. They're also reusable, mixable, etc. thanks to the language typescript.

Maybe what's missing is the generation of a real CSS class, but to be honest, I've lost the utility of that now that inline styles are leveraged with TS/React.

cfcosta:
@jotak you can pass styles around real JS objects, create queries and so on, making it a lot more composable. It also generates unique classes for each of the object that vary dinamically, so that there's no way to poke around and override class properties in other elements.

mtho11:
I'm marking this as Do not merge until we have more discussion.

mtho11:
@jotak I was unaware of `React.CSSProperties` for typescript let's investigate a bit more. I don't see it's use anywhere in the project though.

jotak:
@mtho11 it's used implicitly when sending an object to `style` in any DOM element via JSX.
Now maybe it doesn't cover everything that usual CSS does. Like `:hover` `:focus` and so on, I haven't seen if it's possible to do it via `CSSProperties`. Maybe that's where a dedicated lib brings more advantages.

PR #281, opened 2018-04-25
KIALI-598 - don't hardcode pf colors everywhere

jmazzitelli:
We want to use PF colors everywhere. But we don't want to hardcode the actual #RGB values everywhere either. Since we use these colors mainly in the JS/react code, this creates an enum with the colors that can be used in place of the '#rrggbb' notation.


jmazzitelli:
do not merge this yet, I need to go through the rest of the code and replace all the hardcoded colors with the new enum.

jmazzitelli:
All code has been changed to use the patternfly colors enum. All tests pass. This should be good now for review and merge.

src/components/Badge/__tests__/__snapshots__/Badge.test.tsx.snap
```diff
@@ -70,7 +70,7 @@ ShallowWrapper {
           fontSize={8.8}
         >
           <text
-            fill="#010101"
+            fill="#030303"
```
jotak: snapshot shows the new colors are not exactly as before, but I guess the new ones are correct?

jmazzitelli: yes, I updated all snapshots that should correspond to the new colors. #030303 is patternfly "black"... #010101 is not an official patternfly color.

src/components/Pf/PfColors.tsx
```diff
@@ -0,0 +1,90 @@
+export enum PfColors {
```
jotak: Just to understand how you ended up with that file: is it copy-pasted from somewhere? Or created from scratch?
Maybe it would worth have it pushed to patternfly itself, if it doesn't already exist in that form.

abonas: @serenamarie125 do you know if there is a best practice from pf around this?

jotak: I see that we have access to the palette in `node_modules/patternfly/dist/js/patternfly.js`
Which we can import and use as:
`import { patternfly } from 'patternfly-react';`
`patternfly.pfPaletteColors.green`

But the problem then is that we do not have types on it (it's not written in typescript). Which means no static checks, no IDE autocompletion. In the end I prefer to replicate it in a typescript module as @jmazzitelli did.

josejulio: I think we should use these patternfly constants (`patternfly.pfPaletteColors.green`) instead of writing directly the '#cfe7cd'
I don't think these constants are changed, but maybe more explicit the 1-1 relation

```typescript
Green = '#cfe7cd',
```
We could use (if possible)
```typescript
Green = patternfly.pfPaletteColors.green,
```

src/components/Pf/PfColors.tsx
```diff
@@ -0,0 +1,90 @@
...
+  Cyan = '#007a87', // Cyan400
+  Gold = '#f0ab00', // Gold400
+  Green = '#3f9c35', // Green400
+  LightBlue = '#00b9e4', // LightBlue400
+  LightGreen = '#92d400', // LightGreen400
+  Orange = '#ec7a08', // Orange400
+  Red = '#cc0000', // Red100
+
+  // Kiali colors that use PF colors
+  Gray = '#72767b' // Black600
```
josejulio: Why don't we use Black600 instead? is for convenience?

src/components/Pf/PfColors.tsx
jmazzitelli: I'll see if that would work.

jmazzitelli: Can't use those patternfly values (patternfly.pfPaletteColors.XYZ) inside the enum - compile error. IDE tells me "[ts] Computed values are not permitted in an enum with string valued members." and the npm run build tells me: 

```
/source/kiali/kiali-ui/src/components/Badge/__tests__/Badge.test.tsx
(13,28): Type '{ scale: number; style: string; color: PfColors; leftText: string; rightText: string; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<Badge> & Readonly<{ children?: ReactNode; }> & Rea...'.
  Type '{ scale: number; style: string; color: PfColors; leftText: string; rightText: string; }' is not assignable to type 'Readonly<BadgeProps>'.
    Types of property 'color' are incompatible.
      Type 'PfColors' is not assignable to type 'string'.
```

So unless there are other objections, going to leave or typescript enum as-is.

serenamarie125: @ jeff-phillips-18 could you help here?  we'd like to reference the PF colors, maybe you can help share what you've done in the past?

serenamarie125: We want to safeguard from PF changing colors, and having to come back and update your values if possible.

jotak: +1 to leave it as is for the moment, but in the same time we should try to see with PF team if we can have a TS definition file within patternfly.

jeff-phillips-18: I haven't had the opportunity to attempt to use the PF Colors within typescript classes before. It had been brought up before to include the type definitions in patternfly before but there wasn't a strong need for it. If we feel it is important, please open an issue in https://github.com/patternfly/patternfly and maybe we can address it there.

jmazzitelli: Well, I tell you fellas, we need more typedefs... I gotta have more typedefs - :smiley: so here's the issue @jeff-phillips-18 - https://github.com/patternfly/patternfly/issues/1040 

I will fix the conflicts and hopefully we can merge this now since there is nothing currently available in patternfly today.

PR #284, opened 2018-04-25
KIALI-622: incorrect default for CBs flag and make sure CytoscapeLayout aware of it

vnugent:
[x] correctly initialized CBs when parsing url params
[x] shortened Circuit Breakers button label 
[x] check CBs flag in `CytoscapeLayout.shouldComponentUpdate()`  
[x]  misc clean up

---
CBs = OFF
cb-off

CBs = ON
cb-on

### DIFF
--- a/src/components/GraphFilter/GraphFilter.tsx
+++ b/src/components/GraphFilter/GraphFilter.tsx
@@ -35,7 +35,7 @@ export default class GraphFilter extends React.Component<GraphFilterProps, Graph
     this.props.onRefresh();
   };
 
-  handleToggleCBs = (newValue: boolean) => {
+  handleToggleCBs = (event: any) => {
     this.props.onBadgeStatusChange({ hideCBs: !this.props.badgeStatus.hideCBs });
   };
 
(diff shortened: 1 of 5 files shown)


src/components/GraphFilter/GraphFilter.tsx
```diff
@@ -35,7 +35,7 @@ export default class GraphFilter extends React.Component<GraphFilterProps, Graph
     this.props.onRefresh();
   };
 
-  handleToggleCBs = (newValue: boolean) => {
+  handleToggleCBs = (event: any) => {
```
vnugent: `Switch.onChange()` passes the whole underlying JS object, not a true|false

src/components/ToolbarDropdown/ToolbarDropdown.tsx
```diff
@@ -34,7 +34,7 @@ export class ToolbarDropdown extends React.Component<ToolbarDropdownProps, Toolb
     return (
       <div className="form-group">
         {this.props.nameDropdown && <label style={{ paddingRight: '0.5em' }}>{this.props.nameDropdown}:</label>}
-        <DropdownButton title={this.state.currentName} onSelect={this.onKeyChanged}>
+        <DropdownButton title={this.state.currentName} onSelect={this.onKeyChanged} id="kiali-dropdown">
```
vnugent: React warns about accessibility issue when dropdown doesn't have an Id

PR #287, opened 2018-04-26
KIALI-628 Use the ACE editor on Mixer Rules

lucasponce:
This PR introduces the ACE editor on the Mixer Rules area.

As we have several and different configuration per Mixer Rule, this approach makes clickable the specific handler or instance in the Mixer Rule list:

image

Then we move to the editor in read-only

image

We don't miss the navigation within the same Mixer Rule we can select other handler or instances:

image

We are reusing same proposal made in the OpenShift console in terms of how to use the editor.

When we enable the modification of these files we can add the Save, Cancel buttons, but as today everything is read only I thought that would cause more confusion.


src/pages/IstioRuleDetails/IstioRuleInfo.scss
```diff
@@ -0,0 +1,5 @@
+.IstioAceEditor {
```
jotak: Here again we have a problem of naming homogeneity: in general CSS classes are low-case with dashes. But some of us started with upper camel case, other stick with the norm... Definitely a problem to solve by @aljesusg in KIALI-599 ;-)
[So, no change request here until we have guidelines.]

src/pages/IstioRuleDetails/IstioRuleInfo.tsx
```diff
@@ -52,16 +72,47 @@ class IstioRuleInfo extends React.Component<RuleId, RuleInfoState> {
...
-  render() {
-    let actionList: any = [];
-    for (let i = 0; i < this.state.actions.length; i++) {
-      actionList.push(
-        <Col key={'ruleAction' + i}>
-          <IstioRuleInfoAction action={this.state.actions[i]} />
-        </Col>
-      );
+  parseSearch(): ParsedSearch {
+    if (this.props.search) {
```
jotak: Can you add a comment about the expected string in this.props.search?

src/pages/IstioRuleDetails/IstioRuleInfo.tsx
```diff
@@ -1,21 +1,41 @@
...
+
+interface RuleDetailsId extends RuleId {
+  search?: string;
+}
+
+interface ParsedSearch {
+  type?: string;
+  name?: string;
+}
+
```
aljesusg: Other thing I wanna add to the style code is use Proptypes. But LGFM until we have guidelines

src/pages/IstioRuleDetails/IstioRuleInfo.tsx
lucasponce: This helper is more a private method to the class where we know the expected parameters.
In absence of them, the default behaviour is correct.
I'll add a comment.

src/pages/IstioRuleDetails/IstioRuleInfo.tsx
```diff
@@ -52,16 +72,47 @@ class IstioRuleInfo extends React.Component<RuleId, RuleInfoState> {
...
-          <IstioRuleInfoAction action={this.state.actions[i]} />
-        </Col>
-      );
+  parseSearch(): ParsedSearch {
+    if (this.props.search) {
+      let firstParams = this.props.search
+        .split('&')[0]
+        .replace('?', '')
+        .split('=');
+      return {
```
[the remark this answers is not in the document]
lucasponce: Yeah, a regex would work too, but I guess as we expect just a single param, perhaps a split is even simpler.
If there is a global helper method I can reuse it.

src/pages/IstioRuleDetails/IstioRuleInfo.tsx
lucasponce: Can you add more details here ?
If you show me an example I can try to add it and follow the style.

aljesusg: Here is an example 

```js
import PropTypes from 'prop-types';

class MyComponent extends React.Component {
  render() {
    // This must be exactly one element or it will warn.
    const children = this.props.children;
    const name = this.props.name;
    return (
      <div>
        The name is { name }
        {children}
      </div>
    );
  }
}

MyComponent.propTypes = {
  name: PropTypes.string,
  children: PropTypes.element.isRequired
};

// Specifies the default values for props:
MyComponent.defaultProps = {
  name: 'Stranger'
};
```

Here is an example documenting the different validators provided:

```js
import PropTypes from 'prop-types';

MyComponent.propTypes = {
  // You can declare that a prop is a specific JS type. By default, these
  // are all optional.
  optionalArray: PropTypes.array,
  optionalBool: PropTypes.bool,
  optionalFunc: PropTypes.func,
  optionalNumber: PropTypes.number,
  optionalObject: PropTypes.object,
  optionalString: PropTypes.string,
  optionalSymbol: PropTypes.symbol,

  // Anything that can be rendered: numbers, strings, elements or an array
  // (or fragment) containing these types.
  optionalNode: PropTypes.node,

  // A React element.
  optionalElement: PropTypes.element,

  // You can also declare that a prop is an instance of a class. This uses
  // JS's instanceof operator.
  optionalMessage: PropTypes.instanceOf(Message),

  // You can ensure that your prop is limited to specific values by treating
  // it as an enum.
  optionalEnum: PropTypes.oneOf(['News', 'Photos']),

  // An object that could be one of many types
  optionalUnion: PropTypes.oneOfType([
    PropTypes.string,
    PropTypes.number,
    PropTypes.instanceOf(Message)
  ]),

  // An array of a certain type
  optionalArrayOf: PropTypes.arrayOf(PropTypes.number),

  // An object with property values of a certain type
  optionalObjectOf: PropTypes.objectOf(PropTypes.number),

  // An object taking on a particular shape
  optionalObjectWithShape: PropTypes.shape({
    color: PropTypes.string,
    fontSize: PropTypes.number
  }),

  // You can chain any of the above with `isRequired` to make sure a warning
  // is shown if the prop isn't provided.
  requiredFunc: PropTypes.func.isRequired,

  // A value of any data type
  requiredAny: PropTypes.any.isRequired,

  // You can also specify a custom validator. It should return an Error
  // object if the validation fails. Don't `console.warn` or throw, as this
  // won't work inside `oneOfType`.
  customProp: function(props, propName, componentName) {
    if (!/matchme/.test(props[propName])) {
      return new Error(
        'Invalid prop `' + propName + '` supplied to' +
        ' `' + componentName + '`. Validation failed.'
      );
    }
  },

  // You can also supply a custom validator to `arrayOf` and `objectOf`.
  // It should return an Error object if the validation fails. The validator
  // will be called for each key in the array or object. The first two
  // arguments of the validator are the array or object itself, and the
  // current item's key.
  customArrayProp: PropTypes.arrayOf(function(propValue, key, componentName, location, propFullName) {
    if (!/matchme/.test(propValue[key])) {
      return new Error(
        'Invalid prop `' + propFullName + '` supplied to' +
        ' `' + componentName + '`. Validation failed.'
      );
    }
  })
};
```

aljesusg: Yes, it is not necessary, my point of view is that there is some views with proptypes and others with not so....or we use proptype in all files or we should remove it

jotak: If I understand correctly Proptypes is something that is primarily meant for vanilla JS, although it works as well with typescript but there is some overlap then. Maybe for all that is covered by typescript we can stick with normal typing, and use Proptypes only when we want more advanced validation?

lucasponce:
@jotak @aljesusg I think I addressed your comments. Pls, let me know if you miss something.
Thx !

PR #288, opened 2018-04-26
[KIALI-610] Shared app duration option in toolbars

aljesusg:
Shared duration of application [KIALI-610](https://issues.jboss.org/browse/KIALI-610).

duration_stored

### DIFF
--- a/src/components/GraphFilter/GraphFilter.tsx
+++ b/src/components/GraphFilter/GraphFilter.tsx
@@ -4,28 +4,14 @@ import { Toolbar, Button, ButtonGroup, Switch, Icon, FormGroup } from 'patternfl
 import { GraphFilterProps, GraphFilterState } from '../../types/GraphFilter';
 import { ToolbarDropdown } from '../ToolbarDropdown/ToolbarDropdown';
 import AutoUpdateNamespaceList from '../../containers/AutoUpdateNamespaceList';
+import { config } from '../../config';
 
 export default class GraphFilter extends React.Component<GraphFilterProps, GraphFilterState> {
   // TODO:  We should keep these mappings with their corresponding filtering components.
   // GraphFilter should be minimal and used for assembling those filtering components.
-  static INTERVAL_DURATION = [
-    [60, '1 minute'],
-    [600, '10 minutes'],
-    [1800, '30 minutes'],
-    [3600, '1 hour'],
-    [14400, '4 hours'],
-    [28800, '8 hours'],
-    [86400, '1 day'],
-    [604800, '7 days'],
-    [2592000, '30 days']
-  ];
-  static GRAPH_LAYOUTS = [
-    ['breadthfirst', 'Breadthfirst'],
-    ['cola', 'Cola'],
-    ['cose', 'Cose'],
-    ['dagre', 'Dagre'],
-    ['klay', 'Klay']
-  ];
+  static INTERVAL_DURATION = config().toolbar.intervalDuration;
+  static GRAPH_LAYOUTS = config().toolbar.graphLayouts;
+
   constructor(props: GraphFilterProps) {
     super(props);
   }
@@ -33,6 +19,7 @@ export default class GraphFilter extends React.Component<GraphFilterProps, Graph
   updateDuration = (value: number) => {
     if (this.props.graphDuration.value !== value) {
       // notify callback
+      sessionStorage.setItem('appDuration', String(value));
       this.props.onFilterChange({ value: value });
     }
   };
@@ -85,11 +72,11 @@ export default class GraphFilter extends React.Component<GraphFilterProps, Graph
             disabled={this.props.disabled}
             onClick={this.updateDuration}
             nameDropdown={'Duration'}
-            initialValue={this.props.graphDuration.value}
+            initialValue={Number(sessionStorage.getItem('appDuration')) || this.props.graphDuration.value}
             initialLabel={String(
-              GraphFilter.INTERVAL_DURATION.filter(elem => {
-                return elem[0] === Number(this.props.graphDuration.value);
-              })[0][1]
+              GraphFilter.INTERVAL_DURATION[
+                Number(sessionStorage.getItem('appDuration')) || config().toolbar.defaultDuration
+              ]
             )}
             options={GraphFilter.INTERVAL_DURATION}
           />
@@ -98,11 +85,7 @@ export default class GraphFilter extends React.Component<GraphFilterProps, Graph
             onClick={this.updateLayout}
             nameDropdown={'Layout'}
             initialValue={this.props.graphLayout.name}
-            initialLabel={String(
-              GraphFilter.GRAPH_LAYOUTS.filter(elem => {
-                return elem[0] === String(this.props.graphLayout.name);
-              })[0][1]
-            )}
+            initialLabel={String(GraphFilter.GRAPH_LAYOUTS[this.props.graphLayout.name])}
             options={GraphFilter.GRAPH_LAYOUTS}
           />
           <Toolbar.RightContent>
--- a/src/components/MetricsOptions/MetricsOptionsBar.tsx
+++ b/src/components/MetricsOptions/MetricsOptionsBar.tsx
@@ -1,8 +1,9 @@
 import * as React from 'react';
-import { Toolbar, ToolbarRightContent, DropdownButton, MenuItem, Icon } from 'patternfly-react';
-
+import { Toolbar, ToolbarRightContent, Icon } from 'patternfly-react';
+import { config } from '../../config';
 import ValueSelectHelper from './ValueSelectHelper';
 import MetricsOptions from '../../types/MetricsOptions';
+import { ToolbarDropdown } from '../ToolbarDropdown/ToolbarDropdown';
 
 interface Props {
   onOptionsChanged: (opts: MetricsOptions) => void;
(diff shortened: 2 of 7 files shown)


src/pages/ServiceDetails/ServiceMetrics.tsx
```diff
@@ -226,9 +243,14 @@ class ServiceMetrics extends React.Component<ServiceId, ServiceMetricsState> {
   dismissAlert = () => this.setState({ alertDetails: undefined });
 
   render() {
+    const { store } = this.context;
+    store.subscribe(() => {
```
jotak: Does it really has to be subscribed in `render`?
`render` can be called many times in the component's lifecycle. I guess subscription has to be called only once. So, constructor or `componentDidMount` is a better fit, no?

src/pages/ServiceDetails/ServiceMetrics.tsx
```diff
@@ -49,6 +52,9 @@ type NamedHistogram = {
 };
 
 class ServiceMetrics extends React.Component<ServiceId, ServiceMetricsState> {
+  static contextTypes = {
```
jotak: What is it for?

src/pages/ServiceDetails/ServiceMetrics.tsx
aljesusg: I don't know ^^ I am new in redux, This is one of my questions to @mtho11 and @vnugent how to handle with the subscribe(), I don't like to have them in the render, but They should be in each component that we wanna update if the store change

src/pages/ServiceDetails/ServiceMetrics.tsx
aljesusg: You can't access to the store of the provider if you not define store in the contextTypes, you can't access with this.context . 
React only populates the context if you specify contextTypes in your component. You will however notice that the props in your component is populated from the store, as you specified in your mapStateToProps function. If you wanna access them without use contextTypes you must handle the subscriptions,... on the root level (where you create the store)

aljesusg:
I know but you said that the toolbars were so similars...I disagree about to merge this toolbars ^^ they are so different

jotak:
@aljesusg ah, I see, there's probably been some misunderstanding.
I think it would be nice to "connect" the toolbars via redux, but I didn't mean to create an unique component.
Each toolbar can subscribe to duration individually, no?

src/pages/ServiceDetails/ServiceMetrics.tsx
mtho11: It doesn't have to be subscribed to in render. A connected component will automatically re-render itself for the `mapStateToProps` state fields.

src/pages/ServiceDetails/ServiceMetrics.tsx
mtho11: You really shouldn't need to access the store you get whatever state you want from the store just by specifying the props in `mapStateToProps` and using `connect`.

src/components/GraphFilter/GraphFilter.tsx
```diff
@@ -33,6 +35,7 @@ export default class GraphFilter extends React.Component<GraphFilterProps, Graph
   updateDuration = (value: number) => {
     if (this.props.graphDuration.value !== value) {
       // notify callback
+      sessionStorage.setItem('appDuration', JSON.stringify(value));
```
lucasponce: I'm not familiar with sessionStorage, so perhaps some comments from my side are pretty obvious.
Why to stringify the value ? Is it a necessary serialization ?

src/components/GraphFilter/GraphFilter.tsx
```diff
@@ -74,10 +77,12 @@ export default class GraphFilter extends React.Component<GraphFilterProps, Graph
             disabled={this.props.disabled}
             onClick={this.updateDuration}
             nameDropdown={'Duration'}
-            initialValue={this.props.graphDuration.value}
+            initialValue={Number(sessionStorage.getItem('appDuration')) || this.props.graphDuration.value}
```
lucasponce: Number() is a built-in conversion helper. isn't it

src/components/GraphFilter/GraphFilter.tsx
aljesusg: You can save a number in the sessionStorage, must be a string (https://developer.mozilla.org/en-US/docs/Web/API/Storage/setItem) so I set stringify to convert and avoid future errors (https://www.w3schools.com/js/js_json_stringify.asp) we could cast the number to string only too but I think that this will be a json object in a future.

src/components/GraphFilter/GraphFilter.tsx
aljesusg: Yes, the result of getItem from the storage is a string and I need to cast to number.

src/components/MetricsOptions/MetricsOptionsBar.tsx
```diff
@@ -96,6 +97,7 @@ export class MetricsOptionsBar extends React.Component<Props, MetricsOptionsStat
   };
 
   onDurationChanged = (key: number) => {
+    sessionStorage.setItem('appDuration', JSON.stringify(key));
```
lucasponce: Same here, I wonder converting the object to JSON is needed.

aljesusg: It's not necessary, only convert to string

src/components/GraphFilter/GraphFilter.tsx
```diff
@@ -9,15 +9,17 @@ export default class GraphFilter extends React.Component<GraphFilterProps, Graph
   // TODO:  We should keep these mappings with their corresponding filtering components.
   // GraphFilter should be minimal and used for assembling those filtering components.
   static INTERVAL_DURATION = [
```
xeviknal: I might arrive late into this PR. I am wondering if both duration arrays can be unified into one. If kiali is going to share state of duration, would it make sense to share same DURATION array?

aljesusg: Yes, but I don't how I shoudl put this array, I don't know If have sense to store it in config.ts

src/components/GraphFilter/GraphFilter.tsx
vnugent: @aljesusg  Similar logic for restoring params from local storage is in `ServiceGraphRouteHandler`.   May I suggest that `appDuration` is initialized there as well and is passed down to child components as prop instead?

this may be useful: https://github.com/kiali/kiali-ui/pull/243#issuecomment-383694057

aljesusg: Tolbar metrics need appDuration too, we shouldn't pass this like props when in each toolbar is setting  a new appDuration in sessionStorage when user select from dropdown.

src/components/GraphFilter/GraphFilter.tsx
```diff
@@ -4,35 +4,22 @@ import { Toolbar, Button, ButtonGroup, Switch, Icon, FormGroup } from 'patternfl
...
-    [2592000, '30 days']
-  ];
-  static GRAPH_LAYOUTS = [
-    ['breadthfirst', 'Breadthfirst'],
-    ['cola', 'Cola'],
-    ['cose', 'Cose'],
-    ['dagre', 'Dagre'],
-    ['klay', 'Klay']
-  ];
+  static INTERVAL_DURATION = config().toolbar.intervalDuration;
```
lucasponce: To have centralized this in a single place is good.
Just minor question due perhaps I'm not familiar with best practices.
Why a config() function instead of a plain exported object ?
(Non blocker).

src/components/MetricsOptions/MetricsOptionsBar.tsx
```diff
@@ -122,22 +108,27 @@ export class MetricsOptionsBar extends React.Component<Props, MetricsOptionsStat
...
-          </DropdownButton>
-          <DropdownButton id="pollInterval" title="Polling interval" onSelect={this.onPollIntervalChanged}>
-            {MetricsOptionsBar.PollIntervals.map(r => (
-              <MenuItem key={r[0]} active={r[0] === this.state.pollInterval} eventKey={r[0]}>
-                {r[1]}
-              </MenuItem>
-            ))}
-          </DropdownButton>
-        </div>
+        <ToolbarDropdown
```
lucasponce: Just to document, I didn't see in my previous review that there is a change in the component DropdownButton -> ToolbarDropdown. I guess it's same component but optimized for the Toolbar layout ?
I looks a good change, does it has any difference between the previous one ?

src/components/ToolbarDropdown/ToolbarDropdown.tsx
```diff
@@ -26,18 +27,18 @@ export class ToolbarDropdown extends React.Component<ToolbarDropdownProps, Toolb
...
 
   render() {
     return (
       <div className="form-group">
         {this.props.nameDropdown && <label style={{ paddingRight: '0.5em' }}>{this.props.nameDropdown}:</label>}
-        <DropdownButton title={this.state.currentName} onSelect={this.onKeyChanged} id="kiali-dropdown">
-          {this.props.options.map(r => (
-            <MenuItem key={r[0]} active={r[0] === this.state.currentValue} eventKey={r}>
-              {r[1]}
+        <DropdownButton title={this.state.currentName} onSelect={this.onKeyChanged} {...this.props}>
```
lucasponce: I see here this is not changed, compared with the previous one.
It seems ok to me but just leaving a note as perhaps others may wonder the same.

src/components/MetricsOptions/MetricsOptionsBar.tsx
aljesusg: Yes, ToolbarDropdown is a optimized component used in Graphtoolbar too, it's so similar for MetricOptions toolbar so I changed it, now it's cleaner

src/components/ToolbarDropdown/ToolbarDropdown.tsx
aljesusg: I added this.props to the component to propage to child....for example "id", with this you don't need to remove the tests 

https://github.com/kiali/kiali-ui/blob/master/src/components/MetricsOptions/__tests__/MetricsOptionsBar.test.tsx#L24

src/components/GraphFilter/GraphFilter.tsx
aljesusg: I saw in another projects that they have the config file like this maybe we need to add some methods in the configuration in a future.

aljesusg:
@jotak is in PTO I remove his review

src/components/GraphFilter/GraphFilter.tsx
```diff
@@ -70,6 +57,7 @@ export default class GraphFilter extends React.Component<GraphFilterProps, Graph
   };
 
   render() {
+    console.log(this.props.graphDuration.value);
```
xeviknal: Just wondering, is that an deliberate console.log or perhaps you forgot it?

aljesusg: Changed !!! Thanks @xeviknal

PR #290, opened 2018-04-26
KIALI-614: always refresh namespace list when dropdown is open

vnugent:
Previously the list was only updated when user switches to different namespace.  Now it updates every time users click on the dropdown.

---
# Before adding a new namespace
kiali-614-before
# After adding a new one
kiali-614-after


josejulio:
I guess that in the future we could indicate somehow that we are pulling the namespaces from the server, although it happens so fast now that is currently not an issue

vnugent:
@josejulio we can connect the error state to the global error message center at some point

PR #292, opened 2018-04-27
Package libraries versions bump

mtho11:
Update the fixed package.json versions to newer versions.


package.json
```diff
@@ -46,22 +46,22 @@
...
     "@types/jest": "22.2.3",
-    "@types/lodash": "4.14.107",
+    "@types/lodash": "4.14.108",
     "@types/node": "9.6.5",
-    "@types/react": "16.3.11",
+    "@types/react": "16.3.12",
     "@types/react-dom": "16.0.5",
     "@types/react-router-dom": "4.2.6",
     "@types/url-search-params": "0.10.1",
-    "awesome-typescript-loader": "4.0.1",
```
josejulio: awesome-typescript-loader 5.0.0 is not compatible with current webpack version 3.11.0

mtho11: @josejulio I'm not seeing that incompatibility. The following commands all work:
* `npm run build`
* `npm start`
* `npm run test`

josejulio: I recall now, this dependency is used to allow storybooks to use typescript. But since we are removing it...

PR #294, opened 2018-04-27
KIALI-637 Fix broken navigation when click on Mixer Rule

lucasponce:


lucasponce:
+1
The motivation was due an issue in the Patternfly React component

https://github.com/patternfly/patternfly-react/issues/239

I haven't checked yet if that has been solved.

The workarounds I saw were: a) to not use this component and create your own nav in plain div b) use this component but parse the url to update the state (all of these on the patternfly-react examples that were around there).

b) was less code and better to maintain, but yes, worth to see if this issue is gone in new versions.

PR #297, opened 2018-04-27
KIALI-627 add namespace label to graphfilter and bring in left margin

jshaughn:
screenshot

### DIFF
--- a/src/components/GraphFilter/GraphFilter.tsx
+++ b/src/components/GraphFilter/GraphFilter.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { Toolbar, Button, ButtonGroup, Switch, Icon } from 'patternfly-react';
+import { Toolbar, Button, ButtonGroup, Switch, Icon, FormGroup } from 'patternfly-react';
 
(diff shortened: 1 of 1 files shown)


src/components/GraphFilter/GraphFilter.tsx
```diff
@@ -73,13 +73,14 @@ export default class GraphFilter extends React.Component<GraphFilterProps, Graph
     return (
       <>
         <Toolbar>
-          <div className="form-group">
+          <FormGroup style={{ paddingLeft: 0 }}>
```
mtho11: Any `style={{ attribute: value }}` should be considered an anti-pattern just as in normal html framework dev you don't add individual styles but classes. In React it's even easier to inline styles https://zhenyong.github.io/react/tips/inline-styles.html (and with TypeStyle there is even typesafety and css name hashing for global namespace preservation).

jshaughn: Alright, sorry, it's difficult to keep up with the speed of change we have with the code styling as well as the actual problem we're trying to solve.  This looked like an easy one but apparently nothing is easy anymore :)  I'll investigate the new way of doing this and generate another PR.

PR #298, opened 2018-04-27
KIALI-609: Don't allow an initial large zoom for the graph

israel-hdez:
Nodes look too big if there are only a few nodes in the graph. This will prevent large zooms when the graph loads. Thereafter, the user can zoom as he desires.


josejulio:
Before:
scree nshot_2018-04-27_13-36-03

vs

After:
screenshot_2018-04-27_13-36-23

PR #299, opened 2018-04-27
KIALI-635: Fix namespace chart lost

israel-hdez:
After the graph is loaded, when the user clicks the background,
properties are pushed to the component only once. Because of this, the
side panel was not loading the required data.

Changing some code to ensure that "final" properties are always pushed
when the side panels are mounted. This allows to load chart data once
the component is mounted.

### DIFF
--- a/src/pages/ServiceGraph/ServiceGraphPage.tsx
+++ b/src/pages/ServiceGraph/ServiceGraphPage.tsx
@@ -1,28 +1,24 @@
 import * as React from 'react';
 
 import Namespace from '../../types/Namespace';
-import { GraphParamsType } from '../../types/Graph';
+import { GraphParamsType, SummaryData } from '../../types/Graph';
 import { Duration, Layout, BadgeStatus } from '../../types/GraphFilter';
 
 import SummaryPanel from './SummaryPanel';
 import CytoscapeLayout from '../../components/CytoscapeLayout/CytoscapeLayout';
 import GraphFilter from '../../components/GraphFilter/GraphFilter';
 import PfContainerNavVertical from '../../components/Pf/PfContainerNavVertical';
-// import PfHeader from '../../components/Pf/PfHeader';
 import PfAlerts from '../../components/Pf/PfAlerts';
 import * as API from '../../services/Api';
 import { computePrometheusQueryInterval } from '../../services/Prometheus';
 
-// summaryData will have two fields:
-//   summaryTarget: The cytoscape element
-//   summaryType  : one of 'graph', 'node', 'edge', 'group'
 type ServiceGraphPageState = {
   alertVisible: boolean;
   alertDetails: {
     namespaceAlert?: any;
     graphLoadAlert?: any;
   };
-  summaryData: any;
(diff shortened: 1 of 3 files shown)


src/types/Graph.ts
```diff
@@ -1,8 +1,13 @@
 import { Duration, Layout, BadgeStatus } from './GraphFilter';
 import Namespace from './Namespace';
```
jshaughn: You can remove the comment at ServiceGraphPage:15 and move it here.

src/pages/ServiceGraph/ServiceGraphPage.tsx
```diff
@@ -22,7 +21,7 @@ type ServiceGraphPageState = {
     namespaceAlert?: any;
     graphLoadAlert?: any;
   };
-  summaryData: any;
+  summaryData?: SummaryData | null;
```
jshaughn: Just curious, what is the purpose of the  ' | null' ?

src/pages/ServiceGraph/ServiceGraphPage.tsx
```diff
@@ -47,7 +46,6 @@ export default class ServiceGraphPage extends React.Component<ServiceGraphPagePr
       isReady: false,
```
jshaughn: Not something you did, but 'isReady: true' should be added to setState in onReady(), please add this.

src/pages/ServiceGraph/ServiceGraphPage.tsx
mtho11: @jshaughn it allows setting the value to `null` otherwise it is expected to be defined.

mtho11: The real question here for @israel-hdez , is do we ever actually set it to `null`? Or can it just be `undefined`?

src/types/Graph.ts
israel-hdez: Done!

src/pages/ServiceGraph/ServiceGraphPage.tsx
israel-hdez: Of course. I added it in handleReady() which is the function for onReady event.

src/pages/ServiceGraph/ServiceGraphPage.tsx
israel-hdez: You can see in my changes that I replaced some:
```javascript
this.setState({
          // ...snip...
          summaryData: { summaryType: 'graph', summaryTarget: null },
          // ...snip...
        });
```

with:

```javascript
this.setState({
          // ...snip...
          summaryData: null,
          // ...snip...
        });
```

This was only to make simpler the conditional that decides if any side-panel should be present.

israel-hdez: The conditional is here: https://github.com/kiali/kiali-ui/pull/299/files#diff-bf5d5931e8a56754ac7cdfce71034a24R128

PR #301, opened 2018-04-28
[WIP] KIALI-341 gracefully refresh the graph when possible.

jmazzitelli:


jmazzitelli:
This is trying to get the graph to update gracefully, rather than having it destroyed and created every time something changes (like hiding badges or changing the time duration).

I have the badges changing gracefully - when you now zoom and pan the graph, then click the switches to hide and show the different badges, you will see the badges hide and show without resetting its pan and zoom.

The rest "should" still work - the code is in there - but it doesn't work.

The problem is the drop downs - specifically duration (though all the dropdowns do this).

Change the duration dropdown (from 1m to 10m or whatever). The ReactCytoscape component is getting unmounted and mounted rather than simply updated which means all its internal data goes away and we lose the state of the cy graph (which means I HAVE to re-create it, I do not have the option of re-using the existing graph which is what we need to get this all to work properly).

The problem appears to be here - does anyone know how to change the URL without causing this unmount?

https://github.com/kiali/kiali-ui/blob/b7d8b10a512c46e3e9613839fdfa40138238c950/src/pages/ServiceGraph/ServiceGraphRouteHandler.tsx#L97

vnugent:
I see where the problem is now.  

https://github.com/kiali/kiali-ui/blob/master/src/components/CytoscapeLayout/CytoscapeLayout.tsx#L140

```
 <ReactCytoscape
              containerID="cy"
              cyRef={cy => {   // will always run 
                this.cyRef(cy);
              }}
              elements={this.props.elements}
```

In `CytoscapeLayout.render()' we reset the reference to the underlying Cytoscape object each time we want the graph to update.  Maybe caching that ref somehow.

jmazzitelli:
1. That `this._isMounted = false` code isn't new or part of this PR - that's been around a while.
2. In  ReactCytoscape, you can see I also have console.log messages (all prefixing the message with "CY:") to show when the unmounting and updating happens. The ReactCytoscape component gets unmounted/mounted whenever those drop downs are changed.
3. The cy reference is already being cached in the ReactCytoscape component see https://github.com/kiali/kiali-ui/pull/301/files#diff-ea9ca45724aa88eabf87c9afab923616R38 . But once that component is unmounted, that cached reference obviously goes away and that's why things aren't working like we want.

So, in short, yes the cy object is being cached, but because (for some reason) the ReactCytoscape/CytoscapeLayout is getting unmounted things break.

I see things unmounting whenever the drop downs change.

screenshot from 2018-04-29 17-40-41

vnugent:
I think the conditional rendering in Patternfly Spinner may be culprit here.

Our code:
https://github.com/kiali/kiali-ui/blob/master/src/components/CytoscapeLayout/CytoscapeLayout.tsx#L132-L149

PatternFfy Spinner:
https://github.com/patternfly/patternfly-react/blob/master/src/components/Spinner/Spinner.js#L14

I can confirm that if you remove <Spinner/> and directly render ReactCytoscape the remounting won't happen.

jmazzitelli:
Yes, OK, taking out the Spinner does, indeed, avoid the remounting.

However, that seems to introduce a new problem.  The summary panel becomes completely blank when I do this (its simply rendered with all white space). So it looks like the summary panel has special code during the mount/unmount in order for it to show data. We would need to work on fixing that if we do indeed want to remove this spinner.

src/components/CytoscapeLayout/CytoscapeLayout.tsx
```diff
@@ -118,13 +97,47 @@ export default class CytoscapeLayout extends React.Component<CytoscapeLayoutProp
         this.handleMouseOut(cytoscapeEvent);
       }
     });
+
+    // when the graph is fully populated and ready, we need to perform additional things
     this.cy.ready((evt: any) => {
       if (!this.props.isReady) {
+        // Don't allow a large zoom if the graph has a few nodes (nodes would look too big).
+        if (this.cy.zoom() > 2.5) {
```
mtho11: Should this be a constant perhaps in the `config` object?

src/components/CytoscapeLayout/CytoscapeLayout.tsx
```diff
@@ -118,13 +97,47 @@ export default class CytoscapeLayout extends React.Component<CytoscapeLayoutProp
...
+  processGraphUpdate() {
+    console.log('CY: graph was updated and needs to be processed further');
+
+    // TODO: refactor the GraphBadge.XXX classes so the constructor doesn't take an
+    //       element. Pass ele to the buildBadge and destroyBadge funcs so we don't have to
+    //       create so many objects - we should just create one object per badge type
+    //       and let it build/destroy multiple badges.
+    this.cy.startBatch();
+    this.cy.nodes().forEach(ele => {
+      if (!this.props.badgeStatus.hideCBs && ele.data('hasCB') === 'true') {
```
mtho11: Truthy makes `=== 'true'` redundant.

src/components/CytoscapeLayout/CytoscapeLayout.tsx
jmazzitelli: This isn't my code - it came from another PR that tragically I had to move around because of git conflicts. So its not really for this PR, but its here because of git conflict resolution. I told you people to stop editing these files while I am in here :) See PR #298

src/components/CytoscapeLayout/CytoscapeLayout.tsx
jmazzitelli: this code is going to change in the near future (see the TODO a couple lines above), but ... duly noted.

jmazzitelli:
I want to show what this PR currently has. First, what appears to be working - notice the pan and zoom remain in effect and selected elements remain selected (and summary panel remains loaded) and the badges are shown/hidden when appropriate (see the "v1" and "ratings v1" nodes - the badges are on those) even after changing the toggles:

update-graph1 ogv

Notice when changing the drop downs, some things are weird. First, changing the drop down of the graph layout is probably OK, changing the complete layout should reset zoom and pan since the entire layout of where the nodes and edges go will be different. But look at the summary panel. The some numbers don't change (the table does seem to change, but the bar graph and spark lines look static most of the time):

update-graph2 ogv

jmazzitelli:
This is  likely going to be replaced with PR #317 - if so, this PR will eventually be closed without merging. Leaving here as-is for now in case something really bad goes wrong in that other PR and I want to come back to this one.

jmazzitelli:
this is obsolete - PR #329 took its place.

PR #302, opened 2018-04-30
[KIALI-610] Fix onClick in Toolbar Component

aljesusg:
There is a bug when you click the dropdown....it's loading the graph/metrics, this is because I propaged this.props to the select component so when you do the action onClick, this trigger the action.

## Before
before_onclick

## After
after_onclick

### DIFF
--- a/src/components/ToolbarDropdown/ToolbarDropdown.tsx
+++ b/src/components/ToolbarDropdown/ToolbarDropdown.tsx
@@ -8,7 +8,7 @@ type ToolbarDropdownProps = {
   nameDropdown?: string;
   initialValue: number | string;
   initialLabel: string | undefined;
-  onClick: PropTypes.func;
+  handleSelect: PropTypes.func;
   options: PropTypes.Map<string | number, string>;
 };
 
@@ -28,7 +28,7 @@ export class ToolbarDropdown extends React.Component<ToolbarDropdownProps, Toolb
 
   onKeyChanged = (key: any) => {
     this.setState({ currentValue: key, currentName: this.props.options[key] });
-    this.props.onClick(key);
+    this.props.handleSelect(key);
   };
 
   render() {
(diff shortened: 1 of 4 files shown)


aljesusg:
Thanks @lucasponce

src/components/ToolbarDropdown/ToolbarDropdown.tsx
```diff
@@ -8,7 +8,7 @@ type ToolbarDropdownProps = {
   nameDropdown?: string;
   initialValue: number | string;
   initialLabel: string | undefined;
-  onClick: PropTypes.func;
+  handleSelect: PropTypes.func;
```
mtho11: Just FYI: PropTypes are not needed with Typescript unless you are creating libraries consumed by js libraries. Since all of our stuff is Typescript it is not necessary. There is also a runtime penalty for using PropTypes (both runtime size and execution time).
https://stackoverflow.com/questions/41746028/proptypes-in-a-typescript-react-application

The exception is `PropTypes.context`

mtho11: This has been a confusing issue for me as well.

aljesusg: The problem was that the drop-down inside toolbardropdown component  has {...this.props} so drop-down had onChange and OnClick property.

aljesusg: About proptype @mtho11 I think that we should use them (I wrote about in the guide). I don't know what do you think about use proptype, flux or none but we shouldn't use proptype a in some components and no in others, it should be consistent in all the app (my opinion)

mtho11: From my understanding, PropTypes are not necessary with Typescript. I'll send out an email and we can discuss more. Maybe there is something I'm missing.

PR #303, opened 2018-04-30
KIALI-629 Use the ACE editor on Services Istio objects

lucasponce:
This PR introduces the ACE editor on Service Istio Object level.
In a preliminar iteration it follows style defined in the OpenShift Console.

From the Istio name object we can navigate to the editor like this:
image

image

### DIFF
--- a/src/pages/IstioRuleDetails/IstioRuleInfo.tsx
+++ b/src/pages/IstioRuleDetails/IstioRuleInfo.tsx
@@ -72,19 +72,62 @@ class IstioRuleInfo extends React.Component<RuleDetailsId, RuleInfoState> {
       });
   }
 
+  // Handlers and Instances have a type attached to the name with '.'
+  // i.e. handler=myhandler.kubernetes
+  validateParams(parsed: ParsedSearch): boolean {
+    if (!parsed.type || !parsed.name || this.state.actions.length === 0) {
+      return false;
+    }
+    let validationType = ['handler', 'instance'];
+    if (parsed.type && validationType.indexOf(parsed.type) < 0) {
+      return false;
+    }
+    for (let i = 0; i < this.state.actions.length; i++) {
+      let splitName = parsed.name.split('.');
+      if (splitName.length !== 2) {
+        continue;
+      }
+      // i.e. handler=myhandler.kubernetes
+      // innerName == myhandler
+      // innerType == kubernetes
+      let innerName = splitName[0];
+      let innerType = splitName[1];
+      if (
+        parsed.type === 'handler' &&
+        this.state.actions[i].handler.name === innerName &&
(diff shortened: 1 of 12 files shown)


src/pages/IstioRuleDetails/IstioRuleInfo.tsx
```diff
@@ -72,19 +72,62 @@ class IstioRuleInfo extends React.Component<RuleDetailsId, RuleInfoState> {
...
+    }
+    let validationType = ['handler', 'instance'];
+    if (parsed.type && validationType.indexOf(parsed.type) < 0) {
+      return false;
+    }
+    for (let i = 0; i < this.state.actions.length; i++) {
+      let splitName = parsed.name.split('.');
+      if (splitName.length !== 2) {
+        continue;
+      }
```
aljesusg: ...How is this working?  Don't you need to set else {break;} ? we

lucasponce: Good catch, yes, the split should be out of the loop and break if it fails, not iterate on all the actions.

lucasponce: @aljesusg fixed

aljesusg: Waiting for Travis to merge @lucasponce

src/pages/ServiceDetails/ServiceDetailsPage.tsx
```diff
@@ -39,6 +73,113 @@ class ServiceDetails extends React.Component<RouteComponentProps<ServiceId>, Ser
...
+  validateParams(parsed: ParsedSearch): boolean {
+    if (!parsed.type || !parsed.name) {
+      return false;
+    }
+    // Check we have the right parameter
+    let validateTypes = ['routerule', 'destinationpolicy', 'virtualservice', 'destinationrule'];
+    if (parsed.type && validateTypes.indexOf(parsed.type) < 0) {
+      return false;
+    }
+    if (parsed.type === 'routerule' && this.state.serviceDetailsInfo.routeRules) {
```
xeviknal: Is there a way to avoid repeating this if-else and for statement? In typescript can we do not-typed methods?
Thinking about a method `function find(untypedIstioObjectList, name) boolean`.

lucasponce: Yes, but then it means to not use types and treat those as plain javascript.
I considered that, but I prefer to maintain the strong types here as there are different concepts.

aljesusg: I was thinking about a switch case but there is a different loop in each condition so...

PR #305, opened 2018-04-30
KIALI-627 nameaspace label (part two)

jshaughn:
- use inline styles
- line up the two option toolbars


jshaughn:
screenshot

jshaughn:
By the way, originally @gbaufake had asked to have all whitespace removed left of the 'Namespace' label.  This was difficult because it is embedded in a sub-dive of the Toolbar component.  Also, if we're going to use the Toolbar component we should probably not change the l+f.  Instead, I opted to wrap the toggles in their own Toolbar, thus providing a consistent look and maintaining the Toolbar as-is.

PR #308, opened 2018-05-01
KIALI-632 graph UX changes

jshaughn:


jshaughn:
kiali-632

jmazzitelli:
If the goal was to make a UI look boring, congratulations, you succeeded :)

mwringe:
When selecting the service box, its not highlighting it like the document says (I think the background is suppose to be that light blue and the nodes grey).

I also can't easily see what is the selected node, it looks to similar to the other nodes (eg it has a light blue outline instead of a light grey outline). This is especially apparent if you zoom out a bit.

mwringe:
I think we are also missing the hover over being the same as selected, when I hover over a node, its outline doesn't change to the light blue.

jshaughn:
@mwringe As per feedback, I updated border-width from 1px to  2px for the selected element, to help better distinguish it from other highlighted nodes.

screenshot

jshaughn:
> When selecting the service box, its not highlighting it like the document says (I think the background is suppose to be that light blue and the nodes grey).

Yes, I made a comment about this is the UX doc.  That doesn't play with the node highlighting, you'll get everything in the box shaded the same blue.  Special casing the nodes in the box to be highlighted differently is, I think, a prohibitive amount of work.

jshaughn:
> I think we are also missing the hover over being the same as selected, when I hover over a node, its outline doesn't change to the light blue.

Also commented in the doc, doing that makes it almost impossible to distinguish between a selected and highlighted node.

mwringe:
Merging for now as we have most of the improvements requested by @serenamarie125 but I think we will need some more refining as we go forward.

PR #309, opened 2018-05-01
KIALI-381: new manual refresh button for metrics screen

vnugent:
- [x]  Added a manual refresh button.  Disable button when loading is in progress.
- [x]  Updated unit test and snapshots

image


src/components/MetricsOptions/MetricsOptionsBar.tsx
```diff
@@ -8,6 +8,7 @@ import { ToolbarDropdown } from '../ToolbarDropdown/ToolbarDropdown';
 interface Props {
```
mtho11: @aljesusg this is exactly what I think should be used instead of React PropTypes

aljesusg: Ok , My point is use or no use proptype in all the app and keep all similar. I need feedback about to improve the style code guide

PR #310, opened 2018-05-02
Kiali-654: Bring other Service Graph visibility layers into Redux

mtho11:
Enable the other service graph visibility layers to be Redux enabled.
Also, add some utility functions for eliminating some of the verbose-ness for reducer operations.
And add more type-safety.
This is all under the covers refactoring no real screen changes.


mtho11:
1. In a different PR I will add disabling the toolbar (as the 'toolbar' will change drastically in the next few days to something that looks much better).
2. More declarative format of toolbar (but just like above this will change much in next few days)

Right now just want to merge this step in and continue.

jmazzitelli:
I'm going to merge this. I see the TODOs, I assume you will be adding back in the URL params that were removed here. Also, the default for the "show circuit breaker" toggle is now false - I assume you meant to do this. But just in case you did not, I mention it here. Not a big deal to me either way.

PR #313, opened 2018-05-03
KIALI-504 Replace deprecated 'componentWillReceiveProps' 1/2

aljesusg:
Jira : [KIALI-504](https://issues.jboss.org/browse/KIALI-504)

Replace deprecated componentWillReceiveProps.

See Discussion in [comments](https://issues.jboss.org/browse/KIALI-504)

Documentation : [Link](https://reactjs.org/blog/2018/03/29/react-v-16-3.html)

Reporter cc @mtho11

### DIFF
--- a/src/components/ServiceHealth/HealthIndicator.tsx
+++ b/src/components/ServiceHealth/HealthIndicator.tsx
@@ -18,22 +18,24 @@ interface Props {
   rateInterval: number;
 }
 
-export class HealthIndicator extends React.PureComponent<Props, {}> {
+interface HealthState {
   globalStatus: H.Status;
   info: string[];
+}
+
+export class HealthIndicator extends React.PureComponent<Props, HealthState> {
+  static updateHealth = (health?: Health) => {
+    let infoArr: string[] = [];
+    return { info: infoArr, globalStatus: H.computeAggregatedHealth(health, info => infoArr.push(info)) };
+  };
+
+  static getDerivedStateFromProps(props: Props, state: HealthState) {
+    return HealthIndicator.updateHealth(props.health);
+  }
 
   constructor(props: Props) {
     super(props);
-    this.updateHealth(props.health);
-  }
-
-  componentWillReceiveProps(nextProps: Props) {
-    this.updateHealth(nextProps.health);
-  }
-
-  updateHealth(health?: Health) {
-    this.info = [];
-    this.globalStatus = H.computeAggregatedHealth(health, info => this.info.push(info));
+    this.state = HealthIndicator.updateHealth(props.health);
   }
 
   render() {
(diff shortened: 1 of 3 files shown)


aljesusg:
I think is all correct but I wanna wait to the reporter ^^

src/components/ServiceHealth/HealthIndicator.tsx
```diff
@@ -28,8 +28,8 @@ export class HealthIndicator extends React.PureComponent<Props, {}> {
     this.updateHealth(props.health);
   }
 
-  componentWillReceiveProps(nextProps: Props) {
-    this.updateHealth(nextProps.health);
+  componentDidUpdate() {
```
mtho11: Why replace `componentWillReceiveProps` with `componentDidUpdate` According to the React blog on [Component Lifecycle changes](https://reactjs.org/blog/2018/03/29/react-v-16-3.html#component-lifecycle-changes) it should be replaced with `getDerivedStateFromProps`
I have concerns with `componentDidUpdate` as it happens after the `render` method where the `componentWillReceiveProps` happens before `render`
See Diagram:
reactjs-components-life-cycle2

mtho11: @vnugent what do you think here?

aljesusg: I comment his in the JIRA, the point is that you can't call another method from a static method so or we make the fetch inside this static method or we need to use another lifecycle method @mtho11, I read this in the doc. Maybe I am wrong but IDE and debugger don't let me to do it ^^

src/pages/ServiceDetails/ServiceInfo.tsx
```diff
@@ -85,8 +85,10 @@ class ServiceInfo extends React.Component<ServiceDetailsId, ServiceInfoState> {
     return parsed;
   };
 
-  componentWillReceiveProps(props: ServiceDetailsId) {
-    this.setState(this.parseState(props.serviceDetails));
+  componentDidUpdate(prevProps: ServiceDetailsId) {
+    if (prevProps !== this.props) {
+      this.setState(this.parseState(this.props.serviceDetails));
```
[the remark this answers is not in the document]
josejulio: I think `parseState` and `sortRouteRulesByPrecedence` can be static also. 
The point here is that (not 100% sure, might need to read more) after the update is complete, `componentDidUpdate` will be called, and if you do a setState, another update will be called. I don't think we want that.

src/components/ServiceHealth/HealthIndicator.tsx
vnugent: I just looked at the previous version with  `componentWillReceiveProps()`.  If all `updateHealth()` does is computing various new values for render(), `updateHealth()` can just be called in `componentDidMount()`.  

For future improvement I'd change `updateHealth(Health)` to arrow function to make it really clear what `this` is referring to, and break it up for readability.  Without knowing what Health argument is, it'd be hard for someone else to debug the function.

src/pages/ServiceDetails/ServiceInfo.tsx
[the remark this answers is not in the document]
aljesusg: Me neither ^^ just playing with the lifecycle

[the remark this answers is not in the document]
josejulio: And yes, a call to `setState` will make the component update again [1].

> An update can be caused by changes to props or state. These methods are called when a component is being re-rendered:

[1] https://reactjs.org/docs/react-component.html#updating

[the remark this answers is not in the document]
mtho11: That is exactly what I was afraid of. Good confirmation @josejulio

mtho11:
Where are we sitting with this now?

aljesusg:
I am going to work in this now @mtho11 I as focused in the login

aljesusg:
About the removal of componentWillReceiveProps: we should be able to handle  with a combination of getDerivedStateFromProps and componentDidUpdate, in some cases only with componentDidUpdate [React Blog](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#examples)

aljesusg:
@mth11 @josejulio  Can you review this? I am going to make this in 2 steps because IstioConfigDetailsPage & ItemDescription have some promise related with componentWillReceiveProps method and they need more changes.

src/pages/ServiceGraph/ServiceGraphRouteHandler.tsx
```diff
@@ -88,8 +76,26 @@ export default class ServiceGraphRouteHandler extends React.Component<
         edgeLabelMode: nextEdgeLabelMode
       };
       sessionStorage.setItem(SESSION_KEY, JSON.stringify(newParams));
-      this.setState({ ...newParams });
+      return { ...newParams };
     }
+    return {};
```
josejulio: I think you should return null here:

> It should return an object to update the state, or null to update nothing.

see https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops

aljesusg: You are right , Changed !!! Thanks @josejulio

josejulio:
@mtho11 do you have anything to add?

aljesusg:
Close by https://github.com/kiali/kiali-ui/pull/486

PR #314, opened 2018-05-03
[KIALI-534] Login

aljesusg:
Require https://github.com/kiali/kiali/pull/218
I sent the PR for feedback
[Kiali-534](https://issues.jboss.org/browse/KIALI-534)

login

Links of Login Page
Github [Link](https://github.com/kiali)
Youtube [Link](https://www.youtube.com/channel/UCcm2NzDN_UCZKk2yYmOpc5w)

I wanna set this in configuration too.

cc @mtho11 @mwringe

### DIFF
--- a/src/__itests__/App.browser.test.tsx
+++ b/src/__itests__/App.browser.test.tsx
@@ -3,6 +3,15 @@ import puppeteer from 'puppeteer';
 let testBrowser: puppeteer.Browser;
 let page: puppeteer.Page;
 
+const host = 'localhost';
+const port = 5003;
+const username = 'admin';
+const password = 'admin';
+
+const getUrl = () => {
+  return 'http://' + host + ':' + port;
+};
+
 jest.setTimeout(30000);
 
 describe('browser_smoke', () => {
@@ -14,7 +23,18 @@ describe('browser_smoke', () => {
     });
     page = await testBrowser.newPage();
     await page.setViewport({ width: 1440, height: 900 });
-    await page.goto('http://admin:admin@localhost:5003/console/service-graph/bookinfo', { waitUntil: 'networkidle2' });
+    await page.goto(getUrl(), { waitUntil: 'load' });
+    await page.waitFor('input[name=username]');
+    await page.waitFor('input[name=password]');
+
+    await page.type('input[name=username]', username);
+    await page.type('input[name=password]', password);
+
+    await page.click('button[type="submit"]');
+  });
+
+  afterAll(() => {
+    testBrowser.close();
   });
 
   it('should load the app', async () => {
--- a/src/components/Nav/Navigation.tsx
+++ b/src/components/Nav/Navigation.tsx
@@ -10,11 +10,14 @@ import MessageCenter from '../../containers/MessageCenterContainer';
 import IstioConfigPage from '../../pages/IstioConfigList/IstioConfigListPage';
 import IstioConfigDetailsPage from '../../pages/IstioConfigDetails/IstioConfigDetailsPage';
 import HelpDropdown from './HelpDropdown';
+import UserDropdown from '../../containers/UserDropdownContainer';
 import ServiceDetailsPage from '../../pages/ServiceDetails/ServiceDetailsPage';
 import ServiceGraphRouteHandler from '../../pages/ServiceGraph/ServiceGraphRouteHandler';
 import ServiceListPage from '../../pages/ServiceList/ServiceListPage';
 import ServiceJaegerPage from '../../pages/ServiceJaeger/ServiceJaegerPage';
 import PfSpinnerContainer from '../../containers/PfSpinnerContainer';
+import LoginPage from '../../containers/LoginPageContainer';
+import store from '../../store/ConfigStore';
 
 const istioConfigPath = '/istio';
 export const istioConfigTitle = 'Istio Config';
@@ -33,6 +36,7 @@ const istioConfigRx = /\/namespaces\/(.*)\/istio\/(.*)/g;
 
 type PropsType = {
   location: any;
+  authenticated: boolean;
 };
 
 type StateType = {
@@ -100,7 +104,10 @@ class Navigation extends React.Component<PropsType, StateType> {
   };
 
   render() {
-    return (
+    store.subscribe(() => {
+      document.documentElement.className = this.props.authenticated ? 'layout-pf layout-pf-fixed' : 'login-pf';
+    });
+    return this.props.authenticated ? (
       <>
         <VerticalNav
           setControlledState={this.setControlledState}
@@ -113,6 +120,7 @@ class Navigation extends React.Component<PropsType, StateType> {
             <VerticalNav.IconBar>
               <MessageCenter.Trigger />
               <HelpDropdown />
+              <UserDropdown />
             </VerticalNav.IconBar>
             <MessageCenter drawerTitle="Message Center" />
           </VerticalNav.Masthead>
@@ -137,6 +145,8 @@ class Navigation extends React.Component<PropsType, StateType> {
           <Redirect to={serviceGraphPath} />
         </SwitchErrorBoundary>
       </>
+    ) : (
+      <LoginPage />
     );
   }
 
--- a/src/components/Nav/UserDropdown.tsx
+++ b/src/components/Nav/UserDropdown.tsx
@@ -1,0 +1,29 @@
+import * as React from 'react';
+import { Dropdown, Icon, MenuItem } from 'patternfly-react';
+type UserProps = {
+  username: string;
+  logout: () => void;
+};
+
+type UserState = {};
+
+class UserDropdown extends React.Component<UserProps, UserState> {
+  render() {
+    return (
+      <>
+        <Dropdown componentClass="li" id="user">
+          <Dropdown.Toggle useAnchor={true} className="nav-item-iconic">
+            <Icon type="pf" name="user" /> {this.props.username}
+          </Dropdown.Toggle>
+          <Dropdown.Menu>
+            <MenuItem id="usermenu_logout" onClick={() => this.props.logout()}>
+              <Icon type="pf" name="key" /> Logout
+            </MenuItem>
+          </Dropdown.Menu>
+        </Dropdown>
+      </>
+    );
+  }
+}
+
+export default UserDropdown;
--- a/src/pages/Login/LoginPage.tsx
+++ b/src/pages/Login/LoginPage.tsx
@@ -1,0 +1,121 @@
+import React from 'react';
+import { Alert, Row, Col, Form, FormGroup, FormControl, Button, HelpBlock } from 'patternfly-react';
+import PropTypes from 'prop-types';
+import { KEY_CODES } from '../../config';
+import SocialLink from '../../components/SocialLink/SocialLink';
+
+const kialiTitle = require('../../assets/img/kiali-title.svg');
+
+type LoginProps = {
+  user: { username: string; password: string } | undefined;
+  logging: boolean;
+  error: any;
+  message: string;
+  authenticate: (username: string, password: string) => void;
+};
+
+type LoginState = {
+  username: string;
+  password: string;
+};
+
+export default class LoginPage extends React.Component<LoginProps, LoginState> {
+  static contextTypes = {
+    store: PropTypes.object
+  };
+  constructor(props: LoginProps) {
+    super(props);
+
+    // reset login status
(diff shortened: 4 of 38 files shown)


mwringe:
> if authentication is false, there is not login page and app use user/password of config file, when authentication is true, after login username/passwords is stored in session and app get them.

I don't know if I understand this. What does authentication being true/false change?

I don't understand why disabling it changes where the username/password comes from. Isn't the username and password for basic auth always coming from the config file?

aljesusg:
If "devs" set authentication to false they'll work like now without enter user and password, the idea is let the dev an environment to work without login page ^^, so in this case the app get the username and password from config file.

If you set authentication to true so you'll have the login view each time you start the app and you'll need to set the username and password in the login form.

jmazzitelli:
Why do we want to take away the login page for devs? You know what is going to happen - we'll all have it set to false, the login page will break and we'll never know for weeks :)

I see no purpose for turning off authentication/login just for developers. It isn't a big deal to me anyway - I just use the browser feature that stores the username/password.

mwringe:
A couple of comments:

1) hitting enter doesn't work when trying to login. You currently have to click the login button

2) we probably don't want to have 'PatternFly' as the title of the login page. We can fix this simply by using the existing Kiali title we have, but its not perfect as is:
[login.TXT](https://github.com/kiali/kiali-ui/files/1995602/login.TXT)

Other than that, this looks awesome :) Can't wait for it to be ready to be merged once we get the green light from the CI

aljesusg:
I'm trying to improve the coverage but it's hard test redux ^^ I am gonna fix the enterkey event ,tabkey and title. Thanks for feedback @mwringe

josejulio:
Does this breaks the screenshot feature that @vnugent added?

aljesusg:
I sent an email to kiali-dev about login and kiali-bot.

aljesusg:
@vnugent Kiali-bot is working now ;)

aljesusg:
Since last rebase Travis gets stuck in the App test giving a timeout. Is there any change related with?

aljesusg:
yarn itest en my local environment the screenshot output is this
kiali-browser-test

vnugent:
Fyi itest in Travis uses the public Kiali for the backend.

src/__itests__/App.browser.test.tsx
```diff
@@ -3,18 +3,40 @@ import puppeteer from 'puppeteer';
...
+    sessionStorage.setItem('user', 'admin');
     testBrowser = await puppeteer.launch({
       headless: true,
       slowMo: 300,
       args: ['--no-sandbox', '--disable-setuid-sandbox']
     });
     page = await testBrowser.newPage();
     await page.setViewport({ width: 1440, height: 900 });
-    await page.goto('http://admin:admin@localhost:5003/console/service-graph/bookinfo', { waitUntil: 'networkidle2' });
+    await page.goto(getUrl(), { waitUntil: 'load' });
```
lucasponce: Just curiosity. How do these async tests work ? They wait until the condition is matched for a timeout ?.

aljesusg: you can set the reserved word await or not, If you set wait the code wait until the page go to URL in this case

aljesusg: HEre is the documentation about https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md so I added aditional requirement that is wait until load the full page, (this is { waitUntil: 'load' }) because I need the form rendered for fill the inputs with the username and password

src/actions/UserAction.ts
```diff
@@ -0,0 +1,39 @@
...
+export const success = user => {
+  return { type: userConstants.LOGIN_SUCCESS, user };
+};
+
+export const failure = error => {
+  return { type: userConstants.LOGIN_FAILURE, error };
+};
+
+export const login = (username: string, password: string) => {
+  return dispatch => {
```
lucasponce: I'm mostly learning with this review, as I haven't worked with Redux yet. So, this "dispatch" function here is the convention of redux for actions. Am I right ?

aljesusg: Yes, you can see an example here https://egghead.io/lessons/react-redux-store-methods-getstate-dispatch-and-subscribe about getstate dispatch and subscribe

src/components/Nav/Navigation.tsx
```diff
@@ -99,41 +101,52 @@ class Navigation extends React.Component<PropsType, StateType> {
...
+              <Route path={servicesJaegerPath} component={ServiceJaegerPage} />
+              <Route path="/namespaces/:namespace/services/:service" component={ServiceDetailsPage} />
+              <Route path={istioConfigPath} component={IstioConfigPage} />
+              <Route path="/namespaces/:namespace/istio/:objectType/:object" component={IstioConfigDetailsPage} />
+              <Redirect to={serviceGraphPath} />
+            </SwitchErrorBoundary>
+          </div>
+        ) : (
+          <div>
+            <ConnectedLoginPage dispatch={null} />
```
lucasponce: Perhaps I am wrong, I do not remember how ended the discussion about using null vs undefined.
I just see this one, but please, ignore it if it doesn't make sense.

aljesusg: No .... you're right I remember some discussion about I'll change this

src/components/Nav/UserDropdown.tsx
```diff
@@ -0,0 +1,34 @@
...
+type UserProps = {
+  dispatch?: any;
+};
+
+class UserDropdown extends React.Component<UserProps, {}> {
+  userLogout = () => {
+    // const { dispatch } = this.props;
+    sessionStorage.removeItem('user');
+    // dispatch(UserAction.logout());
+    window.location.reload();
```
lucasponce: Invoking the window is always like access to bytecode.
Perhaps this can be done through the React Router ?

aljesusg: I asked about this when I opened the PR but nobody answer me about How I can do it? Because I tried but I couldn't

aljesusg: Should I create a static context and call to this.context.router.transitionTo(/path-to-link); ?

aljesusg: I fixed...I don't know why I couldn't use it before

src/pages/Login/LoginPage.tsx
```diff
@@ -0,0 +1,170 @@
...
+          messageError: actualState.message
+        });
+      }
+      if (actualState.loggedIn) {
+        this.setState({
+          submitted: false,
+          showError: false
+        });
+        window.location.reload();
+      }
```
aljesusg: I'll change this too

aljesusg:
I did the changes @lucasponce

src/components/Nav/Navigation.tsx
```diff
@@ -99,41 +101,52 @@ class Navigation extends React.Component<PropsType, StateType> {
   };
 
   render() {
+    let authenticated = sessionStorage.getItem('user');
```
lucasponce: I have one security comment here. So, authentication state is stored in sessionStorage. Is it possible to add this step from the login page and bypass this ? If I understood, with the debug tools of redux, I can locally submit an item called 'user' and that would be make the page render ?

aljesusg: The idea is in the next steps store this in redux. Because we need username and password for all the API calls I opened https://issues.jboss.org/browse/KIALI-763 for this, the next step is prepare the backend to make the authentication through openshift/kubernetes and work with a token.

aljesusg:
Session moved to redux @lucasponce ;)

src/components/Nav/Navigation.tsx
```diff
@@ -98,42 +103,68 @@ class Navigation extends React.Component<PropsType, StateType> {
...
-        >
-          <Route path="/service-graph/:namespace" component={ServiceGraphRouteHandler} />
-          <Route path={servicesPath} component={ServiceListPage} />
-          <Route path={servicesJaegerPath} component={ServiceJaegerPage} />
-          <Route path="/namespaces/:namespace/services/:service" component={ServiceDetailsPage} />
-          <Route path={istioConfigPath} component={IstioConfigPage} />
-          <Route path="/namespaces/:namespace/istio/:objectType/:object" component={IstioConfigDetailsPage} />
-          <Redirect to={serviceGraphPath} />
-        </SwitchErrorBoundary>
+        {this.state.authenticated ? (
```
jotak: I think you can remove some unnecessary div here to reduce the DOM depth, with something like:
```
    return this.state.authenticated ? (
      <>
        <VerticalNav
           blablabla
      </>
    ) : <ConnectedLoginPage dispatch={undefined} />;
```

src/components/Nav/UserDropdown.tsx
```diff
@@ -0,0 +1,57 @@
...
+type UserState = {
+  username: string;
+};
+
+class UserDropdown extends React.Component<UserProps, UserState> {
+  static contextTypes = {
+    router: PropTypes.object
+  };
+
+  constructor(props: any) {
```
jotak: props is UserProps instead of any?

src/pages/Login/LoginPage.tsx
```diff
@@ -0,0 +1,169 @@
...
+                      <Button
+                        type="submit"
+                        onKeyPress={this.handleKeyPress}
+                        className="btn btn-primary btn-block btn-lg"
+                      >
+                        Log In
+                      </Button>
+                    </Form>
+                  </div>
+                  <footer className={'login-pf-page-footer'}>
```
jotak: Suggestion, not blocking (we can do it via a later PR): we could group all these "social links" into a dedicated component, that we would reuse in the about page too.
It could also include twitter and medium (we have a medium, don't we?)

src/pages/Login/LoginPage.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+const kialiTitle = require('../../assets/img/kiali-title.svg');
+
+type LoginProps = {
+  dispatch: any;
+};
+
+type LoginState = {
+  username: string;
+  password: string;
```
jotak: I wonder how storing password in state and/or redux could be vulnerable to some XSS attack. Shouldn't we store a cookie instead? Or is it planned for a future task?

aljesusg:
@jotak what do you think about this Social component ?  You can add new socialMedia in the config file

jotak:
@aljesusg nice!

vnugent:
@aljesusg  I checked out your PR and was able to reproduce the failure.  It's not happening in master though.  Still investigating...

aljesusg:
Thanks @vnugent , if you remove the access to store from services/API the test pass....anyway the UI runs without problems...so I don't know where is the problem. It has no sense.

josejulio:
Could you rebase? We removed coverage support (for now) to prevent travis to hang and fail.

aljesusg:
the problem was the Graphfilter test @josejulio not the coverage , it fail when I try to access redux from services/API I don't know why. Anyway I wanna refactor this PR and make some container to mapping the props and state too.

aljesusg:
Code refactored, still investigating about graphfiltertoolbar

aljesusg:
I found the problem.

NamespaceAction use store and call to services/api that call again to store ...it's a cyclomatic problem so we need to move services/api to action or pass in each view authentication

aljesusg:
@vnugent  we should deploy @kiali-bot after merge this.

mtho11:
@aljesusg I think @vnugent is out for a couple more days.

src/components/Nav/Navigation.tsx
```diff
@@ -99,7 +104,12 @@ class Navigation extends React.Component<PropsType, StateType> {
   };
 
   render() {
-    return (
+    store.subscribe(() => {
```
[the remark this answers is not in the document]
aljesusg: I did the changes, what do you think about change the style in the reducer?

aljesusg:
@mtho11 I did all the changes. Can we merge this?

aljesusg:
Rebased

mtho11:
I tested login/logout looks good.

aljesusg:
Thanks @mtho11

mtho11:
There is one important thing that needs to get addressed in another PR:
Inside `src/reducers/LoginState.ts`. There can only be state changes in a reducer no side-effects because they are pure reducer functions.  So changing the UI `document.documentElement.className` is a no-no. The LoginState reducer should only mutate state and make no changes to the UI ever.

PR #315, opened 2018-05-03
KIALI-642 Connects MessageCenter with Redux

josejulio:
peek 2018-05-04 12-51

This shows the error replaced from the "graph" page.
Are we moving those alerts to the MessageCenter?

//cc @mwringe @vnugent


josejulio:
Not sure if i missed with something but i'm getting this error:
> sh: 1: react-scripts-ts: not found

josejulio:
I updated all the pages errors (that found):

peek 2018-05-04 12-51

jmazzitelli:
FYI: I am reviewing this now. Going to look at this in depth (mainly so I can continue to immerse myself in more redux :) so it will take me a while. Should be done with the review today though.

src/pages/ServiceList/ServiceListPage.tsx
```diff
@@ -1,43 +1,20 @@
...
-    this.setState({ alertVisible: true, alertDetails: error });
-  };
-
-  dismissAlert = () => {
-    this.setState({ alertVisible: false });
+    MessageCenter.add(error);
   };
 
   render() {
     let alertsDiv = <div />;
```
jmazzitelli: since alerts are gone, should we remove this div completely (or at minimum take it out and just put \<div/> directly in the return value?

src/pages/ServiceList/ServiceListPage.tsx
```diff
@@ -1,43 +1,20 @@
...
-    if (this.state.alertVisible) {
-      alertsDiv = (
-        <div>
-          <Alert onDismiss={this.dismissAlert}>{this.state.alertDetails.toString()}</Alert>
-        </div>
-      );
-    }
     return (
       <div className="container-fluid container-pf-nav-pf-vertical">
         <h2>Services</h2>
```
jmazzitelli: see the {alertsDiv} below - as per my comment above, this should at minimum be renamed or just hardcode a \<div/> here - or I think maybe even just remove this entirely.

src/reducers/MessageCenter.ts
```diff
@@ -0,0 +1,101 @@
...
+      messages: []
+    }
+  ],
+  hidden: true,
+  expanded: false,
+  expandedGroupId: 'default'
+};
+
+const mergeToState = (prevState, toMerge) => {
+  return Object.assign({}, prevState, toMerge);
```
jmazzitelli: can this be shortened to just the following?

```const mergeToState = (prevState, toMerge) => ( { ...prevState, toMerge } )```

File: src/actions/MessageCenterActions.ts [hunk 4]
```diff
@@ -0,0 +1,84 @@
...
+        dispatch(MessageCenterActions.hideMessageCenter());
+      }
+      return Promise.resolve();
+    };
+  },
+  markGroupAsRead: (groupId: number) => {
+    return (dispatch, getState) => {
+      const state = getState();
+      const foundGroup = state.messageCenter.groups.find(group => group.id === groupId);
+      console.log(foundGroup);
```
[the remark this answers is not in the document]
josejulio: Actually, I'm not planing to log that out, but maybe i could log if i didn't find a group.

src/pages/ServiceList/ServiceListPage.tsx
josejulio: Missed that one, yep, we should remove that one.

src/reducers/MessageCenter.ts
josejulio: Yes! to `const mergeToState = (prevState, toMerge) => ( { ...prevState, ...toMerge } )`

src/components/MessageCenter/MessageCenterTrigger.tsx
```diff
@@ -0,0 +1,23 @@
+import * as React from 'react';
+import * as PfReact from 'patternfly-react';
+
+type StateType = {};
+type PropsType = {
+  newMessagesCount: number;
+  toggleMessageCenter: () => void;
+};
+
+export default class MessageCenterTrigger extends React.PureComponent<PropsType, StateType> {
```
mtho11: I personally prefer for stateless components:
`const PfHeader: React.SFC<PfHeaderProps> = props => {
  return <div className="page-header">{props.children}</div>;
};
export default PfHeader`
That way you aren't even allowed to add state. Which most all of the state will be in Redux now anyway.

josejulio:
> In addition to my comments, I would also say that I think you missed a spot. See:
> 
> https://github.com/kiali/kiali-ui/blob/master/src/pages/ServiceGraph/ServiceGraphPage.tsx#L16-L20

I didn't, see: https://github.com/kiali/kiali-ui/pull/315/files#diff-bf5d5931e8a56754ac7cdfce71034a24L16

src/containers/MessageCenterContainer.tsx
```diff
@@ -0,0 +1,49 @@
...
+    onHideDrawer: () => dispatch(MessageCenterActions.hideMessageCenter()),
+    onToggleGroup: group => dispatch(MessageCenterActions.toggleGroup(group.id)),
+    onMarkGroupAsRead: group => dispatch(MessageCenterActions.markGroupAsRead(group.id)),
+    onClearGroup: group => dispatch(MessageCenterActions.clearGroup(group.id)),
+    onNotificationClick: message => dispatch(MessageCenterActions.markAsRead(message.id)),
+    onDismissNotification: () => console.log('dismiss')
+  };
+};
+
+const mapStateToPropsMCT = state => {
```
mtho11: What does MCT stand for? Might as well spell it out. It will get compiled out at production so there is no penalty for it.

src/reducers/MessageCenter.ts
mtho11: Personally, I prefer the ES6 spread operator because the syntax seem cleaner (just my preference though). 
We could add `tassign` if we want `Object.assign` to be typed.

src/components/MessageCenter/MessageCenterTrigger.tsx
josejulio: I didn't know what was our stand for using functional components.
I can change that if is OK to mix then both (functional components vs class components)

src/containers/MessageCenterContainer.tsx
josejulio: MessageCenterTrigger, I can be more verbose sure.

src/reducers/MessageCenter.ts
josejulio: Used the spread operator in latest commit.

PR #318, opened 2018-05-08
Update Project packages for this sprint 5.

mtho11:
Had to back out an upgrade to Redux 4.0 as things were failing. So still at Redux 3.7.2


josejulio:
> Ok, things work locally but not on travis?

`Error: package-lock.json is not in sync with package.json`

Can you try to run `npm install` again and see if your -lock file changes? It seems that travi's npm doesn't agree with the current package-lock.json

mtho11:
@josejulio after `npm install`my lock file doesn't change so nothing new to check-in here

mtho11:
Maybe its because I'm using a different version of npm? I'm using npm version 6.0.0

mtho11:
I deleted `node_modules` and did an `npm i`. No change.

josejulio:
> Maybe its because I'm using a different version of npm? I'm using npm version 6.0.0

Might be that, travis also use 5.6.0
https://travis-ci.org/kiali/kiali-ui/builds/376145463#L451

mtho11:
So it was the npm version. Npm version 6.0.0 produces a different `lock` file than 5.6.0.

mtho11:
Ok someone can push the green button.

PR #320, opened 2018-05-08
[KIALI-692] Tests for ToolbarDropdown Component

aljesusg:
Tests for this component
JIRA: [KIALI-692](https://issues.jboss.org/browse/KIALI-692)

### DIFF
--- a/src/components/ToolbarDropdown/ToolbarDropdown.tsx
+++ b/src/components/ToolbarDropdown/ToolbarDropdown.tsx
@@ -46,3 +46,5 @@ export class ToolbarDropdown extends React.Component<ToolbarDropdownProps, Toolb
     );
   }
 }
+
+export default ToolbarDropdown;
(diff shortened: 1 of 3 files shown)


abonas:
I'm wondering how test coverage didn't change if the PR is adding tests, do you have any idea?

aljesusg:
@abonas  n the Coverage build ToolbarDropdown is 100% but in master build coverage is 100% too so....why this happens?

Maybe the snapshoot is taking 100% coverage ...so if there is a component that has a toolbardropdown like a children component, this component will be covered. I'll do some tests about this.

captura de pantalla 2018-05-09 a las 16 46 12

josejulio:
Coveralls gets the report from running:
`npm test -- --coverage`
Which creates the file `./coverage/lcov.info` and is sent to coveralls
`cat ./coverage/lcov.info | coveralls` [1]

Can you check if that command is reporting it unexpected values too?

[1] https://github.com/kiali/kiali-ui/blob/master/.travis.yml#L16

josejulio:
https://github.com/kiali/kiali-ui/blob/master/src/components/MetricsOptions/__tests__/MetricsOptionsBar.test.tsx#L27 may be indirectly testing some of that code.

aljesusg:
Yes...but I think that we should merge this tests, if the MetricOptions view is removed the coverage will decrease so much because include this component.

aljesusg:
Thanks @josejulio

PR #324, opened 2018-05-09
Add ActionCreator tests to test the Actions 

mtho11:
Test the Redux ActionCreators to ensure the actions they create are what we think.

The testing part of https://issues.jboss.org/browse/KIALI-654

### DIFF
--- a/src/actions/ServiceGraphFilterActions.ts
+++ b/src/actions/ServiceGraphFilterActions.ts
@@ -8,7 +8,7 @@ export enum ServiceGraphFilterActionKeys {
   TOGGLE_GRAPH_CIRCUIT_BREAKERS = 'TOGGLE_GRAPH_CIRCUIT_BREAKERS',
   TOGGLE_GRAPH_ROUTE_RULES = 'TOGGLE_GRAPH_ROUTE_RULES',
   // Disable Actions
-  DISABLE_GRAPH_LAYERS = 'DISABLE_GRAPH_LAYERS'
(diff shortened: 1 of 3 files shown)


src/actions/ServiceGraphFilterActions.ts
```diff
@@ -8,17 +8,19 @@ export enum ServiceGraphFilterActionKeys {
   TOGGLE_GRAPH_CIRCUIT_BREAKERS = 'TOGGLE_GRAPH_CIRCUIT_BREAKERS',
   TOGGLE_GRAPH_ROUTE_RULES = 'TOGGLE_GRAPH_ROUTE_RULES',
   // Disable Actions
-  DISABLE_GRAPH_LAYERS = 'DISABLE_GRAPH_LAYERS'
+  ENABLE_GRAPH_FILTERS = 'ENABLE_GRAPH_FILTERS'
 }
 
+const keys = ServiceGraphFilterActionKeys;
```
josejulio: Why is this indirection preferred over the full name?

mtho11: It is totally not needed. I have seen many of you trying to shorten the names so I guess this is me caving in. I actually don't like it.
I say spell everything out. But I know this is not the practice in Go.

josejulio: Then let's spell everything out! Lets not cave in to bad practices.

PR #327, opened 2018-05-10
Kiali 589 remove toggle button filters and replace with popover checkboxes

mtho11:
The current toggle button visibility filters doesn't look so great on laptop screens and there are some issues with clicking around fast.
This PR deletes the filter toolbar and just places the checkboxes in a popover to conserve screen space (and make it more appealing).

checkbox-filters

https://issues.jboss.org/browse/KIALI-589


mtho11:
We know this will most likely get ripped out when UXD is back. However, this solves the hideous toggles bars dominating the screen (and the other clicking issues associated with them). And the issue with the toggles resizing to different sizes at seeming random times.

PR #328, opened 2018-05-10
[WIP] Add PfSpinner Redux Component since it is used in multiple places.

mtho11:
This is the first step of Reduxifying the spinner.

- [ ]  Add Redux to spinner

### DIFF
--- a/src/components/Pf/PfSpinner.tsx
+++ b/src/components/Pf/PfSpinner.tsx
@@ -1,0 +1,12 @@
+import React from 'react';
+
+type PfSpinnerProps = {
+  // none yet
+};
+
+const PfSpinner: React.SFC<PfSpinnerProps> = props => {
+  // It is more than likely it won't have any children; but it could.
+  return <div className="spinner spinner-sm left-spinner">{props.children}</div>;
+};
+
+export default PfSpinner;
--- a/package-lock.json
+++ b/package-lock.json
@@ -4422,7 +4422,6 @@
       "version": "1.2.3",
       "resolved": "https://registry.npmjs.org/fsevents/-/fsevents-1.2.3.tgz",
       "integrity": "sha512-X+57O5YkDTiEQGiw8i7wYc2nQgweIekqkepI8Q3y4wVlurgBt2SuwxTeYUYMZIGpLZH3r/TsMjczCMXE5ZOt7Q==",
-      "optional": true,
       "requires": {
         "nan": "2.10.0",
         "node-pre-gyp": "0.9.1"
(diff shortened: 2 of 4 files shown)


src/components/Pf/PfSpinner.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+type PfSpinnerProps = {
+  // none yet
+};
+
+const PfSpinner: React.SFC<PfSpinnerProps> = props => {
```
josejulio: I would like to note that this is preferred over functional components for type safety.
I didn't know about this class, just read about it

https://medium.com/@iktakahiro/react-stateless-functional-component-with-typescript-ce5043466011

josejulio:
are we going to replace [patternfly-react.Spinner](https://rawgit.com/patternfly/patternfly-react/gh-pages/index.html?knob-Loading=true&knob-Size=md&selectedKind=patternfly-react%2FWidgets&selectedStory=Spinner&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs) in favor of this one?

josejulio:
Also, about adding Redux to the spinner, isn't it an overhead? I think the spinner shouldn't have any state, but only props at most to know if is "spinning" and manage the logic outside.
Maybe you meant to add the Redux to the loaders?

mtho11:
@josejulio Yes, it will just be a redux containerized wrapper around the patternfly-react spinner.

mtho11:
@josejulio no overhead to adding redux to the spinner.
The intention is to have an intelligent component that is just focused around reacting to anything setting the `isLoading` redux state. This removes all the logic from every screen having if/then logic and trying to manipulate the loading indicator. Instead, just include the component in your JSX and set the `isLoading` global state and the component knows what to display (and its managed all in one place).
So all this logic is gone from individual pages:
`if (this.state.loading) {
      ruleListComponent = <PfSpinner />;
    } else {
      ruleListComponent = (
        <div>
....
`
And instead replaced by a single component that knows what to do based on state.
DRY's everything up.

josejulio:
The only thing that I don't like (if i understood correctly) is that everyone will be using the same "isLoading" state, thus if something is loading in the sidebar, the graph will also be in loading state.

mtho11:
@josejulio using the same global 'isLoading' state is where we would need `Promise.all` or more advanced features of Rx.js `observables` to coordinate the combination of promises and their completion.

mtho11:
The real point of this PR is to allow our middleware to handle displaying the spinner and not have every screen section implement this.

mtho11:
In an effort to escape rebasing hell, we should probably merge sections when they are ready. This is a small section that could be merged now (although not finished yet obviously).

PR #329, opened 2018-05-11
KIALI-341 gracefully update graph

jmazzitelli:
This is basically working. There are some additional things that we might want to do now, or get this PR merged and do them later:

The layout dropdown works now, so we don't have to wait for reduxification of the layout dropdown - but when it is reduxified, there are some small changes to be made (we need to add the layout to the map state-to-props function).

We should investigate if we want to (and even if its possible) get the graph to be reused when changing Duration dropdown and Namespace dropdown. I suspect changing Namespace we would want the graph to completely get recreated since presumably the graph will be very different and we just want to "start fresh". It is possible the same thing is true for Duration since changing from 1 minute to 7 days might drastically change the graph and again we might want to "start fresh" in that case too.

The Refresh button also re-creates the cy graph. We might want to look into that, too.


jmazzitelli:
the layout dropdown now works. when we redux-ify the layout dropdown there is some minor changes to make. doesn't have to be in this PR though.

jmazzitelli:
Miracle of miracles - it is green.

src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -41,53 +40,100 @@ export interface CytoscapeMouseOutEvent extends CytoscapeBaseEvent {}
...
     return (
       this.props.isLoading !== nextProps.isLoading ||
       this.props.graphLayout !== nextProps.graphLayout ||
       this.props.showEdgeLabels !== nextProps.showEdgeLabels ||
       this.props.showNodeLabels !== nextProps.showNodeLabels ||
       this.props.showCircuitBreakers !== nextProps.showCircuitBreakers ||
-      this.props.showRouteRules !== nextProps.showRouteRules
+      this.props.showRouteRules !== nextProps.showRouteRules ||
+      this.props.elements !== nextProps.elements ||
+      this.props.graphLayout !== nextProps.graphLayout
```
josejulio: If we use `React.PureComponent` we get this for free (not a blocker!)

https://reactjs.org/docs/react-api.html#reactpurecomponent

src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -41,53 +40,100 @@ export interface CytoscapeMouseOutEvent extends CytoscapeBaseEvent {}
...
+    console.log('CY: updating graph component');
+    this.processGraphUpdate(this.getCy());
+  }
+
+  render() {
+    console.log('CY: rendering graph component');
+
+    return (
+      <div id="cytoscape-container" style={{ marginRight: '25em', height: '100%' }}>
+        <Spinner loading={this.props.isLoading}>
```
josejulio: I think this is the reason of not having the ref in the first didMount, it was "loading", there wasn't any graph yet.

jmazzitelli:
Here's the current state of this PR - I think its ready. Any issues are probably minor and can be fixed in followup PRs. But since this has alot of changes, we should merge and get it into master to avoid conflicts with others.

graceful-update-graph

mtho11:
We probably need some more testing here but that can wait for another PR.

PR #330, opened 2018-05-11
KIALI-661: Graph badges for missing sidecars

israel-hdez:
Depends on https://github.com/kiali/kiali/pull/188

- Show a badge if API reports missing sidecars for a service. Badge is red to denote a problematic state.
image

- List the badge of missing sidecars in the node summary.
image

- Use CSS 'em' units instead of CSS 'transform' to let the browser so vector-based rendering instead of pixel-based rendering of the badges. Now graph badges are sharper regardless of zoom:
image

### DIFF
--- a/src/components/CytoscapeGraph/graphs/GraphBadge.ts
+++ b/src/components/CytoscapeGraph/graphs/GraphBadge.ts
@@ -3,6 +3,7 @@ import { PfColors } from '../../../components/Pf/PfColors';
 
 const FLASH_BADGE: string = 'fa fa-bolt';
 const ROUTE_BADGE: string = 'fa fa-code-fork';
+const SIDECAR_BADGE: string = 'pf pficon-blueprint';
 
 // Each node that has a badge will have custom data associated with it.
 // Each entry in the custom data is keyed on the badge type; an entry is itself
@@ -44,7 +45,7 @@ class GraphBadge {
 
     const setScale = () => {
       const zoom = node.cy().zoom();
-      div.style.transform = div.style.transform + `scale(${zoom},${zoom})`;
+      div.style.fontSize = `${zoom}em`;
     };
 
     const popper = node.popper({
@@ -125,5 +126,11 @@ export class CircuitBreakerBadge extends GraphBadge {
 export class RouteRuleBadge extends GraphBadge {
   constructor() {
     super(ROUTE_BADGE, PfColors.Purple300, 'top');
+  }
+}
+
+export class MissingSidecarsBadge extends GraphBadge {
+  constructor() {
+    super(SIDECAR_BADGE, PfColors.Red100, 'bottom');
   }
 }
--- a/src/pages/ServiceGraph/SummaryPanelNode.tsx
+++ b/src/pages/ServiceGraph/SummaryPanelNode.tsx
@@ -130,7 +130,7 @@ export default class SummaryPanelNode extends React.Component<SummaryPanelPropTy
               color={PfColors.Green500}
             />
           </div>
-          {this.renderBadgeSummary(node.data('hasCB'), node.data('hasRR'))}
+          {this.renderBadgeSummary(node.data('hasCB'), node.data('hasRR'), node.data('hasMissingSidecars').toString())}
         </div>
         <div className="panel-body">
           <InOutRateTable
@@ -170,9 +170,10 @@ export default class SummaryPanelNode extends React.Component<SummaryPanelPropTy
     );
   };
 
-  private renderBadgeSummary = (hasCB: string, hasRR: string) => {
+  private renderBadgeSummary = (hasCB: string, hasRR: string, hasMissingSC: string) => {
     const displayCB = hasCB === 'true';
     const displayRR = hasRR === 'true';
+    const displayMissingSC = hasMissingSC === 'true';
     return (
       <>
         {displayCB && (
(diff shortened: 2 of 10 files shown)


jshaughn:
@israel-hdez , nice, the badges look much better.

src/components/CytoscapeGraph/graphs/GraphBadge.ts
```diff
@@ -127,3 +128,9 @@ export class RouteRuleBadge extends GraphBadge {
     super(ROUTE_BADGE, PfColors.Purple300, 'top');
   }
 }
+
+export class MissingSidecarsBadge extends GraphBadge {
+  constructor() {
+    super(SIDECAR_BADGE, PfColors.Red100, 'bottom');
```
jshaughn: I'm not UX but I would suggest keeping the position the same as the other badges, otherwise it gets noisy and badges float around all around the node.  Also, I think I prefer keeping the colors the same, but then again, this badge indicates a problem, so perhaps red is OK.

abonas: @serenamarie125 ?

pilhuhn:
Is that case even possible? Istio recording traffic to a service without sidecar?

jotak:
It seems we already have an icon used to show presence of istio sidecar: see https://github.com/kiali/kiali-ui/pull/228
That would be 2 different icons for very similar purpose?

israel-hdez:
@pilhuhn IIRC, it's not possible if all pods backing a service have missing sidecars. However, if _some_ (not all) of the pods have missing sidecars, only some traffic will be logged. Then, the goal of the badge is to inform this condition.

@jotak Oh! good point. So, the badge can be a red boat :)

pilhuhn:
On 14 May 2018, at 20:03, Edgar Hernández wrote:

> @jotak Oh! good point. So, the badge can be a red boat :)

I was thinking the same as @jotak. A red boat -- or better a boat with
a red bar through it (think no-parking sign). Red vs blue boat only
may be bad for visually impaired people.

-- 
hwr@pilhuhn.de  m:0179/207 4919  b:http://pilhuhn.blogspot.com

src/components/CytoscapeGraph/graphs/GraphBadge.ts
israel-hdez: @jshaughn Yes, we need some UX feedback here.
I think to fix badge positioning we will need a Jira, because only fixed positioning is available now. With the current code, they will look too far if there aren't any other badges or with a space in between if it's only the CI visible:
image

I can use this positioning for now if you think it's better while we work with UX in a new Jira.

jshaughn: @israel-hdez Maybe someone changed something, After adding the route rule badge I thought I had it such that badges would pull left, and then line up to the right.  I don't care too much for the icons/badging to be honest.

abonas:
regarding red boat, not only could it be an issue for accessibility, but potentially an issue with branding (changing color might be an issue)

src/components/CytoscapeGraph/graphs/GraphBadge.ts
israel-hdez: @jshaughn I'll leave the positioning as is for now... We could think about changing it when it becomes a problem due the amount of badges or when we have UX feedback.

israel-hdez:
@jshaughn True. I added it in the summary and updated PR description with the image.

israel-hdez:
@abonas I think we do the opposite way because in the graph we want to make it easy to spot problems or  misconfigurations. If we show badges when sidecars are present, most nodes of the graph would have it making it harder to spot the ones without the badge.

If I understand correctly, a pod without sidecar can be problem because we cannot correctly report traffic, topology, health, etc for that pod and the service backed by the pod. 

BTW, the badges doesn't currently support images, but only Patternfly icons. So, I would like to get this PR in and change the icon in a follow up PR if we resolve that the badge should be something else.

src/pages/ServiceGraph/SummaryPanelNode.tsx
```diff
@@ -170,7 +170,7 @@ export default class SummaryPanelNode extends React.Component<SummaryPanelPropTy
     );
   };
 
-  private renderBadgeSummary = (hasCB: string, hasRR: string) => {
+  private renderBadgeSummary = (hasCB: string, hasRR: string, hasMissingSC: boolean) => {
```
jshaughn: Not a blocker but not sure why 'hasMissingSC' didn't follow the same pattern as the other flag fields.

jmazzitelli: that is confusing.

The "string" is there (as opposed to "boolean") becaise IIRC we support "unknown" as a value (to let someone reading the metadata know that it could not be determined if the value is true or false).

I would recommend just keeping things consistent, otherwise, it will be confusing in the future to see different types.

israel-hdez: It comes as a boolean from the API, so no need to check for strings.
@jmazzitelli Ok, changed it to receive a string. However, I had to do `.toString()` in the caller.

serenamarie125:
Lots of questions here in my mind.  We chatted about this before - how many badges are possible, and what are the possible states.  IMO they should have fixed positions, consistency is important.  As for color, i don't know why we would introduce separate colors for different badges, unless they are indicated some level of importance.  Can we get a document going around possible badges and there meanings?

PR #332, opened 2018-05-14
KIALI-721 fix the graph highlighter when switching layouts.

jmazzitelli:
When switching layouts and something is selected, the graph highlighter erronenously "unfreezes" - meaning the dimming and highlighting gets re-enabled when the graph should still be "frozen" (dimming and highlighting should not be affected on hover). This fixes that problem.

This also fixes multiple callbacks being triggered - turns out for each time the layout was switched, we were calling the callbacks (if we switched layouts 3 times, the callbacks were getting called 3 times). This fixes that, too. Now the callbacks are only ever called at most one time per event.

### DIFF
--- a/src/components/CytoscapeGraph/CytoscapeGraph.tsx
+++ b/src/components/CytoscapeGraph/CytoscapeGraph.tsx
@@ -134,7 +134,7 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
       return;
     }
 
-    // Convienently, the graph highlighter caches the cy instance that is currently in use.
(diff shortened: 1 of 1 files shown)


src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -134,6 +134,15 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
       return;
     }
 
+    // Convienently, the graph highlighter caches the cy instance that is currently in use.
```
jshaughn: Fix this typo! Rejected!

PR #333, opened 2018-05-14
KIALI-447 Ad latency to edge summary

jshaughn:
This PR adds a latency sparkline chart to the edge summary.  It presents our four standard histogram metric values: average, median, 95th and 99th percentiles.  By default the median shows but the legend is clickable to allow selection of the other values.  Hover shows the actual values.

screenshot

### DIFF
--- a/src/pages/ServiceGraph/SummaryPanelEdge.tsx
+++ b/src/pages/ServiceGraph/SummaryPanelEdge.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import Badge from '../../components/Badge/Badge';
 import RateTable from '../../components/SummaryPanel/RateTable';
 import RpsChart from '../../components/SummaryPanel/RpsChart';
+import LatencyChart from '../../components/SummaryPanel/LatencyChart';
 import { SummaryPanelPropType } from '../../types/Graph';
 import * as API from '../../services/Api';
 import * as M from '../../types/Metrics';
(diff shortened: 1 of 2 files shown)


src/pages/ServiceGraph/SummaryPanelEdge.tsx
```diff
@@ -132,13 +141,46 @@ export default class SummaryPanelEdge extends React.Component<SummaryPanelPropTy
...
+          sourceVersion
+        );
+        const lat99 = this.getDatapoints(
+          histograms['request_duration_in']['percentile99'],
+          '99th',
+          sourceService,
+          sourceVersion
+        );
+
+        // REMOVE dump latency
```
jmazzitelli: did you forget to do this removal?

src/pages/ServiceGraph/SummaryPanelEdge.tsx
```diff
@@ -132,13 +141,46 @@ export default class SummaryPanelEdge extends React.Component<SummaryPanelPropTy
...
+        console.log('METRICS: ' + JSON.stringify(metrics));
+        console.log('HISTOS: ' + JSON.stringify(histograms));
```
jmazzitelli: should we remove these debug messages? Seems like it might produce noisy logs, json-ifying metric and histrogram data.

jshaughn: yes, that is why I left that "REMOVE" in there, because I'd likely forget...  They are gone now...

PR #334, opened 2018-05-14
KIALI-679 Allows graph to show latency in the edges

josejulio:
Got a first pass on this, I had to update some redux actions to better match with this, but not yet using them I would need to connect the GraphFilter to redux and decided that other PR would better fit it.

So i setup the latency scenario like this, using this yml for openshift:
```yaml
apiVersion: config.istio.io/v1alpha2
kind: RouteRule
metadata:
  name: ratings-delay-abort
  namespace: bookinfo
spec:
  destination:
    name: ratings
  match:
    source:
      name: reviews
  route:
  - labels:
      version: v1
  httpFault:
    delay:
      fixedDelay: 1.000s
      percent: 100
  precedence: 2
```

Delay of 1 second from reviews to rating, I suspect that [this](https://github.com/istio/istio/issues/4323) issue prevents us from seeing the delay from reviews to rating (see gif), but is definetly happening and we can see it from productpage to review.

peek 2018-05-14 18-20

### DIFF
--- a/src/components/GraphFilter/GraphFilter.tsx
+++ b/src/components/GraphFilter/GraphFilter.tsx
@@ -115,7 +115,7 @@ export default class GraphFilter extends React.Component<GraphFilterProps, Graph
             id={'graph_filter_edges'}
             disabled={this.props.disabled}
             handleSelect={this.updateEdges}
-            nameDropdown={'Edge Mode'}
+            nameDropdown={'Edge Labels'}
             initialValue={this.props.edgeLabelMode}
             initialLabel={GraphFilter.EDGE_LABEL_MODES[this.props.edgeLabelMode]}
             options={GraphFilter.EDGE_LABEL_MODES}
--- a/src/components/GraphFilter/GraphFilter.tsx
+++ b/src/components/GraphFilter/GraphFilter.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import { Toolbar, Button, Icon, FormGroup } from 'patternfly-react';
 
-import { Duration, Layout } from '../../types/GraphFilter';
+import { Duration, Layout, EdgeLabelMode } from '../../types/GraphFilter';
 import { ToolbarDropdown } from '../ToolbarDropdown/ToolbarDropdown';
 import NamespaceDropdownContainer from '../../containers/NamespaceDropdownContainer';
 import { config } from '../../config';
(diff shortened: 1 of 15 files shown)


josejulio:
> @josejulio My personal preference would be to add this to the existing filters checkboxes 

I don't think it can fit well inside the checkboxes, unless I add the dropdown there. 

> but maybe @serenamarie125 can give some guidance. 

Yes please!

> The additional dropdown may be OK for now. Also, going forward we may want to expand on this to let the user set the latency setting (the percentile shown) because right now it is fixed at 95th percentile even thought we do support a query param setting.

I can update that in other PR if nothing is resolved about the dropdown (passing the query param)

> If we keep the dropdown a couple suggestions:
> 
> Edge mode -> Edge Labels
> ** None -> Hide
> ** Request rate -> Requests per second
> ** Latency -> Latency 95th percentile

Will update the code.

> note, we could later add other latency settings to the drop down. Perhaps if we kept the "Edge Labels" toggle in the current filter, it could pop up a dialog where the user selected the desired labels. I don't know, just an idea.

I'm not sure that a popover over a popover would look good, but @serenamarie125 should know better, in any case, this is an initial proposal to have something to work on top of.

josejulio:
Ah, welcome to rebase hell

josejulio:
> Screenshot for build 1007

@vnugent I rebased and the screenshot is "broken". I'm not sure why is broken, running it locally yields a screenshot

```
npm run build
npm run kiali &
npm run itest
```

kiali-browser-test


Is there anything that i'm missing?

josejulio:
@jshaughn updated with your suggestions and rebased, please re-review.

src/components/GraphFilter/GraphFilter.tsx
```diff
@@ -97,6 +111,15 @@ export default class GraphFilter extends React.Component<GraphFilterProps, Graph
             initialLabel={String(GraphFilter.GRAPH_LAYOUTS[this.props.graphLayout.name])}
             options={GraphFilter.GRAPH_LAYOUTS}
           />
+          <ToolbarDropdown
+            id={'graph_filter_edges'}
+            disabled={this.props.disabled}
+            handleSelect={this.updateEdges}
+            nameDropdown={'Edge Labes'}
```
jshaughn: Typo.

PR #344, opened 2018-05-16
kiali-733:  Improve handling of larger screenshot file

vnugent:
- [x] Pipe image file to curl instead of passing base64 data as an argument
- [x] parameterize upload and comment function to make it easier to post multiple images in the future


vnugent:
@josejulio @israel-hdez can you please review?

PR #346, opened 2018-05-17
KIALI-360 request percentage as new edge label option

jshaughn:
other:
- switch to camel case for edge label options for 'prettier' urls
- fix bug in EdgeLabelMode.fromString
- show latency < 1s as 'ms'
- change edge summary latency chart to default to 95th percentile to
  match current graph setting

### DIFF
--- a/src/types/GraphFilter.ts
+++ b/src/types/GraphFilter.ts
@@ -7,9 +7,10 @@ export interface Duration {
 }
 
 export enum EdgeLabelMode {
-  HIDE = 'HIDE',
-  REQUESTS_PER_SECOND = 'REQUESTS_PER_SECOND',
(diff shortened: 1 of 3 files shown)


src/types/GraphFilter.ts
```diff
@@ -18,13 +19,14 @@ export namespace EdgeLabelMode {
...
-      return EdgeLabelMode[value] as EdgeLabelMode;
+    const key = Object.keys(EdgeLabelMode).find(k => EdgeLabelMode[k] === value);
+    if (key) {
+      return EdgeLabelMode[key];
     }
     if (!defaultValue) {
       throw TypeError(`${value} is not a EdgeLabelMode`);
     }
-
+    console.log('REMOVE" returning default for [' + value + ']');
```
jmazzitelli: please remove this ... and what's up with the double-quote in there?

PR #350, opened 2018-05-18
KIALI-739: filter duration erroneously set to 0

vnugent:
Only write to and read from sessionStorage in top most component to avoid inconsistent props.  To verify the fix please remove appDuration and service-graph-params data in browser storage (Chrome -> developer tools -> Application)

@jshaughn  @mtho11 please review when you get a chance

---
session-storage


aljesusg:
This break KIALI-610 about shared the duration option between graphfilter and metrics toolbar of service. Why you remove this @vnugent ?

vnugent:
@aljesusg that's an oversight on my part.  I'll submit another PR.

PR #352, opened 2018-05-19
Reduxify the PfSpinner

mtho11:
The goal of this is to make the Spinner turn on/off automatically without explicitly calling client code. By hooking into the axios interceptor we can intercept the calls to request/response and automatically turn the spinner on/off based off of Redux state `globalState.isLoading`

- [x] Make PfSpinner a Redux connected component that shows/hides the spinner based on Redux state `globalState.isLoading`
- [x] Remove existing client code logic and replace with Smart Spinner component that knows how to turn on/off
- [x] Replace client use of PfSpinner to not have logic in client but just call to ActionCreator
- [x] Add Tests
- [x] Hook up to Axios to automatically control this behavior (no need to set any state) by dispatching events via request/response interceptors.

Of course, this spinner component can be manually dispatched as well.

https://issues.jboss.org/browse/KIALI-714


src/containers/PfSpinnerContainer.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { connect } from 'react-redux';
+
+type PfSpinnerProps = {
+  isLoading: boolean;
+};
+
+const mapStateToProps = (state: KialiAppState) => ({
+  isLoading: state.globalState.isLoading
+});
+export const PfSpinner: React.SFC<PfSpinnerProps> = props => {
```
jmazzitelli: if this really isn't a PatternFly spinner, and if the implementation is independent of Patternfly, should we rename this container and the exported variables so they don't have "Pf" in the names?

src/reducers/GlobalState.ts
```diff
@@ -0,0 +1,25 @@
+import { GlobalState } from '../store/Store';
+import { updateState } from '../utils/Reducer';
+import { GlobalActionKeys } from '../actions/GlobalActions';
+
+const INITIAL_GLOBAL_STATE: GlobalState = {
+  isLoading: false
+};
+
+// This Reducer allows changes to the 'serviceGraphFilterState' portion of Redux Store
```
jmazzitelli: This comment looks like a copy-n-paste from another file and should be removed or re-worded, I think. This reducer doesn't change serviceGraphFilterState - it changes the global state. no?

src/pages/IstioConfigList/IstioConfigListComponent.tsx
```diff
@@ -287,40 +287,37 @@ class IstioConfigListComponent extends React.Component<IstioConfigListComponentP
...
-            itemCount={this.state.istioItems.length}
-            onPageSet={this.pageSet}
-            onPerPageSelect={this.pageSelect}
-          />
-        </div>
-      );
-    }
+    ruleListComponent = (
+      <div>
+        <PfSpinnerContainer />
```
jmazzitelli: looking at the code this new code replaces, the old code only shows the NamespaceFilter <div> when not loading. This new code seems to always have the NamespaceFilter that follows the spinner. Is this what we want, or do we want to hide the rest of the content if isLoading is true? Does this namespace filter code break somehow or do we not want it to show if we are currently loading?

src/pages/ServiceList/ServiceListComponent.tsx
```diff
@@ -343,44 +343,41 @@ class ServiceListComponent extends React.Component<ServiceListComponentProps, Se
...
-                isAscending={this.state.isSortAscending}
-                onClick={this.updateSortDirection}
-              />
-            </Sort>
-            <RateIntervalToolbarItem
-              rateIntervalSelected={this.state.rateInterval}
-              onRateIntervalChanged={this.rateIntervalChangedHandler}
+    serviceListComponent = (
+      <div>
+        <PfSpinnerContainer />
```
jmazzitelli: same comment as my other one - is it OK that we now always return the namespace filter here even if we are loading?  The old code did not return the namespace filter and only showed the spinner when currently loading.

mtho11: Yes, it's a smart component. Just place it wherever you want it and it looks at the Redux state for `isLoading` and determines if it needs to be rendered or not. And with the Axios interceptors hooked in, it knows whenever an HTTP request/response is made and updates the redux state automatically.
 This eliminates all of the client logic in multiple screens and consolidates it in the component itself.  For instance, all of this:
```
if (this.state.loading) {
  serviceListComponent = <PfSpinner />;
} else {
  serviceListComponent = (
    <div>
...
```
Is eliminated, and just becomes:
```
<PfSpinner />
```
It moves us from procedural logic in every `*Page` to a declarative smart component that you can place and it will work properly.

src/containers/PfSpinnerContainer.tsx
mtho11: Good point. I need to think on this one.

src/reducers/GlobalState.ts
mtho11: Yes, cut-n-paste error.

src/pages/IstioConfigList/IstioConfigListComponent.tsx
mtho11: Ok, good point. Changing.

mtho11:
@jmazzitelli ok I think I have addressed all of the changes

PR #354, opened 2018-05-21
KIALI-669: Dispatch Redux action to load side panel detailed info

israel-hdez:


src/containers/ServiceGraphPageContainer.ts
```diff
@@ -7,9 +7,15 @@ import ServiceGraphPage from '../pages/ServiceGraph/ServiceGraphPage';
...
-  graphTimestamp: state.serviceGraphDataState.timestamp,
-  graphData: state.serviceGraphDataState.graphData,
-  isLoading: state.serviceGraphDataState.isLoading
+  graphTimestamp: state.serviceGraph.dataState.timestamp,
+  graphData: state.serviceGraph.dataState.graphData,
+  isLoading: state.serviceGraph.dataState.isLoading,
+  summaryData: state.serviceGraph.selectedObject
+    ? {
+        summaryTarget: state.serviceGraph.selectedObject.obj,
+        summaryType: state.serviceGraph.selectedObject.type
```
jmazzitelli: Is "type" a language keyword? I think we should consider renaming this - "type" just seems too "keyword-y". I dunno if this is a problem or not.

src/reducers/index.ts
```diff
@@ -1,16 +1,13 @@
 import { combineReducers } from 'redux';
-
 import namespaces from './Namespaces';
-import serviceGraphFilterState from './ServiceGraphFilterState';
 import { KialiAppState } from '../store/Store';
 import messageCenter from './MessageCenter';
-import serviceGraphDataState from './ServiceGraphDataState';
+import ServiceGraphReducer from './ServiceGraph';
```
jmazzitelli: should these names start with lowercase now? @mtho11  I seem to remember you changing alot of these names to be all lowercase as their first character (??)

src/actions/ServiceGraphActions.ts
```diff
@@ -0,0 +1,19 @@
...
+import { CytoscapeClickEvent } from '../components/CytoscapeGraph/CytoscapeGraph';
+
+export enum ServiceGraphActionKeys {
+  GRAPH_OBJECT_CLICKED = 'GRAPH_OBJECT_CLICKED',
+  GRAPH_READY = 'GRAPH_READY'
+}
+
+// synchronous action creators
+export const ServiceGraphActions = {
+  graphObjectClicked: createAction(ServiceGraphActionKeys.GRAPH_OBJECT_CLICKED, (event: CytoscapeClickEvent) => ({
```
mtho11: Is `graphObjectClicked` the right name for this? That is the thing that dispatches this Redux action not what the actual action does. The name should be what the redux action does.

src/actions/ServiceGraphActions.ts
```diff
@@ -0,0 +1,19 @@
...
+    type: ServiceGraphActionKeys.GRAPH_OBJECT_CLICKED,
+    ...event
+  })),
+  graphReady: createAction(ServiceGraphActionKeys.GRAPH_READY, (cy: any) => ({
+    type: ServiceGraphActionKeys.GRAPH_READY,
```
mtho11: Let's not use `type` it's a very overloaded term and it's a typescript keyword for `type` alias

src/containers/ServiceGraphPageContainer.ts
mtho11: Yes, `type` is a typescript keyword.

src/pages/ServiceGraph/ServiceGraphPage.tsx
```diff
@@ -12,7 +12,7 @@ import { computePrometheusQueryInterval } from '../../services/Prometheus';
 import { style } from 'typestyle';
 
 type ServiceGraphPageState = {
-  summaryData?: SummaryData | null;
```
mtho11: Awesome, this change makes ServiceGraphPage *stateless*

src/store/Store.ts
```diff
@@ -30,13 +30,21 @@ export interface ServiceGraphDataState {
   graphData: any;
 }
 
+export interface ServiceGraphState {
+  dataState: ServiceGraphDataState;
```
mtho11: `dataState` is redundant in a Redux store everything is dataState. Let's be more precise here. The protocol we have been following here is just to name it the same as the type, so `serviceGraphDataState`

src/store/Store.ts
```diff
@@ -30,13 +30,21 @@ export interface ServiceGraphDataState {
...
+  filterState: ServiceGraphFilterState;
+  selectedObject: {
```
mtho11: `Object` is another one of those general/overloaded terms. Almost everything is an object. The name should convey what kind of _selectedObject_ it is.

src/actions/ServiceGraphActions.ts
```diff
@@ -0,0 +1,19 @@
+import { createAction } from 'typesafe-actions';
+import { CytoscapeClickEvent } from '../components/CytoscapeGraph/CytoscapeGraph';
+
+export enum ServiceGraphActionKeys {
+  GRAPH_OBJECT_CLICKED = 'GRAPH_OBJECT_CLICKED',
+  GRAPH_READY = 'GRAPH_READY'
```
mtho11: Remember that these redux events should self-describe themselves if one is reading the Redux action log. If one comes across a global event called 'GRAPH_READY' it doesn't tell me *how* this event is going to change the state and to which graph.

mtho11: The Redux events aren't namespaced and since they are global they need to be understandable on their own.

israel-hdez:
@mtho11 I hope to have fixed the comments. Added tests for the actions and the reducers.

israel-hdez:
@mtho11 @jmazzitelli Can you re-review?

PR #360, opened 2018-05-23
KIALI-767 Show pods information + some typesafety refactoring

jotak:
Also now hasIstioSidecar function runs from pods rather than deployments, since not all pods may come from deployments.

Refactoring:
- Create an interface for service details model coming from API. It should be temporary (I'll create a JIRA to merge it with the model class used in ts _[edit: [KIALI-768](https://issues.jboss.org/browse/KIALI-768)]_). It's "ServiceDetailsInfoFromAPI"
- In model coming from the API, replace some "Map" fields with "raw" key-value maps ({ [key: string]: string }). Using Map is incorrect and leads to potential bugs as functions defined in the Map class are not present in runtime, leading to compile-time/runtime mismatch. There was actually a bug where "Map.size" was invoked and never returned the intended result so service labels were never displayed in service details.
- Remove 90% of the state in ServiceInfo component since it was duplicated from props and never modified, so useless as state.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 
 # testing
 /coverage
+kiali-browser-test.png
 
 # production
 /build
--- a/src/pages/ServiceDetails/ServiceDetailsPage.tsx
+++ b/src/pages/ServiceDetails/ServiceDetailsPage.tsx
@@ -9,7 +9,12 @@ import { NamespaceFilterSelected } from '../../components/NamespaceFilter/Namesp
 import { ActiveFilter } from '../../types/NamespaceFilter';
 import * as API from '../../services/Api';
 import * as MessageCenter from '../../utils/MessageCenter';
-import { hasIstioSidecar, ServiceDetailsInfo, ObjectValidation } from '../../types/ServiceInfo';
+import {
+  hasIstioSidecar,
+  ServiceDetailsInfo,
+  ServiceDetailsInfoFromAPI,
+  ObjectValidation
+} from '../../types/ServiceInfo';
 import AceEditor, { AceOptions } from 'react-ace';
 import 'brace/mode/yaml';
 import 'brace/theme/eclipse';
@@ -66,8 +71,8 @@ class ServiceDetails extends React.Component<RouteComponentProps<ServiceId>, Ser
     NamespaceFilterSelected.setSelected([activeFilter]);
   };
 
-  parseServiceDetailsInfo = data => {
-    let parsed: ServiceDetailsInfo = {
+  parseServiceDetailsInfo = (data: ServiceDetailsInfoFromAPI): ServiceDetailsInfo => {
+    return {
       labels: data.labels,
       name: data.name,
       created_at: data.created_at,
@@ -75,7 +80,8 @@ class ServiceDetails extends React.Component<RouteComponentProps<ServiceId>, Ser
       type: data.type,
       ports: data.ports,
       endpoints: data.endpoints,
-      istio_sidecar: hasIstioSidecar(data.deployments),
+      istio_sidecar: hasIstioSidecar(data.pods),
+      pods: data.pods,
       deployments: data.deployments,
       dependencies: data.dependencies,
       routeRules: data.route_rules,
@@ -85,7 +91,6 @@ class ServiceDetails extends React.Component<RouteComponentProps<ServiceId>, Ser
       ip: data.ip,
       health: data.health
     };
-    return parsed;
   };
 
   validateParams(parsed: ParsedSearch): boolean {
@@ -184,7 +189,7 @@ class ServiceDetails extends React.Component<RouteComponentProps<ServiceId>, Ser
       });
     API.getServiceDetail(this.props.match.params.namespace, this.props.match.params.service)
       .then(response => {
-        let data = response['data'];
+        const data: ServiceDetailsInfoFromAPI = response['data'];
         this.setState({
           serviceDetailsInfo: this.parseServiceDetailsInfo(data)
         });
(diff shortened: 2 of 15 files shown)


jotak:
Please read the refactoring explanation above :-)
I've been "incited" to do these changes as I was adding pod information, but if you prefer that I split the work and open a different PR for the refactoring I can do so.

jotak:
[do not merge yet => just noticed I haven't displayed the annotations yet, in particular "created-by" and istio containers]

jotak:
Screenshot:
pods details

jotak:
@lucasponce I think it can be a first step for this task but then I'll enter a new JIRA to go further on that:

The idea would be to remove the "deployment" tab and update the "created by" part in pods to be able to show related deployment information. It may require some update on the backend too.
Maybe if we find some deployments that are not associated with any pod we could add a section like "Unused Deployments", or such, below the list of pods.

Also, we are showing labels, maybe we should show the Deployment selectors too

lucasponce:
> Maybe if we find some deployments that are not associated with any pod we could add a section like "Unused Deployments", or such, below the list of pods.

Yes, I think that also having pods grouped per version would help from Istio perspective, as version -> deployment should be a 1 to 1 relationship.

But that can be iterated in following future work, and we can decide if the Pods/Deployments are just a single tab.

A minor but very useful comment, it would be possilbe to add the images ? I mean, the istio-proxy container will exist on productized vs upstream distribution, but the image would flag if we are under one or another.

If it is something simple, it could be a nice to have for this one.

jotak:
(Last commit is related to backend PR https://github.com/kiali/kiali/pull/209 which should be merged at the same time)

lucasponce:
> Using Map is incorrect and leads to potential bugs as functions defined in the Map class are not present in runtime, leading to compile-time/runtime mismatch. There was actually a bug where "Map.size" was invoked and never returned the intended result so service labels were never displayed in service details.

Can you add more details here ?

lucasponce:
> Create an interface for service details model coming from API. It should be temporary (I'll create a JIRA to merge it with the model class used in ts [edit: KIALI-768]). It's "ServiceDetailsInfoFromAPI"

Not sure if I understand this. I guess this is because there is another WIP to touch same file, and this is a temporal workaround, but not sure if I like. I tend to think that modifications should be done in the files without using these ones, I find it more hard to understand.

jotak:
>> Using Map is incorrect and leads to potential bugs as functions defined in the Map class are not present in runtime, leading to compile-time/runtime mismatch. There was actually a bug where "Map.size" was invoked and never returned the intended result so service labels were never displayed in service details.
>
> Can you add more details here ?

Sure. When the data is fetched from the backend with axios, the typescript compiler has no way to validate the type of the data, it is necessarily something that should in theory be performed at runtime, ie. not by typescript but by javascript, but JS being typeless and not performing this kind of check, the data can be propagated everywhere without any no guarantee that it's of the type it's supposed to be. Typescript has no choice but to believe us, so better not give it a false indication :-)

Here we say the incoming data is of type Map. It's not accurate. There's a difference between the Map class, which has some functions like "size()", "has(...)", "get(...)" etc. and a raw json object where these functions are not present. That's why I was saying there's a runtime vs compile-time mismatch. We get then a false sense of security by the compiler who says "hey you can call `size`, `has` or `get` on your map object", whereas it's wrong and it would fail at runtime.

I see two ways to fix that: first would be to convert json objects to Map explicitly as soon as the data is returned from the rest call, using Map constructor which, [as you can see](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) doesn't take a simple object as parameter but a array of tuples. The second option which seems simpler to me is what I did, not using Map but the actual type as it is recognized by typescript in such situation, which is `{[key: string]: string}`.

Now I noticed that what I said in my previous comment was not totally accurate, the bug I mentioned is something I introduced myself while coding this PR, it wasn't in the existing code.

Although I've found another one: [here](https://github.com/kiali/kiali-ui/pull/360/files#diff-7cdcbd07abd513dbb734e4cfdb5aa81aL69) labels was set as an empty Map and not as a copy/reference of the labels coming from props. And if you start your kiali without this PR you will see: the Label field at the top of the service details page never contains any label...

lucasponce:
Ok, so, basically, using Map() object can introduce runtime errors when fetching from the backend.
I was curious as I think until today we have been using the Map object and perhaps we didn't see/notice the error.
Perhaps the rebase conflicts are due this change and Map() should be updated from other places, too.

jotak:
>> Create an interface for service details model coming from API. It should be temporary (I'll create a JIRA to merge it with the model class used in ts [edit: KIALI-768]). It's "ServiceDetailsInfoFromAPI"
>
> Not sure if I understand this. I guess this is because there is another WIP to touch same file, and this is a temporal workaround, but not sure if I like. I tend to think that modifications should be done in the files without using these ones, I find it more hard to understand.

So I created this class in the same effort to enforce better typing, for similar reasons than explained above. Without this class, the data coming from the REST API isn't typed until it's transformed into `ServiceDetailsInfo`. Which is sad because until then we don't get the benefit of the compiler checks. One example I experienced: when I changed the signature of [hasIstioSidecar](https://github.com/kiali/kiali-ui/pull/360/files#diff-9cbf89d446475c4f25d14193db594233R354) to work with Pods instead of Deployments, I had no compiler error [where the call is performed](https://github.com/kiali/kiali-ui/pull/360/files#diff-a5b0179a9adf158be7103b26db8c9b84R82) because the `data` variable was untyped. That's sad. Types are useful for that, I think we should use types wherever we can.

Note that I wish I could use the class `ServiceDetailsInfo` instead of `ServiceDetailsInfoFromAPI`, but there are some differences between the two ; that's why I opened the ticket KIALI-768. As soon as this PR is merged I'll work on KIALI-768 and we will get rid of `ServiceDetailsInfoFromAPI`. But til then, to me it's better to have a bad-looking typed data than a bad-looking untyped data :-)

jotak:
(I'll look at the conflicts tomorrow)

jotak:
I rebased and fixed conflicts.
Note I'm not fixing all the inaccurate types here. I've just created a JIRA to continue the task: https://issues.jboss.org/browse/KIALI-793

PR #362, opened 2018-05-24
KIALI-769 Group IstioValidations per objectType and name in the UI

lucasponce:
This is the counterpart of https://github.com/kiali/kiali/pull/208 to adapt the new grouped validations into the UI.

Please, don't merge until backend is ready, but it can be reviewed and tested with the backend PR.


lucasponce:
Thx @xeviknal going to check why travis failed, perhaps I need to format some code.

jmazzitelli:
This PR causes a regression - the service details page crashes. PR #364 works around it, but I don't think it works around it the correct way. Someone needs to re-visit this PR #362 and make sure it all works.

PR #364, opened 2018-05-24
fix npe

jmazzitelli:
This works around the bug introduced in this PR: https://github.com/kiali/kiali-ui/pull/362

We probably want to fix this properly - it doesn't seem right that .get() is being reported as "not a function" when this should be a Map object. But for now, this is working.

### DIFF
--- a/src/pages/ServiceDetails/ServiceInfo.tsx
+++ b/src/pages/ServiceDetails/ServiceInfo.tsx
@@ -107,7 +107,9 @@ class ServiceInfo extends React.Component<ServiceDetailsId, ServiceInfoState> {
   }
 
   validationsFor(objectType: string) {
-    return this.state.validations.get(objectType);
+    return this.state.validations && typeof this.state.validations.get === 'function'
+      ? this.state.validations.get(objectType)
+      : null;
   }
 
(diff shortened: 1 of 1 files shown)


src/pages/ServiceDetails/ServiceInfo.tsx
```diff
@@ -107,7 +107,9 @@ class ServiceInfo extends React.Component<ServiceDetailsId, ServiceInfoState> {
   }
 
   validationsFor(objectType: string) {
-    return this.state.validations.get(objectType);
+    return this.state.validations && typeof this.state.validations.get === 'function'
```
mtho11: This has a bit of a smell to it. But let's fix this now and cleanup later as tomorrow is the sprint demo.

josejulio: probably related to this: https://github.com/kiali/kiali-ui/pull/360#issuecomment-391791684

lucasponce:
Yes, this bug was introduced due bad use of Map() object, as discussed in #360, it is refactored in that PR.
Is it master stable with this PR, can it wait until #360 is addressed ?

PR #370, opened 2018-05-28
Remove console.logs

jotak:
Note, there's still some debug logs caused by the use of redux-logger.

### DIFF
--- a/src/components/CytoscapeGraph/CytoscapeGraph.tsx
+++ b/src/components/CytoscapeGraph/CytoscapeGraph.tsx
@@ -77,18 +77,14 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
   }
 
   componentDidMount() {
-    console.log('CY: mounting graph component');
     this.cyInitialization(this.getCy());
   }
 
   componentDidUpdate() {
-    console.log('CY: updating graph component');
     this.processGraphUpdate(this.getCy());
   }
 
   render() {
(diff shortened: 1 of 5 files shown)


src/containers/MessageCenterContainer.tsx
```diff
@@ -37,7 +37,7 @@ const mapStateToPropsMessageCenterTrigger = state => {
     badgeDanger: boolean;
   };
   const dangerousMessageTypes = [MessageType.ERROR, MessageType.WARNING];
-  const messageCenterTriggerPropsToMap = state.messageCenter.groups
+  return state.messageCenter.groups
```
josejulio: thanks!

src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -77,18 +77,14 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
   }
 
   componentDidMount() {
-    console.log('CY: mounting graph component');
```
josejulio: I find them useful when working on the graph. I suppose we could add in the future something like this:
https://github.com/mreuvers/typescript-logging

jotak: Yes, I agree. Some logs can be useful in certain situation but annoying the rest of the time. Categorising logs + changing level locally would help to satisfy both situations.

PR #373, opened 2018-05-29
KIALI-768, KIALI-767 sanitize model

jotak:
- Unified json names (camel case)
- Remove Map from model
- Better typing; start typing from API.ts


src/actions/ServiceGraphDataActions.ts
```diff
@@ -38,7 +38,7 @@ export const ServiceGraphDataActions = {
       const restParams = { duration: duration + 's' };
       API.getGraphElements(namespace, restParams).then(
         response => {
-          const responseData = response['data'];
+          const responseData: any = response['data'];
```
lucasponce: Just for me, as I see in the code that the use of any is optional. What are the benefits to add the any here ?

jotak: It isn't optional in this case because I've changed API functions to return Promise<Response<T>> ; when T is unknown (as here) typescript decides it's an empty object `{}`, and here I got compile errors so I had either to declare a proper type (but I'm lacking that knowledge) or cast to `any`

src/pages/ServiceDetails/ServiceInfo/ServiceInfoRouteRules/RouteRuleRoute.tsx
```diff
@@ -10,7 +10,7 @@ import './RouteRuleRoute.css';
 interface RouteRuleRouteProps {
   name: string;
   route: DestinationWeight[];
-  validations: Map<string, ObjectValidation>;
+  validations: { [k: string]: ObjectValidation };
```
lucasponce: Same here, a "key" I think it helps to not miss we are coding a map. I see "key" is used in other files, too.

src/pages/ServiceDetails/ServiceInfo/ServiceInfoRoutes.tsx
```diff
@@ -3,7 +3,7 @@ import { Col, Row } from 'patternfly-react';
 import { Link } from 'react-router-dom';
 
 interface ServiceInfoRoutesProps {
-  dependencies?: Map<string, string[]>;
+  dependencies?: { [key: string]: string[] };
```
lucasponce: +1

src/services/Api.ts
```diff
@@ -11,8 +23,8 @@ const auth = (user: string, pass: string) => {
   };
 };
 
-let newRequest = (method: string, url: string, queryParams: any, data: any) => {
-  return new Promise((resolve, reject) => {
+let newRequest = <T>(method: string, url: string, queryParams: any, data: any) => {
```
lucasponce: Good change to introduce the type from here.

src/types/Metrics.ts
```diff
@@ -1,6 +1,6 @@
 export interface Metrics {
-  metrics: Map<String, MetricGroup>;
-  histograms: Map<String, Histogram>;
+  metrics: { [k: string]: MetricGroup };
```
lucasponce: I would like to see a "key" instead of "k" but I know this is trivial.
For me "key" heps.

src/types/ServiceInfo.ts
```diff
@@ -362,42 +362,24 @@ export interface ServiceDetailsInfo {
...
-  deployments?: Deployment[];
-  route_rules?: RouteRule[];
-  destination_policies?: DestinationPolicy[];
-  virtual_services?: VirtualService[];
-  destination_rules?: DestinationRule[];
-  dependencies?: Map<string, string[]>;
-  health?: Health;
-}
+// validations are grouped per 'objectType' first in the first map and 'name' in the inner map
+export type Validations = { [k1: string]: { [k2: string]: ObjectValidation } };
```
lucasponce: k1 -> objectType
k2 -> name
I guess that in case of multiple maps, it helps to get the scenario.
But as it is commented out, at least I would use key1, key2.

PR #374, opened 2018-05-29
KIALI-822: Show error on service graph when encountered

mwringe:
Current we incorrectly display the empty service graph page even if the reason we don't have results is because we encountered an error.

This updates to the service graph page to display an error message instead, which should make things more clear to the user.

### DIFF
--- a/src/components/CytoscapeGraph/CytoscapeGraph.tsx
+++ b/src/components/CytoscapeGraph/CytoscapeGraph.tsx
@@ -18,6 +18,7 @@ import { ServiceGraphActions } from '../../actions/ServiceGraphActions';
 type CytoscapeGraphType = {
   elements?: any;
   isLoading?: boolean;
+  error?: any;
   edgeLabelMode: EdgeLabelMode;
   showNodeLabels: boolean;
   showCircuitBreakers: boolean;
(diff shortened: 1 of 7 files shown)


jshaughn:
@serenamarie125  It is possible to set the arrow head to the same color as the arrow line.  I'm not sure why we don't do that, I'll create a separate Jira to make that change.

I wanted to point out that the screen shots added by 'kiali-bot' are not usually relevant to the PR, it's actually just a testing mechanism to make sure the PR didn't break the page.  It's sort of verbose and confusing because it generates a lot of tangential screenshots.  We're actively discussing how to make this 'bot thing better.  I think it was throwing you off based on your comments, those shots above are not really representing anything in the PR.

israel-hdez:
@mwringe This is duplicating the error in the message center. Perhaps, we could not push the error to the message center to avoid the duplication of the error?

mwringe:
Ugh, I forgot about this PR :/ I will update it to bring in the latest from master.

@israel-hdez yes, it is duplicating the error message. We have the room and I think its more clear to duplicate the error message here than to only display the error message in the message centre (the message centre stores a lot of errors and its not always clear what caused which error, I think its more clear this way so that people know why the service graph didn't load). If people think this is a problem, we can always just have it display a generic error message and ask to look at the message centre.

israel-hdez:
@mwringe Perhaps #436 has solved this.

mwringe:
Awesome

PR #378, opened 2018-05-30
KIALI-837 Fix metrics page spinner

jotak:
- Use same spinner as in other places (pf spinner)
- Use global loading state
- <strike>Duplicate the PfSpinnerContainer component into a similar one able to show a "refresh" button when it isn't loading</strike>


jotak:
@mtho11 can you have a look please?
I'm basically duplicating the class you just created `PfSpinnerContainer`, with some specific behaviour as it embeds a refresh button.
I could have modified the existing one as well, like by adding a boolean props `showRefreshButton` but I preferred to keep it as simple as possible. It still can be changed.

jotak:
Actually I purely removed the spinner as it is now global

jotak:
Thanks @aljesusg

PR #383, opened 2018-05-30
Kiali-848 Remove Redux Logger (too verbose)

mtho11:
Remove the redux logger since it is too verbose.
We always have Redux Dev tools which does soo much more.

https://issues.jboss.org/browse/KIALI-848


mtho11:
@aljesusg ah yes good catch!

PR #385, opened 2018-05-30
[KIALI-534] Login Change Style, remove change in reducers

aljesusg:
Move the change of the style of login from reducer to components.
login-logout


mtho11:
@aljesusg just need to resolve the conflict

PR #386, opened 2018-05-30
 KIALI-745 Autoupdate the service graph

josejulio:
The main concern of this PR is to add the polling interval to the Service Graph, however other issues happened when removing the "loading" spinner that made the graph "jump" when loading the data.

#382 Also did the same (removed the loading spinner) so this fixes some issues after that.

- Edges not being up to date after refreshes
  - This was was caused because the `cy.json` only diffs the values received, it doesn't replace data and the backend doesn't return empty data (so the empty data is never replaced)
- Badges not being removed between namespace changes
  - Badges were removed AFTER updating the json, if we had any badge on an element that would be removed, it would stay there (because we remove the element first and the badge later)

Issues not fixed:
- If going from a one-node layout to a multiple-node namespace, the nodes look all stacked on top of each other. Manually triggering a new layout fixes this. Tried to force the relayout in code, but didn't fixed it (however, if forcing the relayout and going from `istio-system` to `bookinfo` (looks stacked) then to `all` it looks good, there must be some re-render issue (tried `cy.forceRender`) I assume i'm doing something wrong
- Need to reload the sidepanel after each refresh of the graph (Todo)

It might be an option to merge this PR and fix the issues on a separate PR (because current graph is sort of buggy)

### DIFF
--- a/src/components/Filters/PollIntervalFilter.tsx
+++ b/src/components/Filters/PollIntervalFilter.tsx
@@ -1,0 +1,38 @@
+import * as React from 'react';
+
+import { config } from '../../config';
+import { ToolbarDropdown } from '../ToolbarDropdown/ToolbarDropdown';
+import { PollInterval } from '../../types/GraphFilter';
+
+type PollIntervalFilterProps = {
+  id: string;
+  disabled: boolean;
+  onPollIntervalChanged: (interval: number) => void;
+  selected?: PollInterval;
+};
+
+const PollIntervals = config().toolbar.pollInterval;
+const DefaultPollInterval = config().toolbar.defaultPollInterval;
+
+const PollIntervalFilter: React.SFC<PollIntervalFilterProps> = props => {
+  const selectedLabel = PollIntervals[props.selected!.value]
+    ? PollIntervals[props.selected!.value]
+    : `${props.selected!.value} ms`;
+  return (
+    <ToolbarDropdown
+      id={props.id}
+      disabled={props.disabled}
+      handleSelect={props.onPollIntervalChanged}
+      nameDropdown="Poll Interval"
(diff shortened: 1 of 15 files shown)


src/components/Filters/PollIntervalFilter.tsx
```diff
@@ -0,0 +1,38 @@
...
+      nameDropdown="Poll Interval"
+      initialValue={props.selected!.value}
+      initialLabel={selectedLabel}
+      options={PollIntervals}
+    />
+  );
+};
+
+PollIntervalFilter.defaultProps = {
+  selected: { value: DefaultPollInterval }
```
mtho11: I don't know that the extra object inside `selected` property is necessary, referring to the `value`. Unless there are multiple properties (like: `unitOfMeasure` or others) it just seems like it's not adding much.  The `.value` portion seems redundant. `defaultProps.selected` is all we need (which would have the type `DefaultPollInterval`)
It's a minor nit, and maybe I'm just not thinking of something.
I think there are no changes needed here as this has been done in other work -- you were just following one precedent -- although there is a mixture.
We can address this separately in a Kiali-dev mailing.

mtho11:
I'm totally for putting these things in piecemeal for a more agile approach.

src/types/GraphFilter.ts
```diff
@@ -6,6 +6,10 @@ export interface Duration {
   value: number;
 }
 
+export interface PollInterval {
```
mtho11: I think of `PollInterval` as being a `type` and not an interface since it doesn't have multiple properties inside of it.

jmazzitelli:
I'm merging this now, we can iterate in another PR if additional changes are needed.

PR #388, opened 2018-05-31
emergency patch

jmazzitelli:


pilhuhn:
@jmazzitelli Can you change the title and/or add a description of what this does?
The title is not descriptive at all

jmazzitelli:
This was reverted yesterday anyway. The problem was when you logged in you immediately got an error in the startup graph page due to an undefined variable being accessed. This patch quickly performed a "null check" on that offending variable. @josejulio reverted this and a couple other commits to put us back into a good state in master. He's getting a new PR ready that will fix the original problem.

PR #391, opened 2018-05-31
Kiali-853 Bump package versions for sprint 6

mtho11:
I coudn't bump redux thunk because of some errors -- but no big deal.


aljesusg:
Should we add the redux-store here or after bump?

vnugent:
Do we remove node_modules dir when bumping packages? The screen shot indicates an actual problem with starting kiali.

PR #392, opened 2018-05-31
KIALI-662: Fix zoom adjustment with Cose layout.

israel-hdez:
fixlayoutzoom

### DIFF
--- a/src/components/CytoscapeGraph/CytoscapeGraph.tsx
+++ b/src/components/CytoscapeGraph/CytoscapeGraph.tsx
@@ -223,10 +223,13 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
     if (this.updateLayout) {
       cy.layout(LayoutDictionary.getLayout(this.props.graphLayout)).run();
       // Don't allow a large zoom if the graph has a few nodes (nodes would look too big).
-      if (cy.zoom() > 2.5) {
-        cy.zoom(2.5);
+      setTimeout(() => {
+        cy.fit();
+        if (cy.zoom() > 2.5) {
+          cy.zoom(2.5);
+        }
         cy.center();
-      }
+      });
       this.updateLayout = false;
     }
 


israel-hdez:
This will fix the issue with the cose layout, but is delaying the zoom adjustment.
So, the browser will show for a very, very, veeeery brief instant a not adjusted graph and then do the adjustment.

That very brief instant is comparable to the shutter of camera when taking a picture at 1/60. In the viewfinder you can see darkness for an instant.

So, not sure if we really want this PR. It fixes something with a drawback.

mtho11:
@israel-hdez can you add a little video as well

src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -250,10 +250,14 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
 
     // Don't allow a large zoom if the graph has a few nodes (nodes would look too big).
     // TODO: only do this if there is a small number of nodes - let the user zoom in large graphs
-    if (cy.zoom() > 2.5) {
-      cy.zoom(2.5);
+    // TODO: Remove setTimeout. It was added because zoom adjustment is not working fine with Cose layout.
+    setTimeout(() => {
```
josejulio: Cytoscape layout have events, one is triggered when the layout starts running, one when all the nodes are in position and one when the layout stops. I think we can hook on that. There is currently a PR open (#393) that does some changes on when this code is called.
Can you put a follow after that one is merged? or if you want, i can put it, i have a rough idea of what needs to be done, instead of the setTimeout i would suggest a [pon](http://js.cytoscape.org/#cy.promiseOn) or [one](http://js.cytoscape.org/#cy.one) when is really [needed](https://github.com/kiali/kiali-ui/pull/393/files#diff-04b15b4602d6b9ccf1eb82d15e07535fR231).

jmazzitelli:
I don't like the looks of that, FWIW.

PR #393, opened 2018-05-31
KIALI-861 Prepare service graph for refreshing

josejulio:
This should fix the following issues:

- When changing from namespace (all nodes on top of each other)
- When receiving rates/latencies, they are not updated accurately and sometimes it shows a different one nor they are removed when the rate/latency is empty.
- Edge labels are not loaded accurately, if selecting one when the metrics don't have rate or latency in place
- Refresh side panel info with the current graph
- Badges are not removed when changing from namespace

### DIFF
--- a/src/actions/ServiceGraphDataActions.ts
+++ b/src/actions/ServiceGraphDataActions.ts
@@ -45,12 +45,20 @@ const decorateGraphData = (graphData: any) => {
       hasMissingSidecars: undefined
     }
   };
-  if (graphData && graphData.elements) {
-    if (graphData.elements.nodes) {
-      graphData.elements.nodes = graphData.elements.nodes.map(node => ({ ...elementsDefaults.nodes, ...node }));
+  if (graphData) {
+    if (graphData.nodes) {
+      graphData.nodes = graphData.nodes.map(node => {
+        const decoratedNode = { ...node };
+        decoratedNode.data = { ...elementsDefaults.nodes, ...decoratedNode.data };
+        return decoratedNode;
+      });
     }
-    if (graphData.elements.edges) {
-      graphData.elements.edges = graphData.elements.edges.map(edge => ({ ...elementsDefaults.edges, ...edge }));
+    if (graphData.edges) {
+      graphData.edges = graphData.edges.map(edge => {
+        const decoratedEdge = { ...edge };
+        decoratedEdge.data = { ...elementsDefaults.edges, ...decoratedEdge.data };
+        return decoratedEdge;
+      });
     }
   }
   return graphData;
--- a/src/actions/ServiceGraphDataActions.ts
+++ b/src/actions/ServiceGraphDataActions.ts
@@ -14,6 +14,48 @@ export enum ServiceGraphDataActionKeys {
   HANDLE_LEGEND = 'HANDLE_LEGEND'
 }
 
+// When updating the cytoscape graph, the element data expects to have all the changes
+// non provided values are taken as "this didn't change", similar as setState does.
+// Put default values for all fields that are ommited.
+const decorateGraphData = (graphData: any) => {
+  const elementsDefaults = {
+    edges: {
+      rate: undefined,
+      rate3XX: undefined,
+      rate4XX: undefined,
+      rate5XX: undefined,
+      percentErr: undefined,
+      percentRate: undefined,
+      latency: undefined,
+      isUnused: undefined
+    },
+    nodes: {
+      version: undefined,
+      rate: undefined,
+      rate3XX: undefined,
+      rate4XX: undefined,
+      rate5XX: undefined,
+      rateSelfInvoke: undefined,
+      hasCB: undefined,
+      hasRR: undefined,
+      isDead: undefined,
+      isGroup: undefined,
+      isRoot: undefined,
+      isUnused: undefined,
+      hasMissingSidecars: undefined
+    }
+  };
+  if (graphData && graphData.elements) {
+    if (graphData.elements.nodes) {
+      graphData.elements.nodes = graphData.elements.nodes.map(node => ({ ...elementsDefaults.nodes, ...node }));
+    }
+    if (graphData.elements.edges) {
+      graphData.elements.edges = graphData.elements.edges.map(edge => ({ ...elementsDefaults.edges, ...edge }));
+    }
+  }
+  return graphData;
+};
+
 // synchronous action creators
 export const ServiceGraphDataActions = {
   getGraphDataStart: createAction(ServiceGraphDataActionKeys.GET_GRAPH_DATA_START),
@@ -22,7 +64,7 @@ export const ServiceGraphDataActions = {
     (timestamp: number, graphData: any) => ({
       type: ServiceGraphDataActionKeys.GET_GRAPH_DATA_SUCCESS,
       timestamp: timestamp,
-      graphData: graphData
+      graphData: decorateGraphData(graphData)
     })
   ),
   getGraphDataFailure: createAction(ServiceGraphDataActionKeys.GET_GRAPH_DATA_FAILURE, (error: any) => ({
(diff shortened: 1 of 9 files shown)


josejulio:
Still need to remove some changes that were not needed.

src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -206,57 +204,66 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
 
     this.trafficRenderer.stop();
 
+    const isTheGraphSelectd = this.cy.$(':selected').length === 0;
```
jmazzitelli: I hate doing this but... missing the last "e" in the variable name ("Selected" not "Selectd").

jmazzitelli:
Edge labels are out of sync with the drop-down menu:
edge-labels-bug ogv

mtho11:
Let's go slower this time.

jmazzitelli:
The filter selections are out of sync with the graph:

node-labels-bug ogv

node-labels-bug-2 ogv

src/components/CytoscapeGraph/CytoscapeGraph.tsx
josejulio: will fix, thanks

josejulio:
> This is really close. I'm going to ask for "Request Changes" if only to fix that tiny typo, but also to have more discussion on the edge labels and filter checkboxes not working (I'll attach some animated gifs to show the issues in additional comments). If you think we can fix it in this PR, let's do that. Or, we can just approve this now and fix those issues in another PR.

I think i can finish that (or at least most of that)

src/components/GraphFilter/GraphFilter.tsx
```diff
@@ -131,9 +131,11 @@ export default class GraphFilter extends React.Component<GraphFilterProps, Graph
               <Button onClick={this.props.onLegend}>{this.props.hideLegend ? 'Show' : 'Hide'} Legend</Button>
             </div>
             <div className="form-group">
-              <Button disabled={this.props.disabled} onClick={this.handleRefresh}>
-                <Icon name="refresh" />
-              </Button>
+              <Spinner loading={this.props.disabled}>
```
mtho11: Is the spinner over the refresh button really back?

mtho11: This is a toolbar not a status area.

mtho11: Having multiple areas 'spin' is not really a good user experience to me.

src/pages/ServiceGraph/SummaryPanelNode.tsx
```diff
@@ -52,9 +53,9 @@ export default class SummaryPanelNode extends React.Component<SummaryPanelPropTy
     this.fetchRequestCountMetrics(this.props);
   }
 
-  componentWillReceiveProps(nextProps: SummaryPanelPropType) {
-    if (nextProps.data.summaryTarget && nextProps.data.summaryTarget !== this.props.data.summaryTarget) {
-      this.fetchRequestCountMetrics(nextProps);
+  componentDidUpdate(prevProps: SummaryPanelPropType) {
```
mtho11: Seems like this would not be necessary with Redux. If an interested property were changed it would re-render.

src/components/GraphFilter/GraphFilter.tsx
josejulio: This is just for the transition, I need to have a spinner in the meantime. The PR for the loading in the header was reverted and will put back after this PR

src/pages/ServiceGraph/SummaryPanelNode.tsx
josejulio: I want to do a request to fetch new data to the server, but i don't want to do it always, only when is really needed.
If adding redux, this test would still apply i think, unless the redux part also takes care of the fetching the data (but this test would move to the redux container I think)

src/components/GraphFilter/GraphFilter.tsx
josejulio: There is only one Spinner this time, but will go away in the next PR.

josejulio:
peek 2018-05-31 22-49

Note: I added `Last 20 secs` in config.ts just to speed my tests
Note2: Sorry i forgot to "show" my cursor.

jmazzitelli:
this works very nicely

src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -206,57 +198,66 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
...
-    if (this.newLayout) {
-      cy.layout(LayoutDictionary.getLayout(this.newLayout)).run();
-      this.newLayout = '';
+    // update the layout if needed
+    if (this.updateLayout) {
+      cy.layout(LayoutDictionary.getLayout(this.props.graphLayout)).run();
+      // Don't allow a large zoom if the graph has a few nodes (nodes would look too big).
+      if (cy.zoom() > 2.5) {
+        cy.zoom(2.5);
+        cy.center();
```
jshaughn: Sometimes it doesn't seem like we center the graph outside of this use case.  Perhaps it is supposed to be centered by default, but it doesn't always seem that way to me when I change layouts.

src/actions/ServiceGraphDataActions.ts
```diff
@@ -14,6 +14,56 @@ export enum ServiceGraphDataActionKeys {
   HANDLE_LEGEND = 'HANDLE_LEGEND'
 }
 
+// When updating the cytoscape graph, the element data expects to have all the changes
+// non provided values are taken as "this didn't change", similar as setState does.
+// Put default values for all fields that are ommited.
```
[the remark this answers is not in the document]
josejulio: Thanks

src/components/CytoscapeGraph/CytoscapeGraph.tsx
josejulio: Maybe is related to #392

src/actions/ServiceGraphDataActions.ts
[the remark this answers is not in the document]
josejulio: Fixed the typo

src/components/CytoscapeGraph/CytoscapeGraph.tsx
jmazzitelli: We don't always want to force a center() - that would reset the pan thus jumping the graph when the user might not want to. If I drag the graph around, and I change the duration, or some other setting, I want the graph to remain there if possible (and this PR now does that!). I don't want the graph to jump to center when I just change the duration.

That said, if you change the LAYOUT, now you are asking Cy to redraw the entire graph. I don't think we can control much of that. IF we want to re-center the graph when changing the LAYOUT, I would be OK with that, since the entire graph is essentially going to change. This is also why I think its OK to reset the zoom, too, when changing the layout. The graph just changes too dramatically to expect the current zoom/pan levels to make sense with a new layout.

I would not worry about this for this PR. Let's get this in, as it is very nice as-is.

jshaughn: +1, it was only a comment.  Let's get this merged as soon as @mtho11 approves.  But as I said above, I'm talking about on layout change, not on every update.

src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -206,57 +198,66 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
 
     this.trafficRenderer.stop();
 
+    const isTheGraphSelected = this.cy.$(':selected').length === 0;
+
     cy.startBatch();
 
+    // Destroy badges
```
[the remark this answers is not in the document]
jmazzitelli: i like the badges. We need stinkin' badges.

PR #394, opened 2018-06-01
KIALI-834 Remove contextual info such as credentials from error messages

jotak:


src/components/NamespaceFilter/NamespaceFilter.tsx
```diff
@@ -68,8 +68,9 @@ export class NamespaceFilter extends React.Component<NamespaceFilterProps, Names
         this.setState({ filterTypeList: initialFilters });
       })
       .catch(error => {
-        console.error(error);
-        this.props.onError('Error fetching namespace list.');
+        const errMsg = API.getErrorMsg('Error fetching namespace list.', error);
```
lucasponce: I imagine this one doesn't come from axios, isn't it ?

jotak: Yes it does, as it's originated from `API.getNamespaces` so an axios call

xeviknal:
@jotak I guess that using `API.getErrorMsg` you are preventing to show the whole response data, right?

jotak:
@xeviknal yes exactly, the whole "error" contains actually a lot of contextual info including request headers (so, credentials/token), whereas with `API.getErrorMsg` it extracts just the error message.

PR #396, opened 2018-06-01
KIALI-866 Group pods of same origin

jotak:


jotak:
Do not merge => need to add unit tests first

jotak:
Screenshot with mixed grouped & ungrouped pods:
screenshot

jotak:
This was not a planned task but, after playing a bit with kiali & scaling up pods with tons of replicas, I thought it was kinda urgent 
Any remark or objection? (cc @lucasponce  @xeviknal  @aljesusg  @abonas @serenamarie125  )

abonas:
as mentioned earlier today, I'd put a consistent "title" for each one, so for example the first one won't be "16 replicas" but something more clear like ` <some pod prefix> (X replicas)`

jotak:
@abonas done
I also set the details of names as a PF Tooltip

screenshot

jotak:
Can now be merged (test added)

abonas:
@serenamarie125 would you mind having a look here?

src/pages/ServiceDetails/ServiceInfo/ServiceInfoPods.tsx
```diff
@@ -1,66 +1,97 @@
...
 import { Pod } from '../../../types/ServiceInfo';
 import { PfColors } from '../../../components/Pf/PfColors';
+import { PodsGroup, groupPods } from './ServiceInfoPodsGrouping';
 
 interface Props {
   pods?: Pod[];
 }
 
 class ServiceInfoPods extends React.Component<Props> {
+  groups: PodsGroup[];
```
lucasponce: Just curiosity. Nothing to change, something to learn.
I see several styles for adding "state" into the component.
Instead to use the proper React state, here we are using a plain js object (which is right).
Why don't use the React state ? Just thinking loud if using this could have some undesired side effect through the React lifecycle. Probably not, it seems quite simple (and correct) use case, but just wondering what is the best practice to use.
I see you use it in several places on the code, and in others people use the react state, which also looks good.

src/pages/ServiceDetails/ServiceInfo/ServiceInfoPods.tsx
```diff
@@ -1,66 +1,97 @@
...
         <Row className="row-cards-pf">
           <Col xs={12} sm={12} md={12} lg={12}>
-            {(this.props.pods || []).map((pod, u) => (
+            {(this.groups || []).map((group, u) => (
               <div className="card-pf-body" key={'pods_' + u}>
-                <h3>{pod.name}</h3>
+                <h3>
+                  {group.numberOfPods > 1 ? (
+                    <OverlayTrigger
+                      // prettier-ignore
```
lucasponce: Instead to add a prettier-ignore, better to modify the prettier rules ?

src/pages/ServiceDetails/ServiceInfo/ServiceInfoPods.tsx
```diff
@@ -1,66 +1,97 @@
...
+                      overlay={<Tooltip id={'pod_names_' + u} title="Pod Names">{group.names.join(', ')}</Tooltip>}
+                      placement="top"
+                      trigger={['hover', 'focus']}
+                    >
+                      <span>{group.commonPrefix + '... (' + group.numberOfPods + ' replicas)'}</span>
+                    </OverlayTrigger>
+                  ) : (
+                    group.commonPrefix
```
lucasponce: I wonder if for UX perspective, it should also be added a "(1 replica)" msg in the single result.
I don't know, users could think that instead of a pod, it's something like a group with no pods.
I see it's redundant but I missed the "(1 replica)" here just to see the same format.

src/pages/ServiceDetails/ServiceInfo/ServiceInfoPods.tsx
```diff
@@ -1,66 +1,97 @@
...
-                    <strong>Created at: </strong>
-                    <LocalTime time={pod.createdAt} />
+                    {group.createdAtStart === group.createdAtEnd ? (
+                      <>
+                        <strong>Created at: </strong>
+                        {new Date(group.createdAtStart).toLocaleString()}
+                      </>
+                    ) : (
+                      <>
+                        <strong>Created between: </strong>
```
lucasponce: +1 nice check

src/pages/ServiceDetails/ServiceInfo/__tests__/ServiceInfoPods.test.tsx
```diff
@@ -21,6 +21,8 @@ const pods: Pod[] = [
   }
 ];
 
+Date.prototype.toLocaleString = jest.fn(Date.prototype.toISOString);
```
lucasponce: Good trick.
I would put a comment, just because I'm quite sure we will this one in the future :-).

src/pages/ServiceDetails/ServiceInfo/ServiceInfoPods.tsx
jotak: Personally I tend to reduce the react state at its minimum, that is, it only contains data that may change through the component's life.

Here `PodsGroup` is directly derived from props and will not change unless props change. So in my understanding it's not a good candidate for state.

There's other related concepts that I don't know very well yet, such as "PureComponents" and "SFC", maybe I should use it there but I'm not sure.

src/pages/ServiceDetails/ServiceInfo/ServiceInfoPods.tsx
jotak: Prettier is sometimes tedious to configure and afaik there's no granularity in how rules are applied, it's just "on" or "off". Here prettier decides that the line should be split after `{`, but tslint refuses spaces after `{`, so they clash.

Personally I would like to completely disable prettier fanciful line-breakings, it's quite bad at that game. But when I suggested that a while ago I didn't get much positive votes :)

src/pages/ServiceDetails/ServiceInfo/ServiceInfoPods.tsx
jotak: I've no objection, @abonas or @serenamarie125 no objection too?

xeviknal: To have (1 replica) for 1-pod group sounds good to me, a little bit more homogeneous :)

src/pages/ServiceDetails/ServiceInfo/ServiceInfoPods.tsx
xeviknal: I don't have much knowledge about prettier, but I guess that we should do something when we find that is a pain for many of us. I think we should react when we have some more occurrences.

src/pages/ServiceDetails/ServiceInfo/ServiceInfoPodsGrouping.ts
```diff
@@ -0,0 +1,88 @@
...
+  createdAtStart: number;
+  createdAtEnd: number;
+  createdBy?: Reference;
+  istioContainers?: ContainerInfo[];
+  istioInitContainers?: ContainerInfo[];
+  numberOfPods: number;
+}
+
+const groupKey = (pod: Pod): string => {
+  return JSON.stringify({
```
xeviknal: Good one :) Open solution and so precise.

src/pages/ServiceDetails/ServiceInfo/ServiceInfoPods.tsx
jotak: What we could do is:
- continue to add "prettier-ignore" commands as we found issues
- when it becomes obvious that prettier is not performing well (for instance, after 3 issues with the same rule) then opt-out the rule.

src/pages/ServiceDetails/ServiceInfo/ServiceInfoPods.tsx
jotak: done

PR #397, opened 2018-06-01
KIALI-870 send the on click message to simulate someone selecting the graph.

jmazzitelli:
This will ensure all processing is done appropriate for selecting the graph.

side note: this also changes "this.cy" to the argument "cy" - we want to reference the "cy" object given to this method directly.


jmazzitelli:
Here's what the bug looks like:

bad

jmazzitelli:
Here's what it now looks like with this fix:

good

PR #398, opened 2018-06-01
KIALI-845: Fix badge positioning when there are browser scrollbars

israel-hdez:


jmazzitelli:
Is there a before-and-after screen snapshot you can provide to show the difference?

israel-hdez:
Before fix:
image

After the fix:
image

jmazzitelli:
Wow. I have never seen the badges look like that before-pix. How did you get it to do that? Is that easy to replicate?

israel-hdez:
Yes, turn the legend on. The browser will show scrollbars. Scroll down and the badges will move away from the nodes.

However, the "Show legend" button is now hidden. That will make it hard to replicate.

jmazzitelli:
I see it now... unmaximize the browser window, make it small, show legend, scroll down, watch badges go screwy.

jmazzitelli:
I can confirm this fixes the problem. when green, I'll merge

PR #399, opened 2018-06-01
Kiali-852 Fix Toolbar

mtho11:
Cleanup the toolbar. There are too many items in it now so it has started overlapping into the next line and doesn't look good. This PR moves the **Show Legend** to a Filter submenu (popover) to conserve space.

legend

https://issues.jboss.org/browse/KIALI-852


jmazzitelli:
I just tested this, looks good. We need the edge labels button to go in the filters next, that would help,.. but that's another PR I believe.

PR #400, opened 2018-06-01
quick fix for legend, until we get a better fix

jmazzitelli:
legend

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ before_install:
 script:
   - yarn prettier --list-different &&
     yarn build &&
-    yarn test --forceExit
+    yarn test --forceExit --maxWorkers=4
   - ./run_itest.sh
 before_deploy:
   - yarn set-snapshot-version $TRAVIS_BUILD_NUMBER
--- a/src/components/GraphFilter/GraphLegend.tsx
+++ b/src/components/GraphFilter/GraphLegend.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import Draggable from 'react-draggable';
-import { Card, Icon } from 'patternfly-react';
+import { Card } from 'patternfly-react';
 import { PfColors } from '../../components/Pf/PfColors';
 import { style } from 'typestyle';
 import { NestedCSSProperties } from 'typestyle/lib/types';
@@ -29,14 +29,14 @@ const Arrow = (props: ArrowProps) => {
   }
   const lineBase = style(line);
   const arrowStyle = style({ marginTop: 20, clear: 'both' });
-  const arrowDiv = style({ width: '120px', float: 'left' });
+  const arrowDiv = style({ width: '175px', float: 'left' });
   const labelStyle = style({ float: 'left', verticalAlign: 'middle', height: '20px' });
   return (
     <div className={arrowStyle}>
       <div className={arrowDiv}>
         <div className={lineBase} />
+        <div className={labelStyle}>{props.label}</div>
       </div>
-      <div className={labelStyle}>{props.label}</div>
     </div>
   );
 };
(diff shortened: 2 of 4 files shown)


abonas:
@jmazzitelli @mtho11 
It is unclear from the PR title nor from the commit message what was the problem with the legend and what was fixed, as well as what is considered a better fix and where is it tracked that one will be issued? In addition, commit message named "see if this helps travis" is also not very clear, both - why (or if) it is related to the legend, as well as what does it contain (remember that in git log no one sees the PR but as a start only the commit message)
cc @aljesusg - not sure if you were aware of any issue with the legend?

aljesusg:
The removed the onclose in the legend and they changed some style. I don't know why they had problem with this...@abonas

abonas:
@serenamarie125 did you review this change?

jmazzitelli:
Sorry - this was JIRA https://issues.jboss.org/browse/KIALI-854 - but this PR wasn't meant to address that JIRA directly, it was just to temporarily get the legend to not look so broken until we get someone to spend time to clean it up. The old legend just didn't look good at all - I'm surprised people are surprised about this.

I'm attaching an animated gif of the old legend - see below - I'd be curious to know what you LIKE about the old legend :)

Here's what's wrong:

1) When you show the legend, it pops up on the bottom - so far down for many people it only shows the legend title and it requires you to scroll down to see the bottom half. We don't want to have a scroll bar on the service graph if we can avoid it, certainly we shouldn't require it just to be able to see the legend.
2) It has an "X" but it does nothing - you click it and it doesn't close the legend. Why is this X here?
3) The labels of the color lines are not consistent - why are some of the labels UNDER the line and some labels TO THE RIGHT of the line (see "Idle")
4) The last line (No Traffic Yet) is outside the border of the legend box. Looks haphazard - the legend info should be fully enclosed in the legend outline, it should not leak outside of it.
5) Why do we have a legend with a title header "Traffic Legend"? I think its obvious it's a legend :)
6) The "Traffic Legend" title header is about 25% of the height of the legend! If we do want a title header saying its a legend, it shouldn't be that big.
7) The "Traffic Legend" title header makes the legend looks like a draggable window - but clicking/dragging the header does nothing - the legend is fixed.

legend

aljesusg:
I'm sorry but I'm confused, when the PR was merged it looked like this, so it seems that something break this @jmazzitelli 
legend

jmazzitelli:
The legend definitely did NOT look like that for the past week or so, if not longer (as you see in my animated gif). At the very least, git commit 607bd258870965f1e4c21732344ac6c8aa1c8aff has the "bad" legend in it.

Feel free to assign yourself that JIRA https://issues.jboss.org/browse/KIALI-854 and fix the legend. I think we still want to make that header smaller - "Traffic Legend" - because it takes up about 25% of the entire height of the legend. Anyway, if its draggable, closable with the X, the labels are consistently to the right, and it doesn't force the scrollbars to show up, then it would be good.

As for why the "Show Legend" button was removed and replaced with the checkbox in the filters, please see: https://issues.jboss.org/browse/KIALI-852 specifically the screen snapshot that illustrates the issue. As we keep adding more and more items to the toolbar, it started wrapping, causing things to look ugly and having the summary panel overlap the bottom of the toolbar. So we need to limit the toolbar.

jmazzitelli:
BTW: here's another JIRA @mattmahoneyrh created for the legend - https://issues.jboss.org/browse/KIALI-857

jmazzitelli:
FWIW: I just git reset my branch to this commit (which I think was the one that introduced the legend):
```
commit d4a062e9c35fe7ebf73d86a83ad9006778b38bc8 (HEAD -> master)
Author: Alberto Gutierrez <aljesusg@gmail.com>
Date:   Fri May 25 11:26:03 2018 -0700
    [KIALI-153] Show Graph legend (#361)
```
and the legend still looks and behaves as my original animated gif above. I don't see it working like yours did, @aljesusg ... I don't know what's wrong. I do know at the very least @mtho11 , @jshaughn , and @mattmahoneyrh  all see the same things I do. I am using Google Chrome 65.0.3325.181.

aljesusg:
I am using Google Chrome too, anyway I tested with Firefox too. I'll try to get back the legend and I'll ping you to check the graph. Thanks @jmazzitelli

jmazzitelli:
Interestingly, it "somewhat" works in my Firefox 52.6.0. I can move it, and the "X" does close it. But the layout of the labels are still off, and it still pops up at the bottom. So, it seems something in here is browser-dependent (even though it still isn't working 100% correctly on my firefox either).

firefox

PR #402, opened 2018-06-02
KIALI-745 Autoupdate the service graph

josejulio:
peek 2018-06-01 19-12

Adds autorefresher, the options are in the refresh button and the default refresh rate is 15s.
~~I'm using a DropdownKebab instead of a SplitButton, because the later doesn't look right inside the Toolbar.~~
Using SplitButton now, had to workaround the issue.

And added back the 5s and 10s, but the default is still 15s

screenshot_2018-06-06_12-33-15

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -18,11 +18,10 @@ export const config = () => {
         604800: 'Last 7 days',
         2592000: 'Last 30 days'
       },
-      defaultPollInterval: 5000,
+      defaultPollInterval: 15000,
       pollInterval: {
         0: 'Pause',
-        5000: '5 seconds',
-        10000: '10 seconds',
+        15000: '15 seconds',
         30000: '30 seconds',
         60000: '1 minute',
         300000: '5 minutes'
--- a/src/config.ts
+++ b/src/config.ts
@@ -21,6 +21,8 @@ export const config = () => {
       defaultPollInterval: 15000,
       pollInterval: {
         0: 'Pause',
+        5000: '5 seconds',
+        10000: '10 seconds',
         15000: '15 seconds',
         30000: '30 seconds',
         60000: '1 minute',
--- a/src/actions/ServiceGraphDataActions.ts
+++ b/src/actions/ServiceGraphDataActions.ts
@@ -87,7 +87,7 @@ export const ServiceGraphDataActions = {
       dispatch(ServiceGraphDataActions.getGraphDataStart());
       const duration = graphDuration.value;
       const restParams = { duration: duration + 's' };
-      API.getGraphElements(authentication(), namespace, restParams).then(
+      return API.getGraphElements(authentication(), namespace, restParams).then(
         response => {
           const responseData: any = response['data'];
           const graphData = responseData && responseData.elements ? responseData.elements : EMPTY_GRAPH_DATA;
--- a/src/actions/ServiceGraphFilterActions.ts
+++ b/src/actions/ServiceGraphFilterActions.ts
@@ -1,6 +1,6 @@
 // Action Creators allow us to create typesafe utilities for dispatching actions
 import { createAction } from 'typesafe-actions';
-import { EdgeLabelMode } from '../types/GraphFilter';
+import { EdgeLabelMode, PollIntervalInMs } from '../types/GraphFilter';
 
 export enum ServiceGraphFilterActionKeys {
   // Toggle Actions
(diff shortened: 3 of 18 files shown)


aljesusg:
I think that user won't know that this option is there.

jmazzitelli:
I'm not sure I like the kebob.  I think I like that split button better, even though there's a small space between the two buttons.

josejulio:
I'll try later today or tomorrow to update it to use the SplitButton and will raise an issue on `patternfly-react` (I want to test that it fails on their storybooks when used inside the Toolbar component)

serenamarie125:
Although i like the functionality this PR adds, i'd like to redesign this.  There are issues with findability ( user would never know where to look) as well as what the menu items refer to.

There is a toolbar design meeting this week, hoping that should help resolve the design direction.

pilhuhn:
Could we merge this one nevertheless (if technical review is ok), so we can get feedback of the technical working from more users and then to the UX change afterwards? 

I fear this would otherwise linger around for too long.

serenamarie125:
@pilhuhn I’m fine with that.  I’m hoping that after we have these design meetings this week, We can create jira issues that will be prioritized to follow the new design.

jmazzitelli:
I just tested this... I selected "30 seconds", saw it refresh, then selected "Pause" but saw it refresh again in 30 seconds. I think now it is paused. I'm still testing, because I want to make sure I saw what I saw, but there might be a problem with this PR.

src/config.ts
```diff
@@ -18,11 +18,10 @@ export const config = () => {
         604800: 'Last 7 days',
         2592000: 'Last 30 days'
       },
-      defaultPollInterval: 5000,
+      defaultPollInterval: 15000,
```
jotak: I agree about increasing the default, but why removing shorter interval options? While monitoring metrics we sometimes want to have a shorter refresh rate. For instance Grafana shortest is 5s .

jmazzitelli: Istio's Prometheus scrape interval is 15s. We'd essentially be making about 2 automatic calls unnecessarily to the backend server ("unnecessary" because we know the data isn't going to change until every 3rd call). To avoid stressing the server with those requests that we know will not provide new data, we limit it to the scrape interval.

jotak: Good point... but scrape interval can still be configured.

jmazzitelli: Configurable by Prometheus, yes, but AFAIK, it is not configurable from Istio (I don't know of a way to configure that from, say, the Istio install methods short of a user going in and manually modifying the ansible/helm scripts OR going in and somehow manually modifying the Prometheus configuration files when already installed - in either case, I don't see many people doing that - certainly the vast majority of Istio users won't be changing the scrape interval).

IMO, I do not believe we should give people the option to auto-refresh any lower than then default scrape interval when, for the vast majority of users, those lower refresh rates would be detrimental to the system performance (by hitting the server with 2 unnecessary requests) while providing no useful value, especially when there is no documented (and presumably supported) way to lower the scrape interval.

jotak: Fair enough :)

jmazzitelli:
BTW: I just noticed, the layout dropdown option is acting very weird with this PR, too. The layouts are changing even if I select the same kind of layout. And Dagre is not looking like the Dagre I've come to know and love. I don't know what could have changed, but this PR seems to have broken the layout stuff.

jmazzitelli:
Hmmm... just shutdown, restarted, and reloaded the UI, and now the layout is working again. Dunno what happened there.

josejulio:
> This isn't working 100% correctly. Try this:
> 
> Set refresh to 15 seconds. Watch every 15 seconds and you will see the spinner - showing that it is correctly refreshing every 15s.
> Set it to Pause
> Notice it will refresh within 15 seconds again, and then stop refreshing.
> Do the same test with 30 seconds - watch the spinner show it refreshing every 30s
> Set it to pause
> Notice the spinner will show it refreshed again within 30 seconds. But then stop
> So it appears when you select "Pause" it doesn't pause it immediately. It will refresh one more time after you selected "Pause" but then it will stop after that. So we need to get it to not do that - when I select Pause, I want it to not refresh automatically after that. But it isn't doing that - it is auto-refreshing one more time after I select Pause.

I though i had this issue fixed. Working on this.

src/config.ts
jotak: Actually... I was wondering why I was seeing updates every 5 seconds on the metrics screen. Turns out the default scrape interval is 5s, not 15: https://github.com/istio/istio/blob/2ac5ccdaa3fcd5dda5b3874b81e7ed7f33ffb80f/install/kubernetes/helm/istio/charts/prometheus/templates/configmap.yaml#L17-L33

jmazzitelli: Interesting. That is the new helm stuff, so it looks like installing on Kubernetes with Helm will set it to 5. Do we know if the same is done when using the ansible installer on OpenShift?

jmazzitelli: Looking at my OpenShift installed via Ansible installer, I also see this 5s - so it looks like they do override that 15s:

```
global:
  scrape_interval: 15s
  scrape_timeout: 10s
  evaluation_interval: 1m
scrape_configs:
- job_name: istio-mesh
  scrape_interval: 5s
  scrape_timeout: 5s
```

So this means we should put "5s" as an option in the refresh drop down... though I would agree that we should still default to something slower, like 15s or 30s. (this PR has it defaulting to 15s which is fine by me).

jmazzitelli:
Just tested... this works now. I'm OK with merging if everyone else is.

jmazzitelli:
Oh, and I like the split button now :)

src/config.ts
josejulio: Will add back the 5s but keep the default on 15s

PR #403, opened 2018-06-03
KIALI-753 Add config validation in IstioConfig

lucasponce:
- Show validations on IstioConfig list view.
- Filtering and sorting by validation state (valid, not valid, not validated).
- Add markers and annotations on editor in IstioConfig and Service Details.
- Add API for fetching validations per object.

image

image

### DIFF
--- a/src/pages/ServiceDetails/ServiceInfo.tsx
+++ b/src/pages/ServiceDetails/ServiceInfo.tsx
@@ -75,49 +75,34 @@ class ServiceInfo extends React.Component<ServiceDetails, ServiceInfoState> {
     const virtualServices = this.props.serviceDetails.virtualServices || [];
     const destinationRules = this.props.serviceDetails.destinationRules || [];
 
-    for (let i = 0; i < routeRules.length; i++) {
-      if (
+    validationChecks.hasRouteRuleChecks = routeRules.some(
+      routeRule =>
         this.props.validations['routerule'] &&
-        this.props.validations['routerule'][routeRules[i].name] &&
-        !this.props.validations['routerule'][routeRules[i].name].valid
-      ) {
-        validationChecks.hasRouteRuleChecks = true;
-        break;
-      }
-    }
+        this.props.validations['routerule'][routeRule.name] &&
+        !this.props.validations['routerule'][routeRule.name].valid
+    );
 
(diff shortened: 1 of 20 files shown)


lucasponce:
I have added a minor change to address KIALI-835, which is to add a missing icon on tab title.
image

lucasponce:
Ops, my last commit didn't update the test.
Please, don't merge, going to fix that.

src/pages/ServiceDetails/ServiceInfo.tsx
```diff
@@ -54,6 +62,65 @@ class ServiceInfo extends React.Component<ServiceDetails, ServiceInfoState> {
...
+      hasVirtualServiceChecks: false,
+      hasDestinationRuleChecks: false
+    };
+
+    const routeRules = this.props.serviceDetails.routeRules || [];
+    const destinationPolicies = this.props.serviceDetails.destinationPolicies || [];
+    const virtualServices = this.props.serviceDetails.virtualServices || [];
+    const destinationRules = this.props.serviceDetails.destinationRules || [];
+
+    for (let i = 0; i < routeRules.length; i++) {
```
xeviknal: What do you think about using [`Array.some()`](https://www.tutorialspoint.com/typescript/typescript_array_some.htm) method?

xeviknal:
@lucasponce tell me If you like using what I proposed. Let me know, and I will merge that.

src/pages/ServiceDetails/ServiceInfo.tsx
jotak: Ah, it's like `Stream.anyMatch()` in java, good to know

xeviknal: Same in Ruby `collection.any?`, pretty powerful.
Also in TS, there is the `collection.all?` method: [`Array.every()`](https://www.tutorialspoint.com/typescript/typescript_array_every.htm). It returns true if all the elements in the array match the condition passed by param. Powerful too.

lucasponce: I am not a big fan of getting rid of the old school 'for' loop :-).
If the same file has some code using them, normally I would maintain it, but on this case, as it was the first loop I change it.

PR #404, opened 2018-06-04
KIALI-269 Warning Message when the Istio version is not supported

aljesusg:
Show a warning message when the Istio version is not supported.
JIRA: [KIALI-269](https://issues.jboss.org/browse/KIALI-269)

**Require**: https://github.com/kiali/kiali/pull/248

- About Modal Refactored
- Now when the user do login an getStatus action is executed and get versions, products and warningMessages (The application no longer asks for a request each time ithe users go to aboutmodal). This trigger to MessageCenter if there is a warningMessage.

_Example with Istio installed 0.7.1 and the Istio required is >= 0.8_

error_version

@abonas I don't know if it is better to show a message in the navigation bar between the spinner and the bell icon or it's ok with this. Should we ask to @serenamarie125 ?

### DIFF
--- a/src/actions/HelpDropdownActions.ts
+++ b/src/actions/HelpDropdownActions.ts
@@ -1,0 +1,50 @@
+import { createAction } from 'typesafe-actions';
+import * as API from '../services/Api';
+import { Component } from '../store/Store';
+import { MessageType } from '../types/MessageCenter';
+import { MessageCenterActions } from './MessageCenterActions';
+
+export enum HelpDropdownActionKeys {
+  STATUS_REFRESH = 'STATUS_REFRESH'
+}
+
+export const HelpDropdownActions = {
+  statusRefresh: createAction(
+    HelpDropdownActionKeys.STATUS_REFRESH,
+    (status: { [key: string]: string }, components: Component[], warningMessages: string[]) => ({
+      type: HelpDropdownActionKeys.STATUS_REFRESH,
+      status: status,
+      components: components,
+      warningMessages: warningMessages
+    })
+  ),
+  refresh: () => {
+    return (dispatch, getState) => {
+      const state = getState();
+
+      API.getStatus('Bearer ' + state.authentication.token.token).then(
+        status => {
+          dispatch(
+            HelpDropdownActions.statusRefresh(
+              status['data']['status'],
+              status['data']['externalServices'],
+              status['data']['warningMessages']
+            )
+          );
+          status['data']['warningMessages'].forEach(wMsg => {
+            dispatch(MessageCenterActions.addMessage(wMsg, 'default', MessageType.WARNING));
+          });
+        },
+        error => {
+          dispatch(
+            MessageCenterActions.addMessage(
+              API.getErrorMsg('Error fetching status.', error),
+              'default',
+              MessageType.WARNING
+            )
+          );
+        }
+      );
+    };
+  }
+};
--- a/src/components/Nav/HelpDropdown.tsx
+++ b/src/components/Nav/HelpDropdown.tsx
@@ -2,9 +2,21 @@ import * as React from 'react';
 import { Dropdown, Icon, MenuItem } from 'patternfly-react';
 
 import AboutUIModal from '../About/AboutUIModal';
+import { Component } from '../../store/Store';
 
-class HelpDropdown extends React.Component {
+type HelpDropdownProps = {
+  status: { [key: string]: string };
+  components: Component[];
+  warningMessages: string[];
+};
+type HelpDropdownState = {};
+
+class HelpDropdown extends React.Component<HelpDropdownProps, HelpDropdownState> {
   about: any;
+
+  constructor(props: HelpDropdownProps) {
+    super(props);
+  }
 
   render() {
     return (
@@ -13,6 +25,8 @@ class HelpDropdown extends React.Component {
           ref={about => {
             this.about = about;
           }}
+          status={this.props.status}
+          components={this.props.components}
         />
         <Dropdown componentClass="li" id="help">
           <Dropdown.Toggle useAnchor={true} className="nav-item-iconic">
--- a/src/containers/HelpDropdownContainer.ts
+++ b/src/containers/HelpDropdownContainer.ts
@@ -1,0 +1,21 @@
+import { KialiAppState } from '../store/Store';
+import { connect } from 'react-redux';
+import HelpDropdown from '../components/Nav/HelpDropdown';
+
+import { HelpDropdownActions } from '../actions/HelpDropdownActions';
+
+const mapStateToProps = (state: KialiAppState) => ({
+  status: state.statusState.status,
+  components: state.statusState.components,
+  warningMessages: state.statusState.warningMessages
+});
+
+const mapDispatchToProps = (dispatch: any) => ({
+  refresh: () => dispatch(HelpDropdownActions.refresh())
+});
+
+const HelpDropdownConnected = connect(
+  mapStateToProps,
+  mapDispatchToProps
+)(HelpDropdown);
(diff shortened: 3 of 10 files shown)


src/store/Store.ts
```diff
@@ -54,13 +54,25 @@ export interface LoginState {
   logged: boolean;
   logging: boolean;
 }
+
+export interface KialiProduct {
```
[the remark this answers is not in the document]
aljesusg: Yes, It could be confusing, I'll change it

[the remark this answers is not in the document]
lucasponce: I see that "products" is used in the context of redux, then perhaps my comment has not sense then.
If "products" is widely used I guess then it is following same terminology.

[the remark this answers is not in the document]
lucasponce: Ah no, I see StatusState.products is introduced on this PR.

So, I re-affirm that "products" sounds weird to me :-).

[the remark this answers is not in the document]
aljesusg: Yes, but I think we should change all or move KialiProduct to Product with the idea to avoid people think that are a components or products of Kiali.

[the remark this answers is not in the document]
aljesusg: The problem is that in backend come like products no components. I'll do a PR to change it in backend

src/reducers/HelpDropdownState.ts
```diff
@@ -0,0 +1,24 @@
...
+import { HelpDropdownActionKeys } from '../actions/HelpDropdownActions';
+
+const INITIAL_STATE: StatusState = {
+  status: {},
+  products: [],
+  warningMessages: []
+};
+
+// This Reducer allows changes to the 'serviceGraphDataState' portion of Redux Store
+const HelpDropdownState = (state: StatusState = INITIAL_STATE, action) => {
```
lucasponce: Missing type of action ?

src/actions/HelpDropdownActions.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// synchronous action creators
+export const HelpDropdownActions = {
+  statusRefresh: createAction(
+    HelpDropdownActionKeys.STATUS_REFRESH,
+    (status: { [key: string]: string }, products: KialiProduct[], warningMessages: string[]) => ({
+      type: HelpDropdownActionKeys.STATUS_REFRESH,
+      status: status,
+      products: products,
```
abonas: product? Kiali is a project, not a product :)
product is something commercial

src/components/Nav/HelpDropdown.tsx
```diff
@@ -2,17 +2,31 @@ import * as React from 'react';
 import { Dropdown, Icon, MenuItem } from 'patternfly-react';
 
 import AboutUIModal from '../About/AboutUIModal';
+import { KialiProduct } from '../../store/Store';
 
-class HelpDropdown extends React.Component {
+type HelpDropdownprops = {
```
lucasponce: Minor, I 'HelpDropdownprops' -> 'HelpDropdownProps' ?

lucasponce: HelpDropdownprops -> HelpDropdownProps ?

src/containers/HelpDropdownContainer.ts
```diff
@@ -0,0 +1,18 @@
...
+  status: state.statusState.status,
+  products: state.statusState.products,
+  warningMessages: state.statusState.warningMessages
+});
+
+const mapDispatchToProps = (dispatch: any) => ({
+  refresh: () => dispatch(HelpDropdownActions.refresh())
+});
+
+const HelpDropdownConnected = connect(mapStateToProps, mapDispatchToProps)(HelpDropdown);
```
lucasponce: This comment is due I'm learning redux, too.
So HelpDropdownConnected is the HelpDropdown component but linked with the reducers, which will bind the data from the store with the component.

src/actions/HelpDropdownActions.ts
aljesusg: ^^ business skills++

src/containers/HelpDropdownContainer.ts
aljesusg: Yes with this you use the component HelpDropdown an connect to him the properties and state  of redux.
In line 7 I am passing to this components status, products and warningMessages like a properties, I asigned them the actual value of the state in redux. In the line 13 I am passing actions to the component like props, in this case I am passing the method refresh of HelpDropdownActions defined in actions/HelpDropdownActions, so in the component I can use this method with this.props.refresh

src/reducers/HelpDropdownState.ts
aljesusg: Not, If you see another reducers there is not type in action because you can store any kind of object.

serenamarie125:
A couple of questions
1. what's the purpose of showing the message
1. should it be warning or error level?
1. rather than using a toast notification should we use a modal?  Does the toast go away after 8 seconds, or are you forcing the user to dismiss it?

aljesusg:
Hi @serenamarie125 about questions.

1. Notify the user that the vesion installed of Istio is not supported by Kiali and parts of the application may fail

2. In the JIRA the requirement is a warning message, I think this is because Kiali could work without problem with a versionnot supported.

3. That was my question, maybe we should show a message in the nav between logo and bell icon, the message not go away the user must click to dismiss this message

src/actions/HelpDropdownActions.ts
```diff
@@ -0,0 +1,52 @@
+import { createAction } from 'typesafe-actions';
+import * as API from '../services/Api';
+import { Component } from '../store/Store';
+import { MessageType } from '../types/MessageCenter';
+import { MessageCenterActions } from './MessageCenterActions';
+
+export enum HelpDropdownActionKeys {
+  STATUS_REFRESH = 'STATUS_REFERSH'
```
israel-hdez: Typo: REF**ER**SH

aljesusg:
@serenamarie125 What do you think about this?

serenamarie125:
@aljesusg i can't see the image :( 

In OpenShift, they have something called System Status ... where they add an item to the top nav banner.  
screen shot 2018-06-13 at 1 57 41 pm

In your case, you could add the icon & provide an info tip when selected which would give the appropriate message (rather than having the text directly on the top banner).

@cshinn what do you think ?

cshinn:
@serenamarie125 I think something like system status makes sense. Especially if there aren't other things that would go in the notification drawer. If we could tell which parts are working and which aren't we could put warning markers next to the broken pieces.

aljesusg:
@serenamarie125 you need to install an istio version lower than supported to see the message. 

About the message, we have the [notification drawer](https://rawgit.com/patternfly/patternfly-react/gh-pages/index.html?knob-is%20Drawer%20Expandable=true&selectedKind=patternfly-react%2FCommunication%2FNotification%20Drawer&selectedStory=Notification%20Drawer&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs) for this. So yes we could add like  openshift the infotip.

serenamarie125:
@aljesusg I think it depends on how prominent this needs to be.

Typically the notification drawer is something that is used to notify the user of something.  If it's a higher level "system status" kind of issue that you want the user to be informed of at all times, I'd think that having a warning dialog in the masthead may be more prominent.  

@abonas @aljesusg thoughts?

aljesusg:
@lucasponce @xeviknal Can you review this please?

aljesusg:
The UX changes about the notificacion will be done in [KIALI-984](https://issues.jboss.org/browse/KIALI-984) about Notification Drawer

serenamarie125:
Looks very nice!  Thank you @aljesusg

PR #405, opened 2018-06-04
KIALI-876 Fix marker set in wrong line

lucasponce:
- Parse correctly offset in multiple weights/labels
- Add more tests

image

### DIFF
--- a/src/types/AceValidations.ts
+++ b/src/types/AceValidations.ts
@@ -105,7 +105,7 @@ const parseCheck = (yaml: string, check: ObjectCheck): AceCheck => {
       - spec/destination
       - spec/name (this is used for destination Rules which maps spec/name into destinationName)
       - spec/precedence/<value>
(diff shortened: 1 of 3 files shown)


src/types/AceValidations.ts
```diff
@@ -104,8 +105,8 @@ const parseCheck = (yaml: string, check: ObjectCheck): AceCheck => {
       - spec/destination
       - spec/name (this is used for destination Rules which maps spec/name into destinationName)
       - spec/precedence/<value>
-      - spec/route/weight/<value>
-      - spec/route[<value>]/labels
+      - spec/route[<nWeight>/weight/<value>
```
xeviknal: I guess there is missing a `]` right here, isn't it?

lucasponce: +1 yeah, it was missing as it is in the comment. Thx.

PR #406, opened 2018-06-04
[POC] KIALI-756 New Overview Page

jshaughn:
This creates a new Overview page that is currently defined to be the new default.  The goal is to provide an initial "traffic light" view for overall "App" health.  Initially an "App" is basically a namespace.  And it is represented by the service[s] from that namespace which handle external requests.  Initially that is constrained to "istio-ingress" and "unknown".  That may need to be expanded if we want to see services called only from other services.  The view also does not look for orphan services (dotted line services). This is currently by design but I'm not sure if it is correct (note, adding it in may not be completely trivial).

The service nodes reflect health both by their coloring and via a text label.  By clicking the service you get the familiar service node summary with the addition of the standard health icon and standard popover.

This relies on the server additions in https://github.com/kiali/kiali/pull/247.

Here is an example (a bit dull with only green bookinfo traffic):

screenshot

### DIFF
--- a/src/actions/ServiceGraphDataActions.ts
+++ b/src/actions/ServiceGraphDataActions.ts
@@ -4,6 +4,7 @@ import { Duration } from '../types/GraphFilter';
 import * as API from '../services/Api';
 import * as MessageCenter from '../utils/MessageCenter';
 import { authentication } from '../utils/Authentication';
+import * as H from '../utils/Health';
 
 const EMPTY_GRAPH_DATA = { nodes: [], edges: [] };
 
@@ -11,6 +12,11 @@ export enum ServiceGraphDataActionKeys {
   GET_GRAPH_DATA_START = 'GET_GRAPH_DATA_START',
   GET_GRAPH_DATA_SUCCESS = 'GET_GRAPH_DATA_SUCCESS',
   GET_GRAPH_DATA_FAILURE = 'GET_GRAPH_DATA_FAILURE',
+  CLEANUP_GRAPH_DATA = 'CLEANUP_GRAPH_DATA',
+  GET_OVERVIEW_GRAPH_DATA_START = 'GET_OVERVIEW_GRAPH_DATA_START',
+  GET_OVERVIEW_GRAPH_DATA_SUCCESS = 'GET_OVERVIEW_GRAPH_DATA_SUCCESS',
+  GET_OVERVIEW_GRAPH_DATA_FAILURE = 'GET_OVERVIEW_GRAPH_DATA_FAILURE',
+  CLEANUP_OVERVIEW_GRAPH_DATA = 'CLEANUP_OVERVIEW_GRAPH_DATA',
   HANDLE_LEGEND = 'HANDLE_LEGEND'
 }
 
@@ -36,13 +42,14 @@ const decorateGraphData = (graphData: any) => {
       rate4XX: undefined,
       rate5XX: undefined,
       rateSelfInvoke: undefined,
+      globalStatus: undefined,
       hasCB: undefined,
+      hasMissingSidecars: undefined,
       hasRR: undefined,
       isDead: undefined,
       isGroup: undefined,
       isRoot: undefined,
-      isUnused: undefined,
-      hasMissingSidecars: undefined
+      isUnused: undefined
     }
   };
   if (graphData) {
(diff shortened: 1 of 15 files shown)


jshaughn:
screenshot from 2018-06-04 16-51-05

abonas:
The JIRA says it is a POC. 
What does it mean in terms of this PR?  didn't see it's indicated as POC here.
Was the design discussed with @serenamarie125 ?

pilhuhn:
Can you create screenshots with more than 1 application deployed?
E.g. bookinfo + 2 synthetic meshes ?

jshaughn:
> The JIRA says it is a POC. What does it mean in terms of this PR? didn't see it's indicated as POC here.

It is labeled as Do Not Merge and the PR exists so that we can talk around it.  If you prefer just prefix POC into the title, it's fine by me.

abonas:
@jshaughn it is labeled as 'do not merge *yet*' :)
which means that if this label is off, it can be merged.

jshaughn:
> which means that if this label is off, it can be merged.

I guess I don't understand, the label only gets removed manually.  I'll add "POC" to the title but remember, if someone removes that prefix this could get merged.  After being approved by someone other than me.

jshaughn:
@pilhuhn Here is another pic.  We need to talk about where we want to go with this stuff...

screenshot from 2018-06-05 16-46-03

jshaughn:
Hi @serenamarie125 , this POC is pretty much the design to be discussed, this PR is just a talking point. It's based on the  [mesh doc](https://docs.google.com/document/d/13l6uA7eP3ndf3gL8FnHBbi10oaxmlxu-Cj8ry-SRdl0/edit?ts=5ade3377#heading=h.jj6p9kpqnpr4).   We have a meeting scheduled today, I think, to discuss health topics so hopefully that will give us a chance to talk more about this and the general topic.

theute:
I'm not super enthousiast about a new overview page again in the shape of a graph
Beside not being something we agreed we would work in that sprint, we have some "traffic light" view in the "Services" tab, and the planned task https://docs.jboss.org/display/KIALI/Sprint+%236 was to reflect the service view in the existing graph, can't we stick to this ?
We'll need some definition of "application" and probably a higher level view, but I count on getting user feedback about how they shape an application, the namespace doesn't seem to be the right thing.

lucasponce:
> Idea of a card-based overview-page

I prefer this "dashboard" view instead of a new graph.
I think the new graph introduces some sort of replication in the features and it is not clear why shouldn't be put together with some "zooming" feature instead of a split in overview/graph.

A card view also it's flexible and it can combine different type of info and get the attention of the user.

jshaughn:
> Beside not being something we agreed we would work in that sprint

Not sure when we didn't agree to do this is S6, A POC was set as a goal in the planning meeting, as a subtask of KIALI-655  and all of the idea generation (one level graph, namespace=app, etc)  was in the doc and associated meetings referenced above.  And my status reports have been talking about this.   You must have missed it.

I honestly don't know why I'm seemingly getting hassled about this work from various people.  It's a POC, the code is there exactly to show it as a proof of concept, to talk around.  It's fine if people don't like the way it looks or if it doesn't pan out.  We can toss it, revise it, etc.   It's an impl of exactly what was discussed and agreed on with @pilhuhn as a first pass.

jshaughn:
Closing this, I think the work here is done, the POC has been able to generate more discussion about what we may want to do in this area.

jotak:
@jshaughn +1 it was clearly referred as a POC so I don't see any issue in having opened this PR, it triggered all the discussions that was necessary (and even a bit more)
Thanks for the work!

pilhuhn:
> We'll need some definition of "application" and probably a 
> higher level view, but I count on getting user feedback 
> about how they shape an application, the namespace doesn't 
> seem to be the right thing.

We have in the light of the mesh that span more than one namespace
document and discussion said, that for now an application is what can 
be reached from the outside (that has traffic from a route or fro Istio-Ingress(*)).
This mesh can then incorporate services from multiple namespaces. 
This is the best we can get now, as there is no consensus anywhere about
what label(-key) to denote for an application. And even we did not come 
up with such a thing.
While the namespace per se is not the right thing, It can help putting a name
on the application. Of course in cases where more than one route leads "into" a
namespace, we need a differentiator and also add the service name or such.

PR #407, opened 2018-06-05
[Kiali-850] Configure Redux-Persist

aljesusg:
With this we can persist the state of redux.

**Require**: https://github.com/kiali/kiali-ui/pull/413

- Keep the session and options after refresh app (F5)
- Let puppeter (Itest) navigate to others page for @abonas proposal 
- Fix Bug [Kiali-850](https://issues.jboss.org/browse/KIALI-850)
- Work in the session timeout [KIALI-533](https://issues.jboss.org/browse/KIALI-533) reported by @mtho11 

Documentation : [Redux-persist ](https://github.com/rt2zz/redux-persist) 5582 starts 398 forks

redux-persist

Things to improve.

- Select another kind of storage and encrypt this information. Currently there is no sensitive information so we can do this after.

## Storage Engines

- **localStorage** `import storage from 'redux-persist/lib/storage'`
- **sessionStorage** `import storageSession from 'redux-persist/lib/storage/session'`
- **AsyncStorage** react-native `import { AsyncStorage } from 'react-native'`
- **[localForage](https://github.com/mozilla/localForage)** recommended for web
- **[electron storage](https://github.com/psperber/redux-persist-electron-storage)** Electron support via [electron store](https://github.com/sindresorhus/electron-store)
- **[redux-persist-filesystem-storage](https://github.com/robwalkerco/redux-persist-filesystem-storage)** react-native, to mitigate storage size limitations in android ([#199](https://github.com/rt2zz/redux-persist/issues/199), [#284](https://github.com/rt2zz/redux-persist/issues/284))
- **[redux-persist-node-storage](https://github.com/pellejacobs/redux-persist-node-storage)** for use in nodejs environments.
- **[redux-persist-sensitive-storage](https://github.com/CodingZeal/redux-persist-sensitive-storage)** react-native, for sensitive information (uses [react-native-sensitive-storage](https://github.com/mCodex/react-native-sensitive-info)).
- **[redux-persist-expo-securestore](https://github.com/Cretezy/redux-persist-expo-securestore)** react-native, for sensitive information using Expo's SecureStore. Only available if using Expo SDK (Expo, create-react-native-app, standalone).
- **[redux-persist-fs-storage](https://github.com/leethree/redux-persist-fs-storage)** react-native-fs engine
- **[redux-persist-cookie-storage](https://github.com/abersager/redux-persist-cookie-storage)** Cookie storage engine, works in browser and Node.js, for universal / isomorphic apps
- **[redux-persist-weapp-storage](https://github.com/cuijiemmx/redux-casa/tree/master/packages/redux-persist-weapp-storage)** Storage engine for wechat mini program, also compatible with wepy
- **custom** any conforming storage api implementing the following methods: `setItem` `getItem` `removeItem`. (**NB**: These methods must support promises)

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "react-scripts-ts": "2.15.1",
     "redux": "3.7.2",
     "redux-mock-store": "1.5.1",
+    "redux-persist": "5.9.1",
     "redux-thunk": "2.2.0",
     "ssri": "6.0.0",
     "typesafe-actions": "1.1.2",
--- a/src/store/ConfigStore.ts
+++ b/src/store/ConfigStore.ts
@@ -11,7 +11,8 @@ declare const window;
 
 const persistConfig = {
   key: 'root',
-  storage: storage
+  storage: storage,
+  whitelist: ['authentication']
 };
 
 const composeEnhancers =
--- a/src/store/ConfigStore.ts
+++ b/src/store/ConfigStore.ts
@@ -1,9 +1,18 @@
 import { createStore, applyMiddleware, compose } from 'redux';
 import { KialiAppState } from './Store';
+import { persistStore, persistReducer } from 'redux-persist';
 import rootReducer from '../reducers';
 import thunk from 'redux-thunk';
 
+// defaults to localStorage for web and AsyncStorage for react-native
+import storage from 'redux-persist/lib/storage';
+
 declare const window;
+
+const persistConfig = {
+  key: 'root',
+  storage: storage
+};
 
 const composeEnhancers =
   (process.env.NODE_ENV === 'development' && window && window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__) || compose;
@@ -13,12 +22,12 @@ const configureStore = (initialState?: KialiAppState) => {
   const middlewares = [thunk];
   // compose enhancers
   const enhancer = composeEnhancers(applyMiddleware(...middlewares));
+  // persist reducers
+  const persistentReducer = persistReducer(persistConfig, rootReducer);
   // create store
-  return createStore(rootReducer, initialState, enhancer);
+  return createStore(persistentReducer, initialState, enhancer);
 };
 
 // pass an optional param to rehydrate state on app start
-const store = configureStore();
-
-// export store singleton instance
-export default store;
+export const store = configureStore();
+export const persistor = persistStore(store);
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -3,10 +3,11 @@ import { Router, withRouter } from 'react-router-dom';
 import { Provider } from 'react-redux';
 import './App.css';
 import Navigation from '../containers/NavigationContainer';
-import store from '../store/ConfigStore';
+import { store, persistor } from '../store/ConfigStore';
 import axios from 'axios';
 import { GlobalActionKeys } from '../actions/GlobalActions';
 import history from './History';
+import { PersistGate } from 'redux-persist/lib/integration/react';
 
 // intercept all Axios requests and dispatch the LOADING_SPINNER_ON Action
 axios.interceptors.request.use(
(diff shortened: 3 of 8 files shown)


aljesusg:
@mtho11 after learning about this we can set which states we wanna persist or not 

```javascript
// BLACKLIST
const persistConfig = {
  key: 'root',
  storage: storage,
  blacklist: ['navigation'] // navigation will not be persisted
};

// WHITELIST
const persistConfig = {
  key: 'root',
  storage: storage,
  whitelist: ['navigation'] // only navigation will be persisted
};
```

but I think that actually we can persist all the information, so What do you think about merge this? Thanks

src/store/ConfigStore.ts
```diff
@@ -13,12 +22,12 @@ const configureStore = (initialState?: KialiAppState) => {
   const middlewares = [thunk];
   // compose enhancers
   const enhancer = composeEnhancers(applyMiddleware(...middlewares));
+  // persist reducers
+  const pReducer = persistReducer(persistConfig, rootReducer);
   // create store
-  return createStore(rootReducer, initialState, enhancer);
+  return createStore(pReducer, initialState, enhancer);
```
mtho11: Can we spell out `persistentReducer` instead of `pReducer`? I don't know what a `pReducer` is without looking around in the code and see how the object was instantiated. No need to be terse, meaningful names is   key. Besides, in production it will get minified anyway.

aljesusg: Changed @mtho11 Thanks

package.json
```diff
@@ -31,6 +31,7 @@
     "react-scripts-ts": "2.15.1",
     "redux": "3.7.2",
     "redux-mock-store": "1.5.1",
+    "redux-persist": "^5.9.1",
```
mtho11: Oh one last thing. Can we change this to fixed versions?

mtho11:
As for future steps, I don't think we want to encrypt (at least not yet). There is value in being able to browse the local browser DB (and grab snapshots).

package.json
aljesusg: Fixed to 5.9.1

aljesusg:
This fail when try to convert object

```json
filterState:{showLegend: false, showNodeLabels: true, showCircuitBreakers: false, showRouteRules: true, showMissingSidecars: true, …}
graphData: {nodes: Array(9), edges: Array(7)}
graphDataTimestamp: 1528318507
hideLegend:true
isLoading: false
sidePanelInfo: {kind: "graph", graphReference: Core}
```

Ir could be the undefined

```json
graphData: 
   edges:Array(7)
        0: data:{rate: "0.818", rate3XX: undefined, rate4XX: undefined, rate5XX: undefined, percentErr: undefined, …}
```

aljesusg:
@mtho11 we avoid this fail if we persist only the authentication , the problem was in the toolbar object it seems there are some converting circular in the toolbar that it's not recommended we should check the ServiceGraphDataActions to avoid this.  

Anyway we can continue persisting only the authenticacion and add others later how globalState, messageCenter and namespaces, but ServiceGraph should be reviewed.
```javascript
// This defines the Kiali Global Application State
export interface KialiAppState {
  // Global state === across multiple pages
  // could also be session state
  globalState: GlobalState;
  // page settings
  authentication: LoginState;
  messageCenter: MessageCenterState;
  namespaces: any;
  serviceGraph: ServiceGraphState;
}

```
. This let us go with KIALI-533 about session timeout. What do you think?

aljesusg:
We can set blacklist to serviceGraph and all would work less the persist state for the serviceGraph @mtho11 
```javascript
const persistConfig = {
  key: 'root',
  storage: storage,
  blacklist: ['serviceGraph']
};
```

mtho11:
@aljesusg sure we can start small and expand. Sounds good!

mtho11:
Tested looks good now.

PR #413, opened 2018-06-06
[KIALI-850] Add redux-persist library

aljesusg:
Added library redux-persist 

[Kiali-850](https://issues.jboss.org/browse/KIALI-850)


aljesusg:
Thanks @mtho11 after merge I'll remove the label in the other PR

PR #414, opened 2018-06-06
KIALI-741 fix chart height issue due to large legend

jotak:
Note, this is certainly not a perfect patch. Charts still appear somewhat inconsistent because they may have different heights - which makes them not always aligned. I'm pretty sure we'll get some complains about that :)

However this is IMO still much better than the current state as described in the jira issue ( https://issues.jboss.org/browse/KIALI-741 ), having charts completely unreadable. And I do not have a better solution so far, but if anyone has ideas please let me know.


jotak:
Screenshot:
Screenshot

PR #417, opened 2018-06-06
KIALI-731: Add button to show YAML of route rules, etc.

israel-hdez:
For route rules, destination policies, destination rules and virtual
services.

This is how it looks for route rules:
image

For destination policies, virtual services and destination rules, it looks like this:
image

### DIFF
--- a/src/pages/ServiceDetails/ServiceInfo/ServiceInfoDestinationPolicies.tsx
+++ b/src/pages/ServiceDetails/ServiceInfo/ServiceInfoDestinationPolicies.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { Col, Row } from 'patternfly-react';
+import { Col, Row, Icon } from 'patternfly-react';
 import { DestinationPolicy, EditorLink } from '../../../types/ServiceInfo';
 import LocalTime from '../../../components/Time/LocalTime';
 import DetailObject from '../../../components/Details/DetailObject';
@@ -23,9 +23,12 @@ class ServiceInfoDestinationPolicies extends React.Component<ServiceInfoDestinat
               return (
                 <div className="card-pf-body" key={'rule' + i}>
                   <div>
-                    <h3>
-                      <Link to={this.props.editorLink + '?destinationpolicy=' + dPolicy.name}>{dPolicy.name}</Link>
-                    </h3>
+                    <h3>{dPolicy.name}</h3>
+                  </div>
+                  <div>
+                    <Link to={this.props.editorLink + '?destinationpolicy=' + dPolicy.name}>
+                      Show Yaml <Icon name="angle-double-right" />
+                    </Link>
                   </div>
                   <div>
                     <strong>Created at</strong>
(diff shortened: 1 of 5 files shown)


israel-hdez:
@lucasponce I've changed the button to an explicit link. I hope this is good enough to be merged. If we get UX feedback, I can change it again.

lucasponce:
@israel-hdez I think there is a problem with this link in terms of UX.
If the assume that the rule is to add this link, we need to refactor all places in Istio Config, too.
UX should use same representation of an element in all places, not only of this, for that we need to @serenamarie125 or @abonas to lead this.

Personally, I think the previous page where the title has the link in a greater style as @xeviknal worked.
So, I don't see the need for this change.

For that, UX should drive this, because there is not an agreement about what should be, and I guess we are all in the same page related we can't mix different styles for same UX use case, that would be more confusing to the end user.

israel-hdez:
@lucasponce Ok, I wasn't seeing the problem. Then, let's wait for UX feedback.
Probably, this PR and the related jira can be closed if the header is enough.

serenamarie125:
@israel-hdez I'm sorry, I'm confused as to what was added as part of this PR - is it the Show YAML link?
And what happens when the link is selected?  Any chance you can use recordit, or some animated gif to show the user flow?

lucasponce:
@serenamarie125, the main discussion on this PR is about the style / UX for the access into the ACE editor.
Today, we can see the yaml from ServiceDetails and from Istio Config.
On ServiceDetails, clicking on the link of the name is the way to access, in the same case it happens when navigating to others objects from the Istio Config; in special Istio Rules, as those objects have multiple instances.

So, main request was to add a "show the yaml" button/link in the ServiceDetails, but that breaks the use case in Istio Config. Then, the main point is to define how to access to that.

We have used patternfly link as standard way, for example, some screenshots in the ServiceDetails:
image
image
And others in the IstioConfig -> Rules:
image
image

So, personally I have no prefence, in OpenShift console, it is used a button "Edit YAML" to mark this action at the right of a list, but on this case, we don't have a "Edit" action, but just a "See" as content is note editable yet. That was the rationale to use a link in first attempt.

Any proposal should take into consideration the Istio Config section to use the same UX solution per same use case (access to the yaml details).

I hope then this discussion is more clear.

lucasponce:
An additional comment: the link pattern is used for similar scenarios as well.
For example, in the tab sources to link with related services.
image
So, we are not adding a "go to the service XYZ" link, it is implicit that clicking in the service name, it navigates to it.

mwringe:
Design issue opened here: https://github.com/kiali/kiali-design/issues/4

Lets discuss it more in there.

lucasponce:
From kiali/kiali-design#4 @serenamarie125 has proposed some mockups that will involve some refactors in the Service Details and Istio Config pages.
We can rebase this PR and accept it as a provisional solution until the UXD changes are addressed.
@israel-hdez let me know what do you think.
Thanks.

israel-hdez:
@lucasponce Rebased, for provisional solution.

PR #420, opened 2018-06-07
Fix login, check the token is still valid after refresh

aljesusg:
Redux-persist store the state in the localstorage, If you try later or with another build the token can fail if pass the expiredTime but you still have this information in localStorage and this avoid the login.

Now there is a validation of the token, If it's not valid you'll need login again. After we can use this "checkCredentials" to validate the session-timeout too


src/config.ts
```diff
@@ -45,6 +45,19 @@ export const config = () => {
 
 export const KEY_CODES = { TAB_KEY: 9, ENTER_KEY: 13, ESCAPE_KEY: 27 };
 
+export const HTTP_CODES = {
```
mtho11: Can we put the `HTTP_CODES` in its own in the `types` directory. I think this would be useful for others as well.

aljesusg: we should put KEY_CODES too I think.

aljesusg: Moved to types/Common

PR #422, opened 2018-06-07
KIALI-817 Change node shape based on namespace.

cfcosta:
This is the ui counterpart to [kiali/kiali#252](https://github.com/kiali/kiali/pull/252), so check there if you need information.

### DIFF
--- a/src/actions/ServiceGraphDataActions.ts
+++ b/src/actions/ServiceGraphDataActions.ts
@@ -41,7 +41,8 @@ const decorateGraphData = (graphData: any) => {
       isGroup: undefined,
       isRoot: undefined,
       isUnused: undefined,
-      hasMissingSC: undefined
+      hasMissingSC: undefined,
+      isOutside: undefined
     }
   };
   if (graphData) {
--- a/src/components/CytoscapeGraph/graphs/GraphStyles.ts
+++ b/src/components/CytoscapeGraph/graphs/GraphStyles.ts
@@ -40,15 +40,22 @@ export class GraphStyles {
         css: {
           // color: PfColors.Black,
           content: (ele: any) => {
+            const service = ele.data('serviceName') || ele.data('service').split('.')[0];
+            const namespace = ele.data('namespace');
             const version = ele.data('version');
+
             if (!ele.data('showNodeLabels')) {
               return '';
             }
+
+            if (ele.data('isOutside')) {
+              return `${service}\n${namespace}`;
+            }
+
             if (ele.data('parent')) {
               return version;
             }
-            const name = ele.data('service') || ele.data('id');
-            const service = name.split('.')[0];
+
             return version && version !== 'unknown' ? service + '\n' + version : service;
           },
           'background-color': PfColors.Black200,
@@ -77,6 +84,12 @@ export class GraphStyles {
         selector: 'node[isRoot]',
         style: {
           shape: 'diamond'
+        }
+      },
+      {
+        selector: 'node[isOutside]',
+        style: {
+          shape: 'pentagon'
         }
       },
       {


jotak:
@cfcosta , it seems that changing colors for namespaces would not play well with another ongoing topic that is displaying health on nodes. The current plan was to change node border colors according to health. Looks like we're conflicting here? Too many colored information would, I think, make everything confusing.

See also https://issues.jboss.org/browse/KIALI-785 
@theute , @serenamarie125 what's your opinion?

theute:
Agreed, do we really need a new color here ? 
(IMO the color line of the shape might reflect the health of the whole namespace, but we're not there yet)

cfcosta:
@jshaughn can you review this, please?

src/actions/ServiceGraphDataActions.ts
```diff
@@ -42,7 +42,8 @@ const decorateGraphData = (graphData: any) => {
       isGroup: undefined,
       isRoot: undefined,
       isUnused: undefined,
-      hasMissingSidecars: undefined
+      hasMissingSidecars: undefined,
+      isOutsideNamespace: undefined
```
jshaughn: You need to rebase to fix conflicts (e.g. 'hasMissingSidecar's is now 'hasMissingSC').  With respect to the server PR, 'isOutsideNamespace' is now just 'isOutside'

src/components/CytoscapeGraph/graphs/GraphStyles.ts
```diff
@@ -31,15 +31,23 @@ export class GraphStyles {
         css: {
           // color: PfColors.Black,
           content: (ele: any) => {
+            const name = ele.data('service') || ele.data('id');
+            const namespace = name.split('.')[1];
+            const service = name.split('.')[0];
```
jshaughn: you should now be able use ele.data('serviceName') and ele.data('namespace') and avoid the split completely.

src/components/CytoscapeGraph/graphs/GraphStyles.ts
```diff
@@ -31,15 +31,23 @@ export class GraphStyles {
...
             const version = ele.data('version');
+
+            if (ele.data('isOutsideNamespace')) {
```
jshaughn: again, 'isOutside'

src/components/CytoscapeGraph/graphs/GraphStyles.ts
```diff
@@ -70,6 +78,12 @@ export class GraphStyles {
           shape: 'diamond'
         }
       },
+      {
+        selector: 'node[isOutsideNamespace]',
```
jshaughn: one more, 'isOutside'

jshaughn:
Adding DNM label.  We do this whenever the ui PR first requires merge of a server PR.

src/components/CytoscapeGraph/graphs/GraphStyles.ts
```diff
@@ -40,15 +40,22 @@ export class GraphStyles {
         css: {
           // color: PfColors.Black,
           content: (ele: any) => {
+            const service = ele.data('serviceName') || ele.data('service').split('.')[0];
+            const namespace = ele.data('namespace');
             const version = ele.data('version');
+
+            if (ele.data('isOutside')) {
```
josejulio: I think this should be *after* the line 53
```
if (!ele.data('showNodeLabels')) {
   return '';
}
```

else the label will be showing even though showNodeLabels is false

cfcosta: True! Fixed.

serenamarie125:
I don't see a screenshot of what the PR explains ... do you have a screenshot of what it looks like?  What shape are you using?  If there are objects from multiple namespaces, are you using the same shape? or different ones?

PR #426, opened 2018-06-08
KIALI-910 Show colored borders to represent health in graph

jotak:
Since this is my first incursion in Graph code maybe I haven't done things as you would have expected, so I'm all ears about comments.

This shows colored borders on nodes for each _service_ (and not _version_).

Example:
health in graph

This is a first step but maybe discussions can lead to other subtasks being created. Like improving performances, reducing number of calls or showing more information or per-version health maybe.

Currently if performs one backend api call per service, asynchronously from the main graph fetching.
I think it has advantages (graph can be displayed without having to wait for all health information being collected) and drawbacks (many http calls), so probably an optimization will be to fetch health information for all services within a namespace at once.

Then, there is also what @jshaughn suggested: restraining health checking to services that are suspected to not perform well (regarding their error rates).

### DIFF
--- a/src/components/CytoscapeGraph/CytoscapeGraph.tsx
+++ b/src/components/CytoscapeGraph/CytoscapeGraph.tsx
@@ -86,7 +86,7 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
   componentDidUpdate(prevProps: CytoscapeGraphProps, prevState: CytoscapeGraphState) {
     const cy = this.getCy();
     this.processGraphUpdate(cy);
-    if (this.props.elements !== prevProps.elements && this.props.namespace.name !== 'all') {
+    if (this.props.elements !== prevProps.elements) {
       this.updateHealth(cy);
     }
   }
@@ -330,8 +330,10 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
     cy.nodes().forEach(ele => {
       const fqService = ele.data('service');
       if (fqService && (ele.data('isGroup') || !ele.data('parent'))) {
-        const service = fqService.split('.')[0];
-        API.getServiceHealth(authentication(), this.props.namespace.name, service, duration).then(r => {
+        const serviceParts = fqService.split('.');
+        const service = serviceParts[0];
+        const namespace = serviceParts[1];
+        API.getServiceHealth(authentication(), namespace, service, duration).then(r => {
           const health = r.data;
           ele.data('health', health);
           const status = H.computeAggregatedHealth(health);
--- a/src/components/CytoscapeGraph/CytoscapeGraph.tsx
+++ b/src/components/CytoscapeGraph/CytoscapeGraph.tsx
@@ -14,7 +14,6 @@ import * as API from '../../services/Api';
 import { KialiAppState } from '../../store/Store';
 import { GraphParamsType } from '../../types/Graph';
 import { EdgeLabelMode } from '../../types/GraphFilter';
-import { Health } from '../../types/Health';
 import { authentication } from '../../utils/Authentication';
 import * as H from '../../utils/Health';
 
@@ -62,7 +61,7 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
     this.updateLayout = false;
   }
 
-  shouldComponentUpdate(nextProps: any, nextState: any) {
+  shouldComponentUpdate(nextProps: CytoscapeGraphProps, nextState: CytoscapeGraphState) {
     this.updateLayout =
       this.props.graphLayout !== nextProps.graphLayout ||
       (this.props.elements !== nextProps.elements &&
@@ -84,8 +83,12 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
     this.cyInitialization(this.getCy());
   }
 
-  componentDidUpdate() {
-    this.processGraphUpdate(this.getCy());
+  componentDidUpdate(prevProps: CytoscapeGraphProps, prevState: CytoscapeGraphState) {
+    const cy = this.getCy();
+    this.processGraphUpdate(cy);
+    if (this.props.elements !== prevProps.elements && this.props.namespace.name !== 'all') {
+      this.updateHealth(cy);
+    }
   }
 
   render() {
@@ -269,17 +272,6 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
     if (this.props.showTrafficAnimation) {
       this.trafficRenderer.start();
     }
-
-    // Asynchronously fetch health
-    cy.nodes().forEach(ele => {
-      const fqService = ele.data('service');
-      if (fqService && (ele.data('isGroup') || !ele.data('parent'))) {
-        const service = fqService.split('.')[0];
-        API.getServiceHealth(authentication(), this.props.namespace.name, service).then(r =>
-          this.updateHealth(r.data, ele)
-        );
-      }
-    });
   }
 
   private handleTap = (event: CytoscapeClickEvent) => {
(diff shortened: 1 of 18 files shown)


jotak:
Second commit (accept or reject - I can split that to another jira as well) shows how we could show additional health information in the side panel. 

side panel

src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -265,6 +269,17 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
     if (this.props.showTrafficAnimation) {
       this.trafficRenderer.start();
     }
+
+    // Asynchronously fetch health
+    cy.nodes().forEach(ele => {
```
josejulio: I'm concerned that multiple layout changes will keep updating this data, even if our refresh interval is every 15 seconds or so.
Can we validate that we are only updating these fields when `this.props.elements` change?

jotak: Hmm... But this.props.elements may not change on graph refresh, no? The health should be IMO updated at every refresh interval.

josejulio: The contents might not change, but this will be true after a graph refresh
`this.props.elements !== prevprops.elements`

jotak:
@serenamarie125 / @pilhuhn / @theute what do you think in terms of UX? Does it look as you expected?

theute:
Yes, this sounds good to me.
Now that it is implemented with a real screenshot, I do realize that it's not as trivial to interprete the graph as I hoped, so we'll need to think harder, but if it reflects the health of the service as we calculate today then it's fine by me.
PS: I can't judge the perfomance impact.

abonas:
it looks good! 
some comments and thoughts:
1. there should be an alternative for color blind. @serenamarie125 ?
2. the 'reviews' box is orange, but services inside are grey. I would match them to be both orange.
3. flow wise, when looking at the bigg(er) picture than one service, I'm trying to understand what story does this tell to the user - looking at the screenshot above, unless it is not a real use case, in that case - please ignore this.
So:
2 versions of "reviews" service are supposedly orange (=warning), send "green" (=everything is working great) traffic to a red (=error/problem) "ratings" service. If the ratings service has an issue, what does the "green" traffic to it mean? and how is that even possible?

jotak:
@abonas :
> the 'reviews' box is orange, but services inside are grey. I would match them to be both orange.

This is because currently, the health is per-service, not per version of service. Some unhealthy behaviour could be caused by only one version of the service, so I think it could be misleading to show them all unhealthy.
I think we could add another subtask to this story to be able to provide per-version health, and this will be solved.

> flow wise, when looking at the bigg(er) picture than one service, I'm trying to understand what story does this tell to the user [...]

Well sometimes traffic is OK but there are still some concerns. If I remember correctly, to generate this screenshot I scaled up some pods and didn't wait that they were all correctly running. So health was reporting some deployment issues, even if the traffic was flowing. This was a temporary unhealthy state and if I had waited one minute or two, all would have been green.

src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -313,6 +320,29 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
       .sort()
       .every((eId, index) => eId === aIds[index]);
   }
+
+  private updateHealth(cy: any) {
```
jshaughn: I didn't expect the health check to be performed at the request of the client.  I thought it would be done on the server-side.  I'm not really sure how I feel about being done this way, I guess it's OK but the HTTP traffic may get intense.  And I'm not sure how this plays with auto-refresh (or manual refresh for that matter).   I guess it can be applied incrementally to the graph.  Something makes me think this may be better on the back-end, we may want to consider it going forward but at least for now I think this is fine.

src/components/CytoscapeGraph/graphs/GraphStyles.ts
```diff
@@ -1,6 +1,7 @@
 import { PfColors } from '../../../components/Pf/PfColors';
 import { EdgeLabelMode } from '../../../types/GraphFilter';
 import { config } from '../../../config';
+import { FAILURE, DEGRADED } from '../../../utils/Health';
```
jshaughn: Really minor, but FAILURE doesn't sound like a health status.  FAILING, UNHEALTHY, SICK, etc :)  Certainly not a blocker.

src/pages/ServiceGraph/SummaryPanelNode.tsx
```diff
@@ -14,6 +14,7 @@ import { Icon } from 'patternfly-react';
 import { authentication } from '../../utils/Authentication';
 import { Link } from 'react-router-dom';
 import { shouldRefreshData } from './SummaryPanelCommon';
+import { HealthIndicator, DisplayMode } from '../../components/ServiceHealth/HealthIndicator';
```
jshaughn: Adding the icon an popover to the summaries is good,  I had added this in my Overview POC as well.  We could consider adding one to the whole-graph summary, as an aggregate of the graph services (not in this PR) .

src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -80,8 +83,12 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
     this.cyInitialization(this.getCy());
   }
 
-  componentDidUpdate() {
-    this.processGraphUpdate(this.getCy());
+  componentDidUpdate(prevProps: CytoscapeGraphProps, prevState: CytoscapeGraphState) {
+    const cy = this.getCy();
+    this.processGraphUpdate(cy);
+    if (this.props.elements !== prevProps.elements && this.props.namespace.name !== 'all') {
```
jshaughn: Why are we excluding 'all' ?

src/components/CytoscapeGraph/graphs/GraphStyles.ts
```diff
@@ -159,6 +160,20 @@ export class GraphStyles {
           opacity: '0.3'
         }
       },
+      {
+        selector: 'node.' + DEGRADED.name,
+        style: {
+          'border-color': DEGRADED.color,
+          'border-width': '2px'
```
jshaughn: I think we could consider a wider border to make these nodes more visible, especially when zoomed out in a larger mesh.

abonas:
and indeed we certainly need @serenamarie125 here :)

src/components/CytoscapeGraph/CytoscapeGraph.tsx
jotak: well this was my biggest concern in terms of architecture as well and the main advantage of doing this way, as I explain in the PR description, is that this way it won't slow down the graph display, since all health information is gathered independently from graph generation. So if fetching health is slow, it will only impact health display.

Both approaches has pros and cons (as you say, higher http traffic), but I think the more we're adding data to fetch, the more it becomes advantageous to fetch extra-data asynchronously.

By the way I created an optimization subtask ( https://issues.jboss.org/browse/KIALI-911 ) that could be used, for instance, to fetch all service's health at once.

src/components/CytoscapeGraph/graphs/GraphStyles.ts
jotak: Ok then, maybe most importantly I should review the associated text because "Item XYZ Failure" is what we currently display to the user

src/pages/ServiceGraph/SummaryPanelNode.tsx
jotak: +1 I thought about something similar too :)

josejulio:
Note, i could only review the code, not test it (I'm having some problems with my env).

src/components/CytoscapeGraph/CytoscapeGraph.tsx
jotak: I'm relying on `this.props.namespace.name` to get health on services. I haven't found the `namespace` info available somewhere else. So obviously, 'all' didn't make it and resulted in empty health for all services.
I guess I should add `namespace` info in all graph nodes data from the backend. Unless someone sees another way to get the information.

But while thinking about it, maybe the other tasks about multi-namespaces feature already did that?

josejulio: I think is being worked here:

https://github.com/kiali/kiali/pull/252/files#diff-89df2dfb7238cfb78adc212f2a0581aeR35

jotak: Oh, you're right, and as @jshaughn commented there we have the namespace as a part of "service" string. I'll fix that then.

src/components/CytoscapeGraph/graphs/GraphStyles.ts
jotak: I'll try with 3px. Anyway that's something easy to fine-tune afterwards.

src/components/CytoscapeGraph/CytoscapeGraph.tsx
josejulio: yes but that's not merged yet

serenamarie125:
Based on our meeting last week, we were working through some designs for health this week . I guess we weren't fast enough, we were hoping to get a proposal complete tomorrow.

src/components/CytoscapeGraph/CytoscapeGraph.tsx
jotak: @josejulio it's something already there: namespace is included in the "fully qualified service name" that we get as input, such as 'productpage.bookinfo.blahblah' so let's use that.

theute:
I'll  merge and then we'll incorporate the changes suggested by @serenamarie125 . (I guess this is ok with you Serena)
Visually this is not a huge change, it will help to see if connector health and service health make sense together and if the health should be computed differently (Several people including me, are confused by the screenshots, but it does reflect the health as we compute it, it just highlight something)

PR #427, opened 2018-06-08
KIALI-757: Add expand feature in the detailed metrics page

israel-hdez:
~~Styles are a little hacky. But I hope it's ok.~~
expandfeature

### DIFF
--- a/src/pages/ServiceDetails/MetricsChartBase.tsx
+++ b/src/pages/ServiceDetails/MetricsChartBase.tsx
@@ -1,20 +1,31 @@
 import * as React from 'react';
-import { LineChart } from 'patternfly-react';
+import { LineChart, Icon } from 'patternfly-react';
 import { TimeSeries } from '../../types/Metrics';
+import { style } from 'typestyle';
 
 type MetricsChartBaseProps = {
   familyName: string;
+  onExpandRequested?: () => void;
 };
+
+const expandBlockStyle = style({
+  marginBottom: '-1.5em',
+  zIndex: 1,
+  position: 'relative',
+  textAlign: 'right'
+});
 
 abstract class MetricsChartBase<Props extends MetricsChartBaseProps> extends React.Component<Props> {
   protected abstract get controlKey(): string;
   protected abstract get seriesData(): any;
 
   protected nameTimeSeries = (groupName: string, matrix: TimeSeries[]): TimeSeries[] => {
+    const tsName = this.props.familyName.replace('Output: ', '').replace('Input: ', '');
+
     matrix.forEach(ts => {
       const labels = Object.keys(ts.metric).map(k => ts.metric[k]);
 
-      ts.name = this.props.familyName + groupName;
+      ts.name = tsName + groupName;
       if (labels.length !== 0) {
         ts.name += '{' + labels.join(',') + '}';
       }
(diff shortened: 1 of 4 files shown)


src/pages/ServiceDetails/MetricsChartBase.tsx
```diff
@@ -1,20 +1,31 @@
...
+import { style } from 'typestyle';
 
 type MetricsChartBaseProps = {
   familyName: string;
+  onExpandRequested?: () => void;
 };
 
+const expandBlockStyle = style({
+  marginBottom: '-1.5em',
+  zIndex: 1,
```
mtho11: Why do we have to change the z-index?

src/pages/ServiceDetails/ServiceMetrics.tsx
```diff
@@ -11,6 +11,31 @@ import { authentication } from '../../utils/Authentication';
 import HistogramChart from './HistogramChart';
 import MetricChart from './MetricChart';
 import { Histogram, MetricGroup } from '../../types/Metrics';
+import history from '../../app/History';
+import { Link } from 'react-router-dom';
+import { style } from 'typestyle';
+
+const expandedChartContainerStyle = style({
+  position: 'absolute',
```
mtho11: This has to be an absolute layout?

israel-hdez: Yes and no. All these styles can be omitted and the chart will look good. However:
* The chart limits itself to 350px height. So, most of the time it wouldn't use the available vertical space. It is too hard (for me) to correctly set the height without the absolute positioning.
* Also, this is a "hack" to place the expanded chart on top of the tabs (and hide the tabs) to gain more space for the expanded chart.

This is why I said that styles are a little hacky. If you don't like it, I'll just remove all these styles and let the chart to limit itself to 350px height, and allowing the tabs to be visible.

src/pages/ServiceDetails/MetricsChartBase.tsx
israel-hdez: You can see that I have a negative `margin-bottom`. This is to place the _Expand_ link close to the chart (where I think it looks better). But this also means that the c3 control and the _Expand_ link are overlapping. The link is not clickable if the `z-index` is not adjusted.

src/pages/ServiceDetails/ServiceMetrics.tsx
mtho11: @israel-hdez yes, I think that would be the better solution. The absolute stuff makes me nervous

jotak: @israel-hdez, maybe if you provide height as C3 chart params instead of style, you will avoid the 350px max-height. It can be provided as props: http://c3js.org/reference.html#size-height

israel-hdez: @jotak The problem is that it isn't possible to set the right height without calculating it in js/ts. But, probably, I'll go this way. Let me try some CSS before doing it via javascript - I don't like too much to do in JS what CSS should be doing :)

jotak: I know exactly what you mean... I already introduced that kind of dubious height calculation here. Hope we can find a better way at some point.

PR #432, opened 2018-06-12
KIALI-750 Add istio mTLS support into service graph

xeviknal:
The aim of this PR is to display which communications are being handled using the istio mutual TLS.

edge-mtls-with-rates

[kiali/kiali#269](https://github.com/kiali/kiali/pull/269) needs to be merged at the same time.

Connected to this PR, I have created a UI JIRA for allowing edges to display icons. This would help us to make this feature fancier: https://issues.jboss.org/browse/KIALI-935

### DIFF
--- a/src/components/CytoscapeGraph/graphs/GraphStyles.ts
+++ b/src/components/CytoscapeGraph/graphs/GraphStyles.ts
@@ -25,6 +25,14 @@ export class GraphStyles {
       return PfColors.Green400;
     };
 
+    const getTLSValue = (ele: any, tlsValue: string, nonTlsValue: string): string => {
+      if (ele.data('enabledmTLS') && ele.data('edgeLabelMode') === EdgeLabelMode.MTLS_ENABLED) {
+        return tlsValue;
+      } else {
+        return nonTlsValue;
+      }
+    };
+
     return [
       {
         selector: 'node',
@@ -116,13 +124,19 @@ export class GraphStyles {
                 return percentRate > 0 ? percentRate.toFixed(0) + '%' : '';
               }
               case EdgeLabelMode.MTLS_ENABLED: {
-                return ele.data('enabledmTLS') ? 'mTLS' : '';
+                return ele.data('enabledmTLS') ? '\ue923' : '';
               }
               default:
                 return '';
             }
           },
           'curve-style': 'bezier',
+          'font-family': (ele: any) => {
+            return getTLSValue(ele, 'PatternFlyIcons-webfont', 'inherit');
+          },
+          'text-rotation': (ele: any) => {
+            return getTLSValue(ele, '0deg', 'autorotate');
+          },
           'font-size': '7px',
           'line-color': (ele: any) => {
             return getEdgeColor(ele);
@@ -135,7 +149,6 @@ export class GraphStyles {
             return getEdgeColor(ele);
           },
           'text-margin-x': '6px',
-          'text-rotation': 'autorotate',
           width: 1
         }
       },
--- a/src/components/CytoscapeGraph/graphs/GraphStyles.ts
+++ b/src/components/CytoscapeGraph/graphs/GraphStyles.ts
@@ -115,6 +115,9 @@ export class GraphStyles {
                 const percentRate = ele.data('percentRate') ? parseFloat(ele.data('percentRate')) : 0;
                 return percentRate > 0 ? percentRate.toFixed(0) + '%' : '';
               }
+              case EdgeLabelMode.MTLS_ENABLED: {
+                return ele.data('enabledmTLS') ? 'mTLS' : '';
+              }
               default:
                 return '';
             }
--- a/src/types/GraphFilter.ts
+++ b/src/types/GraphFilter.ts
@@ -12,7 +12,8 @@ export enum EdgeLabelMode {
   HIDE = 'hide',
   REQUESTS_PER_SECOND = 'requestsPerSecond',
   REQUESTS_PERCENT_OF_TOTAL = 'requestsPercentOfTotal',
-  LATENCY_95TH_PERCENTILE = 'latency95thPercentile'
+  LATENCY_95TH_PERCENTILE = 'latency95thPercentile',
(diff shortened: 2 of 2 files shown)


lucasponce:
@xeviknal isn't this going to clash with current traffic edges labels ?
I mean, if you switch on all edge info, can be visualized correctly ?
If yes, I don't mind this solution until @serenamarie125 or @abonas review the UX aspects.

pilhuhn:
I am not happy with the "word"ing. In my opinion it should be a badge (or set of if TLS can also be one-way) or a different line type (double line for TLS vs single line without).

jshaughn:
Instead of adding a text label, which may conflict with other labeling or be hidden, maybe we should consider changing the source "arrow" shape of the edge.  Cytoscape has several options.  A simple horizontal line could perhaps signal mTLS: like this:

pilhuhn:
Or a double-arrow ?  ---->>

abonas:
I +1 the above suggestions of a visual cue for TLS and etc instead of textual one.
@serenamarie125  your inputs please :pray: ?

jshaughn:
@pilhuhn Right, although I don't think that is not a cytoscape option.  Here are the [arrow styles: ](http://manual.cytoscape.org/en/stable/Styles.html#available-shapes-and-line-styles).

jmazzitelli:
If we do opt for arrowhead shapes to denote this, we'd need to also add it to the legend - no one will remember what a "diamond" arrowhead means versus a regular arrowhead (as an example - I have no idea what the shape should be).

xeviknal:
I've been playing a bit with shapes.

IMO, I totally discard playing with different shapes than arrows:
source-circle
source-tee-target-diamond

But I do like playing with what @jshaughn  said, the pipe:

source-tee
target-triangle-tee

Playing with the concept of the pipe, it came up the idea of having a pipe next to each node that perform authentication. For example, mutual TLS should use two pipes, one for source another for destination. In case of simple TLS, the pipe would be placed next to the destination node.
See the example:

source-tee-target-triangle-tee

(reducing the scale)
source-tee-target-triangle-tee-smaller

In case we want to add support of simple TLS. This would be the graph:
target-triangle-tee-smaller

jshaughn:
LGTM!

abonas:
brief summary of a meeting with @serenamarie125 and Chris (github user for Chris? :))
1. a lock icon will be shown in the middle of the link if there is a secure connection (to simplify as first step, and not get into all the use cases of bidirectional, one way, etc.)
2. the non secured connections will not be indicated to make the graph less cluttered.
3. the lock will be switched on/off similar to other badges/labels.

serenamarie125:
@abonas it's @cshinn ... I think that the above is a good solution, until we get a better understanding of the overall use cases

jshaughn:
Not sure but I don't think icons are supported in that way.

xeviknal:
Finally got that! 
@jshaughn It looks like it is possible to use icons within edges, look at that:

edge-lock-green

@abonas @serenamarie125 @cshinn I think this is close to what we talked. What do you think?

pilhuhn:
While the shape of the lock is not exactly to my taste, I think this is a good solution.
We should do the same for a circuit breaker, as this is something that would open/close the connection between two services, which is represented by the edge.

@xeviknal How does it look when edge labels and lock are present?

jmazzitelli:
> @xeviknal How does it look when edge labels and lock are present?

That's one thing we need to see - make sure the labels don't clash.

I think we should create another JIRA for the CB badge. Remember, for the CB at least, every incoming request would have a CB badge on it if we move away from badging the node and do all the edges. Not sure if it would be too noisy if we do that.

Actually, would it be too noisy with this lock everywhere too? I would want to see what a "true" mesh looks like that uses the TLS stuff - is that lock going to be EVERYWHERE?

xeviknal:
> While the shape of the lock is not exactly to my taste, I think this is a good solution.

Great! Regarding to the lock icon, I am using the patternfly one. It was suggested by UX/UD and make sense to follow as much as we can the icon palette.

> We should do the same for a circuit breaker, as this is something that would open/close the connection between two services, which is represented by the edge.

Just added a JIRA for that. https://issues.jboss.org/browse/KIALI-954

> @xeviknal How does it look when edge labels and lock are present?
> That's one thing we need to see - make sure the labels don't clash.

I added a new option in the edge label dropdown. This way, we don't have two different labels shown at the same time.

edge-filter

> Actually, would it be too noisy with this lock everywhere too? I would want to see what a "true" mesh looks like that uses the TLS stuff - is that lock going to be EVERYWHERE?

Good point. I think that we would need to know if people will use istio with auth enabled by default or not. Any case, revert the situation to show only the non-secured edges, it won't be though. I guess after some more feedback from customers and community we will have it clear. IMO, this is a good starting point.

xeviknal:
> Actually, would it be too noisy with this lock everywhere too? I would want to see what a "true" mesh looks like that uses the TLS stuff - is that lock going to be EVERYWHERE?

I've got an idea for a subsequent PR: Istio allows admins to enable by default mTLS between all its nodes. So what we could do is:

1. If default mTLS enabled: show only those edges that are not secured with mTLS (no TLS and one way TLS).
2. In case of default mTLS disabled: show those edges with mTLS and simple TLS.

jshaughn:
@xeviknal , good job getting the lock icon on the edge.  Although, I have to say I much prefer the arrow styling; it's less noisy, will cause less overlap, and does not get in the way of any other edge labeling.

serenamarie125:
@xeviknal, this is great!  I think that icon should remain the typical PF icon color rather than green.  Regarding the icon design itself, this is the PatternFly icon - if you have concerns on the icon design itself we could ask the team to do some research, but I suggest we stick with the PF icon for consistency.

The arrow styling is definitely less noisy, but conveys nothing to the user imo. The lock does denote secure.  

Before changing the experience going forward, let’s understand upcoming use cases and functionality to be sure we are considering the overall experience of Kiali as a whole.

xeviknal:
Check the final outcome out:

security-layer-graph

PR #435, opened 2018-06-13
KIALI-931: UXD toolbar refactor for Display settings and badges

mtho11:
First steps in UXD changes for Kiali service graph toolbar.

- [x] Implement in UI
- [x] Change to Single _Graph Settings_ dropdown eliminating the _Badges_ dropdown and combining.
- [x] Change edge labels to radio buttons
- [x] Hookup Layout handling
- [X] Hookup Edge label mode

Display Settings:
radio-btn

Jira: https://issues.jboss.org/browse/KIALI-931
Design Issue: https://github.com/kiali/kiali-design/issues/2


abonas:
1. Shouldn't the badges be inside the settings menu like the other options? Why are they separate?
2. It will be helpful to have a description in the PR what does the first step include, and what are the next steps and their content.
3. the JIRA itself also needs to have more info and details about its content and scope.

jmazzitelli:
I think the UXD suggestion of putting badging settings under Display settings, too, makes sense. This is what they suggested (but I like your drop-downs better as it gives us the ability to add more later and not blow up the size of that main Display Settings popup menu)

ui

mtho11:
@abonas there were two template options I was following this option:
template2
I guess you (like mazz) were liking the other option. This easily changeable.

src/containers/DisplaySettingsContainer.tsx
```diff
@@ -90,49 +110,117 @@ export const GraphLayers: React.SFC<GraphLayersProps> = props => {
...
   return (
-    <OverlayTrigger overlay={popover} placement="bottom" trigger={['click']} rootClose={true}>
-      <Button>
-        <Icon name="filter" />
-      </Button>
-    </OverlayTrigger>
+    <>
+      <OverlayTrigger overlay={displaySettingsPopover} placement="bottom" trigger={['click']} rootClose={true}>
+        <Button>
+          Display Settings <Icon name="angle-down" />
```
serenamarie125: this should be "Graph Settings"

serenamarie125:
Please align the first dropdown to be left aligned with the page title, and use 15px spacing between each of the items on the toolbar

src/containers/DisplaySettingsContainer.tsx
mtho11: Ok, thanks @serenamarie125

mtho11:
@serenamarie125 Should the Patternfly-react stuff not have 20px spacing on the Toolbar? Especially on the first element.
pf-react-toolbar

serenamarie125:
@mtho11 from what I understand, that margin is provided in PF3 core as well.  These issues will be addressed in PF ( where individual items will not have margins/spacing built in ).

Sounds like others have solved this by either: remove your margin or set it to -20px

mtho11:
Just fixed a bug where redux values were overwriting the values coming in from Url params (**the url is source of truth**) at strange times. I have commented out the redux portion so we can start using the refactored toolbar. I will enable the redux portion in another PR.

src/containers/GraphSettingsContainer.tsx
```diff
@@ -0,0 +1,244 @@
...
+}
+
+// inherit all of our Reducer state section  and Dispatch methods for redux
+type GraphSettingsProps = ServiceGraphDispatch & ServiceGraphFilterState & GraphParamsType;
+
+// Allow Redux to map sections of our global app state to our props
+const mapStateToProps = (state: KialiAppState) => ({
+  showLegend: state.serviceGraph.filterState.showLegend,
+  showNodeLabels: state.serviceGraph.filterState.showNodeLabels,
+  // edgeLabelMode: state.serviceGraph.filterState.edgeLabelMode,
```
josejulio: Is this disconnected?

josejulio: I tested and `edgeLabelMode` is working. I'm wondering why is this left here commented, can we remove it?

mtho11: Yes, it is disconnected as stated in above comment.

mtho11: I don't want to uncomment because it will get redux connectd in another PR.

PR #436, opened 2018-06-13
KIALI-633 Change Empty Service Graph message on loading or on error

lucasponce:
image


src/components/CytoscapeGraph/EmptyGraphLayout.tsx
```diff
@@ -19,6 +21,23 @@ type EmptyGraphLayoutState = {};
 
 export default class EmptyGraphLayout extends React.Component<EmptyGraphLayoutProps, EmptyGraphLayoutState> {
   render() {
+    if (this.props.isError) {
+      return (
+        <EmptyState className={emptyStateStyle}>
+          <EmptyStateTitle>Error loading Service Graph</EmptyStateTitle>
+          <EmptyStateInfo>
+            Service Graph cannot be loaded. Please access to the Message Center for more details.
```
jotak: I'm not sure if "Message Center" is a name that the user knows about. Maybe just "Service Graph cannot be loaded" is enough, and we can assume the user will see the red alert icon in message center?

jotak: Or even better, can we show a link "click here for details" that opens the message center?

lucasponce: Well, Message Center is how the component is showed:
image

But I think is good idea to add a link to open it, I'm going to add it.

lucasponce: @jotak changed, let me know how it looks.

jshaughn:
LGTM (better late than never...)

PR #437, opened 2018-06-14
KIALI-940 Double-clicking a node from another namespaces navigates to it

cfcosta:


src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -161,10 +164,43 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
...
-      const cytoscapeEvent = getCytoscapeBaseEvent(evt);
-      if (cytoscapeEvent) {
-        this.handleTap(cytoscapeEvent);
+    const injectDoubleTap = fun => {
+      let lastTap, currentTap, tapTimeout;
+
+      return (evt: any) => {
+        currentTap = evt.target;
+
+        tapTimeout = setTimeout(() => (lastTap = undefined), 350);
```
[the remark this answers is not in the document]
josejulio: Also, I think this could trigger the "single" click too (maybe), not sure if it will behave OK
`setTimeout(() => {lastTap = undefined; fun(evt)}, 350);`
So a double click won't trigger a single click (which is how i think it should behave)
The problem is that the delay (350ms) might be noticeable and annoying.

src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -267,6 +306,21 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
     }
   }
 
+  private handleDoubleTap = (event: CytoscapeClickEvent) => {
+    if (!event.summaryTarget.data('isOutsideNamespace')) {
+      return;
+    }
+
+    this.context.router.history.push(
```
josejulio: I think this is going to be replaced by a redux state here: https://github.com/kiali/kiali-ui/pull/435 but not really sure

@mtho11 might know more.

src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -161,10 +164,43 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
...
+        const cytoscapeEvent = getCytoscapeBaseEvent(evt);
+
+        if (cytoscapeEvent) {
+          this.handleTap(cytoscapeEvent);
+        }
+      })
+    );
+
+    cy.on('doubleTap', (evt: any) => {
+      let event = getCytoscapeBaseEvent(evt);
```
josejulio: `const event`

src/components/CytoscapeGraph/CytoscapeGraph.tsx
josejulio: Also, I think this needs to be handled outside, in the [ServiceGraphPage](https://github.com/kiali/kiali-ui/blob/master/src/pages/ServiceGraph/ServiceGraphPage.tsx#L89)
If so, this would only need to do a `this.props.onDoubleClick(event);`

src/components/CytoscapeGraph/CytoscapeGraph.tsx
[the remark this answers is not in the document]
cfcosta: What do you mean by "reduce the scope"? I agree with you that the double click should not trigger the click event twice, will change that.

As for the delay, I agree, maybe 100ms would be better?

src/components/CytoscapeGraph/CytoscapeGraph.tsx
cfcosta: Yeah, it's not reduxified yet, so not sure how the placeholder code should be.

src/components/CytoscapeGraph/CytoscapeGraph.tsx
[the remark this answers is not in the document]
josejulio: Sorry, i meant to say that we should declare `tapTimeout` inside of `return(evt:any) => {` instead of the outer block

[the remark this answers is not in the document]
cfcosta: About declaring the tapTimeout inside the function: I checked the code, the problem doing it is that I need the reference to outlive the function, to catch the second click correctly.

And the delay problem, I think it will not happen, because it waits on its own thread, and if you click anywhere else it resets the function.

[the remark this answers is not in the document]
josejulio: > About declaring the tapTimeout inside the function: I checked the code, the problem doing it is that I need the reference to outlive the function, to catch the second click correctly.

right, missed that.

> And the delay problem, I think it will not happen, because it waits on its own thread, and if you click anywhere else it resets the function.

My concern is that the single click will lag for 350ms ( or 100ms if you change that), but i suppose we can test, thanks.

src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -161,10 +164,43 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
...
 
-    cy.on('tap', (evt: any) => {
-      const cytoscapeEvent = getCytoscapeBaseEvent(evt);
-      if (cytoscapeEvent) {
-        this.handleTap(cytoscapeEvent);
+    const injectDoubleTap = fun => {
+      let lastTap, currentTap, tapTimeout;
+
+      return (evt: any) => {
+        currentTap = evt.target;
```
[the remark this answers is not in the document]
cfcosta: I don't think this should be const, because the whole reference is kept on the injection mechanism and it's changed on every click.

cfcosta:
@josejulio can you review this again please?

cfcosta:
@jshaughn I saw this, I think that this can be merged now and if necessary, we can change it later, it's not a big change.

src/components/CytoscapeGraph/CytoscapeGraph.tsx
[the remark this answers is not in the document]
josejulio: yes, my point is that currentTap is only used on the current scope, we don't share it outside of this (arrow) function, i don't think is a big issue, but we should remove it from the outer declaration.
```typescript
    const injectDoubleTap = fun => {
      let lastTap, currentTap, tapTimeout;

      return (evt: any) => { // currentTap is only used in this scope, it doesn't change (in the scope).
        // unlinke lastTap and tapTimeout that are used outside
        currentTap = evt.target;

        tapTimeout = setTimeout(() => (lastTap = undefined), 350);

        if (lastTap && currentTap === lastTap) {
          clearTimeout(tapTimeout);
          lastTap = undefined;

          currentTap.trigger('doubleTap', evt);
        } else {
          lastTap = currentTap;

          fun(evt);
        }
      };
    };
```

jshaughn:
@cfcosta fine by me if you want to merge, I was just making a comment.  When @josejulio is happy you can proceed.  @serenamarie125 , just tagging you so that you are aware of this new behavior.

serenamarie125:
@jshaughn double clicking will bring you where ? to the service graph with that namespace in context?

jshaughn:
> @serenamarie125 double clicking will bring you where ? to the service graph with that namespace in context?

Yes, I believe that is the intent of this PR,  @cfcosta can confirm.  I think the general idea is this...  When we display a graph for NS1 you can have edge nodes outside of NS1.  Services that make requests to, or are requested by NS1 services are also there.  So, the idea is to be able to "jump" from the edge service, say in NS2, to the NS2 graph as if you had changed the namespace dropdown.

cfcosta:
@jshaughn @josejulio updated to istio 0.8 (and #422 ).

src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -176,10 +179,43 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
       }
     };
 
-    cy.on('tap', (evt: any) => {
+    const injectDoubleTap = fun => {
```
mtho11: `fun` is not the most intuitive variable. Is this 'fun'? no, it's not **fun**, but what does it represent?

mtho11: It is probably a function of some sort -- but what does it do?

cfcosta: Something like `originalCallback`, `tapEventCallback` or renaming the function to `injectDoubleTapToClickEvent` would make it better?

src/components/CytoscapeGraph/CytoscapeGraph.tsx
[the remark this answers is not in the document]
cfcosta: Fixed.

serenamarie125:
I think this is fine for now, but we should consider looking at this again in the future.  Users will need to discover this capability, and I'm not quite sure they will know where there are navigated to immediately.  Not blocking this merge, but think this should be looked at as part of an overall navigation improvement story.

jshaughn:
@serenamarie125 Going forward I think we may have multiple options related to a service.  Here we have this 'jump to namespace'.  But we are also looking at a drill-down such that the graph focuses on the service, showing only the direct requesting and requested services, so in addition to a dbl-click default we may need some sort of ability to select an option, maybe right-click menu?

src/components/CytoscapeGraph/CytoscapeGraph.tsx
josejulio: fun is done :(
`tapEventCallback` sounds good to me

serenamarie125:
@jshaughn i think having a double click act as a shortcut to something on a context menu would make sense.  The question will be is it a right click menu ( doesn't work on touch devices ) or accessed another way.  It's going to take some thinking, since single click already changes context of the properties panel!

serenamarie125:
@cfcosta how do we show the change in context upon the double click?  Just that the namespace dropdown will be updated ?

cfcosta:
@serenamarie125 it navigates to there, so there’s a full page refresh (same when changing namespaces currently). At some point we’re going to reduxify the namespace selection and then it will probably invoke a spinner or something like that.

src/components/CytoscapeGraph/CytoscapeGraph.tsx
cfcosta: Changed.

PR #440, opened 2018-06-14
KIALI-915 UI must now opt-in for istio traffic

jshaughn:
This is pending merge of the [server PR](https://github.com/kiali/kiali/pull/287).  

With both PRs in place we won't see istio infra services unless looking specifically at the istio-system namespace, which will not show all services for all namespaces, because all services interact with the istio infra services.


jshaughn:
@jmazzitelli Please merge as the backend support is already in.

PR #449, opened 2018-06-18
Kiali 924 replace latency with response time

jshaughn:


jshaughn:
@josejulio  This is set to DNM until the server-side work is merged.  The label can be removed as soon as the merge is done.   See https://github.com/kiali/kiali/pull/296.

PR #453, opened 2018-06-19
KIALI-971 Graph animations follows the curve edges

josejulio:
Implements Bezier paths to allow the balls follow the curves.

There is only an issue, we are using "internal" data from cytoscape, due to https://github.com/cytoscape/cytoscape.js/issues/2139 once that is merged and released i can update the code to use public API.

animation-curves-2


jmazzitelli:
This is working and is very nice. @josejulio write a JIRA for that followup work "Update cy and get animation to use the new public stuff rather than the internal workarond"

PR #455, opened 2018-06-20
KIALI-840 Hide sidebar on small windows

cfcosta:
This is the naivest implementation possible, and might be a good idea to
review or find some other implementation that's smarter.

This aims to fix the following problem: when you resize the window to a
smaller size, the sidebar goes over the graph, so we hide it and
add it back when the windows is bigger.

Before:

After:

There are some caveats that I could find:

* Graph dimensions are not responsive, so even removing the sidebar does
not fixes the size of the graph to fill the whole available screen
space. That causes some weird bugs, such as this one:

Ideas on how to fix this are welcome.

### DIFF
--- a/src/pages/ServiceGraph/SummaryPanel.tsx
+++ b/src/pages/ServiceGraph/SummaryPanel.tsx
@@ -5,55 +5,68 @@ import SummaryPanelGraph from './SummaryPanelGraph';
 import SummaryPanelGroup from './SummaryPanelGroup';
 import SummaryPanelNode from './SummaryPanelNode';
 
-type SummaryPanelState = {
-  // stateless
-};
+export default class SummaryPanel extends React.Component<SummaryPanelPropType, {}> {
+  updateSummary = () => {
+    this.forceUpdate();
+  };
 
-export default class SummaryPanel extends React.Component<SummaryPanelPropType, SummaryPanelState> {
+  componentDidMount() {
+    window.addEventListener('resize', this.updateSummary);
+  }
+
+  componentWillUnmount() {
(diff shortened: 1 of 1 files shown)


pilhuhn:
I think generally this is a good idea.
Can we somehow add an indicator that shows that there would be a sidebar? And potentially make it clickable so that the sidebar pops out and overlays the graph (background of the sidebar must not be opaque in that case)

jshaughn:
+1, a collapsible panel may be perfect here.  I suggest @serenamarie125 weigh in on this topic.

serenamarie125:
I think this needs more thought ... for responsive needs in general.  For example

- Vertical nav should be auto hiding
- Discuss responsive behavior of the toolbar (this is the old design, so don't want to waste time here)
- I would think for tablet size, you should be able to show the properties panel as long as the vertical nav is hidden, but the user should be able to dismiss it as well
- I would think for phone size, you should be able to show the properties panel when clicking on an item, and allow dismissing when clicking the "X"

cfcosta:
@serenamarie125 I created KIALI-1042 to cover the other feedback on responsive navigation, so that we can keep track on all changes and make this simpler.

So what I'm thinking is: I'm going to add the indicator to force showing/hiding the sidebar (probably on a checkbox on the filter listing), and keep the automatic behaviour as default when no option is chosen.

josejulio:
Before doing this, we need to allow the graph to horizontally resize, so when we hide the sidebar it shows the graph.
I'm working on that in the scope of KIALI-1082 and after that I will implement the hiding the sidebar via the filter menu (i.e. an element to hide the sidebar)

Closing this.

PR #456, opened 2018-06-20
KIALI-911 optimize graph health queries having one query per namespace

jotak:
(side change: enforce type on api authentication param)

### DIFF
--- a/src/components/CytoscapeGraph/CytoscapeGraph.tsx
+++ b/src/components/CytoscapeGraph/CytoscapeGraph.tsx
@@ -16,6 +16,7 @@ import { GraphParamsType } from '../../types/Graph';
 import { EdgeLabelMode } from '../../types/GraphFilter';
 import { authentication } from '../../utils/Authentication';
 import * as H from '../../utils/Health';
+import { NamespaceHealth } from '../../types/Health';
 
 type CytoscapeGraphType = {
   elements?: any;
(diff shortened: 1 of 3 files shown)


src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -333,15 +334,25 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
       return;
     }
     const duration = this.props.graphDuration.value;
+    const healthPerNamespace = new Map<String, Promise<NamespaceHealth>>();
     // Asynchronously fetch health
     cy.nodes().forEach(ele => {
       const fqService = ele.data('service');
       if (fqService && (ele.data('isGroup') || !ele.data('parent'))) {
         const serviceParts = fqService.split('.');
+        if (serviceParts.length < 2) {
```
lucasponce: It is enough to filter by length ?
is there any other special node besides "unknown" one ?

src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -333,15 +334,25 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
...
+          // Ignore health for special nodes such as "unknown"
+          return;
+        }
         const service = serviceParts[0];
         const namespace = serviceParts[1];
-        API.getServiceHealth(authentication(), namespace, service, duration).then(r => {
-          const health = r.data;
+        let promise = healthPerNamespace.get(namespace);
```
lucasponce: I see in the code that promises are stored (for caching?), I guess to not repeat same query for other service in the same namespace.
If so, I would add a comment line, just to help others to fix it if something is wrong.

src/components/CytoscapeGraph/CytoscapeGraph.tsx
jotak: I think "unknown" is the only one at this time, but since I'm extracting `serviceParts[0]` and `serviceParts[1]` below, this is a way to filter out "unknown" and run a defensive check at the same time.

src/components/CytoscapeGraph/CytoscapeGraph.tsx
jotak: Comment added (next to the map declaration)

PR #458, opened 2018-06-20
KIALI-988 Query the API with the right namespace for the unknown node

israel-hdez:
The unknown node cannot be broken down to {service name, namespace}
data. But the summary panel received the namespace in its properties.
So, if the namespace is not available in the cytoscape node, use the
namespace provided in the properties of the component.

I'm also adding a small fix to the "No traffic legend" that I saw while testing.

### DIFF
--- a/src/pages/ServiceGraph/SummaryPanelNode.tsx
+++ b/src/pages/ServiceGraph/SummaryPanelNode.tsx
@@ -65,7 +65,7 @@ export default class SummaryPanelNode extends React.Component<SummaryPanelPropTy
   }
 
   fetchRequestCountMetrics(props: SummaryPanelPropType) {
-    const namespace = props.data.summaryTarget.data('service').split('.')[1];
+    const namespace = props.data.summaryTarget.data('namespace');
     const service = props.data.summaryTarget.data('service').split('.')[0];
     const version = props.data.summaryTarget.data('version');
     const options: MetricsOptions = {
--- a/src/components/SummaryPanel/RpsChart.tsx
+++ b/src/components/SummaryPanel/RpsChart.tsx
@@ -46,7 +46,12 @@ export default class RpsChart extends React.Component<RpsChartTypeProp, {}> {
   }
 
   private thereIsTrafficData = () => {
-    return this.props.dataRps.length > 0 && this.props.dataRps[0].length > 1;
+    const dataRps: any = this.props.dataRps;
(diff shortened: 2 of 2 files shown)


src/pages/ServiceGraph/SummaryPanelNode.tsx
```diff
@@ -65,7 +65,7 @@ export default class SummaryPanelNode extends React.Component<SummaryPanelPropTy
   }
 
   fetchRequestCountMetrics(props: SummaryPanelPropType) {
-    const namespace = props.data.summaryTarget.data('service').split('.')[1];
+    const namespace = props.data.summaryTarget.data('service').split('.')[1] || this.props.namespace;
```
jshaughn: Not to be a pain, but this doesn't make sense to me.  'Unknown' has nothing to do with the overall namespace.  I think for the 'unknown' service we should have a predictable namespace where the code requires one, either empty or 'unknown'.

cfcosta:
#422 does that.

src/pages/ServiceGraph/SummaryPanelNode.tsx
israel-hdez: The problem is that in order to retrieve the right metrics from the backend, the namespace being shown needs to be passed to the backend.

But you made me think twice about the unknown node... Probably this node is so "special" that perhaps the right approach to solve this is to remove all stats from the side panel (or just don't show the side panel).

jshaughn: The 'unknown' "service" is certainly a pain.  As @cfcosta mentions below, once his PR goes in (which is imminent), the node element will supply a namespace field, and for service 'unknown' the namespace will be set to 'unknown'.  So, I think that should be the namespace name convention used.   I do think the side panel should be shown, for consistency, but it may need some special handling.  Incoming traffic is n/a.  The outgoing traffic could still be useful to see.  If we need special handling server-side, to grab metrics, then we may have to do that, but 'all' should never be in play.

israel-hdez: Ok, now that kiali/kiali#252 is merged, I can use the namespace provided by the node.
The only drawback is that without knowing the "graph namespace", the sparkline will show the overall (or cross-namespace) output traffic for the unknown node. I hope this is OK.

PR #459, opened 2018-06-20
KIALI-986 Long edges no longer make the animation faster than short edges

josejulio:
Notice the long edges and loops

Before:
animation-curves-2

After:
length-of-edge


jmazzitelli:
I think this looks good, and makes it more clear which are the slower channels

PR #460, opened 2018-06-21
[Minor] replace 'new Map' with '{}' to denote empty object

jotak:


jotak:
yeah not only we're not using the map API, but the labels object is not a map anymore so it was not consistent anymore to write `|| new Map()`

PR #461, opened 2018-06-22
KIALI-987 Add ServiceEntry support into Istio Config in the UI

lucasponce:
UI counterpart for https://github.com/kiali/kiali/pull/307


lucasponce:
Ops, there is a missing test without update.
I will fix it.

PR #463, opened 2018-06-22
KIALI-1030 Add a refresh button for ServiceList & IstioConfig

jotak:
Remove also binds in favor of fat arrows
And remove "loading" flag as it is not necessary anymore since the loader was moved to topbar


lucasponce:
Can you add a capture just to check it's same style as graph page ? Thx !

jotak:
It's the same as in metrics page, ie. without the auto-refresh options:

screenshot

PR #464, opened 2018-06-22
KIALI-1029 Fix marker on AceEditor when checking host

lucasponce:
A fix to mark correctly host/hosts field on VS and DR.

### DIFF
--- a/src/types/AceValidations.ts
+++ b/src/types/AceValidations.ts
@@ -103,11 +103,11 @@ const parseCheck = (yaml: string, check: ObjectCheck): AceCheck => {
     Potential paths:
       - <empty, no path>
       - spec/destination
-      - spec/name (this is used for destination Rules which maps spec/name into destinationName)
+      - spec/hosts
+      - spec/host
       - spec/precedence/<value>
       - spec/route[<nWeight>]/weight/<value>
       - spec/route[<nLabel>]/labels
-      - spec/hosts
 
     For this version we are going to mark the first element block.
     In future iterations we can add more granularity and try to mark just the offending word.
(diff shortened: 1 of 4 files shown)


src/types/AceValidations.ts
```diff
@@ -116,9 +116,10 @@ const parseCheck = (yaml: string, check: ObjectCheck): AceCheck => {
     let path = check.path;
     if (path.startsWith('spec/destination')) {
       aceMarker = parseMarker(yaml, 'destination:');
-    } else if (path.startsWith('spec/name')) {
-      // This usecase comes from Destination Rule, the spec/name in the original yaml object is brought into kiali as destinationName field
-      aceMarker = parseMarker(yaml, 'destinationName:');
+    } else if (path.startsWith('spec/hosts')) {
```
abonas: this is just a suggestion and not blocking the PR, but perhaps it would be worth to generalize this part of code? For instance, change parseMarker to receive generically a part of string without hardcoding it? At the moment there is a lot of very similar code (and might get more similar conditions in the future)

src/types/__tests__/AceValidations.test.ts
```diff
@@ -186,7 +186,7 @@ describe('#parseAceValidations in DestinationRule', () => {
       expect(annotation.column).toBe(0);
       expect(annotation.row).toBe(3);
       expect(annotation.type).toBe('error');
-      expect(annotation.text).toBe('Name doesnt have a valid service');
+      expect(annotation.text).toBe('Host doesnt have a valid service');
```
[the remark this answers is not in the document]
lucasponce: Yes. There was some prettier rule that failed using double qoute vs single quote.
```
-          message: 'Destination doesn\'t have a valid service',
+          message: "Destination doesn't have a valid service",
```
That should be fixed now.

src/types/AceValidations.ts
lucasponce: This function should calculate the position of the line from a backend description.
There are some cases that could be unified, but I decided to not do it for readability as there are others that need special handling.

PR #465, opened 2018-06-25
KIALI-1000 Use blue border for selected nodes despite health

jotak:


abonas:
@jotak a screenshot would be very beneficial here for "before/after" the change.
As well as @serenamarie125 input ;)

serenamarie125:
@jotak having a description and associated screenshots will help.  I think that you are saying that this PR just backs out reflecting health via node color.  If that's the case 👍

jotak:
Hmm not sure we're on the same page. The purpose of this PR is actually to NOT show the health color when a node is selected (of course, unselected nodes are still colored). See the JIRA ticket & discussions about it (since there were already some discussions there, if we want to revise the purpose of this PR then I'd suggest to do it on JIRA: https://issues.jboss.org/browse/KIALI-1000 

Screenshots:
**details v1 node selected (healthy)**
details v1 node selected (healthy)

**productpage v1 node selected (degraded)**
productpage v1 node selected (degraded)
_As you can see, the selection color prevails over the health color. Health is still displayed on summary panel_

**Edge in error selected**
Edge in error selected
_As you can see, the same happens when we select an edge in error. This is an existing behaviour, not something impacted by this PR_

jmazzitelli:
I'm +1 with this. However, I'm not convinced this is how it should ultimately be. I just don't know a good way to show health but still show what is selected. To me, it should be obvious what is selected (that's why I created the JIRA - it wasn't obvious what is selected) - I just don't know how to do that effectively :)

jotak:
@jmazzitelli yes I agree, too. This would not be the perfect solution, but perhaps a temporary step to align behaviour with how it works with edges, so we have some consistency. But either health or selection should have another distinctive way to be displayed that doesn't overlap with the other (could be line style (dotted/dashed/...)

PR #466, opened 2018-06-25
KIALI-1021 ServiceList: fetch health async

jotak:
- Split list item descriptions into a dedicated component, which receives health as a promise
- On service update, fetch health async and update promise
- Remove ServiceListOptions, not needed anymore

### DIFF
--- a/src/pages/ServiceList/ServiceListComponent.tsx
+++ b/src/pages/ServiceList/ServiceListComponent.tsx
@@ -14,18 +14,17 @@ import { Link } from 'react-router-dom';
 import PropTypes from 'prop-types';
 
 import { NamespaceFilter, NamespaceFilterSelected } from '../../components/NamespaceFilter/NamespaceFilter';
-import { ActiveFilter, FilterType } from '../../types/NamespaceFilter';
+import { PfColors } from '../../components/Pf/PfColors';
 import * as API from '../../services/Api';
 import Namespace from '../../types/Namespace';
+import { ActiveFilter, FilterType } from '../../types/NamespaceFilter';
 import { Pagination } from '../../types/Pagination';
-import { IstioLogo, ServiceItem, ServiceList, SortField, overviewToItem } from '../../types/ServiceListComponent';
-import { getRequestErrorsRatio } from '../../utils/Health';
-import { HealthIndicator, DisplayMode } from '../../components/ServiceHealth/HealthIndicator';
-
-import ServiceErrorRate from './ServiceErrorRate';
-import RateIntervalToolbarItem from './RateIntervalToolbarItem';
-import { PfColors } from '../../components/Pf/PfColors';
+import { IstioLogo, ServiceItem, ServiceOverview, SortField, overviewToItem } from '../../types/ServiceListComponent';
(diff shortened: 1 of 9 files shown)


src/pages/ServiceList/ServiceListComponent.tsx
```diff
@@ -63,9 +62,9 @@ export const sortFields: SortField[] = [
     title: 'Error Rate',
     isNumeric: true,
     compare: (a: ServiceItem, b: ServiceItem) => {
-      const ratioA = getRequestErrorsRatio(a.health.requests);
-      const ratioB = getRequestErrorsRatio(b.health.requests);
-      return ratioA === ratioB ? a.name.localeCompare(b.name) : ratioA.value - ratioB.value;
+      const ratioA = getRequestErrorsRatio(a.health ? a.health.requests : undefined).value;
```
lucasponce: Isn't more clear to deal directly with 0/-1 here, instead to delegate into the getRequestErrorRation() to resolve the "undefined" value ?
I saw that logic hard to review in a first attemp.

src/types/ServiceListComponent.ts
```diff
@@ -14,15 +13,23 @@ export interface ServiceList {
...
+): ServiceItem => {
+  const item: ServiceItem = {
     name: overview.name,
-    health: overview.health,
     istioSidecar: overview.istioSidecar,
-    namespace: namespace
+    namespace: namespace,
+    healthPromise: healthPromise
   };
+  healthPromise.then(h => (item.health = h));
```
lucasponce: Missing catch ? if this promise fails, where is the error msg handled ?

src/pages/ServiceList/ServiceListComponent.tsx
jotak: +1

src/types/ServiceListComponent.ts
jotak: No, this is just an intermediate operation on the promise, it's not the "terminal" operation. The promise is consumed at the component level, and that's where the error is caught.

src/pages/ServiceList/ServiceListComponent.tsx
```diff
@@ -61,17 +64,37 @@ export const sortFields: SortField[] = [
...
+  services: ServiceItem[],
+  sortField: SortField,
+  isAscending: boolean
+): Promise<ServiceItem[]> => {
+  if (sortField.title === 'Error Rate') {
+    // In the case of error rate sorting, we may not have all health promises ready yet
+    // So we need to get them all before actually sorting
+    const allHealthPromises: Promise<ServiceItemHealth>[] = services.map(item => {
+      return item.healthPromise.then(health => {
+        const withHealth: any = item;
```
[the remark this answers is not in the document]
jotak: Yeah but if I keep it as `ServiceItem` then I cannot add `.health` field in the line below, so I have this little trick to make it easier.
A more "correct" way would have been to declare a new `ServiceItemHealth` object a copy fields one by one from the `ServiceItem`, but... it was little a bit too much for what I want :)

PR #467, opened 2018-06-26
KIALI-1036 Add missing refresh buttons

jotak:
- In Istio rule info
rule

- In Istio config details
istio

- In service details
service

### DIFF
--- a/src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
+++ b/src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { Col, Row } from 'patternfly-react';
+import { Button, Col, Icon, Row } from 'patternfly-react';
 import { Link, RouteComponentProps } from 'react-router-dom';
 import { NamespaceFilterSelected } from '../../components/NamespaceFilter/NamespaceFilter';
 import { ActiveFilter } from '../../types/NamespaceFilter';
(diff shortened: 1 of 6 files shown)


src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
```diff
@@ -51,7 +51,11 @@ class IstioConfigDetailsPage extends React.Component<RouteComponentProps<IstioCo
 
   updateTypeFilter = () => this.updateFilters(true);
 
-  fetchIstioObjectDetails = (props: IstioConfigId) => {
+  fetchIstioObjectDetails = () => {
+    this.fetchIstioObjectDetailsFromProps(this.props.match.params);
```
abonas: is match.params always non empty and won't "crash"?

abonas:
isn't the icon named fa-refresh? (I might be totally wrong here :))
https://www.patternfly.org/styles/icons/
regardless, cc @serenamarie125  - fyi on the refresh button addition. any comments?

src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
jotak: It shouldn't, since these props are not marked optional (thanks typescript for helping here!)

jotak:
@abonas fa-refresh is the one that is used indeed, the "Icon" component prefixes name with "fa-" by default.

aljesusg:
@jotak is right @abonas, by default Icon component set type to fa if you wanna use other type you must set it in type prop.

serenamarie125:
As part of the new design shown in https://github.com/kiali/kiali-design/issues/4 , I've suggested to become more aligned with the direction that OpenShift 4.0 will be taking, with a button bar on the bottom including a Reload and Close button: 
image

abonas:
@serenamarie125 I think there are several use cases involved, and I have a feeling we are not talking about them all here :)
Since the refresh button is present on several different screens in Kiali, and not only in yaml view, where should the Reload button be located then in the graph/service details page and others? And what is the expected behavior of pressing the Close button when:
1. standing in the yaml tab - should it navigate the user to overview tab?
2. standing on the graph main page - close Kiali UI?
3. standing on the service details page

PR #468, opened 2018-06-26
KIALI-1051 Fix health unsynchronized issue in summary panel

jotak:
Using a promise in panel's data makes sure the panel can get the latest up-to-date value for health

### DIFF
--- a/src/pages/ServiceGraph/SummaryPanelGroup.tsx
+++ b/src/pages/ServiceGraph/SummaryPanelGroup.tsx
@@ -13,6 +13,7 @@ import { Link } from 'react-router-dom';
 import { shouldRefreshData } from './SummaryPanelCommon';
 import { HealthIndicator, DisplayMode } from '../../components/ServiceHealth/HealthIndicator';
 import Label from '../../components/Label/Label';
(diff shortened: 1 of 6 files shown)


src/pages/ServiceGraph/SummaryPanelGroup.tsx
```diff
@@ -49,37 +51,45 @@ export default class SummaryPanelGroup extends React.Component<SummaryPanelPropT
...
 
   componentWillUnmount() {
     this._isMounted = false;
   }
 
+  updateHealth() {
+    const healthPromise = this.props.data.summaryTarget.data('healthPromise');
+    if (healthPromise) {
+      healthPromise.then(h => this.setState({ health: h })).catch(err => this.setState({ health: undefined }));
+    }
```
israel-hdez: Perhaps you should do `this.setState({ health: null }))` if there is no `healthPromise`.

jotak:
oh, good catch @israel-hdez thanks

jotak:
@israel-hdez issue fixed. Also added some fixed-width blank to prevent glitch while reloading

PR #471, opened 2018-06-27
KIALI-1056 Workaround for react-ace issue: force annotations refresh

jotak:
See https://github.com/securingsincity/react-ace/issues/300


lucasponce:
Good catch.

PR #474, opened 2018-06-28
KIALI-1015 Add validations to Virtual Services

xeviknal:
The main goal of this PR is move all the visuals from RouteRules tab to Virtual Services. Here the result of this PR changes:

virtual-service-ui-migration-2

This PR needs kiali/kiali#319 to be merged.

### DIFF
--- a/src/components/Details/DetailObject.tsx
+++ b/src/components/Details/DetailObject.tsx
@@ -1,11 +1,18 @@
 import * as React from 'react';
 import Label from '../Label/Label';
+import { Icon } from 'patternfly-react';
 
 interface DetailObjectProps {
   name: string;
   detail: any;
   labels?: string[];
   exclude?: string[];
+  validation?: Validation;
+}
+
+interface Validation {
+  message: string;
+  icon: string;
 }
 
 class DetailObject extends React.Component<DetailObjectProps> {
@@ -37,10 +44,12 @@ class DetailObject extends React.Component<DetailObjectProps> {
   }
 
   canDisplay(name: string) {
-    return this.props.exclude && !this.props.exclude.includes(name);
+    return this.props.exclude == null || !this.props.exclude.includes(name);
   }
 
-  buildList(name: string, value: any, isLabel: boolean): any {
+  // buildList returns a recursive list of all items within value. It shows a validation
+  // only for the first iteration (when depth is 0)
+  buildList(name: string, value: any, isLabel: boolean, depth: number): any {
     if (!this.canDisplay(name)) {
       return '';
     }
@@ -70,14 +79,14 @@ class DetailObject extends React.Component<DetailObjectProps> {
           childrenList.push(<li key={listKey + '_i' + i}>{v}</li>);
         } else {
           Object.keys(v).forEach((key, j) => {
-            let childList = this.buildList(key, v[key], checkLabel);
+            let childList = this.buildList(key, v[key], checkLabel, depth + 1);
             childrenList.push(<li key={listKey + '_i' + i + '_j' + j}>{childList}</li>);
           });
         }
       });
     } else {
       Object.keys(value).forEach((key, k) => {
-        let childList = this.buildList(key, value[key], checkLabel);
+        let childList = this.buildList(key, value[key], checkLabel, depth + 1);
         childrenList.push(<li key={listKey + '_k' + k}>{childList}</li>);
       });
     }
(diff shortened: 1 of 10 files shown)


xeviknal:
@kiali/ux-reviewers I am wondering if you could check this screen out. This should be similar to the Route Rule tab already discussed and implemented before:

route-rules-ui-migration

If you need some more help, let me know and Ill do my best.

src/components/Details/DetailObject.tsx
```diff
@@ -1,10 +1,13 @@
 import * as React from 'react';
 import Label from '../Label/Label';
+import { Icon } from 'patternfly-react';
 
 interface DetailObjectProps {
   name: string;
   detail: any;
   labels?: string[];
+  exclude?: string[];
+  validation?: any;
```
lucasponce: Perhaps validation should have a type here.
I see there is at least a .icon and a .message, so, I guess it would be nice to have a strong type here if possible.

src/components/Details/DetailObject.tsx
```diff
@@ -35,7 +38,15 @@ class DetailObject extends React.Component<DetailObjectProps> {
     return this.props.labels.indexOf(name) > -1;
   }
 
-  buildList(name: string, value: any, isLabel: boolean): any {
+  canDisplay(name: string) {
+    return this.props.exclude == null || !this.props.exclude.includes(name);
+  }
+
+  buildList(name: string, value: any, isLabel: boolean, depth: number): any {
```
lucasponce: Perhaps a comment here or at the top of the class just indicating that validations are rendered at first level can help to understand the new "depth" parameter.

src/components/Details/__tests__/DetailObject.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+      port: {
+        number: 22,
+        name: 'ssh'
+      }
+    },
+    weight: 85
+  };
+
+  const mockRandom = () => {
+    const mockMath = Object.create(global.Math);
```
lucasponce: Why is it necessary to mock the random ?
That means that there is some logic inside the code invoking the random() function ?

src/pages/ServiceDetails/ServiceInfo/ServiceInfoRouteRules/RouteRuleRoute.tsx
```diff
@@ -2,7 +2,7 @@ import * as React from 'react';
 import * as resolve from 'table-resolver';
 import {
   checkForPath,
-  DestinationWeight,
+  DestinationWeightV1Alpha1,
```
lucasponce: +1
In my following PR will remove the old RouteRule object, so this rename will be temporal.

abonas:
Why the word "Load" is used? is this the recommended term here?

src/pages/ServiceDetails/ServiceInfo/ServiceInfoVirtualServices/VirtualServiceRoute.tsx
```diff
@@ -0,0 +1,352 @@
...
+        <li>Host: {destination.host || '-'} </li>
+        <li>Subset: {destination.subset || '-'} </li>
+        <li>Port: {destination.port ? destination.port.number : '-'} </li>
+      </ul>
+    );
+  }
+
+  renderTable(route: any, i: number) {
+    const resolvedColumns = resolve.columnChildren(this.columns());
+    const resolvedRows = resolve.resolve({
```
lucasponce: I'm not familiar with the table-resolver library.
I assume this is just a migration from previous routerule table.

lucasponce:
> Why the word "Load" is used? is this the recommended term here?

I think this comes from the UXD design discussion.
In any case, it is the word using for the current old RouteRule objects, for this PR is just a migration from the same functionality and UI.

abonas:
> I think this comes from the UXD design discussion.
> In any case, it is the word using for the current old RouteRule objects, for this PR is just a migration from the same functionality and UI.

np

src/components/Details/__tests__/DetailObject.test.tsx
xeviknal: Yep, it looks like DetailObject is using a random method:
https://github.com/xeviknal/kiali-ui/blob/4fb935a6d1002e74272efb2a737c6dadfaec65f0/src/components/Details/DetailObject.tsx#L21

Without that mocking test will fail every build.

lucasponce: Oh, true, I didn't remember this logic, yes, it makes sense :-).

src/pages/ServiceDetails/ServiceInfo/ServiceInfoVirtualServices/VirtualServiceRoute.tsx
xeviknal: yep, this is needed for migration. Table-resolver comes from reactabular.js the lib for building tables in PF. It is used in rourerule table too.

PR #476, opened 2018-06-28
KIALI-934 Add a control to allow zooming with buttons

israel-hdez:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "cytoscape-cose-bilkent": "4.0.0",
     "cytoscape-dagre": "2.2.1",
     "cytoscape-klay": "3.1.1",
+    "cytoscape-panzoom": "2.5.2",
     "cytoscape-popper": "1.0.1",
     "deep-freeze": "0.0.1",
     "js-yaml": "3.12.0",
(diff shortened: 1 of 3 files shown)


israel-hdez:
@mtho11 I found this _panzoom_ extension. I hope this works to cover KIALI-934. I find it much better than adding more buttons to the toolbar.

mtho11:
@kiali/ux-reviewers  What do you guys think of this panzoom control instead of +/- buttons on the toolbar?

jmazzitelli:
I like it - and it matches what a lot of people are already used to because this kind of control is found in many other apps.

mtho11:
@israel-hdez can you add a video showing how it's used, so it's clear to others how it works.

lucasponce:
+1 If you don't have a mouse with wheel, it's not obvious how to zoom in/out.
So, this component is useful (or a similar feature).

mtho11:
I suppose one downside is the screen real-estate that it occupies.

lucasponce:
> I suppose one downside is the screen real-estate that it occupies.

Perhaps it can be show on/off ?

israel-hdez:
@mtho11  Here is a gif demoing how it works:
panzoomcontrol

@lucasponce @mtho11 Yes it takes some space, but it's an overlay. I don't see necessary to be able to turn it on/off. But you say.

mtho11:
I like the fact that we still have normal panning on the screen

cshinn:
@israel-hdez I'm looking this over with some other UXDers and I'll get back to you ASAP. Sorry about the delay, I've been out for the past several days.

mtho11:
@cshinn Just FYI: Thursday is our sprint end deadline

cshinn:
@israel-hdez @mtho11 @serenamarie125 This functionality is something that is coming up more frequently in our projects, so we should create a full-featured design that we can standardize across them. This tool has some good functionality and isn't too far from what we have in other places, so I'd say that it makes sense to use it for the time being. I'll create some stories to work on a more general pan/zoom pattern and we can update the visuals at a later date.

Moving the tool to the bottom of the canvas might keep it from covering as much important territory

PR #478, opened 2018-06-29
KIALI-1063 Turn off loading spinner on response error

josejulio:


aljesusg:
I think that spinner is running when an axios request is running so in response error the axios request is finished and spinner should stop, see sec/app/App.tsx from L12, so I think that we don't need this

https://github.com/kiali/kiali-ui/blob/6202fbf4ba5d114f9e409c23898a80cc8857ca57/src/app/App.tsx#L12

josejulio:
> Is it really working in case of several concurrent requests? I would say a counter that increments for each requests and decrements for each response should be the way to go (and it's loading when counter > 0)

That's other bug that i was going to file :)

jotak:
@aljesusg yeah, but wrote *concurrent* requests.
Start REQ1 => spinner spins
Start REQ2 => spinner spins
End REQ1 => spinner stops ... wait, what?
we need to count :) But that's another bug, yes.

josejulio:
> I think that spinner is running when an axios request is running so in response error the axios request is finished and spinner should stop, see sec/app/App.tsx from L12, so I think that we don't need this

Is not doing that on timeout, there is no response when a request times out, so it goes to `error` L36

aljesusg:
Oh I see ... And check if there is any axios pending in the response section and avoid counters?

josejulio:
> Oh I see ... And check if there is any axios pending in the response section and avoid counters?

We will need counters to know all requests finished, but going to do that in a separate PR

josejulio:
> Oh I see ... And check if there is any axios pending in the response section and avoid counters?

Unless we can check if there is any pending requests, can we do that?

mtho11:
You guys are all correct. Nice insights.
I believe the real solution to this is via _Observables_ (Rx.js). This allows us to combine multiple requests (streams) into whatever strategies we want.
Promises have their limitations.
Sorry for hijacking, obviously another sprint ;)

mtho11:
@jotak @aljesusg go to bed. It's Friday night 🛏

josejulio:
https://issues.jboss.org/browse/KIALI-1080

josejulio:
> @jotak @aljesusg go to bed. It's Friday night

Actually i think that is at least Saturday morning for them :)

jotak:
@josejulio now it is, but not on my first post. (and I'm in bed... counting the requests helps to sleep)

aljesusg:
Me too ^^

josejulio:
@jotak The related bug (count the requests) is tracked in [KIALI-1080](https://issues.jboss.org/browse/KIALI-1080), can you review this one and merge if is OK?

Thanks!

PR #482, opened 2018-07-02
KIALI-1025 Set isFetching to false if an error happens

josejulio:
I couldn't consistently reproduce this bug, it happened to me once, but I suspect the issue is that the namespace request fails [1] and `isFetching` remains `true`, blocking the next refresh [2]

[1] https://github.com/kiali/kiali-ui/blob/master/src/actions/NamespaceAction.ts#L34
[2] https://github.com/kiali/kiali-ui/blob/master/src/actions/NamespaceAction.ts#L42

### DIFF
--- a/src/actions/NamespaceAction.ts
+++ b/src/actions/NamespaceAction.ts
@@ -3,7 +3,8 @@ import * as Api from '../services/Api';
 export enum NamespaceActionKeys {
   NAMESPACE_RELOAD = 'NAMESPACE_RELOAD',
   NAMESPACE_START = 'NAMESPACE_START',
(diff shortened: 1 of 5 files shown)


josejulio:
@vnugent can you take a look?

josejulio:
> Can you add an Action and Reducer Tests for this.

Sure

josejulio:
@mtho11 done. 
Had to refactor a bit the actions/reducers to use the typesafe-actions and grouped the actions as in the other code

mtho11:
@josejulio now there is conflicts

josejulio:
> @josejulio now there is conflicts

I hate conflicts!

josejulio:
> @josejulio now there is conflicts

Rebased

PR #484, opened 2018-07-02
KIALI-1089 Clears selection when a relayout is needed (namespace, lay…

josejulio:
…out or elements change)

Removes the selection if changing from one namespace to other (or if a layout change is needed)


src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -280,9 +285,10 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
     // update the entire set of nodes and edges to keep the graph up-to-date
     cy.json({ elements: this.props.elements });
 
-    // update the layout if needed
+    // update the layout if needed and reset to default values
     if (this.updateLayout) {
       cy.layout(LayoutDictionary.getLayout(this.props.graphLayout)).run();
+      clearSelection();
```
jmazzitelli: We don't want to reset the selection just because we changed the layout. In fact, its a nice feature that I could select a node, and change the layout from breadthfirst to Dagre and still see the same node selected as before. There is no reason why we would need or want the selection to go away if we simply change the layout.

We only want to clear the selection when we change the NAMESPACE (which is what is happening when we double-click a pentagon node) because in that case the node that is selected is probably going away anyway.

josejulio:
Updated the code, had to opt-in for the "custom selection behavior" [1] to achieve this.

The reason I had to do this, is because the doubleTap "selected" the node, but it does in the step after changing the namespace, so i couldn't catch it to "unselect" in a clean way.

[1] https://github.com/cytoscape/cytoscape.js/issues/1145#issuecomment-153083828

josejulio:
Whoops, forgot to remove the DNM tag

jmazzitelli:
The selection now behaves as we want. However, I'm seeing odd behavior in some layouts SOME of the time. I do not know if its related to this PR. I think we need more investigation before merging this PR. Here are two examples of what I just saw with this PR:

screenshot from 2018-07-05 12-15-42
screenshot from 2018-07-05 12-16-18

josejulio:
Will check what's wrong

josejulio:
I couldn't reproduce the issue (yet) but saw something that might be related.

When doing a relayout (because layout or namespace change or new nodes are added) it seems that we were reusing the previous positions and the layout algorithm might consider that. At least it seemed that Dagre was using that, because when switching from namespace "all" to "bookinfo" rendered a different dagre graph.

I solved this by reseting all the positions to (0,0) before running the layout.

josejulio:
Thanks!

PR #485, opened 2018-07-02
KIALI-1090 Increase opacity on dim node/edges to make them easier to …

josejulio:
…read

Before:
kiali-1090-before

After (**updated 2018-07-04**):
kiali-1090-after-updated

Using pf-black-500:
kiali-1090-pfblack500

### DIFF
--- a/src/components/CytoscapeGraph/graphs/GraphStyles.ts
+++ b/src/components/CytoscapeGraph/graphs/GraphStyles.ts
@@ -186,7 +186,7 @@ export class GraphStyles {
       {
         selector: 'node.' + DimClass,
         style: {
-          opacity: '0.3'
+          opacity: '0.6'
         }
       },
       {


josejulio:
> I am OK with this. Before merging, I would like to see what other people have to say. For sure I didn't like the really dim opacity of 0.3 because I couldn't read the labels of dimmed nodes and edges (which is odd, because I didn't notice this before - I wonder if something else causes the opacity to be even lower due to a recent change?).

Not really sure, but in your screenshot, the stuff inside the group box seems harder to read than the service outside it.

josejulio:
> I am OK with this. Before merging, I would like to see what other people have to say. For sure I didn't like the really dim opacity of 0.3 because I couldn't read the labels of dimmed nodes and edges (which is odd, because I didn't notice this before - I wonder if something else causes the opacity to be even lower due to a recent change?).

cc: @kiali/ux-reviewers @pilhuhn @mwringe

mwringe:
In the second screen shot I can read the services (circles) better, but the connection lines look too similar to me. Can we get it so the lines are like in the first screenshot?

josejulio:
> Can we get it so the lines are like in the first screenshot?

yes!

josejulio:
Edge opacity is unchanged now, see updated screenshot.

mwringe:
> I wonder if something else causes the opacity to be even lower due to a recent change?

Do we know why there was a change? I wonder if this was a UXD recommendation that we are now undoing.

@kiali/ux-reviewers any input here?

josejulio:
It seems that @jmazzitelli lowered the opacity (by 0.1) when refactoring other stuff :)

Initial commit: https://github.com/josejulio/swsui/commit/555139fac34585a1a1adb42168f8e21619c1c97b#diff-195cf831935737255b1a9df5d8fb4b10R84

Refactoring commit:
https://github.com/josejulio/swsui/commit/8bcad0a70e606eeae53cc878821ddc95d20d4826#diff-195cf831935737255b1a9df5d8fb4b10L85

cshinn:
@josejulio, I agree that this is much easier to read than before, but I'm not sure that even this will be legible to someone who needs higher contrast. Perhaps excluding the font color from the change, or using something like pf-black-500 for the font if the original black is too stark.

josejulio:
> @josejulio, I agree that this is much easier to read than before, but I'm not sure that even this will be legible to someone who needs higher contrast. Perhaps excluding the font color from the change, or using something like pf-black-500 for the font if the original black is too stark.

Tried with pf-black-500 and seems harder to read IMO (see first comment):

I guess since there is an improvement we can merge this one and create a JIRA to try to further improve that.

mwringe:
> I guess since there is an improvement we can merge this one and create a JIRA to try to further improve that.

+1

Please make sure a link is added when creating the jira so that someone can follow the discussion.

jmazzitelli:
merging this - any additional changes we will provide in another PR(s)

PR #486, opened 2018-07-03
KIALI-504 Replace deprecated 'componentWillReceiveProps' 2/2

aljesusg:
Jira : [KIALI-504](https://issues.jboss.org/browse/KIALI-504)

Replace deprecated componentWillReceiveProps.

See Discussion in [comments](https://issues.jboss.org/browse/KIALI-504)

Documentation : [Link](https://reactjs.org/blog/2018/03/29/react-v-16-3.html)


src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
```diff
@@ -25,6 +25,47 @@ interface IstioConfigDetailsState {
...
+        };
+        return newState;
+      })
+      .catch(error => {
+        MessageCenter.add(API.getErrorMsg('Could not fetch IstioConfig details.', error));
+      });
+    return allPromise;
+  };
+
+  static getDerivedStateFromProps(props: RouteComponentProps<IstioConfigId>, currentState: IstioConfigDetailsState) {
```
jotak: Does that work? How is the caller of `getDerivedStateFromProps` able to get the asynchronous result?

Maybe here it's better to use `componentDidUpdate` instead of `getDerivedStateFromProps`, since we're fetching the API, no?

aljesusg: I don't know I am still testing if this work. I read that should work with asynchronous results, if not we should move to componentDidUpdate. This is why I did 2 PR for this tasks

jotak: Ok,
According to the doc it seems recommended to use componentDidUpdate in such case : https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops . It says getDerivedStateFromProps is sort of a last option if other options don't work

aljesusg: Yes, thanks @jotak

src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
```diff
@@ -25,6 +25,47 @@ interface IstioConfigDetailsState {
 class IstioConfigDetailsPage extends React.Component<RouteComponentProps<IstioConfigId>, IstioConfigDetailsState> {
   aceEditorRef: React.RefObject<AceEditor>;
 
+  static fetchIstioObjectDetailsFromProps = (props: IstioConfigId) => {
+    let promiseConfigDetails = API.getIstioConfigDetail(
```
mtho11: Looks like this `let` could be a `const`

src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
```diff
@@ -25,6 +25,47 @@ interface IstioConfigDetailsState {
...
+      authentication(),
+      props.namespace,
+      props.objectType,
+      props.object
+    );
+    let promiseConfigValidations = API.getIstioConfigValidations(
```
mtho11: Same `const` here too.

src/pages/ServiceList/ItemDescription.tsx
```diff
@@ -13,23 +13,44 @@ interface State {
...
-      this.onItemChanged(nextProps.item);
-    }
-  }
-
-  onItemChanged(item: ServiceItem) {
-    item.healthPromise.then(h => this.setState({ health: h })).catch(err => this.setState({ health: undefined }));
+    ItemDescription.onItemChanged(this.props.item)
+      .then((result: State) => {
+        // After promise update with this state
+        this.setState(result);
```
mtho11: Would be nice if this state was moved to Redux. A different PR obviously.

aljesusg:
After test the lifecycle...yes @jotak we need to move this to componentDidUpdate, getDerivedStateFromProps don't work . Thanks.

PR #495, opened 2018-07-05
KIALI-1103 Fix versions components in about and refactor

aljesusg:
Jira: [KIALI-1103](https://issues.jboss.org/browse/KIALI-1103)

- Moved about info to store-persist (Don't need to check everytime the versions)
- Removed authentication header in getStatus (Don't need it in /api and /api/status)
- Added comments to service/Api and refactor headers

### DIFF
--- a/src/services/Api.ts
+++ b/src/services/Api.ts
@@ -14,21 +14,32 @@ interface Response<T> {
   data: T;
 }
 
+/**  Headers Definitions */
+
 const loginHeaders = { 'X-Auth-Type-Kiali-UI': '1' };
-const authHeader = (auth: string) => {
-  return { Authorization: auth };
+
+const authHeader = (auth: string) => ({ Authorization: auth });
+
+/**  Helpers to Requests */
+
+const getHeaders = (auth?: string) => {
+  if (auth === undefined) {
+    return { ...loginHeaders };
+  }
+  return { ...loginHeaders, ...authHeader(auth) };
 };
+
 const basicAuth = (username: string, password: string) => {
   return { username: username, password: password };
 };
 
-let newRequest = <T>(method: string, url: string, queryParams: any, data: any, auth: string) => {
+let newRequest = <T>(method: string, url: string, queryParams: any, data: any, auth?: string) => {
   return new Promise<Response<T>>((resolve, reject) => {
     axios({
       method: method,
       url: url,
       data: data,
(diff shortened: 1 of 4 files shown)


src/services/Api.ts
```diff
@@ -14,21 +14,38 @@ interface Response<T> {
...
+
 const loginHeaders = { 'X-Auth-Type-Kiali-UI': '1' };
-const authHeader = (auth: string) => {
-  return { Authorization: auth };
+
+const authHeader = (auth: string) => ({ Authorization: auth });
+
+/**  Helpers to Requests */
+
+const getHeaders = (auth: string | undefined = undefined) => {
```
jotak: arguments can be simplified: `(auth?: string)` , i think it works the same

src/services/Api.ts
```diff
@@ -14,21 +14,38 @@ interface Response<T> {
...
   return { username: username, password: password };
 };
 
-let newRequest = <T>(method: string, url: string, queryParams: any, data: any, auth: string) => {
+let newRequest = <T>(
+  method: string,
+  url: string,
+  queryParams: any,
+  data: any,
+  auth: string | undefined = undefined
```
jotak: same here

src/services/Api.ts
aljesusg: Yes, you are right

aljesusg: Changed

jotak:
I'm not sure about moving the status to a store. Why wouldn't we want up-to-date information? Especially I see the "Kiali state" field that might be running or not, that could be a changing information that worth keeping up-to-date.

aljesusg:
This was in store, we don't have any tool to check if the server is running only in the login step.

aljesusg:
Maybe we can add in the masthead a status of Kiali server and refresh status every 5 seconds to show if Kiali is working. But the point of move this to store is avoid to request Everytime the versions of components when they won't change ^^

jotak:
@aljesusg hmm, on server-side I saw something about a "running" status but maybe this is something unused then. See https://github.com/kiali/kiali/blob/master/status/status.go#L72
It's a bit annoying to see dead code such as this.

But ok then, +1 to merge. It will still be time to change if we're adding more information in status in the future.

aljesusg:
Ok I'll open a JIRA to make the changes that we dicussed in the chat, about the dead code I don't know when change that ^^

PR #496, opened 2018-07-05
KIALI-533 Session Timeout

aljesusg:
Everytime user refresh the website this sessiontimeout is updated, when the user refresh and the actual time pass the last time + hours in config file  user will need login again.

### DIFF
--- a/src/actions/NamespaceAction.ts
+++ b/src/actions/NamespaceAction.ts
@@ -51,6 +51,9 @@ export const fetchNamespacesIfNeeded = () => {
   return (dispatch, getState) => {
     if (shouldFetchNamespaces(getState())) {
       // Dispatch a thunk from thunk!
+      if (getState()['authentication']['token'] === undefined) {
+        return Promise.resolve();
+      }
       const auth = 'Bearer ' + getState().authentication.token.token;
       return dispatch(asyncFetchNamespaces(auth));
     } else {
--- a/src/pages/Login/LoginPage.tsx
+++ b/src/pages/Login/LoginPage.tsx
@@ -35,6 +35,10 @@ export default class LoginPage extends React.Component<LoginProps, LoginState> {
     };
   }
 
+  componentDidMount() {
+    document.documentElement.className = 'login-pf';
+  }
+
   handleChange = (e: any) => {
     const { name, value } = e.target;
     this.setState({ [name]: value });
--- a/src/actions/LoginActions.ts
+++ b/src/actions/LoginActions.ts
@@ -3,6 +3,9 @@ import * as API from '../services/Api';
 import { Token } from '../store/Store';
 import { HTTP_CODES } from '../types/Common';
 import { HelpDropdownActions } from './HelpDropdownActions';
+import { config } from '../config';
+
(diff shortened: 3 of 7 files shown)


src/actions/NamespaceAction.ts
```diff
@@ -51,6 +51,9 @@ export const fetchNamespacesIfNeeded = () => {
   return (dispatch, getState) => {
     if (shouldFetchNamespaces(getState())) {
       // Dispatch a thunk from thunk!
```
lucasponce: What means this comment ?

josejulio: I think that comment should be next to the `return dispatch(asyncFetchNAmespaces(auth))`

This `NamespaceAction` was the first Redux action file introduced to our code.
That  comment is there to show that we can dispatch a `thunk` from within a `thunk`
https://github.com/reduxjs/redux-thunk#whats-a-thunk

src/pages/Login/LoginPage.tsx
```diff
@@ -35,6 +35,10 @@ export default class LoginPage extends React.Component<LoginProps, LoginState> {
     };
   }
 
+  componentDidMount() {
+    document.documentElement.className = 'login-pf';
```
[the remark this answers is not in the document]
aljesusg: Should be dinamically if we use login or not, should be "layout-pf layout-pf-fixed" or "login-pf"

See componentDidMount of component [Navigation L99-L106](https://github.com/kiali/kiali-ui/blob/master/src/components/Nav/Navigation.tsx#L99-L106)

```typescript
  componentDidMount() {
    // Change layout-pf layout-pf-fixed by
    if (this.props.authenticated) {
      document.documentElement.className = 'layout-pf layout-pf-fixed';
    } else {
      document.documentElement.className = 'login-pf';
    }
}
```

abonas:
if the timeout is configurable, what's the default value and where can Kiali admin configure it? do we need something in the readme about it?

aljesusg:
Default value is 3 hours without refresh the browser, I'll do another PR for the Readme...we should add the configuration file details in the Readme you are right @abonas . Thanks

mtho11:
3 hours seems a bit high for a session timeout. Isn't 30 minutes more reasonable? Ideally, we would have a dialog box come up 1 minute before timeout saying "Your session will timeout in 1 minute. Would you like to extend your session for another 30 minutes?"

aljesusg:
Ok, I'll change it to 30' in the Readme PR, about the dialog box me should make a check every x seconds to control  it, actually when you refresh the browser the sessiontimeout rid updated to another 3 hours in the checkCredentials action.

mtho11:
The only reason I say anything about it is that it is kind of a security thing. When people walk away from their browser others can do bad things; so best to limit this to a smaller time period.

aljesusg:
Yes @mtho11 I agree with you , I was thinking how to do it the check ^^

aljesusg:
Documentation and change to 30 minutes https://github.com/kiali/kiali-ui/pull/500
I'll open a JIRA for the dialog box

aljesusg:
Created JIRA about dialog box https://issues.jboss.org/browse/KIALI-1117

PR #499, opened 2018-07-05
KIALI-840 Avoid sidebar to cover the service graph when resized

josejulio:
kiali-840

// cc: @mwringe

### DIFF
--- a/src/components/CytoscapeGraph/CytoscapeReactWrapper.tsx
+++ b/src/components/CytoscapeGraph/CytoscapeReactWrapper.tsx
@@ -41,7 +41,6 @@ type CytoscapeReactWrapperState = {};
  */
 export class CytoscapeReactWrapper extends React.Component<CytoscapeReactWrapperProps, CytoscapeReactWrapperState> {
   cy: any;
-  panzoom: any;
   divParentRef: any;
 
   constructor(props: CytoscapeReactWrapperProps) {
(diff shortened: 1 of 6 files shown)


src/components/CytoscapeGraph/CytoscapeReactWrapper.tsx
```diff
@@ -93,14 +92,14 @@ export class CytoscapeReactWrapper extends React.Component<CytoscapeReactWrapper
     );
 
     this.cy = cytoscape(opts);
-    this.panzoom = this.cy.panzoom();
```
josejulio: @israel-hdez updated this code, it returns the `cy` instance and receives the init arguments or the `destroy` command:
see: https://github.com/cytoscape/cytoscape.js-panzoom/blob/master/cytoscape-panzoom.js#L561

src/components/CytoscapeGraph/CytoscapeReactWrapper.tsx
```diff
@@ -93,14 +92,14 @@ export class CytoscapeReactWrapper extends React.Component<CytoscapeReactWrapper
...
+    // Receives object with default init values or 'destroy' command and returns cy instance
+    this.cy.panzoom();
+    this.cy.container().querySelector('.cy-panzoom').style.zIndex = -1;
```
israel-hdez: do we want to change z-index?

josejulio: Yes, if we don’t, the panzoom will be over the left menu when scrolling and that looks wrong

israel-hdez: But the graph also goes on top of the controls :/
I'm ok with that if you see it right.

josejulio: Didn’t notice that. Will see if can workaround that.

josejulio:
@israel-hdez updated the code, moved the z-index to other div

PR #501, opened 2018-07-06
KIALI-1112 Do a graph fit outside the batch operation to center the r…

josejulio:
…esult

This might also fix [KIALI-1060](https://issues.jboss.org/browse/KIALI-1060)

kiali-1112

### DIFF
--- a/src/components/CytoscapeGraph/CytoscapeGraph.tsx
+++ b/src/components/CytoscapeGraph/CytoscapeGraph.tsx
@@ -64,12 +64,14 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
   private trafficRenderer: TrafficRender;
   private cytoscapeReactWrapperRef: any;
   private updateLayout: boolean;
+  private refit: boolean;
   private resetSelection: boolean;
   private cy: any;
 
   constructor(props: CytoscapeGraphProps) {
     super(props);
     this.updateLayout = false;
+    this.refit = false;
   }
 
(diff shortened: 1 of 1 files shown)


src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -64,12 +64,14 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
...
   private cytoscapeReactWrapperRef: any;
   private updateLayout: boolean;
+  private refit: boolean;
   private resetSelection: boolean;
   private cy: any;
 
   constructor(props: CytoscapeGraphProps) {
     super(props);
     this.updateLayout = false;
+    this.refit = false;
```
jmazzitelli: Just a question - when is this.refit false again? I see it set to false in the constructor. But is this ever going to false again? I guess what this PR is doing is ensuring a refit is performed on every graph update now?

josejulio: In `processGraphUpdate` : https://github.com/kiali/kiali-ui/pull/501/files#diff-04b15b4602d6b9ccf1eb82d15e07535fR355

josejulio: Refit is performed on graph initialization and when there is a layout change: https://github.com/kiali/kiali-ui/pull/501/files#diff-04b15b4602d6b9ccf1eb82d15e07535fR325

jmazzitelli: ah... ok, i missed that. yeah, refit on every update would be bad, which is why i was confused :)

PR #503, opened 2018-07-09
KIALI-1114 Allow IstioConfig list show warning messages

aljesusg:
Jira: [KIALI-1114](https://issues.jboss.org/browse/KIALI-1114)

peek 2018-07-10 22-16

### DIFF
--- a/src/components/ConfigValidation/ConfigIndicator.tsx
+++ b/src/components/ConfigValidation/ConfigIndicator.tsx
@@ -1,7 +1,8 @@
 import * as React from 'react';
 import { ObjectValidation } from '../../types/ServiceInfo';
 import { PfColors } from '../Pf/PfColors';
-import { Icon, OverlayTrigger, Popover } from 'patternfly-react';
+import { Icon, OverlayTrigger, Popover, ListGroup, ListGroupItem } from 'patternfly-react';
+import { style } from 'typestyle';
 
 interface Props {
   id: string;
@@ -20,6 +21,12 @@ export const NOT_VALID: Validation = {
   icon: 'error-circle-o'
 };
 
+export const WARNING: Validation = {
+  name: 'Warning',
+  color: PfColors.Gold100,
+  icon: 'warning-triangle-o'
+};
+
 export const VALID: Validation = {
   name: 'Valid',
   color: PfColors.Green400,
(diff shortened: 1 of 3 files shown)


lucasponce:
Is the filter updated ?

aljesusg:
No @lucasponce I sent the Pull Request because I wanna make a question ^^ What do you think about show  instead number of issues , the number of warnings and number of errors?

lucasponce:
> No @lucasponce I sent the Pull Request because I wanna make a question ^^ What do you think about show instead number of issues , the number of warnings and number of errors?

+1 sounds good.

abonas:
@aljesusg IMO number of errors and number of warnings sounds much better than general number of issues.
cc @kiali/ux-reviewers what do you think?

aljesusg:
Great @abonas  so I'll remove number of total issues in that case, @lucasponce if there is only warnings issues the config is valid or not valid??

serenamarie125:
👍🏼 To including the actual number @aljesusg

lucasponce:
> Great @abonas so I'll remove number of total issues in that case, @lucasponce if there is only warnings issues the config is valid or not valid??

Filter should be updated to manage four states then, error, warning, valid and not validated. To let user work with all combination of results.
(Today it manages three, but we should split the not valid in the error and warning ones).

aljesusg:
That's my point @lucasponce now there are Istio objects with warning issues and no errors and the config is not valid.

aljesusg:
@serenamarie125 the total issues number in addition to number of errors issues and number of warning issues?

cshinn:
@aljesusg are there other types of issues besides warnings and errors? If not, I don't think having a total would be more helpful than listing them separately

aljesusg:
Updated
peek 2018-07-09 21-11

aljesusg:
Last update removing lines @cshinn 

peek 2018-07-09 23-45

aljesusg:
mmm maybe let me check it. Thanks @lucasponce

aljesusg:
Updated 

peek 2018-07-10 22-16

cshinn:
@aljesusg looks great!

lucasponce:
@aljesusg I'm testing locally master but I have this

image

I guess the expected behaviour would be to show the warning icon.

PR #505, opened 2018-07-10
KIALI-1113 Avoid overlap of node labels

josejulio:
Note: Cose still has some sort of overlap, but it seems more related to the overlapping of the group boxes

### Layout cola
Before:
before-cola

After:
after-cola

### Layout Breadthfirst
Before:
before-breadthfirst

After:
after-breadthfirst

### Layout Cose
Before:
before-cose

After:
after-cose-2

after-cose


jshaughn:
@josejulio THANK YOU!

PR #506, opened 2018-07-10
KIALI-1119 Nodes from other namespaces now show their version

josejulio:
Service name and namespace are now grouped into the box and single nodes have their version appended (if any)

kiali-1119-box

kiali-1119-services

Depends on https://github.com/kiali/kiali/pull/347

### DIFF
--- a/src/components/CytoscapeGraph/graphs/GraphStyles.ts
+++ b/src/components/CytoscapeGraph/graphs/GraphStyles.ts
@@ -54,15 +54,20 @@ export class GraphStyles {
               return '';
             }
 
-            if (ele.data('isOutside')) {
-              return `${service}\n${namespace}`;
-            }
-
             if (ele.data('parent')) {
               return version;
             }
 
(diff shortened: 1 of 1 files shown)


jmazzitelli:
@jshaughn didn't you have code in the other area where composite nodes are created where you can disable composite nodes if configured to? Is that also supported here in this PR?

I mentioned it because I think its a 50% chance composite nodes will go away in the future. We should make sure the code supports graphs with no composite nodes

jshaughn:
@jmazzitelli It's true that compound (not composite) nodes are optional although we always enable them.  But even if they are disabled this PR looks correct because the individual nodes will then include the name/namespace/version.

Having said that, I'm not sure I love putting namespace on a node at all.  It can be verbose as seen in the screenshot.  But that is for UX to decide.

josejulio:
> Having said that, I'm not sure I love putting namespace on a node at all. It can be verbose as seen in the screenshot. But that is for UX to decide.

Yes, but is only for the nodes outside the current namespace

PR #508, opened 2018-07-11
KIALI-1147 [UX] Change the term 'Undeployed' to 'Not deployed' for Istio sidecars indication

aljesusg:
[UX] Change the term 'Undeployed' to 'Not deployed' for Istio sidecars indication

peek 2018-07-11 10-04


aljesusg:
Updated
peek 2018-07-11 10-04

PR #510, opened 2018-07-11
KIALI-1129 Show virtuals services in a table within service details page

xeviknal:
Here you have the update:

screenshot of kiali console

Let me point something:
- I would like to hear from you about the columns shown in the table. Which would be the meaningful ones?
- Values in name column aren't links. They will be added into [KIALI-1130](https://issues.jboss.org/browse/KIALI-1130)

cc @kiali/ux-reviewers

### DIFF
--- a/src/pages/ServiceDetails/ServiceInfo/ServiceInfoVirtualServices.tsx
+++ b/src/pages/ServiceDetails/ServiceInfo/ServiceInfoVirtualServices.tsx
@@ -70,6 +70,16 @@ class ServiceInfoVirtualServices extends React.Component<ServiceInfoVirtualServi
           cell: {
             formatters: [this.cellFormat]
           }
+        },
+        {
+          property: 'actions',
+          header: {
+            label: 'Actions',
+            formatters: [this.headerFormat]
+          },
+          cell: {
+            formatters: [this.cellFormat]
+          }
         }
       ]
     };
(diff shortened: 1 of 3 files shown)


lucasponce:
> I would like to hear from you about the columns shown in the table. Which would be the meaningful ones?

VirtualServices as other CRD object used in the Istio context has two kind of properties: those coming from the ObjectMeta (generic to k8s), coming from the Spec, specific to the context.

As we split this view from table to detail, maintaining the generic ones: "name", "createdAt", "resourceVersion" makes sense. If we bring more Spec fields, we would break this "why this yes or why nt this other one". Also, adding the validation makes sense to me.

One thing I would add here is a hover in the validation icon at least to show the # of warnings and # of errors in a similar way of IstioConfig.

Then, when we enter in the details, we can add more value showing the details of the validations too.

src/pages/ServiceDetails/ServiceInfo/ServiceInfoVirtualServices.css.map
```diff
@@ -0,0 +1,7 @@
+{
```
lucasponce: Do we commit the .map files ?
Those are for debugging ?
I don't know just wonder what is the usual rule.

xeviknal: yep, I totally forgot to remove them. Probably, we should update .gitignore to ignore all .map files

src/pages/ServiceDetails/ServiceInfo/ServiceInfoVirtualServices.tsx
```diff
@@ -1,42 +1,99 @@
...
-  globalStatus(rule: VirtualService) {
-    let validation = this.validation(rule);
+  globalStatus(vs: VirtualService, idx: number) {
+    let validation = this.validation(vs);
     let checks = globalChecks(validation);
     let severity = validationToSeverity(validation);
     let iconName = severityToIconName(severity);
     let color = severityToColor(severity);
     let message = checks.map(check => check.message).join(',');
+    console.log(color);
```
lucasponce: Missed console.log there

xeviknal:
here it comes the version with the warning/error summary:

screenshot of kiali console 1

It is using the same component as istio config list do.

abonas:
I think that viewyaml link has to have its own column with a separator, otherwise it might give the impression it is yaml of only one property which it shares thr column with

abonas:
I am also not sure resource version is that valuable here, do users really follow and remember resource versions of all their objects? Maybe an indicator when it was updated will be more helpful if we have that info. Or is it the created field used for updated too?

lucasponce:
Resource versions are necessary. As it doesn't have a modification date.
So, if a resource is updated, the only change is in resourceVersion (need to check with k8s >1.9, but before that was the case).

xeviknal:
@abonas I thought quite the same, but in serena's mockup was like that:
https://github.com/kiali/kiali-design/issues/4#issuecomment-400154688

Should I add a new column named, for example, `actions` with that only link? Should I stick to serenas?

xeviknal:
screenshot of kiali console 2

Do we prefer something like that? @kiali/ux-reviewers thoughts?

lucasponce:
> Do we prefer something like that?
+1
Being purist and looking into patternfly http://www.patternfly.org/pattern-library/content-views/table-view/
I would adjust/short the actions column, just to differenciate from the "data/info".

xeviknal:
screenshot of kiali console 3

Give me some thumbs up on this :)

PR #511, opened 2018-07-11
Ignore .css.map files from git

xeviknal:
do you think this is a good idea?
Perhaps I am not using project properly, but I found myself committing one .css.map file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,7 @@ package-lock.json
 
 # compiled patternfly assets
 src/**/*.css
+src/**/*.css.map
 !src/index.css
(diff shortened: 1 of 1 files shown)


jeff-phillips-18:
I'm not sure of a best practice, but mostly I've seen the map files get ignored.

lucasponce:
I don't see an explicit ignore line on https://github.com/openshift/console but in the code there is not any .map file.

xeviknal:
It doesn't sound natural to me the fact of debugging into production environment.

xeviknal:
@mtho11 glad that we had this conversation! thx!

PR #512, opened 2018-07-11
KIALI-1162 Missing warning icon on tab under ServiceDetails

aljesusg:
JIRA: [KIALI-1162](https://issues.jboss.org/browse/KIALI-1162)
screenshot from 2018-07-11 21-00-48

@lucasponce I changed the note **Not all checks passed** too, I think that this should show if there is an error or warning too.

### DIFF
--- a/src/pages/ServiceDetails/ServiceInfo.tsx
+++ b/src/pages/ServiceDetails/ServiceInfo.tsx
@@ -18,7 +18,7 @@ import ServiceInfoDescription from './ServiceInfo/ServiceInfoDescription';
 import ServiceInfoPods from './ServiceInfo/ServiceInfoPods';
 import ServiceInfoDeployments from './ServiceInfo/ServiceInfoDeployments';
 import ServiceInfoRoutes from './ServiceInfo/ServiceInfoRoutes';
-import { ServiceDetailsInfo, severityToIconName, Validations } from '../../types/ServiceInfo';
+import { ServiceDetailsInfo, severityToIconName, validationToSeverity, Validations } from '../../types/ServiceInfo';
 import ServiceInfoVirtualServices from './ServiceInfo/ServiceInfoVirtualServices';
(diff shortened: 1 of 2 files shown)


src/pages/ServiceDetails/ServiceInfo.tsx
```diff
@@ -130,11 +140,11 @@ class ServiceInfo extends React.Component<ServiceDetails, ServiceInfoState> {
                     <NavItem eventKey={2}>{'Source Services (' + Object.keys(dependencies).length + ')'}</NavItem>
                     <NavItem eventKey={3}>
                       {'Virtual Services (' + virtualServices.length + ')'}
-                      {validationChecks.hasVirtualServiceChecks ? errorIcon : undefined}
+                      {validationChecks.hasVirtualServiceChecks ? getValidationIcon() : undefined}
                     </NavItem>
                     <NavItem eventKey={4}>
                       {'Destination Rules (' + destinationRules.length + ')'}
-                      {validationChecks.hasDestinationRuleChecks ? errorIcon : undefined}
+                      {validationChecks.hasDestinationRuleChecks ? getSeverityIcon() : undefined}
```
lucasponce: DestinationRules might have warnings as well.
Not today, but in next step, so I think this logic should be generic for both.

aljesusg: So should we prepare the code or we should wait until next update?

lucasponce: Instead to have a getValidationIcon() and getSeverityIcon() for the same task.
I would unify it now to make it generic.
Indeed, previous logic was generic, but it didn't consider 'warning' case.

xeviknal: why are you using `getValidationIcon()` for virtual services and `getSeverityIcon()` for DestinationRules? what is the purpose?

aljesusg:
Changed to a generic method

PR #513, opened 2018-07-11
KIALI-1159 FIX Refreshing kiali always redirects to service graph

aljesusg:
- Refactor
  - Now you manage the routes in  /src/components/routes
  - Reduce the code and complexity 
- Fix Refresh

refresh

### DIFF
--- a/src/components/Nav/Navigation.tsx
+++ b/src/components/Nav/Navigation.tsx
@@ -1,38 +1,23 @@
 import * as React from 'react';
 import { VerticalNav } from 'patternfly-react';
 import PropTypes from 'prop-types';
-import { Route, Redirect } from 'react-router-dom';
+import { routes } from '../../routes';
+import RenderPage from './RenderPage';
+import { matchPath } from 'react-router';
+import _ from 'lodash';
 
-import SwitchErrorBoundary from '../SwitchErrorBoundary/SwitchErrorBoundary';
-import PfContainerNavVertical from '../Pf/PfContainerNavVertical';
 import MessageCenter from '../../containers/MessageCenterContainer';
-
-import IstioConfigPage from '../../pages/IstioConfigList/IstioConfigListPage';
-import IstioConfigDetailsPage from '../../pages/IstioConfigDetails/IstioConfigDetailsPage';
 import HelpDropdown from '../../containers/HelpDropdownContainer';
 import UserDropdown from '../../containers/UserDropdownContainer';
-import ServiceDetailsPage from '../../pages/ServiceDetails/ServiceDetailsPage';
-import ServiceGraphRouteHandler from '../../pages/ServiceGraph/ServiceGraphRouteHandler';
-import ServiceListPage from '../../pages/ServiceList/ServiceListPage';
-import ServiceJaegerPage from '../../pages/ServiceJaeger/ServiceJaegerPage';
 import LoginPage from '../../containers/LoginPageContainer';
 import { store } from '../../store/ConfigStore';
 import PfSpinnerContainer from '../../containers/PfSpinnerContainer';
 
-const istioConfigPath = '/istio';
 export const istioConfigTitle = 'Istio Config';
-const serviceGraphPath = '/service-graph/all';
-export const serviceGraphTitle = 'Graph';
-const servicesPath = '/services';
-const servicesJaegerPath = '/jaeger';
 export const servicesTitle = 'Services';
-export const servicesJaeger = 'Distributed Tracing';
 
 const pfLogo = require('../../img/logo-alt.svg');
 const pfBrand = require('../../assets/img/kiali-title.svg');
-
-const servicesRx = /\/namespaces\/(.*)\/services\/(.*)/g;
-const istioConfigRx = /\/namespaces\/(.*)\/istio\/(.*)/g;
 
 type PropsType = {
   location: any;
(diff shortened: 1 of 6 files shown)


src/components/Nav/Navigation.tsx
```diff
@@ -106,25 +48,42 @@ class Navigation extends React.Component<PropsType, StateType> {
...
     }
   };
 
+  renderMenuItems() {
+    const { location } = this.props;
+    const menuItems = routes();
+    const activeItem = menuItems.find(item => {
+      let isRoute = _.startsWith(location.pathname, item.to);
+      if (!isRoute && item.pathsActive) {
+        isRoute = _.filter(item.pathsActive, path => path.test(location.pathname)).length > 0;
```
josejulio: I think you can use [matchPath](https://reacttraining.com/react-router/web/api/matchPath), this might be safer than our current `startsWith` approach.
Also, maybe this example can be useful: https://reacttraining.com/react-router/web/example/custom-link

src/components/Nav/Content.tsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import { Redirect, Route } from 'react-router-dom';
+import SwitchErrorBoundary from '../SwitchErrorBoundary/SwitchErrorBoundary';
+import { routes, pathRoutes } from '../../routes';
+import PfContainerNavVertical from '../Pf/PfContainerNavVertical';
+
+class Content extends React.Component {
```
mtho11: `Content` is a really general term. Everything is Content.
Also, it looks like this `React.Component` could be a [ReactPureComponent](https://reactjs.org/docs/react-api.html#reactpurecomponent) and would save some memory and performance.

src/routes.tsx
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * Return array of objects that describe vertical menu
+ * @return {array}
+ */
+const routes = () => [
+  {
+    iconClass: 'fa pficon-topology',
+    title: 'Graph',
+    to: '/service-graph/all',
+    redirect: true,
```
josejulio: Can we have a test that checks that we only have one item with `redirect` ? If we accidentally add more it will produce something undesired. Other option is to have a separate item in `routes.ts` e.g. `const defaultRoute = routes[0]` or more explicity:
```typescript
const serviceGraphRoute = {
  iconClass: 'fa pficon-topology',
  title: 'Graph',
  to: '/service-graph/all',
  redirect: true,
  component: ServiceGraphRouteHandler,
  pathsActive: [/\/service-graph\/(.*)\//g]
};
const defaultRoute = serviceGraphRoute;

const routes = () => [
  serviceGraphRoute,
  // etc ...
];
```

src/components/Nav/Content.tsx
aljesusg: mmm it's not working with PureComponent

aljesusg: it seems there are problems in Purecomponent with routes

src/routes.tsx
aljesusg: Changed

src/components/Nav/Navigation.tsx
aljesusg: Changed

src/components/Nav/Content.tsx
josejulio: I don't think is really a `PureComponent`
> If your React component’s render() function renders the same result given the same props and state, you can use React.PureComponent for a performance boost in some cases.

But this one renders different depending the url

src/routes.tsx
```diff
@@ -0,0 +1,60 @@
+import ServiceGraphRouteHandler from './pages/ServiceGraph/ServiceGraphRouteHandler';
+import ServiceListPage from './pages/ServiceList/ServiceListPage';
+import IstioConfigPage from './pages/IstioConfigList/IstioConfigListPage';
+import ServiceJaegerPage from './pages/ServiceJaeger/ServiceJaegerPage';
+import ServiceDetailsPage from './pages/ServiceDetails/ServiceDetailsPage';
+import IstioConfigDetailsPage from './pages/IstioConfigDetails/IstioConfigDetailsPage';
+
+const baseName = '/';
```
israel-hdez: Is this actually used?
If not, please remove. I find this confusing given that we have:
https://github.com/kiali/kiali-ui/blob/83aa48c2d87b60fbc07e486fb81078da76616d85/src/app/History.tsx#L2

src/routes.tsx
```diff
@@ -0,0 +1,60 @@
...
+    redirect: true,
+    component: ServiceGraphRouteHandler,
+    pathsActive: [/\/service-graph\/(.*)\//g]
+  },
+  {
+    iconClass: 'fa pficon-service',
+    title: 'Services',
+    to: '/services',
+    component: ServiceListPage,
+    pathsActive: [/\/namespaces\/(.*)\/services\/(.*)/g]
```
israel-hdez: Hmm... All these regex should start with `^` in order to work like the old code (that is using `startsWith`). Also, I think that the `g` flag shouldn't be used.

src/routes.tsx
aljesusg: Yes, I'll remove it, I was looking to fix the error in console with path

src/routes.tsx
lucasponce: +1 
Using `^` will improve performance as well.

src/components/Nav/Navigation.tsx
```diff
@@ -106,25 +49,42 @@ class Navigation extends React.Component<PropsType, StateType> {
...
     }
   };
 
+  renderMenuItems() {
+    const { location } = this.props;
+    const menuItems = routes();
+    const activeItem = menuItems.find(item => {
+      let isRoute = matchPath(location.pathname, { path: item.to, exact: true, strict: false }) ? true : false;
+      if (!isRoute && item.pathsActive) {
+        isRoute = _.filter(item.pathsActive, path => path.test(location.pathname)).length > 0;
```
lucasponce: loadsh is used because we don't know the type and we can filter correctly ?
Or what is the use case ?

aljesusg: The case is know which menu item should be active so if path not match with the "normal paths" stored in each item.path in store I'll check if there is a pathsActive that match.

This is to have active services for example when the path is /\/namespaces\/(.*)\/services\/(.*)/g

src/routes.tsx
aljesusg: Maybe move this to the same file to configure all related with routes in the same file

aljesusg: Changed

src/routes.tsx
aljesusg: Changed

src/components/Nav/Navigation.tsx
aljesusg: Changed

src/routes.tsx
```diff
@@ -0,0 +1,60 @@
...
+    path: '/namespaces/:namespace/services/:service',
+    component: ServiceDetailsPage
+  },
+  {
+    path: '/namespaces/:namespace/istio/:objectType/:object',
+    component: IstioConfigDetailsPage
+  }
+];
+
+export { baseName, routes, pathRoutes };
```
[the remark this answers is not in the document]
aljesusg: Used now ^^

aljesusg:
@lucasponce , @mtho11 can you review it? Thanks !!!

aljesusg:
Thanks !!!

PR #514, opened 2018-07-11
KIALI-1058 Present an error view if the CytoscapeGraph fails

josejulio:
This won't block the controls, allowing the user to select a different layout if needed

Bonus:
Simplified the refit operation and this maybe fixed the "all nodes appearing on top of each other".
It was a mix of [KIALI-1166](https://issues.jboss.org/browse/KIALI-1166) and the graph not doing an updateLayout as the first operation (only doing the refit)

kiali-1058

### DIFF
--- a/src/components/CytoscapeGraph/CytoscapeGraph.tsx
+++ b/src/components/CytoscapeGraph/CytoscapeGraph.tsx
@@ -64,14 +64,12 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
   private trafficRenderer: TrafficRender;
   private cytoscapeReactWrapperRef: any;
   private updateLayout: boolean;
-  private refit: boolean;
   private resetSelection: boolean;
   private cy: any;
 
   constructor(props: CytoscapeGraphProps) {
     super(props);
     this.updateLayout = false;
-    this.refit = false;
   }
 
   shouldComponentUpdate(nextProps: CytoscapeGraphProps, nextState: CytoscapeGraphState) {
@@ -130,8 +128,6 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
             ref={e => {
               this.setCytoscapeReactWrapperRef(e);
             }}
-            elements={this.props.elements}
-            layout={this.props.graphLayout}
           />
         </EmptyGraphLayout>
       </div>
@@ -179,7 +175,7 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
       return;
     }
     this.cy = cy;
-    this.refit = true;
+    this.updateLayout = true;
 
     this.graphHighlighter = new GraphHighlighter(cy);
     this.trafficRenderer = new TrafficRender(cy, cy.edges());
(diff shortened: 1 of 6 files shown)


mtho11:
@jmazzitelli looks like we are waiting for you on this one

israel-hdez:
@jmazzitelli I'm seeing the 'all' namespace problem in master. I guess this PR isn't introducing the issue.

josejulio:
> @jmazzitelli I'm seeing the 'all' namespace problem in master. I guess this PR isn't introducing the issue.

Yes, probably is something that I broke outside this PR :)

jmazzitelli:
what about the "unknown" as the default?

jmazzitelli:
I created https://issues.jboss.org/browse/KIALI-1176 for this namespace dropdown bug. I will approve this PR since that bug is not related here.

abonas:
cc @kiali/ux-reviewers 
1. Why can't the message about the reason for not loading the graph just presented on the screen? I mean - why is the user sent to "the message center" if the main screen anyway presents half of the message that the graph cannot be loaded?
2. Shouldn't the error "Error loading graph" be with the proper error icon and in red color?

mwringe:
> Why can't the message about the reason for not loading the graph just presented on the screen? I mean - why is the user sent to "the message center" if the main screen anyway presents half of the message that the graph cannot be loaded?

> Shouldn't the error "Error loading graph" be with the proper error icon and in red color?

I don't think this PR touches on how the error page handles things. There was a PR for the error page that included all of this, but another PR got merged first.

I think we can make these changes to this PR without having to create an additional jira to handle it.

abonas:
@mwringe so I misunderstand what this PR is about :)
my comments still stand, whether handled here or later/after, I don't want to complicate things here

josejulio:
> Why can't the message about the reason for not loading the graph just presented on the screen? I mean - why is the user sent to "the message center" if the main screen anyway presents half of the message that the graph cannot be loaded?

I'm not really sure why the message wasn't passed to this component, but the message is being show in the top right for 6 seconds before hiding into the MessageCenter. Haven't looked how easy/hard is to put that message there.

> Shouldn't the error "Error loading graph" be with the proper error icon and in red color?

I guess so, If you want I can change the color and put an error icon there. If UXD needs to check this, can we open a Jira or issue at https://github.com/kiali/kiali-design/issues to not block this longer than needed? 

> There was a PR for the error page that included all of this

This PR added the isError part: https://github.com/kiali/kiali-ui/pull/436

mwringe:
> This PR added the isError part: #436

And the other one which had the icon and error message as part of error page: https://github.com/kiali/kiali-ui/pull/374

It should be fairly simple to add in the icon and error message.

cshinn:
@abonas I agree, if changing the layout scheme will solve the problem, then rather than opening the message center, that large button should change the layout scheme to one that works. 

We could use an error icon and some red in this case, but given that nothing is broken except that the user needs to change the layout, I think it's too strong of a signal. Perhaps an informational icon (the i in the circle) would better convey the severity of the issue.

mwringe:
> We could use an error icon and some red in this case, but given that nothing is broken except that the user needs to change the layout, I think it's too strong of a signal. Perhaps an informational icon (the i in the circle) would better convey the severity of the issue.

The service graph cannot be loaded due to an error,

cshinn:
@mwringe sorry, I thought this was only for a specific case. If we're talking about a general error case, then the red error icon makes sense. If we expect that we're likely to run into the "need to change the display scheme" issue frequently, then it might be worth having a special case for that one. Or better yet, catch it and do the display change automatically.

josejulio:
I created a JIRA for improving how the error message appears

https://issues.jboss.org/browse/KIALI-1198

josejulio:
If there is no further comments on this, can we merge it?

PR #516, opened 2018-07-11
KIALI-949 Document our new release process

cfcosta:
This is a tentative PR to gather some feedback on how we should do stuff. Right now we have a v0.5.0 branch on the UI for the latest release, but it was not updated on `package.json`. This is bad for many reasons, one being that there's no way to actually know which snapshot of 0.2.10 is the correct one for the version you're running.

So what's I'm suggesting on this (and we can merge if everyone agrees) is to bump the version on master after the sprint starts, and release the final version after it ends. This is the way most other OSS projects do it, afaik (either that or they keep the latest version number until the release is ready, but I don't like this because it creates a disconnect from what master refers to in comparison to the tag).

What do you guys think?

### DIFF
--- a/RELEASING.adoc
+++ b/RELEASING.adoc
@@ -15,8 +15,8 @@ This document also uses 1.0.0 as the new version, just as an example.
 
 == Preparing for a new version
 
-After a release happens, the first step for a new "batch" of modifications is
-to update link:package.json[] with the new version, and creating a PR for it:
+After a release, the first step for a new "batch" of modifications is to update
+link:package.json[] with the new version, and creating a PR for it:
 
 [source, diff]
 ----
@@ -34,7 +34,7 @@ index 7f699b1..08bc0a8 100644
    },
 ----
 
-This is to signalize everyone that everything that goes into master from that
+This is to signal to everyone that everything that goes into master from that
 point forward belongs to that version. Note that this is supposed to be the
 same version that is going to be released for the core, so they are in sync.
 
@@ -43,14 +43,14 @@ link:http://semver.org[semver], but not as strict, given both that we are not
 1.0 yet and that we are the end product, in the sense that nothing should
 depend on us to do anything. For now, and that might change in the future, our
 API is private and we make no guarantees that compatibility is going to be
-mantained.
+maintaned.
 
 For our versioning scheme, the `major` part means a major release, and that's
 decided in advance by the team. The `minor` means new features are included
 with the release, and the `patch` version covers only bugfixes.
 
 The UI and the core might be out of sync in `patch` versions, but they should
-be in sync for `minor`. That means that, if the ui is 0.5.10, then any core
+be in sync for `minor`. That means that, if the UI is 0.5.10, then any core
 version on 0.5.X should work with it, but that are no guarantees that it will
 work with core 0.4.X or core 0.6.X.
 
@@ -62,14 +62,14 @@ might work just fine, we just don't give any guarantees that it will.
 After the work that goes into a new version is done, the next step is to create
 the tag and push it. Everything else will be taken care by Travis.
 
+Note that only tags or branches in the format `v\#.#.#[.Label]` will trigger
+release tag/branch builds.
+
 [source, bash]
 ----
 $ git tag v1.0.0 <hash-of-last-commit-of-release>
 $ git push origin v1.0.0
 ----
-
-And that's it. Note that only tags or branches in the format `v\#.#.#[.Label]`
-will trigger release tag/branch builds.
 
 == Creating a Hotfix
 
--- a/RELEASING.adoc
+++ b/RELEASING.adoc
@@ -1,0 +1,92 @@
+= Releasing Kiali UI
+
+This document covers the specifics of releasing Kiali UI, and is meant as a
+guide for Kiali release managers.
+
+The main document covering core can be found link:https://github.com/kiali/kiali/blob/master/RELEASING.adoc[here].
+
+== Requirements
+
+You must have write permissions to the public GitHub repository in order to be
+able to push the tags, but that's pretty much it. Most of our release process
+is done automatically by Travis.
+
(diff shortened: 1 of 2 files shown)


cfcosta:
This is related to https://github.com/kiali/kiali/pull/351

cfcosta:
@josejulio @xeviknal @mtho11 hey, just pushed a new commit. I also added a RELEASING document that covers some of the release process that I am proposing.

Following this process, it also means that our next version is going to be 0.6.0, not 0.5.1, so I also updated `package.json` for that.

@theute I think this might be interesting to you as well.

pilhuhn:
Is there a way of doing the tagging/bumping by something like the maven-release-plugin (which is obviously not applicable here)?

cfcosta:
@pilhuhn I don't know, to be honest. We can think about doing it automatically after we define the whole process, though.

RELEASING.adoc
```diff
@@ -0,0 +1,92 @@
...
+might work just fine, we just don't give any guarantees that it will.
+
+== Creating the Tag
+
+After the work that goes into a new version is done, the next step is to create
+the tag and push it. Everything else will be taken care by Travis.
+
+[source, bash]
+----
+$ git tag v1.0.0 <hash-of-last-commit-of-release>
```
mtho11: A more `npm` automated approach could be
```
npm version minor
npm push && npm push --tags
```
Npm will automatically increment the version based on `patch|minor|major'

josejulio: +1
but with yarn :)
https://yarnpkg.com/lang/en/docs/cli/version/

cfcosta: I'm up for doing that, but tbh I'd prefer if this was added after someone did a release with it. I don't know how it works, and do not have the permissions to do a release on it.

I think this is outside of this scope, but can be added after the next release.

RELEASING.adoc
```diff
@@ -0,0 +1,92 @@
...
+ {
+   "name": "@kiali/kiali-ui",
+-  "version": "0.5.1",
++  "version": "1.0.0",
+   "bugs": {
+     "url": "https://issues.jboss.org/projects/KIALI/issues/"
+   },
+----
+
+This is to signalize everyone that everything that goes into master from that
```
JStickler: This is to signal to everyone that everything that goes into master...

RELEASING.adoc
```diff
@@ -0,0 +1,92 @@
...
+depend on us to do anything. For now, and that might change in the future, our
+API is private and we make no guarantees that compatibility is going to be
+mantained.
+
+For our versioning scheme, the `major` part means a major release, and that's
+decided in advance by the team. The `minor` means new features are included
+with the release, and the `patch` version covers only bugfixes.
+
+The UI and the core might be out of sync in `patch` versions, but they should
+be in sync for `minor`. That means that, if the ui is 0.5.10, then any core
```
JStickler: UI

RELEASING.adoc
```diff
@@ -0,0 +1,92 @@
...
+After the work that goes into a new version is done, the next step is to create
+the tag and push it. Everything else will be taken care by Travis.
+
+[source, bash]
+----
+$ git tag v1.0.0 <hash-of-last-commit-of-release>
+$ git push origin v1.0.0
+----
+
+And that's it. Note that only tags or branches in the format `v\#.#.#[.Label]`
```
JStickler: I would move the sentence that starts with "note" to above the command.

theute:
Sounds good.
What's the "next version" would be ? At the moment we bump the minor (y in x.y.z) in each release after each sprint (next is 0.6.0), but in the case here we really screw up a minor release (say that 0.6.0 doesn't even start) we would likely want to release a 0.6.1 patch release ASAP after.

So should we be conservative and set to 0.6.1 and change again right betore a release to 0.7.0 or is it ok to change to 0.7.0 and then change to 0.6.1 only if needed ? (PS: I don't know how dependency management works in this UI world, the latest sounds fine if it doesn't screw things up)

And I definitely agree that we need to define the process then automate, since we'll do a release at least every 3 weeks.

RELEASING.adoc
```diff
@@ -0,0 +1,92 @@
...
+
+You must have write permissions to the public GitHub repository in order to be
+able to push the tags, but that's pretty much it. Most of our release process
+is done automatically by Travis.
+
+This document also uses 1.0.0 as the new version, just as an example.
+
+== Preparing for a new version
+
+After a release happens, the first step for a new "batch" of modifications is
```
abonas: not sure how to fix this, but "after a release happens" does not sound like proper English. @JStickler ?

RELEASING.adoc
```diff
@@ -0,0 +1,92 @@
...
+This is to signalize everyone that everything that goes into master from that
+point forward belongs to that version. Note that this is supposed to be the
+same version that is going to be released for the core, so they are in sync.
+
+The versioning scheme that we are following is similar to
+link:http://semver.org[semver], but not as strict, given both that we are not
+1.0 yet and that we are the end product, in the sense that nothing should
+depend on us to do anything. For now, and that might change in the future, our
+API is private and we make no guarantees that compatibility is going to be
+mantained.
```
abonas: mantained -> maintained

cfcosta:
@theute in your case, we start working on v0.7.0, but if there's the need for a backport/bugfix, we start a new branch from v0.6.0, cherry-pick the commits we need and release it after that.

As for versioning on UI, it follows semver, so we don't need to worry if we release v0.6.1 after v0.7.0, any new builds/checkouts would start at v0.7.0, unless someone is bound to v0.6 for any reason, in that case it'd be safe to update to v.0.6.1, as we'd only add fixes.

RELEASING.adoc
cfcosta: Changed it to "After a release", what do you think?

theute:
Sounds good thanks

PR #520, opened 2018-07-16
KIALI-1132 IstioConfig: MixerRules UX refactoring

lucasponce:
- Present MixerRules in a list/table instead of a list of links
- Separate AceEditor from list/table view for MixerRules

istioconfig ogv

### DIFF
--- a/src/pages/IstioConfigDetails/IstioRuleInfo.tsx
+++ b/src/pages/IstioConfigDetails/IstioRuleInfo.tsx
@@ -135,10 +135,7 @@ class IstioRuleInfo extends React.Component<IstioRuleInfoProps> {
             formatters: [this.headerFormat]
           },
           cell: {
-            formatters: [this.cellFormat],
-            props: {
-              align: 'text-center'
-            }
+            formatters: [this.cellFormat]
           }
         },
         {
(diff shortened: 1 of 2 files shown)


abonas:
Shouldn't all the fields in the table be aligned to left instead of center?

src/pages/IstioConfigDetails/IstioRuleInfo.tsx
```diff
@@ -162,10 +156,7 @@ class IstioRuleInfo extends React.Component<IstioRuleInfoProps> {
             formatters: [this.headerFormat]
           },
           cell: {
-            formatters: [this.cellFormat],
-            props: {
-              align: 'text-center'
-            }
+            formatters: [this.cellFormat]
```
abonas: where does it take the alignment style from? I see you deleted the text-center, so the left alignment is taken from somewhere else?

lucasponce:
Table columns fixed:
image

lucasponce:
@xeviknal Travis seems happy now.

src/pages/IstioConfigDetails/IstioRuleInfo.tsx
lucasponce: It is the default alignment style from the component.
Is it looking different locally to you ?

PR #521, opened 2018-07-16
KIALI-1159 FIX refresh in deployment container problem with baseName

aljesusg:
Fix problem refresh in kiali inside the container


aljesusg:
@hhovsepy @lucasponce I am still checking if I am missing some link but this fix the problem with the baseName in links inside the container. Can you test it Hayk?

lucasponce:
cc @hhovsepy this for the reported one, could you also test it ? thanks

hhovsepy:
building it now

lucasponce:
I have tested manually and it works in an openshift deployment and locally using a proxy.
Going to take a look into the code now.

hhovsepy:
Tested on OC cluster, refresh is working now. And it refreshes current page. Links are working as well.

lucasponce:
This fix works, but I wonder why we can't play with the baseName inside the Router.
https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/docs/api/BrowserRouter.md

Something is missing to me, but I have not enough background in react/routers to check.
Before to merge I would take a look if something with baseName can be used instead to build the full url and use the helpers.
If it can't be apply, then I guess we can merge this one.

aljesusg:
@lucasponce there is a configuration for this in [src/app/History.tsx](https://github.com/kiali/kiali-ui/blob/28d382bcb5dc87c3578c0f2e6236bc32cd56ec1d/src/app/History.tsx) but I don't know why is not working.

```typescript
import { createBrowserHistory } from 'history';
import { baseName } from '../routes';

const history = createBrowserHistory({ basename: baseName });
export default history;
```

lucasponce:
Perhaps is missing here  `<Router history={history}>`  in App.tsx ?

aljesusg:
No, 

```typescript
<Provider store={store}>
        <PersistGate loading={<Loading />} persistor={persistor}>
          <Router history={history}>
            <Sidebar />
          </Router>
        </PersistGate>
      </Provider>
```

lucasponce:
Not sure, I understand that baseName in history, helps to move inside the history object.
But when it refreshes, the route move/ignore the baseName to the "/", then the basename used "/console" is missed.
At least is what it looks like.

israel-hdez:
This PR is working like a workaround.

We shouldn't be prefixing routes with the basename.
The real problem is in https://github.com/kiali/kiali-ui/blob/master/src/app/History.tsx file. When the history object is created at: 
https://github.com/kiali/kiali-ui/blob/78a0df20616b0e07550460dd11817e455858bf83/src/app/History.tsx#L4

the `baseName` constant is `undefined` at that moment. If I revert this PR and change that line to
`const history = createBrowserHistory({ basename: '/console' });` everything works normally.

lucasponce:
@israel-hdez Do you mind to prepare a PR with that ?
Instead reverting the PR from @aljesusg I would suggest just to add a new one with these changes.
We can also ping QE to test as well.

israel-hdez:
Of course, I'll prepare it.
But, not sure if I'll get it by the end of your day... I hope I will.

aljesusg:
Mmm I wonder why baseName is not defined in History when it's a constant... I can make the PR with the redo if you can't

PR #522, opened 2018-07-16
KIALI-1177 Fix refresh issue. Basepath was not being set.

israel-hdez:
There was a loop over import statements:

routes.tsx ->
  pages/ServiceDetails/ServiceDetailsPage.tsx ->
  pages/ServiceDetails/ServiceMetrics.tsx ->
  components/MetricsOptions/MetricsOptionsBar.tsx ->
  components/MetricsOptions/ValueSelectHelper.tsx -> app/History.tsx ->
  routes.tsx

In this loop, routes.tsx and History.tsx are involved.

This loop was casuing the transpiler (babel) to generate JS with code in
arbitary order. So, at the time the app/History.tsx code was being run,
the file routes.tsx haven't, effectively setting an "undefined"
basename. This causes react-router to work without right the basename of
the application.

This commit:

* Reverts PR #521, because it is a workaround. This PR solves the root
  problem by moving the basename to History.tsx (since it's not used in
  any other part of the code, given the reversion).
* Adds tslint-no-circular-imports linter extension to detect import
  loops and avoid similar issues in the future.
* Solves all other import loops detected.

The other loops found are:

* actions/ServiceGraphActions.ts ->
    components/CytoscapeGraph/CytoscapeGraph.tsx ->
    actions/ServiceGraphActions.ts
* components/CytoscapeGraph/CytoscapeGraph.tsx ->
    components/CytoscapeGraph/graphs/GraphHighlighter.ts ->
    components/CytoscapeGraph/CytoscapeGraph.tsx
* components/GraphFilter/GraphFilter.tsx ->
    containers/GraphSettingsContainer.tsx ->
    components/GraphFilter/GraphFilter.tsx
* store/ConfigStore.ts -> reducers/index.ts ->
    reducers/ServiceGraphDataState.ts ->
    actions/ServiceGraphDataActions.ts ->
    utils/MessageCenter.ts ->
    store/ConfigStore.ts

(and some others).


gbaufake:
cc @mattmahoneyrh

src/routes.tsx
```diff
@@ -17,55 +14,55 @@ const navItems: MenuItem[] = [
   {
     iconClass: 'fa pficon-topology',
     title: 'Graph',
-    to: kialiRoute('/service-graph/all'),
-    pathsActive: [/\/service-graph\/(.*)/]
+    to: '/service-graph/all',
+    pathsActive: [/^\/service-graph\/(.*)\//]
```
josejulio: Can you remove the `\/`?
It doesn't match an url like (notice after the namespace there is no `/`):
`http://localhost:3000/console/service-graph/kiali-test-breadth-0?layout=dagre&duration=60&edges=hide`

package.json
```diff
@@ -83,6 +83,7 @@
     "npm-snapshot": "1.0.3",
     "prettier": "1.13.4",
     "puppeteer": "1.3.0",
+    "tslint-no-circular-imports": "^0.5.0",
```
josejulio: Please fix the version (i.e. `0.5.0`)

src/actions/ActionKeys.ts
```diff
@@ -0,0 +1,62 @@
...
+  SET_GRAPH_EDGE_LABEL_MODE = 'SET_GRAPH_EDGE_LABEL_MODE',
+  // Disable Actions
+  ENABLE_GRAPH_FILTERS = 'ENABLE_GRAPH_FILTERS',
+  // Refresh Rate
+  SET_GRAPH_REFRESH_RATE = 'SET_GRAPH_REFRESH_RATE'
+}
+
+export enum UserSettingsActionKeys {
+  NAV_COLLAPSE = 'NAV_COLLAPSE'
+}
```
josejulio: This looks OK to me, but alternatives are to use e.g. `GlobalActionsKeys.ts`, having them next to the actions or inside `action/keys`.
cc: @mtho11

src/routes.tsx
israel-hdez: Yes... ops...

package.json
israel-hdez: Fixed.

mattmahoneyrh:
Refresh worked fine for these tests:
 - Browser refresh
 - F5
 - Service Graph Refresh button


Service Graph 'all' namespace is also now properly displaying all namespace's.

src/actions/ActionKeys.ts
mtho11: This refactor of ActionKeys is something totally separate and not required to fix this bug. 
I personally don't agree with this change. The actions keys were co-located with their actions which is the way I believe it should be -- no need to import extra files for the keys. We don't put all the reducers in a single file, we organize them by logical groups (not by kind). As the application grows I think this method of organization makes much more sense.

src/pages/ServiceDetails/ServiceDetailsPage.tsx
```diff
@@ -194,13 +193,13 @@ class ServiceDetails extends React.Component<RouteComponentProps<ServiceId>, Ser
         <div className="page-header">
           <h2>
             Service{' '}
-            <Link to={kialiRoute('/services')} onClick={this.updateFilter}>
+            <Link to="/services" onClick={this.updateFilter}>
```
mtho11: Are we sure hard-coded links to routes are a good idea?

package.json
```diff
@@ -83,6 +83,7 @@
     "npm-snapshot": "1.0.3",
     "prettier": "1.13.4",
     "puppeteer": "1.3.0",
+    "tslint-no-circular-imports": "0.5.0",
```
mtho11: This is new to me, I have not seen this: tslint-no-circular-imports. So I guess a circular import was added and this was needed to fix this bug? Just trying to relate how this is related to the basepath not being set?

src/actions/ActionKeys.ts
josejulio: I think is needed for removing the circular dependencies of the imports

src/pages/ServiceDetails/ServiceDetailsPage.tsx
josejulio: This wasn’t introduced by this PR and the route is relative to the basepath (/console)
Remember that this PR is reverting the workaround introduced by other PR

package.json
josejulio: I think you are missing Edgar’s first comment :)

mtho11: Ah ok @josejulio thanks for the clarification.

aljesusg: Great, I was worried by that, I didn't understand why it wasn't loading the value.

src/actions/ActionKeys.ts
israel-hdez: There is an import cycle that I didn't list in the main description:
<pre>
  utils/Authentication.ts ->
    store/ConfigStore.ts ->
    reducers/index.ts ->
    reducers/ServiceGraphDataState.ts ->
    actions/ServiceGraphDataActions.ts ->
    utils/Authentication.ts
</pre>

The file `reducers/ServiceGraphDataState.ts` depends on `actions/ServiceGraphDataActions.ts` but it only needs `ServiceGraphDataActionKeys` enum. And `actions/ServiceGraphDataActions.ts` has API calls that need the authentication token, thus, the dependency on `utils/Authentication.ts` making the import-cycle. This is the point I chose to break the cycle. By extracting `ServiceGraphDataActionKeys` the cycle is broken.

I extracted all other Action Keys only to make the code uniform. But if you don't like it, I can extract only what is needed.

package.json
lucasponce: Wow, this was a nice catch. I didn't understand the root case as well.
Just curiosity, how was this spotted ? looking into the generated code on a debug session from the browser ?
If there is some setup for this, it can be worth to share.
Thanks !

src/actions/ActionKeys.ts
josejulio: Maybe only split the ServiceGraphDataActionKeys in its own file? 

I feel other way to break this dependency could be to refactor the way utils/Authentication works by using `axios.interceptor` like here:  https://github.com/kiali/kiali-ui/blob/master/src/app/App.tsx#L24
 instead of passing the token to every method.

israel-hdez: Ok, I reverted the refactor. Now it only changes what is needed to fix the cycles.

package.json
israel-hdez: I saw this comment: https://github.com/babel/babel/issues/4094#issuecomment-245410649
So, that made me suspect about a cycle. In order to check order if execution, simple `console.log` statements at the beginning of each file shown that `app/History.tsx` was being run first.
So, no complex setup.

lucasponce: Good catch.
I personally admit I didn't put attention into circular dependencies, for some reason I thought the compiler was already checking it.
Thanks for sharing the details @israel-hdez .

israel-hdez:
@mtho11 Can you re-review and approve if you are ok?

PR #523, opened 2018-07-17
KIALI-1141 Adjust breadcrumbs on Services/Istio Config

lucasponce:
This is a preliminar work to change the provisional `h2` titles and "somehow" breadcrumb feature with the proper patternfly breadcrumb element.
adjust_breadcrumbs ogv

Also, it adjust the breadcrumb messages to clarify user where is located and it doesn't remove any previous feature (clicking to jump into the namespace/type filtering).


src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
```diff
@@ -132,21 +209,44 @@ class IstioConfigDetailsPage extends React.Component<RouteComponentProps<IstioCo
 
   render() {
     const istioRoute = kialiRoute('/istio');
+    const parsedRuleParams = this.parseRuleSearchParams();
+    let titleBreadcrumb: any[] = [];
```
xeviknal: I like to split thinks out and I guess the breadcrumb generation it could be moved into another function. What do you think?

src/pages/ServiceDetails/ServiceDetailsPage.tsx
```diff
@@ -191,22 +195,24 @@ class ServiceDetails extends React.Component<RouteComponentProps<ServiceId>, Ser
     let to = '/namespaces/' + this.props.match.params.namespace + '/services/' + this.props.match.params.service;
     return (
       <>
-        <div className="page-header">
-          <h2>
-            Service{' '}
+        <Breadcrumb title={true}>
```
xeviknal: I think here I would do the same with the breadcrumb generation. Thinking in OO, I would tend to user render as a distributor, placing the layout and calling to internal calls. Object it would be the view, and methods would be: breadcrumb, header, table of whatever, and so on. IMO, this way we can avoid long render methods and will help to have better testing for each view.

abonas:
I think that metrics/info should not be part of the breadcrumbs, only the namespace/service name, etc.
Breadcrumbs should allow navigation back and forth, so in the case here, if a user is standing on "metrics", and then goes "one level back/up" - which page should the user arrive to? It is unclear to me.
@cshinn @serenamarie125 wdyt?
(disclaimer :) - in any case we should redesign navigation as part of https://github.com/kiali/kiali-design/issues/9, so this PR should be a temporary partial improvement)

src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
```diff
@@ -84,6 +99,68 @@ class IstioConfigDetailsPage extends React.Component<RouteComponentProps<IstioCo
...
+    }
+    return false;
+  };
+
+  // Helper method to extract search urls with format
+  // ?handler=name.handlertype or ?instance=name.instancetype
+  // Those url are expected to be received on this page.
+  parseRuleSearchParams = (): ParsedSearch => {
+    let parsed: ParsedSearch = {};
+    if (this.props.location.search) {
```
xeviknal: Just a thought, do you think it would be useful to you the usage of [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) class?

lucasponce:
> I think that metrics/info should not be part of the breadcrumbs, only the namespace/service name, etc.
Breadcrumbs should allow navigation back and forth, so in the case here, if a user is standing on "metrics", and then goes "one level back/up" - which page should the user arrive to? It looks like it navigates to the "info" page of the service, which is represented in the breadcrumbs by both the service itself and then "the next level" "info" too.
So this behavior is unclear to me.

Well, I can see this but the "current" breadcrumb is not only dealing with "levels", as we had introduced "filtering" as part of the "old" behaviour. So, until a new fully design is in place, I would maintain this, at same way we introduced the namespace filtering as part of the "old" breadcrumb.

Anyway, this is just provisional, a intermediate step to discuss about it. It doesn't intend to be the final solution.

src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
lucasponce: Breadcrumb depends on the page, splitting would add a little bit more of complexity in the calls. IMO.

src/pages/ServiceDetails/ServiceDetailsPage.tsx
lucasponce: I am afraid that when per breadcrumb you should add an onClick, splitting will add more complexity, so I decided to placed in the same location where it was the previous `h2` "breadcrumb".

src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
lucasponce: Yeah, it makes sense, I will refactor with that.

abonas:
@lucasponce I think that I am missing what is the fix/improvement in this PR?

lucasponce:
@abonas as commented, before this PR we had this custom-breadcrumb feature:
image
For Services and Istio Config pages (list and details).
Clicking on the "breadcrumb" move the user to the lists with filtering applied per namespace and type.

This PR moves from this "custom breadcrumb" into the patternfly "breadcrumb" component.
Also, it enhances the wording, maintaining the same links used and improving the messages showed.

abonas:
@lucasponce thanks for clarification

src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
xeviknal: Yep, depends on the page, sure. Moving it outside the render method but within the component would increase also the complexity?

cshinn:
I think this new version looks better than the "custom" widget, but agree that the general structure should be re-visited in a different story

PR #524, opened 2018-07-17
KIALI-1130 [UX] ServiceDetails: Introduce a Overview/YAML to show VirtualServices

xeviknal:
This is the next step of [KIALI-1129](https://issues.jboss.org/browse/KIALI-1129).
YAML view and details of virtual services has moved to a fullsize, double-tab area.
Here you have the output:

screen recording 2

Let me note, that when switching tabs this is stored into the browser history.

JIRA: [KIALI-1130](https://issues.jboss.org/browse/KIALI-1129)


xeviknal:
@abonas I think it is a visual effect. Service detail page has the 3 tabs: info/metrics/traces whereas the Virtual Service page has two: Overview and YAML. I think, that both firsts tabs are in the same position and it might gives a confused vision of it, is it possible?

abonas:
@xeviknal exactly :) 
I think that the name should be the same in all of them - either Info in all entity types, or Overview in all entity types. For the sake of consistency. I think Overview is more correct than Info (info = data -> what kind of data?). @jstickler @cshinn @serenamarie125 thoughts? (@xeviknal if this discussion results with changing Info to Overview, then of course it should NOT be part of this PR or block it. it is just easier to discuss it here)

lucasponce:
I started to test/review.
A first comment:
image
There is some problem with the component, when maximizing, the scroll bars are enabled as it doesn't fit in the screen.
I haven't investigated but it sounds like some problem.

xeviknal:
If I understood correctly, this is how the breadcrumb will be:

### Service info:
screenshot of kiali console 4

### Once inside Object details page:
screenshot of kiali console 5
Clicking to 'Virtual Service: reviews' redirects to same page, showing Overview tab.
Clicking to 'Service Info' or 'Service: reviews' redirects to Service Details page / Info tab.

### Once in YAML tab:
screenshot of kiali console 6


WRT 'back to rule' link, where should that point to? Back to overview tab?

lucasponce:
> Clicking to 'Virtual Service: reviews' redirects to same page, showing Overview tab.
> Clicking to 'Service Info' or 'Service: reviews' redirects to Service Details page / Info tab.

+1

> WRT 'back to rule' link, where should that point to? Back to overview tab?

I introduced that in the Rules page to not rely only in the breadcrumb to go back, I guess that "Back to service" could suit here to move up to the Service Info level.

WDYT ?

lucasponce:
Another comment: on this image
image
We rely on the "breadcrumb" to now which object it is, I guess it deserves to put back a "VirtualService" before the name or similar just to know which type is.
If not, when we introduce the same with Destination Rules perhaps it could be confusing.

xeviknal:
Comments addressed, check them out please:

1. Horizontal scroll fixed.
2. Heading added into Istio Object Detail page.
3. Back to service link added above right the ACEditor.

Here the screenshots with the outcome:

screenshot of kiali console 7

screenshot of kiali console 8

lucasponce:
I think that my comment for this picture:
image
was to change "reviews" per "VirtualService: reviews" just to not introduce an `h2` between the breadcrumb and the overview/YAML.
For the YAML, perhaps it can be added after the line, just before the YAML.
In that case, we maintain the suggested design from UXD but we mark the name of the object.

Your approach is also good, too, but I think that is introduced an extra header.

abonas:
why in the breadcrumbs there is VirtualService: review -> VirtualService YAML ?
The review is already a specific one, so the "next" level sounds more general
Perhaps VirtualService:review -> yaml? or at least review:yaml?

lucasponce:
I guess "VirtualService Overview" or "VirtualService YAML" sounds more consistency with the previous "Service Info" / "Service Metrics".

Both options sounds ok to me, if this PR is just for the VirtualServices I would suggest to maintain the same.
But it is a matter of how to decide the criteria, my point is that all should look more or less similar.

xeviknal:
I believe per HTML semantics, having a h1 means having the main title of the page. Adding only the virtual service in the overview tab makes that on the yaml tab you dont know where you are, you lose your reference.

IMO, every page in kiali should have a h1. And specially a single page with few info as we have here. This page is clearly a Virtual Service details page.

lucasponce:
The original recommendation was something like this:
image (from https://github.com/kiali/kiali-design/issues/4).

So, I wouldn't open about the title for this one, as I guess that means more generic discussion not only on this but in all the pages.

xeviknal:
Comment addressed:

screenshot of kiali console 9

xeviknal:
Last refinement:
screenshot of kiali console 11

PR #526, opened 2018-07-17
Istio 1.0

jshaughn:
This is a rebase of istio-1.0 on the latest in master.  There have been no commits to istio-1.0 branches yet and from what I can tell the latest master changes work against the existing 1.0 kiali branch.  We need some of the fixes in master for 1.0.


cfcosta:
This PR moved KIALI-949 back to In Progress, maybe this is something to keep in mind?

josejulio:
> This PR moved KIALI-949 back to In Progress, maybe this is something to keep in mind?

I think it moved every KIALI-### in the commit messages back to in progress because of the jira automation

jshaughn:
@mwringe This is going to force a bunch of jiras back to in progress, did we ever come up with a way to avoid that issue?

If not we should merge it...

mwringe:
I think @theute is the only one who knows what can be done with the jira automation.

jotak:
@jshaughn , I did update istio-1.0 several times already, but not going through PR as it has these unwanted side effects. Since there were no commits, IMO it's ok to push upstream directly.

jotak:
@jshaughn by the way, I think we should proceed with direct upstream push even when there's commits on istio-1.0 - maybe, unless there are conflicts that need to be carefully reviewed.

PR #527, opened 2018-07-18
KIALI-1193 Re-fits the graph on every resize if we are on the "initial" pan/zoom level

josejulio:
This is one of the 3 ways I think we can solve this issue, by doing a `fit` if we are in the "initial" position

If we move away (pan or zoom) of that initial position, it won't do a `fit` again

Other ways to solve this would be to provide an [Empty State](https://www.patternfly.org/pattern-library/communication/empty-state/) side panel that says something like "Load a graph to show its information".

kiali-1193
note: There is "Hello world" on the top right of this gif when the graph is "Loading", this is a quick test of the "Empty State side panel" that forgot to remove previous to recording the gif

### DIFF
--- a/src/pages/ServiceGraph/ServiceGraphPage.tsx
+++ b/src/pages/ServiceGraph/ServiceGraphPage.tsx
@@ -37,7 +37,7 @@ const containerStyle = style({
   height: 'calc(100vh - 60px)' // View height minus top bar height
 });
 
-const cytoscapeGraphContainerStyle = style({ flex: '1', minWidth: '350px', zIndex: 0 });
+const cytoscapeGraphContainerStyle = style({ flex: '1', minWidth: '350px', zIndex: 0, paddingRight: '5px' });
 
 const makeCancelablePromise = (promise: Promise<any>) => {
   let hasCanceled = false;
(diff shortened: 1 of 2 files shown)


jshaughn:
Small comment, I think "fit" should leave a little margin so that the graph doesn't run all the way to a border (like not allowing a node's circle to touch the side-panel rectangle).

src/pages/ServiceGraph/ServiceGraphPage.tsx
```diff
@@ -37,7 +37,7 @@ const containerStyle = style({
   height: 'calc(100vh - 60px)' // View height minus top bar height
 });
 
-const cytoscapeGraphContainerStyle = style({ flex: '1', minWidth: '350px', zIndex: 0 });
+const cytoscapeGraphContainerStyle = style({ flex: '1', minWidth: '350px', zIndex: 0, paddingRight: '5px' });
```
jshaughn: This is good but why only the right side and node all around?

josejulio: There is already a padding/margin on the top, bottom and left

jshaughn:
Should we re-target this to istio-1.0 or leave it for master?

josejulio:
> Should we re-target this to istio-1.0 or leave it for master?

I'm not really sure, there is people working on `master` and `istio-1.0`. I initially targeted `master` because this change is not related to `istio-1.0`. But I'm fine targeting this to `istio-1.0` (and i already did that).

PR #528, opened 2018-07-20
Create PULL_REQUEST_TEMPLATE.md

pilhuhn:


PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,16 @@
+** Describe the change **
+_explantation of what it does_
+
+** Issue reference **
+
+* If this is a fix for a JIRA, please put JIRA-reference in the PR title like _[KIALI-7] New icons_
+* If this is a fix for a GH-issue, please link it here
+
+** Backwards in compatible? **
```
lucasponce: Mmm for the UI the term "backwards incompatibility" it can understand differently for different people (including me).
"Backward compatibility" I guess applies more in the backend, where we define strong API, and after all, it can be checked if an API is different.

For the UI, even if I change someting, I don't have a strong API to check the backward compatibility and I guess it can lead in a lot of cases that it could be or not depending of who triage it.

pilhuhn: You are probably right that this does not make too much sense here.
What about "Change in behaviour? If the UI behaves differently after change, describe the difference". But then I can see this in the description already. 
@JStickler wdyt?

pilhuhn:
@JStickler Yes, the above format with brackets was wrong

PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,17 @@
+** Describe the change **
+_explantation of what it does_
```
josejulio: explanation

PULL_REQUEST_TEMPLATE.md
josejulio: `** Behavior changes **` ? `** Expected changes **`? 
Or maybe just merge this section with the initial `** Describe the change **`

PR #530, opened 2018-07-22
KIALI-1220 Fix link to yaml in DestinationRule

aljesusg:
Fix the link in the Destination Rule.
jul-22-2018 18-26-59

### DIFF
--- a/src/pages/ServiceDetails/IstioObjectDetails.tsx
+++ b/src/pages/ServiceDetails/IstioObjectDetails.tsx
@@ -107,22 +107,34 @@ export default class IstioObjectDetails extends React.Component<IstioObjectDetai
     );
   }
 
+  typeIstioObject() {
+    if ('tcp' in this.props.object && 'http' in this.props.object) {
+      return 'VirtualService';
+    }
+    return 'DestinationRule';
+  }
+
   overviewTab() {
-    const virtualService: VirtualService = this.props.object as VirtualService;
-    return (
-      <TabPane eventKey="overview">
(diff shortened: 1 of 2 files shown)


src/pages/ServiceDetails/IstioObjectDetails.tsx
```diff
@@ -107,22 +107,34 @@ export default class IstioObjectDetails extends React.Component<IstioObjectDetai
...
+  }
+
   overviewTab() {
-    const virtualService: VirtualService = this.props.object as VirtualService;
-    return (
-      <TabPane eventKey="overview">
-        <VirtualServiceDetail virtualService={virtualService} validations={this.props.validations['virtualservice']} />
-      </TabPane>
-    );
+    const istioObj: VirtualService | DestinationRule = this.props.object as VirtualService | DestinationRule;
```
xeviknal: I guess this is not necessary anymore because this.props.object is type of 'VirtualService | DestinationRule' already, so we dont need to do this casting.

aljesusg: This.props.object is not defined, when you try to access to the class or type the result is object not VirtualService...I don't know why

jotak: I pulled your PR locally and it seems to be fine without the `as ...` cast. Maybe your IDE needs a refresh?

aljesusg: Maybe, anyway I talked with @xeviknal about refactor this in another JIRA,anyway the priority is fix the bug Thanks for test it @jotak

PR #531, opened 2018-07-22
KIALI-1201 ServiceDetails: Move DestinationRules into a Table

aljesusg:
Required : #530 [KIALI-1220](https://issues.jboss.org/browse/KIALI-1220)

JIRA: [KIALI-1201](https://issues.jboss.org/browse/KIALI-1201)

### Before
captura de pantalla 2018-07-22 a las 18 36 10

### After

captura de pantalla 2018-07-22 a las 18 48 23

### DIFF
--- a/src/pages/ServiceDetails/ServiceInfo/ServiceInfoDestinationRules.tsx
+++ b/src/pages/ServiceDetails/ServiceInfo/ServiceInfoDestinationRules.tsx
@@ -1,7 +1,9 @@
 import * as React from 'react';
-import { DestinationRule, EditorLink } from '../../../types/ServiceInfo';
-import { Col, Row, Icon } from 'patternfly-react';
+import { DestinationRule, EditorLink, Subset } from '../../../types/ServiceInfo';
+import { Col, Row, Table } from 'patternfly-react';
+import * as resolve from 'table-resolver';
 import LocalTime from '../../../components/Time/LocalTime';
+import Label from '../../../components/Label/Label';
 import DetailObject from '../../../components/Details/DetailObject';
 import { Link } from 'react-router-dom';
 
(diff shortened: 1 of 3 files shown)


pilhuhn:
I think I would move the created at/source version to the right and the 
policy/subsets in their current place, as they are much more important 
than especially the resource version.

aljesusg:
Moved, thanks @pilhuhn

xeviknal:
@aljesusg I guess this table is an intermediate step to move DR's details to another page, right? Otherwise, I would remove subsets and traffic policy from the table.
However, there is this row title repetition '[host]' or 'subsets' that I think there are redundant. What do you think?

aljesusg:
Changed target to istio-1.0
Removed redundant.
screenshot from 2018-07-30 09-13-32

abonas:
@kiali/ux-reviewers fyi. any comments?

cshinn:
@aljesusg @abonas do subsets ever have any traits besides a name and a set of labels? If not, I think it might be clearer (and more space efficient) to remove the word "labels" put the name inline with any associated labels (wrapping to a second line as needed) 

Maybe something like this

group 14

abonas:
@cshinn yes, subsets might have also trafficPolicy optional section.
The structure is as follows: name and labels are required (there could be more than 1 label) and trafficPolicy is optional - if it exists, it overrides a "higher level" traffic policy
https://istio.io/docs/reference/config/istio.networking.v1alpha3/#Subset

@aljesusg Why is the name of the subset appearing after the labels and with [] brackets - is that intentional? 
I'd also make it bold and reverse the order of labels and name, unless, again there is some logic for this particular order that I've missed.

aljesusg:
Yes @abonas it's heredate from the component and how we had the information, we can change it.

aljesusg:
Changed @cshinn  @abonas . thanks
screenshot from 2018-07-31 11-33-45

abonas:
It seems that the names are removed now? Is this intentional?

aljesusg:
yes @abonas because it has no sense I think maybe @xeviknal or @jotak can tell us why is name field in subsets
```json
subsets: [
  {
     labels: {
        version:"v1"
     },
     name: "v1"
   },
   {
     labels: {
         version:"v2"
     },
     name: "v2"
    },
...
]
```

abonas:
@aljesusg this information is coming from Istio, so I don't think it should be removed in Kiali. I think the reason for naming the subsets is to help distinguishing them more easily.

aljesusg:
Ok.Thanks @abonas Changed it
screenshot from 2018-07-31 13-08-46

abonas:
@aljesusg great :)
Do you have an example how it would look with trafficPolicy? (an optional field for a subset)

abonas:
@aljesusg isn't this an optional property of each subset separately?

jotak:
@aljesusg the git history seems wrong here. Is it supposed to go in master branch or in istio-1.0 ? There's a lot of commits in that PR that are obviously not related to your task. Can you fix it? (I think the simplest would be to reset to upstream master and cherry-pick your commit.)

jotak:
@aljesusg  @abonas for what I can see, traffic policy can be set both at destination rule's root and in subsets. So I guess on the screenshot above, it is set at rule's root, not on subset.
How does it look like with a rule such as:

```
kind: DestinationRule
metadata:
  name: bookinfo-ratings
spec:
  host: ratings.prod.svc.cluster.local
  trafficPolicy:
    loadBalancer:
      simple: LEAST_CONN
  subsets:
  - name: testversion
    labels:
      version: v3
    trafficPolicy:
      loadBalancer:
        simple: ROUND_ROBIN
```

(taken from istio doc https://istio.io/docs/reference/config/istio.networking.v1alpha3/ )

aljesusg:
Fixed @jotak it seems that I get the last commits when I changed the base branch. Thanks !!!

aljesusg:
@jotak Is the backend showing the traffic policy of each subset or we must do it?

jotak:
@aljesusg I have no idea, it's more @xeviknal 's part I guess :)

src/pages/ServiceDetails/ServiceInfo/ServiceInfoDestinationRules.tsx
```diff
@@ -14,42 +16,155 @@ class ServiceInfoDestinationRules extends React.Component<ServiceInfoDestination
...
+  generateKey() {
+    return (
+      'key_' +
+      Math.random()
+        .toString(36)
+        .substr(2, 9)
+    );
+  }
+
+  generateSubsets(subsets: Subset[]) {
```
xeviknal: Here, as @jotak mention, is missing the displaying of a trafficPolicy for each subset.
If you see in the definition of [Subset](https://istio.io/docs/reference/config/istio.networking.v1alpha3/#Subset), it can option have a trafficPolicy.
For adding it, I would use the `<DetailObject name="trafficPolicy" detail={subset.trafficPolicy} />`.

PR #534, opened 2018-07-24
KIALI-1173 Refreshing ServiceList, IstioConfig misses the filtering, sorting and pagination

xeviknal:
Tracking filters into URL from service list:
screen recording 4

Tracking object tabs from Service Details page:
screen recording 5


xeviknal:
Both Service List and IstioConfig list share a lot of behaviour regarding the management of the filters, sort params and so on. I think, that for the next listing page (thinking of Workloads list), we should try to abstract somehow that code for reduction of code maintenance.

xeviknal:
Thank you @jotak for the review :)
Good guide to follow. I did tried to follow that. I see, I missed some ones, though :)

abonas:
LGTM from functional POV, didn't dive into the code. imo can be merged.

xeviknal:
@jotak and @abonas I realized that both list of items aren't well setup to refresh their content after 'going back in browser history'. So you can play around with the filters and the URL is changed on each action but when you are going to previous page, URL is changed but the component is not updated (doesn't update the services/istio objects inside).

After some research, I think that components are no updated because any of the component prop has been change (since we only send functions).
Is that okay to you if I put that as a bug instead of placing here the solution. I guess that behavior was happening before and the JIRA expectations are satisfied.

abonas:
@xeviknal I discussed the gif with @serenamarie125 and @cshinn .  
There should *not* be a multiple selection on service name, so the correct behavior should be:
typing the second service should clear the first one selected.
(there's also no strong use case to support multiple services selection at the moment)

xeviknal:
@abonas super. I understand that it should be exactly the same with the other filters (service name, namespace, istio sidecar), right?
In terms of implementation, I would take this into another JIRA since the scope of this one was only about persist the filters/sorting/pagination selected into the URL to let the user share the page properly: [JIRA-1173](https://issues.jboss.org/browse/KIALI-1173)

aljesusg:
Anyway we'll have the same problem that @xeviknal reported, when you change a param this call to OnParamChange that make a push to the history and after call to updateFilters.

When you refresh the browser this history change but the updateFilters is not called so the result is the same.

To make the changes that @abonas proposed you need to change the method filterAdded called in onValueKeyPress in form and filterValueSelected in the filter

```typescript
if (currentFilterType.filterType === 'select') {
      return (
        <Filter.ValueSelector
          filterValues={currentFilterType.filterValues}
          placeholder={currentFilterType.placeholder}
          currentValue={currentValue}
          onFilterValueSelected={this.filterValueSelected}
        />
      );
    } else {
      return (
        <FormControl
          type={currentFilterType.filterType}
          value={currentValue}
          placeholder={currentFilterType.placeholder}
          onChange={e => this.updateCurrentValue(e)}
          onKeyPress={e => this.onValueKeyPress(e)}
        />
      );
    }
```


So you need to change in that method this in https://github.com/kiali/kiali-ui/blob/master/src/components/NamespaceFilter/NamespaceFilter.tsx#L96-L100 

```typescript

    activeFilter.label = filterText;
    activeFilters.push(activeFilter);
    this.setState({ activeFilters: activeFilters });
    NamespaceFilterSelected.setSelected(activeFilters);
    this.props.onFilterChange(activeFilters);
```


by 

```typescript
    activeFilter.label = filterText;
    /** Check if Filter already exists */
    let filters = activeFilters.filter(filter => filter.category === activeFilter.category);
    if (filters.length > 0) {
      filters[0].label = activeFilter.label;
      filters[0].value = activeFilter.value;
    } else {
      activeFilters.push(activeFilter);
    }

    this.setState({ activeFilters: activeFilters });
    NamespaceFilterSelected.setSelected(activeFilters);
    this.props.onFilterChange(activeFilters);
```

peek 2018-07-30 13-29

xeviknal:
[Going back and forward in the browser history got fixed](https://github.com/kiali/kiali-ui/pull/534#issuecomment-407787911) :)

screen recording

src/components/ListPage/ListPage.ts
```diff
@@ -0,0 +1,60 @@
+import * as React from 'react';
+import { RouteComponentProps } from 'react-router';
+import * as MessageCenter from '../../utils/MessageCenter';
+
+export namespace ListPage {
+  export interface URLParameter {
+    name: string;
+    value: string;
+  }
+
```
aljesusg: This should be in the types folder, maybe we'll need this in another views. What do you think?

PR #536, opened 2018-07-24
KIALI-1250 Fixes test errors

josejulio:
Made some changes to the Enum utility and fixed a bug when searching the graphType in the Enum. This affected the EdgeLabelMode

kiali-1250

### DIFF
--- a/src/pages/ServiceGraph/ServiceGraphRouteHandler.tsx
+++ b/src/pages/ServiceGraph/ServiceGraphRouteHandler.tsx
@@ -53,10 +53,9 @@ export default class ServiceGraphRouteHandler extends React.Component<
       urlParams.get('edges'),
       ServiceGraphRouteHandler.graphParamsDefaults.edgeLabelMode
     );
-    const urlParamGraphType = urlParams.get('graphType') ? urlParams.get('graphType').toUpperCase() : '';
     const _graphType = Enum.fromValue(
       GraphType,
-      urlParamGraphType,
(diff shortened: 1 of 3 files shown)


src/pages/ServiceGraph/ServiceGraphRouteHandler.tsx
```diff
@@ -53,7 +53,7 @@ export default class ServiceGraphRouteHandler extends React.Component<
       urlParams.get('edges'),
       ServiceGraphRouteHandler.graphParamsDefaults.edgeLabelMode
     );
-    const urlParamGraphType = urlParams.get('graphType') ? urlParams.get('graphType').toUpperCase() : '';
+    const urlParamGraphType = urlParams.get('graphType') ? urlParams.get('graphType') : '';
```
jmazzitelli: why do you perform this ?: check but not for edge label mode? _edgeLabelMode is the same pattern (uses Enum.forValue, gets the urlParams.get('edges') but it does not perform this same check. I think we should be consistent - either do both the way _edgeLabelMode is done, or both the way _graphType is done.

josejulio: It was required when doing the `.toUpperCase()`, now is no longer required, deleting...

josejulio: Removed and now i'm using directly the urlParams.get('graphType')

PR #537, opened 2018-07-24
KIALI-1228 - use graph type to determine node name

jmazzitelli:


jmazzitelli:
Wait for PR #536 before merging this

src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -147,15 +148,11 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
...
-      edge.data('edgeLabelMode', value);
-    });
+    cy.scratch(CytoscapeGlobalScratchNamespace).edgeLabelMode = value;
   };
 
   private turnNodeLabelsTo = (cy: any, value: boolean) => {
-    cy.nodes().forEach(node => {
-      node.data('showNodeLabels', value);
-    });
+    cy.scratch(CytoscapeGlobalScratchNamespace).showNodeLabels = value;
```
josejulio: I was going to say that we don't need this one anymore, but then i recall there was a bug, that the fix was to turn this on during a layout and set it to the desired value after all the layout magic is done.

src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -147,15 +148,11 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
   };
 
   private turnEdgeLabelsTo = (cy: any, value: EdgeLabelMode) => {
-    cy.edges().forEach(edge => {
-      edge.data('edgeLabelMode', value);
-    });
+    cy.scratch(CytoscapeGlobalScratchNamespace).edgeLabelMode = value;
```
josejulio: We probably don't need this one anymore

src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -289,6 +286,15 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
       this.resetSelection = false;
     }
 
+    const globalScratchData: CytoscapeGlobalScratchData = {
+      edgeLabelMode: this.props.edgeLabelMode,
+      graphType: this.props.graphType,
+      showNodeLabels: this.props.showNodeLabels,
+      versioned: this.props.versioned
+    };
+    console.debug('cy global scratch: ' + JSON.stringify(globalScratchData));
```
josejulio: Is the `console.debug` left here on purpose?

src/components/CytoscapeGraph/graphs/GraphStyles.ts
```diff
@@ -2,6 +2,7 @@ import { PfColors } from '../../../components/Pf/PfColors';
 import { EdgeLabelMode } from '../../../types/GraphFilter';
 import { config } from '../../../config';
 import { FAILURE, DEGRADED } from '../../../utils/Health';
+import { GraphType, CytoscapeGlobalScratchNamespace, CytoscapeGlobalScratchData } from '../../../types/Graph';
```
josejulio: not related to this PR, but I hate these relative `../../../../../../paths/to/files`

src/components/CytoscapeGraph/CytoscapeGraph.tsx
jmazzitelli: this is called on line 330 - we need this to let GraphStyles know how to render the edge label -- see GraphStyles line 142

src/components/CytoscapeGraph/CytoscapeGraph.tsx
jmazzitelli: we still need this because GraphStyles now uses this to know if node labels should be on or off.

src/components/CytoscapeGraph/CytoscapeGraph.tsx
jmazzitelli: I actually had this to debug this problem we were having in the Enum stuff :) I will remove

src/components/CytoscapeGraph/CytoscapeGraph.tsx
josejulio: Yes, but is already set on line 290 with exactly the same value (`this.props.edgeLabelMode`)

src/components/CytoscapeGraph/CytoscapeGraph.tsx
josejulio: damn Enum stuff!

josejulio:
This is missing the group name (reviews):

group-name

josejulio:
It looks like a server problem, here is the related json for that compound node.
```json
{
        "data": {
          "id": "6320efe386bc4fbe51ee63f53222b153",
          "namespace": "bookinfo",
          "workload": "",
          "app": "",
          "service": "reviews.bookinfo",
          "serviceName": "reviews",
          "isGroup": "version"
        }
      },
```

jmazzitelli:
I see the box label - @josejulio you might have an old server

josejulio:
After updating to latest server I can see the box label
group-name-after-updating

PR #538, opened 2018-07-24
fix the if-stmt when not workload graph

jmazzitelli:
@jshaughn this fixes some of those unknown nodes (most of them anyway). I still see one odd one that I think is this one:

```
      {
        "data": {
          "id": "ad921d60486366258809553a3db49a4a",
          "namespace": "istio-system",
          "workload": "unknown",
          "app": "unknown",
          "version": "unknown",
          "service": "unknown",
          "serviceName": "istio-policy",
          "hasMissingSC": true,
          "isGroup": "version"
        }
      }
```

### DIFF
--- a/src/components/CytoscapeGraph/graphs/GraphStyles.ts
+++ b/src/components/CytoscapeGraph/graphs/GraphStyles.ts
@@ -72,7 +72,7 @@ export class GraphStyles {
 
             // use the workload name unless app label was defined
             let content = workload;
-            if (getCyGlobalData(ele).graphType !== GraphType.WORKLOAD && (app || app !== 'unknown')) {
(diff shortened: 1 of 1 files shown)


src/components/CytoscapeGraph/graphs/GraphStyles.ts
```diff
@@ -72,7 +72,7 @@ export class GraphStyles {
 
             // use the workload name unless app label was defined
             let content = workload;
-            if (getCyGlobalData(ele).graphType !== GraphType.WORKLOAD && (app || app !== 'unknown')) {
+            if (getCyGlobalData(ele).graphType !== GraphType.WORKLOAD && app && app !== 'unknown') {
```
jshaughn: This is better but I'm not sure it's correct.  I'll still merge.  But if graphType is APP then I think you should use 'app' in all cases.

PR #540, opened 2018-07-25
KIALI-1246 Remove all usages of cyNode.data('service')

josejulio:
UI is untouched by this PR. 

Most of this is about the sidebar.
I only replaced all the uses of `node.data('service')` by `node.data('app')` (or by `node.data('namespace')` where the actual intend was to get to parse the namespace).

I'm not yet using `workload` anywhere in the sidebar, but we could potentially add that term there 
```
 app: [link to app]
 workload: [link to workload?]
```

Also, still using the term "service":

kiali-1246-2
kiali-1246-1
kiali-1246-0

### DIFF
--- a/src/components/CytoscapeGraph/CytoscapeGraph.tsx
+++ b/src/components/CytoscapeGraph/CytoscapeGraph.tsx
@@ -20,7 +20,8 @@ import {
   CytoscapeMouseOutEvent,
   GraphParamsType,
   CytoscapeGlobalScratchNamespace,
-  CytoscapeGlobalScratchData
+  CytoscapeGlobalScratchData,
+  GraphType
 } from '../../types/Graph';
 import { EdgeLabelMode } from '../../types/GraphFilter';
 import { authentication } from '../../utils/Authentication';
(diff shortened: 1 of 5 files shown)


josejulio:
I'm still checking if i'm doing right the health queries

jmazzitelli:
we'll definitely need to add code to support "workloads"  but I think that can come later in another PR because I think we need to do add server-side APIs to get workload-specific data.

jmazzitelli:
@josejulio do you think we can merge this now? I didn't run it, so I can't say if its working - but I'll blame...er... trust you if you merge it. ;)

src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -456,34 +457,38 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
...
-          // Ignore health for special nodes such as "unknown"
-          return;
-        }
-        const service = serviceParts[0];
-        const namespace = serviceParts[1];
+      const namespace = ele.data('namespace');
+      const app = ele.data('app');
+      const workload = ele.data('workload');
+
+      if ((workload !== 'unknown' || app !== 'unknown') && (ele.data('isGroup') || !ele.data('parent'))) {
```
jotak: It looks correct but I think it's still need some changes later on. `API.getNamespaceHealth` will have to get a parameter (or be renamed) so that we know if we're fetching app-health, service-health or workload-health for the whole namespace. But that will come later.

PR #541, opened 2018-07-25
KIALI-1231 Refactors panel to handle metrics by app or workload

josejulio:
Refactored the logic to handle the metrics by app(versioned or not) and workload.

Currently it shows the sparklines in the app (versioned and not versioned) when the label app is set. It doesn't (yet) handle when we need to access the workload metrics. 

I'm looking at the server code to create that service or use one if available.

TODO:
- [x] Check old label (source_service)

This requires: https://github.com/kiali/kiali/pull/385

### DIFF
--- a/src/pages/ServiceGraph/SummaryPanelEdge.tsx
+++ b/src/pages/ServiceGraph/SummaryPanelEdge.tsx
@@ -3,13 +3,18 @@ import RateTable from '../../components/SummaryPanel/RateTable';
 import RpsChart from '../../components/SummaryPanel/RpsChart';
 import ResponseTimeChart from '../../components/SummaryPanel/ResponseTimeChart';
 import { NodeType, SummaryPanelPropType } from '../../types/Graph';
-import * as API from '../../services/Api';
 import * as M from '../../types/Metrics';
 import graphUtils from '../../utils/Graphing';
-import MetricsOptions from '../../types/MetricsOptions';
-import { authentication } from '../../utils/Authentication';
-import { shouldRefreshData, nodeData, getServicesLinkList } from './SummaryPanelCommon';
+import {
+  shouldRefreshData,
+  nodeData,
+  getServicesLinkList,
+  getNodeMetrics,
+  NodeMetricType,
+  getNodeMetricType
+} from './SummaryPanelCommon';
 import Label from '../../components/Label/Label';
+import { MetricGroup, Metric } from '../../types/Metrics';
 
 type SummaryPanelEdgeState = {
   loading: boolean;
@@ -118,21 +123,54 @@ export default class SummaryPanelEdge extends React.Component<SummaryPanelPropTy
     );
   }
 
+  private getByLabelsIn = (nodeMetricType: NodeMetricType) => {
+    switch (nodeMetricType) {
+      case NodeMetricType.WORKLOAD:
+        return ['source_workload'];
+      case NodeMetricType.APP:
+        return ['source_app'];
+      default:
+        // Unreachable code, but tslint disagrees
+        // https://github.com/palantir/tslint/issues/696
+        throw new Error(`Unknown NodeMetricType: ${nodeMetricType}`);
+    }
+  };
+
+  private getNodeDataPoints = (m: MetricGroup, title: string, nodeMetricType: NodeMetricType, node: any) => {
+    const data = nodeData(node);
+    let comparator;
+    switch (nodeMetricType) {
+      case NodeMetricType.APP:
+        comparator = (metric: Metric) => {
+          return metric['source_app'] === data.app;
+        };
+        break;
+      case NodeMetricType.WORKLOAD:
+        comparator = (metric: Metric) => {
+          return metric['source_workload'] === data.workload;
+        };
+        break;
+      default:
+        // Unreachable code, but tslint disagrees
+        // https://github.com/palantir/tslint/issues/696
+        throw new Error(`Unknown NodeMetricType: ${nodeMetricType}`);
+    }
+    return this.getDatapoints(m, title, comparator);
+  };
+
   private updateCharts = (props: SummaryPanelPropType) => {
     const edge = props.data.summaryTarget;
-    const source = nodeData(edge.source());
-    const dest = nodeData(edge.target());
+    const source = edge.source();
+    const nodeMetricType = getNodeMetricType(source);
 
-    const options: MetricsOptions = {
-      version: dest.version,
-      byLabelsIn: ['source_service', 'source_version'],
-      queryTime: props.queryTime,
-      duration: +props.duration,
-      step: props.step,
-      rateInterval: props.rateInterval,
-      filters: ['request_count', 'request_duration', 'request_error_count']
-    };
-    API.getServiceMetrics(authentication(), dest.namespace, dest.app, options)
+    if (!nodeMetricType) {
+      return;
+    }
+
+    const filters = ['request_count', 'request_duration', 'request_error_count'];
+    const byLabelsIn = this.getByLabelsIn(nodeMetricType);
+
+    getNodeMetrics(nodeMetricType, edge.target(), props, filters, undefined, byLabelsIn)
       .then(response => {
         if (!this._isMounted) {
           console.log('SummaryPanelEdge: Ignore fetch, component not mounted.');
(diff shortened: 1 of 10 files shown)


src/pages/ServiceGraph/SummaryPanelEdge.tsx
```diff
@@ -118,20 +115,13 @@ export default class SummaryPanelEdge extends React.Component<SummaryPanelPropTy
   }
 
   private updateCharts = (props: SummaryPanelPropType) => {
+    const filters = ['request_count', 'request_duration', 'request_error_count'];
+    const byLabelsIn = ['source_service', 'source_version'];
```
jmazzitelli: I do not know what these are used for, however, if this is somehow related to the Prometheus labels (they look like them at least), there is no such thing as "source_service" label anymore.

josejulio: Thanks for the heads up. I have no idea either but is sent to `api/namespace/{namespace}/service/{service}`. I'll check what is that used for.

jshaughn:
@josejulio You should work with @jotak to figure out how to call the metrics API, he has been working on the changes and will know what is there, what is coming, and what is missing.

src/pages/ServiceGraph/SummaryPanelEdge.tsx
```diff
@@ -117,54 +115,59 @@ export default class SummaryPanelEdge extends React.Component<SummaryPanelPropTy
...
+  private getByLabelsIn = (node: any) => {
+    const globalData = nodeGlobalData(node);
+    const data = nodeData(node);
+    if (globalData.graphType === GraphType.APP && data.app !== 'unknown') {
+      if (globalData.versioned) {
+        return ['source_app', 'source_version'];
+      }
+      return ['source_app'];
+    } else {
+      return ['source_workload'];
```
josejulio: @jmazzitelli Can you please check if these labels make sense?

josejulio:
@jmazzitelli Did what you said in your review

josejulio:
@jotak @jshaughn please test

src/pages/ServiceGraph/SummaryPanelCommon.ts
```diff
@@ -29,11 +44,48 @@ export const updateHealth = (summaryTarget: any, stateSetter: (hs: HealthState)
...
+  node: any,
+  props: SummaryPanelPropType,
+  filters: Array<string>,
+  includeIstio?: boolean,
+  byLabelsIn?: Array<string>
+) => {
+  // Determine if is by workload or by app
+  const data = nodeData(node);
+  const globalData = nodeGlobalData(node);
+  const isByApp = globalData.graphType === GraphType.APP && data.app !== 'unknown';
```
jotak: Are you sure about this check? I have a similar check to perform for health, see on my PR: https://github.com/kiali/kiali-ui/pull/546/files#diff-04b15b4602d6b9ccf1eb82d15e07535fR485
If an app box is selected, in workload-based graph, I guess it should show app-based metrics.

src/pages/ServiceGraph/SummaryPanelCommon.ts
```diff
@@ -29,11 +44,48 @@ export const updateHealth = (summaryTarget: any, stateSetter: (hs: HealthState)
...
+    duration: props.duration,
+    step: props.step,
+    rateInterval: props.rateInterval,
+    includeIstio: includeIstio,
+    filters: filters,
+    byLabelsIn: byLabelsIn
+  };
+
+  if (globalData.versioned && data.version && data.version !== 'unknown') {
+    options.version = data.version;
```
jotak: I thought I removed options.version (maybe only on backend, and it wasn't changed on frontend?)
Basically the idea is that a "versioned app" is actually a workload.

src/pages/ServiceGraph/SummaryPanelEdge.tsx
```diff
@@ -117,54 +115,61 @@ export default class SummaryPanelEdge extends React.Component<SummaryPanelPropTy
...
+      }
+    } else {
+      return ['source_workload'];
+    }
+  };
+
+  private getNodeDataPoints = (m: MetricGroup, title: string, node: any) => {
+    const globalData = nodeGlobalData(node);
+    const data = nodeData(node);
+    if (globalData.graphType === GraphType.APP && data.app !== 'unknown') {
```
jotak: Same as above, I'm not sure about that

src/types/Graph.ts
```diff
@@ -5,7 +5,7 @@ import Namespace from './Namespace';
 //   summaryTarget: The cytoscape element
 //   summaryType  : one of 'graph', 'node', 'edge', 'group'
 export interface SummaryData {
-  summaryType: string;
+  summaryType: 'graph' | 'node' | 'edge' | 'group';
```
jotak: nice

src/pages/ServiceGraph/SummaryPanelCommon.ts
josejulio: > I thought I removed options.version (maybe only on backend, and it wasn't changed on frontend?)

Right, it wasn't changed in frontend and i think this is causing this one https://issues.jboss.org/browse/KIALI-1276



> A "versioned app" is not a workload - in the "versioned app" graph, for example, we don't show "workload name" in the node, we should "app" label with "version" label. Also, app/version pairs can span workloads (I can have workload "A-1" and "B-2" both labeled with "appA-versionV").

That's what i understood

src/pages/ServiceGraph/SummaryPanelCommon.ts
josejulio: I think there isn't any "app box" in workload-based graph (at least not yet).
If there is plans for one I could update this check.

src/pages/ServiceGraph/SummaryPanelCommon.ts
jotak: Ok, I talked with mazz and jay today, this needs a little more clarification and I asked this on istio-dev: https://groups.google.com/forum/#!topic/istio-dev/Ll9AxJjZYaY

My initial assumption was that the app-based graph was workload-based under the hood, then aggregated by app. But if "version" still plays a role then, you're probably right and we'll have to put "version" back in the query params server-side.

src/pages/ServiceGraph/SummaryPanelCommon.ts
jotak: Fair enough, this is still because I had wrong assumptions on how the graph was generated.

josejulio:
Rebased and updated to use the new nodeType attribute.
@jotak @jmazzitelli please review

src/pages/ServiceGraph/SummaryPanelCommon.ts
```diff
@@ -29,11 +46,56 @@ export const updateHealth = (summaryTarget: any, stateSetter: (hs: HealthState)
...
 };
+
+export const getNodeMetricType = (node: any) => {
+  const data = nodeData(node);
+  if (data.nodeType === NodeType.WORKLOAD || (data.nodeType === NodeType.APP && data.hasParent)) {
+    return NodeMetricType.WORKLOAD;
+  } else if (data.nodeType === NodeType.APP) {
+    return NodeMetricType.APP;
+  }
+  return null;
```
jotak: iirc we want to favor "undefined" over "null", or did the rule change?

josejulio: I don't think the rule changed, I'm going to update to use "undefined"

jotak:
LGTM again (retested after master rebase)

pilhuhn:
On 7 Aug 2018, at 17:06, Josejulio Martínez wrote:

> I don't think the rule changed, I'm going to update to use "undefined"

If we have those rules - are they documented somewhere?
That would help folks look them up if uncertain.

src/pages/ServiceGraph/SummaryPanelCommon.ts
```diff
@@ -1,5 +1,22 @@
...
+  app: string;
+  version: string;
+  workload: string;
+  nodeType: NodeType;
+  hasParent: boolean;
+}
+
+export enum NodeMetricType {
+  APP = 1,
+  WORKLOAD = 2
```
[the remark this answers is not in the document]
josejulio: Is to quickly decide how i'm going to load the metrics without doing the check everywhere.
I do the check once and use it everywhere i need to know if load the metrics using the workload API or the APP api

https://github.com/kiali/kiali-ui/pull/541/files#diff-feb27b2764c1979765c68d3ca55c1e55R62

josejulio:
> If we have those rules - are they documented somewhere?
> That would help folks look them up if uncertain.

mmh, I can't find it, @mtho11 do you know?

PR #543, opened 2018-07-26
Use npm instead of yarn for the build process

cfcosta:
This is necessary to remove external dependencies from the build process when `node_modules` is vendored on the repository.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,14 +47,14 @@
     "prettier": "prettier --write \"src/**/*.{js,jsx,ts,tsx,json}\"",
     "lint": "tslint --project ./tsconfig.json",
     "lintfix": "tslint --fix --project ./tsconfig.json",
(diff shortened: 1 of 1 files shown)


josejulio:
This looks good, I'm going to target it to branch `istio-1.0`

josejulio:
ugh, can you rebase to `istio-1.0` ?

mtho11:
@cfcosta can you explain a bit more why this was necessary? I don't understand from the comments above.

mtho11:
Is there are Jira for this?

cfcosta:
Sure, for productization we only have access to stuff that’s available on rpm (the dependencies are vendored beforehand), so I did some testing and saw that the result is the same for the build scripts, because the dependencies are already installed beforehand.

Is this causing any problems to you?

Cheers,
Cainã Costa
http://www.cfcosta.com

mtho11:
No problem. I just wanted to understand the problem that it was addressing.

PR #544, opened 2018-07-27
KIALI-1100 Show charts for TCP metrics

jotak:
api-server
Screenshot above shows an application workload sending queries to mongodb (TCP Sent) and receiving responses (TCP Received). The charts are in the "outgoing" column, as the application workload is making the queries.

mongo
This screenshot shows the related mongodb workload that receives the queries (TCP Received) and sending back responses (TCP Sent). The charts are in the "incoming" column, as mongo is server-side.

### DIFF
--- a/src/pages/ServiceDetails/ServiceMetrics.tsx
+++ b/src/pages/ServiceDetails/ServiceMetrics.tsx
@@ -34,6 +34,10 @@ type ServiceMetricsState = {
   requestDurationOut?: Histogram;
   responseSizeIn?: Histogram;
   responseSizeOut?: Histogram;
+  tcpReceivedIn?: MetricGroup;
+  tcpReceivedOut?: MetricGroup;
+  tcpSentIn?: MetricGroup;
+  tcpSentOut?: MetricGroup;
   grafanaLinkIn?: string;
   grafanaLinkOut?: string;
   pollMetrics?: number;
@@ -44,10 +48,14 @@ const chartDefinitions = {
   requestDurationIn: { familyName: 'Request duration (seconds)', isInput: true, component: HistogramChart },
   requestSizeIn: { familyName: 'Request size (bytes)', isInput: true, component: HistogramChart },
   responseSizeIn: { familyName: 'Response size (bytes)', isInput: true, component: HistogramChart },
+  tcpReceivedIn: { familyName: 'TCP received (bps)', isInput: true, component: MetricChart },
+  tcpSentIn: { familyName: 'TCP sent (bps)', isInput: true, component: MetricChart },
   requestCountOut: { familyName: 'Request volume (ops)', isInput: false, component: MetricChart },
   requestDurationOut: { familyName: 'Request duration (seconds)', isInput: false, component: HistogramChart },
   requestSizeOut: { familyName: 'Request size (bytes)', isInput: false, component: HistogramChart },
-  responseSizeOut: { familyName: 'Response size (bytes)', isInput: false, component: HistogramChart }
+  responseSizeOut: { familyName: 'Response size (bytes)', isInput: false, component: HistogramChart },
+  tcpSentOut: { familyName: 'TCP sent (bps)', isInput: false, component: MetricChart },
+  tcpReceivedOut: { familyName: 'TCP received (bps)', isInput: false, component: MetricChart }
 };
 
 class ServiceMetrics extends React.Component<ServiceId, ServiceMetricsState> {
@@ -70,7 +78,14 @@ class ServiceMetrics extends React.Component<ServiceId, ServiceMetricsState> {
 
   onOptionsChanged = (options: MetricsOptions) => {
     this.options = options;
-    options.filters = ['request_count', 'request_size', 'request_duration', 'response_size'];
+    options.filters = [
+      'request_count',
+      'request_size',
+      'request_duration',
+      'response_size',
+      'tcp_received',
+      'tcp_sent'
+    ];
     const intervalOpts = computePrometheusQueryInterval(options.duration!);
     options.step = intervalOpts.step;
     options.rateInterval = intervalOpts.rateInterval;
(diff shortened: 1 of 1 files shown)


jotak:
@vnugent as you opened the initial JIRA about it, you may want to have a look...

jotak:
@abonas :
**About the order:**
Yes it is on purpose as this order reflects the sequence of the events:
- For incoming metrics, first service A receives a TCP request (received), then it responds (sent)
- For outgoing, first service A sends a TCP request (sent), then it receives the response (received)
For easier understanding we may consider renaming the column labels , see suggestion below about naming.

**About the number of remote-end services:**
Here it's the same as for the other metrics: the page displays the aggregated metrics of all remote-end services, regardless their number, but in the options it is possible to refine grouping and see for instance one timeseries per related service.

**About naming:**
The names used in code take the same conventions as for other metrics: say the metrics are focusing on service A, "xxxIn" are metrics from any source service to destination service A, "xxxOut" for metrics where service A is the source and any other is the destination.

This is strictly related to the semantics in Prometheus labels and I admit it can be confusing because having "tcpReceivedOut" seems contradictory at first (if something is "received", then it doesn't seem natural that it goes "out") , however it does actually mean something, it corresponds to the TCP bytes responded consecutive to a client request.

Maybe in this context it is easier to understand "xxxIn" as metrics when service A is the server, and "xxxOut" as metrics when service A is the client. We could eventually rename the top labels "Incoming", "Outgoing" on the metrics page to "Server metrics" and "Client metrics".

abonas:
thanks @jotak for the detailed explanation.
I think that naming server and client can also be misleading, because if we take the example of 'A calls B calls C', then B can be both a client and a server.

jotak:
@abonas
> if we take the example of 'A calls B calls C', then B can be both a client and a server.

Indeed and we would have all 4 TCP metrics for B then, so I think it makes sense. We would have:
- [Server] TCP received from "A call B" request.
- [Server] TCP sent from "A call B" response.
- [Client] TCP sent from "B call C" request.
- [Client] TCP received from "B call C" response.

abonas:
@jotak don't know what to say here, client/server in microservices world sounds like not the best option. maybe it'd be best to hear more opinion on this.

jmazzitelli:
@jotak @abonas - regarding joel's comment https://github.com/kiali/kiali-ui/pull/544#issuecomment-408768480 I would want to chime in and say it is incorrect to use the word "service" in that description. For example:

```
For incoming metrics, first service A receives a TCP request....
For outgoing, first service A sends a TCP request...
The page displays the aggregated metrics of all remote-end services...
...say the metrics are focusing on service A, "xxxIn" are metrics from any source service to destination service A...
```

Strictly speaking, these are NOT "service" metrics, they are workload metrics. And these metrics are produced by the sidecar proxy sitting in the workload instances (aggregated by workload - so all sidecar proxies running in pods A, B, C in Workload W have their metrics aggregated into Workload W metrics).

I think it is easier to understand these metrics when you remember they represent the data coming from the proxies and do not represent metrics from a "service".

A proxy sitting in a workload instance performs this:

Incoming request enters the proxy - the proxy pre-processes that request and then passes that request on to its destination. The proxy then receives the response from the destination, pre-processes the response, and then sends the response back to the originator of the incoming request (e.g. a browser).

There are actually four edges when you think about that:

`Browser ---> proxy ---> destination ---> back to proxy ---> back to browser`

If I understand the graphs correctly that @jotak shows, I believe each of those graphs represent one of those four edges.

I suspect we should re-label these graphs to make this more clear. For example, I agree having two graphs each with the title "TCP sent" is confusing.

@jotak - please confirm that what I said above is all true :) I think I understand this, but let me know if you think this description of the graph metrics is incorrect.

jotak:
@jmazzitelli that's right "service" wasn't the right term, damn it brings more confusion while I was trying to clarify :)

But your picture about the 4 graphs is incorrect

> `Browser ---> proxy ---> destination (workload A) ---> back to proxy ---> back to browser`

Here I don't really care about proxy, that's more an implementation detail, what we show on such request, in these metrics is:
`Browser ---> destination ---> back to browser`

So from "workload A" point of view, here you would have only charts in the left column (that is currently named "Incoming" but could stand for "server" as well). Nothing on the right column.

> I suspect we should re-label these graphs to make this more clear. For example, I agree having two graphs each with the title "TCP sent" is confusing.

But _there is_ actually two "TCP sent " metrics for a given workload: the one involving a request that is originated by the workload ("client metrics") and the one involving a response for a request originated by another workload ("server metrics").

> client/server in microservices world sounds like not the best option
Well, initially I would have thought it's a good way to use well known concepts to distinguish _who_ originates the request, but the confusion seems to prove I'm wrong

jmazzitelli:
> But there is actually two "TCP sent " metrics for a given workload: the one involving a request that is originated by the workload ("client metrics") and the one involving a response for a request originated by another workload ("server metrics").

Right. OK. This is why there are 2 metrics with all the same labels except one has reporter=source and one has reporter=destination (Istio recently renamed these labels from client/server to source/destination).

We probably need UXD help here - because this terminology is going to be very confusion for users (its confusing to us!).

abonas:
@jmazzitelli I think this is more a question to docs than UX.

JStickler:
@abonas , yes I've  been trying to follow the various apps/services/workload conversations because it's pretty obvious that we're going to have to clearly document the distinction since there's so much confusion around the terminology.

israel-hdez:
@jotak Not sure if there is some issue in my env. The backend is returning the same set of TCP metrics for all services in the namespace. Even for the ones that does not generate any TCP traffic (e.g. the frontend svc of the petstiogram app shows tcp traffic in kiali). I hope you can check this.

israel-hdez:
Hmm... yes I think terminology used in this page can be confusing. But TCP metrics have an analog HTTP metric. Thinking in the metrics of the left column, I guess:
* "TCP sent" metric is analog to "Response size" metric.
* "TCP received" metric is analog to "Request size" metric.
For the right column, this matching is inverted.

So, if any rename needs to be done, I guess the "Input" and "Output" headings are the ones that should be renamed. The idea is that the left column (Input) is traffic that is reaching the service that is currently on the screen and the connection was initiated "remotely"; while the right column (Output) is traffic for which the connection is initiated by the service currently on the screen and is reaching other services.

jotak:
@israel-hdez thanks

> [...] I hope you can check this.

Having a look right now

jotak:
@israel-hdez I don't reproduce your issue. When I navigate to metrics tab for "frontend", TCP metrics are just empty. Going back and forth from frontend to mongodb doesn't seem to leave any incorrect state.

Can you give an accurate scenario to reproduce?

israel-hdez:
@jotak Hmm... In my side, the petsiogram app shows this behavior.
I can check if for you works correctly.

PR #545, opened 2018-07-30
add experimental notice

jmazzitelli:
This adds the following notice to the login screen:

image


lucasponce:
I think with the latest discussion we are not going to add an "experimental" label on the release.
Closing this PR to not make it noise (we can revisit this idea in the future).

PR #546, opened 2018-07-31
KIALI-1215 , istio-1.0, workload/app/service health

jotak:
- New endpoints for workload-based health and app-based health
- Standardise schema of Health Details, with "HealthItem"
- Graph able to display workload & app health
- Refactoring Health model in a more OOP way (easier to manage polymorph health)


jotak:
Relative to backend: https://github.com/kiali/kiali/pull/394

abonas:
is this only a refactoring or also introducing new UI?
if the latter, a screenshot will be helpful. and perhaps splitting refactoring from new stuff into separate PRs (first review and merge refactor, then introduce new stuff separately), or I misunderstood here something?

jotak:
It's 95% of refactoring, in fact the health is still displayed in the same way but the visual differences are:

- The graph will now show workload-based health when clicking on a workload node in the summary panel. Before, only app-based health was displayed, so when you clicked in a node that was in an app box, there was no health information.

- Workload-based health is similar to app-based health, except it doesn't have the envoy health:
workload-health

src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -475,44 +474,50 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
...
     // Asynchronously fetch health
     cy.nodes().forEach(ele => {
       const namespace = ele.data('namespace');
       const app = ele.data('app');
       const workload = ele.data('workload');
 
-      if ((workload !== 'unknown' || app !== 'unknown') && (ele.data('isGroup') || !ele.data('parent'))) {
-        const appBasedHealth = this.props.graphType === GraphType.APP && app !== 'unknown';
-        let promise = healthPerNamespace.get(namespace);
+      if (app !== undefined && app !== 'unknown' && (ele.data('isGroup') || !ele.data('parent'))) {
```
jmazzitelli: You should rely on the cy's global scratch data to determine the graph type, don't assume graph type based on node details. At minimum it will be consistent with the rest of the graph code.

jmazzitelli:
@jotak in your screen snapshot, the summary panel shows "Service" ("Service: stadium"). This isn't a service right?

josejulio:
> @jotak in your screen snapshot, the summary panel shows "Service" ("Service: stadium"). This isn't a service right?

We haven't updated that in the UI
I'm still unsure if we should show all the info we have (workload(s) and app) or only depending the graphType (workload based, only show workload; app based show app, if no app, use workload)

src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -475,44 +474,53 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
...
-      const app = ele.data('app');
-      const workload = ele.data('workload');
-
-      if ((workload !== 'unknown' || app !== 'unknown') && (ele.data('isGroup') || !ele.data('parent'))) {
-        const appBasedHealth = this.props.graphType === GraphType.APP && app !== 'unknown';
-        let promise = healthPerNamespace.get(namespace);
-        if (!promise) {
-          promise = API.getNamespaceHealth(authentication(), namespace, duration).then(r => r.data);
-          healthPerNamespace.set(namespace, promise);
+      if (ele.data('isWorkload')) {
```
josejulio: I suppose this only appears (isWorkload) when the graphType === workload or when there is no "app" label (regardless of graphType). is that right?

jmazzitelli: @josejulio Yes, I believe that is true.

Note that this "isWorkload" attribute is soon going away. @jshaughn is refactoring this - "isWorkload" will soon be replaced with "nodeType" attribute whose value will be something like "app", "workload", or "service". @jshaughn can give specifics, but that will be the basic idea.

jotak: That's great. I'd need one confirmation about it:
In app-versioned graph, a node that is inside a group will be of type "workload", right? And a node that is not in group is of type "app". Correct? (cc @jshaughn )

jshaughn: @jotak , @jmazzitelli , no I don't think that is correct.   For a versionedApp graph the nodes inside a group should be NodeTypeApp where  n.app and n.version are both set.  The n.app would be the name on the group and the n.version would be the name displayed for the node.  The only time a node would be NodeTypeWorkload in an app graph is if the workload is not properly labeled and so we can't set the app/version fields.

But, for a versionedApp graph the nodes of type NodeTypeApp will still have n.workload set.  This is because, unless misconfigured, the unique app+version should equate to a unique workload.  That workload name can then be used for metric gathering and health, as opposed to the app+version labels.

jotak: Ok so it won't help me, I'll have to go back to my first idea which was to detect if the node is in group or not.

jshaughn: I guess I don't understand why you would expect it to be a workload, the groups are specifically for boxing app versions.  But I also don't know why it doesn't help you, the workload field is set on the node.  Also, if you need to know if a node is in a group you can look for ele.data('parent').  We should talk this out because my intent was to support your work as best as possible.

josejulio: > Ok so it won't help me, I'll have to go back to my first idea which was to detect if the node is in group or not.

I think you can use `ele.data('workload')`

jotak: Let's talk about it on monday :)

jotak:
@jshaughn @josejulio i've updated, can you check the last commit?

For merging, it will have to be done in sync with https://github.com/kiali/kiali/pull/394 since there are changes in health model.

PR #549, opened 2018-08-03
Kiali 1287 Introduce NodeType

jshaughn:


jshaughn:
There is some weird prettier issue but @jmazzitelli and I ran this for a long time testing it out.  I'm going to merge because the changes here are required to sync with changes already merged server-side.  I'll look at the travis issue later.  @mtho11 , any idea what the issue is?

PR #553, opened 2018-08-06
KIALI-1279 Avoid Kiali lock-up by multiple chart refreshes when page …

josejulio:
…is hidden

** Describe the change **

 - There seems to be an issue with the C3Charts, when refreshing multiple times
   when the page is hidden, it doesn't seem to be processed right away and they
   build up over time. When the page is visible again, they start processing
   (taking 100% cpu) and the page becomes unresponsive. After all the processing
   is done the page becomes responsive again. The more time the page is in
   background, the more time it takes to become responsive.

   Using Visibility API to detect if page is hidden to hide the C3Charts.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,8 @@
     "ssri": "6.0.0",
     "typesafe-actions": "1.1.2",
     "typestyle": "2.0.1",
-    "url-search-params": "0.10.0"
+    "url-search-params": "0.10.0",
+    "visibilityjs": "2.0.1"
   },
   "script-comments": [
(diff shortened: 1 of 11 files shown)


josejulio:
Travis is happy again

jotak:
@josejulio so was it about C3 charts as you suspected? Maybe the same patch should be done on the metrics page?

josejulio:
> @josejulio so was it about C3 charts as you suspected? Maybe the same patch should be done on the metrics page?

Right, I had to hide the C3charts to avoid this. I'll try to reproduce this bug in the metrics page.

PR #554, opened 2018-08-06
KIALI-1289 NameSpaceFilter is not working when the initialFilters is []

aljesusg:
** Describe the change **
When initialFilters is an empty [] NamespaceFilter should show the Namespacefilter only.

This is not working well because the currentFilterType.filterValues is empty, we need to set the currentFilterType after the method updateNamespaces()

** Issue reference **

[KIALI-1289](https://issues.jboss.org/browse/KIALI-1289)

This blocks to [KIALI-1209](https://issues.jboss.org/browse/KIALI-1209)

### Before
before

### After
after


mtho11:
This will not be needed once _namespace_ is moved to the app header; but fine for now and since it is a one line change...

aljesusg:
Yep I  know but we are working in workload list page and we only we'll have the namespace filter while we finish validation.

PR #555, opened 2018-08-06
KIALI-1216 Add a new Workload list page.

aljesusg:
Add workload List to Kiali

Need [#554 KIALI-1289](https://github.com/kiali/kiali-ui/pull/554)

JIRA: [KIALI-1216](https://issues.jboss.org/browse/KIALI-1216)

workloads


aljesusg:
Thanks @serenamarie125

lucasponce:
@aljesusg I have a minor UX suggestion.
I find "Labels:" as something ambiguous on this context, as it is not showing the labels as an user could expect but it is checking that this workload has an "app" and a "version" label.
So, perhaps in the future we want to refactor and add all labels at this level (Services, Workloads, future Apps), but for this PR I would suggest to rename "Labels:" -> "Label validation:".
I think on this way, user has more context that this is to mark which label is present (or not).
A nice thing could be to add all labels, but that should be done also in Services and would involve refactoring on the backend, so for this PR, I would be ok just with that renaming.

aljesusg:
Changed @lucasponce 
screenshot from 2018-08-14 13-24-48

src/components/NamespaceFilter/a
```diff
@@ -0,0 +1,34 @@
+{currentFilterType: {…}, filterTypeList: Array(4), activeFilters: Array(0), currentValue: ""}
```
lucasponce: I don't understand if this file is an error, or if you want to extract some filtering to a different file.
In any case this should be reviewed.

lucasponce:
@aljesusg my bad: I realized that the style is with both words first letter upper case, could you do that fix "Label validation" -> "Label Validation".
Thx

src/types/Workload/WorkloadListComponent.ts
```diff
@@ -0,0 +1,147 @@
...
+          workload.versionLabel
+        );
+        if (workloadItem) {
+          this.workloads.push(workloadItem);
+        }
+      });
+    }
+
+    getDeploymentItems(): WorkloadItem[] {
+      let workloadsItems: WorkloadItem[] = [];
```
jotak: could be simplified into a simple `return this.workloads.map(...)`

src/types/Workload/WorkloadComponent.tsx
```diff
@@ -0,0 +1,123 @@
+import Namespace from '../Namespace';
```
jotak: This file shouldn't be under `src/types` I think. Basically, types shouldn't contain any TSX / rendering stuff. There's a `components` directory more suited for that.

src/types/Workload/WorkloadComponent.tsx
```diff
@@ -0,0 +1,123 @@
...
+    constructor(name: string, namespace: string, istioSidecar: boolean, appLabel: boolean, versionLabel: boolean) {
+      super(name, namespace, istioSidecar, appLabel, versionLabel);
+    }
+
+    getKind = (): string => {
+      return DEPLOYMENT;
+    };
+
+    getLink = (): string => {
+      return '';
```
jotak: something is missing here? (dependant on another task?)

src/services/Api.ts
```diff
@@ -244,6 +244,10 @@ export const getIstioConfigValidations = (
   );
 };
 
+export const getWorkloads = (auth: string, namespace: String): Promise<Response<Validations>> => {
```
lucasponce: I don't think this is correct, and I am afraid it can propagate some extra code.
Just following other API methods, the Response is forced to have the type:
```
i.e.
Response<ServiceList>
Response<IstioConfigList
```
I think for workloads this adds extra logic in the classes used to map the types.

So, I guess it should more or less follow the same pattern unless I'm missing something.

lucasponce:
PR works well locally and it does the expected functional changes.
A couple of comments related the code and some unification of how to map API types.
Probably my last comment could lead in another change around types.

src/types/Workload/WorkloadListComponent.ts
```diff
@@ -0,0 +1,147 @@
...
+      this.namespace = ns;
+      this.workload = workload;
+    }
+  }
+  /** WorkloadNamespaceResponse object */
+  export class WorkloadNamespaceResponse {
+    namespace: Namespace;
+    workloads: (Workload.Deployment)[];
+
+    constructor(response: any) {
```
jotak: I think "response" shouldn't be "any" here, as it comes from the rest API which should provide the right typing.

src/components/NamespaceFilter/a
aljesusg: It's an error ^^

src/types/Workload/WorkloadComponent.tsx
aljesusg: It's tax by the react component returned in the render method.

src/types/Workload/WorkloadComponent.tsx
aljesusg: There is not link waiting for workload details page

lucasponce:
Another issue I can see locally.
Now if I go to Istio Config and I select a namespace filter, that is updated in the URL and in the page.
For example, moving from Istio Config to Services:
image
image
But when navigating to workloads, the filter is missing in the URL (but is still preserved in the local state).
So, I think that the logic that updates the filter to the URL is not 100% correct in the new workloads list.
image

aljesusg:
@lucasponce comments addressed, thanks !!!

lucasponce:
@aljesusg after your last commit I see some problems in the Namespace filter (I think you commented to me) but just to leave a message here for the record.
When I try to select two namespaces, it only takes one (the latest one).
In other pages it works correctly.

aljesusg:
mmm I thought that it had been decided that it was replaced, I'll change it.

lucasponce:
> mmm I thought that it had been decided that it was replaced, I'll change it.

Nope, filtering by multiple namespace is a feature and we don't want to miss it.
Discussions around it are still open and meanwhile there is not clear consensus we should not introduce regressions.

lucasponce:
@aljesusg another issue
image
I think is missing to sort by Workload Name, in the same way we can sort services by name or istio config objects by name too.

lucasponce:
@aljesusg also to search/filter by Workload Name too
image

aljesusg:
Added, thanks.-

aljesusg:
Thanks @lucasponce

PR #556, opened 2018-08-06
Kiali 1282 use consistent padding when doing a graph "fit"

jshaughn:


mtho11:
Just FYI, some editors like **Webstorm** will auto-detect prettier and give you IDE hints and 'Reformat with Prettier' options.

jshaughn:
Updated format, merging...

PR #558, opened 2018-08-07
KIALI-1180 Update grafana link for istio-1 dashboards

jotak:
Note that there isn't in and out dashboards anymore. Just one dashboard for a given service.


jotak:
test is failing, I'm updating right now ...

PR #560, opened 2018-08-07
KIALI-1292 Back link in VS/DR Yaml page goes to corresponding list

israel-hdez:
This avoids the annoyance of re-clicking on the virtualservice/destinationrule tab if the user is browsing the list.

### DIFF
--- a/src/pages/ServiceDetails/IstioObjectDetails.tsx
+++ b/src/pages/ServiceDetails/IstioObjectDetails.tsx
@@ -86,7 +86,7 @@ export default class IstioObjectDetails extends React.Component<IstioObjectDetai
           <Row className={'card-pf-body'} key={'virtualservice-yaml'}>
             <Col xs={12}>
               <div className={'pull-right'}>
-                <Link to={{ pathname: this.props.servicePageURL }}>Back to Service</Link>{' '}
+                <Link to={this.props.servicePageURL}>Back to Service</Link>{' '}
               </div>
               <AceEditor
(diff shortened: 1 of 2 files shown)


abonas:
@israel-hdez a screenshot is worth a thousand words :)
Also, the navigation/breadcrumbs are reworked by @cshinn and @serenamarie125 (for https://github.com/kiali/kiali-design/issues/9) so I'm not sure the behavior introduced here will remain.
cc @lucasponce

israel-hdez:
@abonas Yes, indeed.

So, this is for the back link in the YAML view:
image

That link goes (back) to the service details page. Previously, the "secondary" tab was being reset to the first tab (Pods). Now, if you are seeing a virtualservice/destinationrunle, it will pre-select the virtualservices/destinationrule tab:
image

abonas:
IIRC there should not be such "back" buttons, but I'd like to make sure - @cshinn @serenamarie125 ?

lucasponce:
@abonas as far as I know this is a bugfix not a refactoring leaded from UX recommendation.
This is just to fix an existing bad behaviour, it is not refactoring anything.
So, I would differentiate those in the PR, because UX team and Dev team can be confused.

In that context, I would approve this one.
Once there is a formal change from UX, we can add a dev task to change it.

But I consider this a bugfix, so in that sense.

abonas:
@lucasponce I'm ok with bug fixing, just not introducing more navigation at the moment

PR #561, opened 2018-08-07
KIALI-1291 Fix graph issues with relayout

josejulio:
- Makes it more consistent the layout (for non random layouts, i'm looking at you cose)
- Correctly displays when switching from a non-empty namespace to a empty namespace (and vice versa)

Issues:
  Cose still flashs, I suspect is due to its non discrete nature.


jmazzitelli:
I see this with "cose" layout - but I think we know there is something wrong with that specific layout:

image

josejulio:
I think so, it happens on a browser refresh. Not exactly sure, but probably related to being "not connected"

PR #562, opened 2018-08-07
KIALI-1277 Do not render badges outside the graph area

josejulio:
- Try to put the badges in the last div of the cy container

Solves KIALI-1277 by putting the badges in the last div container (currently it was putting the badges in container[0]. That was the panzoom container).

### DIFF
--- a/src/components/CytoscapeGraph/graphs/GraphBadge.ts
+++ b/src/components/CytoscapeGraph/graphs/GraphBadge.ts
@@ -84,8 +84,11 @@ class GraphBadge {
         modifiers: {
           inner: { enabled: this.inner },
           preventOverflow: {
-            enabled: true,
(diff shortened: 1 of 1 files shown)


jmazzitelli:
almost there. All badges are OK when moving out of the viewing area to the left, right, and down.

But when moving UP, if a badge is on a composite box, that badge pins to the edge. See animated gif - notice the badges on the nodes are behaving OK - its only the badge on the composite box that behaves wrong:

badge-almost ogv

jmazzitelli:
to test with a badge like this one on reviews, run:

oc create -f [bookinfo-reviews-faults.yaml](https://github.com/kiali/kiali/blob/master/hack/istio/bookinfo-reviews-faults.yaml)

PR #564, opened 2018-08-07
KIALI-1295 Fix KIALI locking up in the service details metric page

josejulio:
This is a workaround, the root issue is related to C3Chart

### DIFF
--- a/src/containers/ServiceMetricsContainer.tsx
+++ b/src/containers/ServiceMetricsContainer.tsx
@@ -1,0 +1,11 @@
+import { KialiAppState } from '../store/Store';
+import { connect } from 'react-redux';
+import ServiceGraphPage from '../pages/ServiceDetails/ServiceMetrics';
+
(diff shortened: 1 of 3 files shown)


josejulio:
> Just one question: how do you reproduce the issue (I mean, if I don't have your patch) ? Just let kiali inactive for minutes, for hours, more?
> 
> Revert View details

couple of hours seems fine for some seconds of freeze.
To quickly test i updated config.ts to create a new refreshing interval of 100ms

```typescript
index e9fe3b5..a8a8782 100644
--- a/src/config.ts
+++ b/src/config.ts
@@ -29,6 +29,7 @@ export const config = () => {
       /** Options in refresh */
       pollInterval: {
         0: 'Pause',
+        100: 'Damn fast',
         5000: '5 seconds',
         10000: '10 seconds',
         15000: '15 seconds',
```

With that, 2 minutes is about 16 seconds of freeze (without the patch)

PR #569, opened 2018-08-09
kiali-1059 support custom context path

vnugent:
** Describe the change **
- Frontend changes to support custom context path such as *myapp.com/**kiali***
- By default the app is still served under / unless you specify otherwise during deployment on K8s and OpenShift. (By changing SERVER_WEB_ROOT env variable)
- Please avoid using *absolute* path to reference static assets or API endpoints.  Everything should be relative from `<SERVER_WEB_ROOT>/`  (which can also be "/" ).   
- Adding undefined guards to `AboutUIModal` component to prevent JS error when backend queries fail
- Backend PR https://github.com/kiali/kiali/pull/409

** Issue reference **
- KIALI-1059
- https://github.com/kiali/kiali/issues/321

** Backwards compatible? **
Yes

** Screenshot **
With /kiali

kiali-1059-login

kiali-1059-with-path

404

kiali-1059-404

### DIFF
--- a/package.json
+++ b/package.json
@@ -120,13 +120,9 @@
   ],
   "author": "Red Hat",
   "license": "Apache-2.0",
-  "homepage-comment": [
-    "By default, Create React App produces a build assuming your app is hosted at the server root.",
-    "To override this, specify the homepage in your package.json:",
-    "We don't want this, so don't touch homepage.",
-    "https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#building-for-relative-paths"
-  ],
-  "homepage": "/",
+  "homepage-comment": ["*Do not* change this setting if you wish run Kiali under different server root.",
+                       "Instead update 'web_root' in Kaili config map in OpenShift console."],
+  "homepage": "./",
   "publishConfig": {
     "access": "public"
   },
--- a/public/env.js
+++ b/public/env.js
@@ -1,0 +1,2 @@
+// This file is intentionally left bank.
+// Kiali server may re-generate this file with configuration variables.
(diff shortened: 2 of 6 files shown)


mtho11:
Just waiting until after release on Monday to review.

vnugent:
Can we please put this PR back on Sprint 9?

lucasponce:
Not sure if this one would need backend work to manage the jaeger proxy.
Did you check that ?
I can test it locally later.

vnugent:
@lucasponce I'd suggest to merge as-is and address Jaeger proxy in a separate PR.  The way we set up http server for Jaeger in Kiali backend is not consistent with the API and frontend's.

lucasponce:
@vnugent I don't mind to merge this PR if we have two requisites:
- There is no regressions with the Jaeger integration in normal scenerios.
- If there is some required changed, to have another PR to merge both at same time, before to validate that we don't introduce regressions.

"latest" docker image is being tested for external users so, we need to make an effort to check that everything is looking good.

The Jaeger integration is still a hot topic and it's not something we can leave broken in master, so if you can test and prepare an additional PR, I can review all together and merge them at same time once everything is ok.

jmazzitelli:
One thing I forgot to test ... we need to test this before merging anything. Make sure this still works in "dev" mode (that is, using "yarn start" with the package.json pointing to the server's url).

jmazzitelli:
I tested and this works for me. We should probably get someone else to test also.

vnugent:
> There is no regressions with the Jaeger integration in normal scenerios.
@lucasponce The default behavior of this PR is to serve the API and console under "/" so everything should work just like before unless you change web_root in config map.

PR #571, opened 2018-08-10
KIALI-1300: Change to cose layout to cose-bilkent

mwringe:
** Describe the change **

Change the layout for 'cose' to 'cose-bilkent'.

Whatever the 'cose' layout is, it doesn't seem to be working properly. We can't find a lot of information about it online, nothing of the configuration options listed here seem to have an affect on it: https://github.com/cytoscape/cytoscape.js/blob/master/documentation/demos/cose-layout/code.js

** Issue reference **

This fixes both

https://issues.jboss.org/browse/KIALI-1300
https://issues.jboss.org/browse/KIALI-1297

The url changes from displaying 'cose' to 'cose-bilkent'. Anyone with an old bookmark may have to update it.

** Screenshot **

screenshot from 2018-08-10 10-46-45

### DIFF
--- a/src/components/CytoscapeGraph/graphs/CoseGraph.ts
+++ b/src/components/CytoscapeGraph/graphs/CoseGraph.ts
@@ -3,7 +3,7 @@ import { GraphType } from './GraphType';
 export class CoseGraph implements GraphType {
   static getLayout() {
     return {
-      name: 'cose',
+      name: 'cose-bilkent',
       animate: false,
       nodeDimensionsIncludeLabels: true
     };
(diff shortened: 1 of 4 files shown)


jmazzitelli:
Not as bad as before. I see better visualizations, however, under some situations, nodes do go out in space, but not nearly as bad as before. I think we should merge this.

image

israel-hdez:
Perhaps cose was chosen at first because it's lighter than bilkent. From http://js.cytoscape.org/#layouts/cose:

> The cose layout is very fast and produces good results. The cose-bilkent extension is an evolution of the algorithm that is more computationally expensive but produces near-perfect results.

PR #572, opened 2018-08-10
KIALI-1318 Fix incorrect service links, make istio validation checks never null/undefined

jotak:
This PR comes with https://github.com/kiali/kiali/pull/424
It's not absolutely mandatory (the server-side PR would be sufficient to fix the bug) but IMO it helps by clarifying what we do expect from the json, now enforcing non-null and non-undefined values.

JIRA: https://issues.jboss.org/browse/KIALI-1318


jotak:
@jshaughn / @josejulio I removed the link to service charts from the summary panel (see especially the second commit), is that ok for you?

We'll have to set a link to app / workload metrics when they are available, instead

PR #573, opened 2018-08-10
KIALI-1322 make istio validation checks never null/undefined

jotak:
This PR comes with kiali/kiali#425
It's not absolutely mandatory (the server-side PR would be sufficient to fix the bug) but IMO it helps by clarifying what we do expect from the json, now enforcing non-null and non-undefined values.

JIRA: https://issues.jboss.org/browse/KIALI-1322

### DIFF
--- a/src/components/ConfigValidation/ConfigIndicator.tsx
+++ b/src/components/ConfigValidation/ConfigIndicator.tsx
@@ -53,7 +53,7 @@ const tooltipListStyle = style({
 });
 
 export class ConfigIndicator extends React.PureComponent<Props, {}> {
-  numberOfChecks = (type: string) => (this.props.validation.checks || []).filter(i => i.severity === type).length;
+  numberOfChecks = (type: string) => this.props.validation.checks.filter(i => i.severity === type).length;
 
   getTypeMessage = (type: string) => {
     const numberType = this.numberOfChecks(type);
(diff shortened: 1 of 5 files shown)


src/components/ConfigValidation/ConfigIndicator.tsx
```diff
@@ -53,7 +53,7 @@ const tooltipListStyle = style({
 });
 
 export class ConfigIndicator extends React.PureComponent<Props, {}> {
-  numberOfChecks = (type: string) => (this.props.validation.checks || []).filter(i => i.severity === type).length;
+  numberOfChecks = (type: string) => this.props.validation.checks.filter(i => i.severity === type).length;
```
lucasponce: I see the change.
Ensuring that `this.props.validation.checks` makes code cleaner, but I understand that the former code was correct.
I mean the `(this.props.validation.checks || [])` should also protect against nulls.

Or there was some case not well covered.

P.S.: I'm asking more to learn/understand it, sometimes javascript has some "false friends" and we finish doing some things odd.

jotak: @lucasponce this null-check was ok. The one that wasn't ok was in file `ServiceInfo.tsx` because we were checking against `undefined` while the no-value case was `null`. My change is to say now: ok , the contract is now different, `check` is just non-nullable and non-undefinedable and it is (kinda) guaranteed by the backend.

jotak: PS: this is all guided by the change in the model where `checks?: ObjectCheck[];` now becomes `checks: ObjectCheck[];`.

lucasponce: +1
I like this approach better.
Everywhere where types are fully defined is a win.

PR #581, opened 2018-08-13
KIALI-1339: Temporarily remove the spark charts

mwringe:
Removes the sparkline charts from the graph page.

Temporary until we can figure out and fix all the sparkline graph issues.


israel-hdez:
Not so sure, but apparently, https://github.com/kiali/kiali/pull/428 and https://github.com/kiali/kiali-ui/pull/578 fixed the issue with the sparklines.
In my box, they are working correctly even with the renamed setup of bookinfo.

mwringe:
@jshaughn mentioned that this was also affecting master, even with 428 & 578 applied.

It might be best to move those over to the pre-release branch if they do fix the issue.

PR #583, opened 2018-08-14
KIALI-1343 Service details metrics: show only inbound

jotak:
Simplifies code for state & charts definitions

JIRA: https://issues.jboss.org/browse/KIALI-1343

Screenshot:
screenshot

### DIFF
--- a/src/pages/ServiceDetails/ServiceMetrics.tsx
+++ b/src/pages/ServiceDetails/ServiceMetrics.tsx
@@ -24,37 +24,17 @@ const expandedChartBackLinkStyle = style({
   textAlign: 'right'
 });
 
-type ServiceMetricsState = {
-  alertDetails?: string;
-  requestCountIn?: MetricGroup;
-  requestCountOut?: MetricGroup;
-  requestSizeIn?: Histogram;
-  requestSizeOut?: Histogram;
-  requestDurationIn?: Histogram;
-  requestDurationOut?: Histogram;
-  responseSizeIn?: Histogram;
-  responseSizeOut?: Histogram;
-  tcpReceivedIn?: MetricGroup;
-  tcpReceivedOut?: MetricGroup;
-  tcpSentIn?: MetricGroup;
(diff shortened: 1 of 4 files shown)


lucasponce:
+1
I think that in Workload Details page and Application Details page we will maintain both Inbound and Outboud metrics. Isn't it ?

lucasponce:
I guess this doesn't need any additional work from backend. right ?

src/pages/ServiceDetails/ServiceMetrics.tsx
```diff
@@ -114,19 +103,17 @@ class ServiceMetrics extends React.Component<ServiceMetricsProps, ServiceMetrics
...
+          Object.keys(prevState.charts).forEach(k => {
+            const chart = prevState.charts[k];
+            const histo = metrics.histograms[k];
+            if (histo) {
+              chart.metrics = histo;
+            } else {
+              chart.metrics = metrics.metrics[k];
+            }
+          });
+          return prevState;
```
israel-hdez: IIRC, `prevState` should be used as a constant. Here you are changing some values.
I tried the code and it works. But not sure if React may do weird things under some circumstances.

josejulio: You are right @israel-hdez 
from the [docs](https://reactjs.org/docs/react-component.html#setstate):
> prevState is a reference to the previous state. It should not be directly mutated. Instead, changes should be represented by building a new object based on the input from prevState
> ...
> The output of the updater is shallowly merged with prevState

jotak: ok I'll fix that

jotak:
@lucasponce :
> I think that in Workload Details page and Application Details page we will maintain both Inbound and Outboud metrics. Isn't it ?
Yes, and as we mentioned previously, maybe it would be nice then to show them in separate tabs

> I guess this doesn't need any additional work from backend. right ?
Correct

PR #584, opened 2018-08-14
KIALI-1194 Node and Edge styling per UX

jshaughn:
- solve issues with selection hiding health
- use the same thresholds for health and graph

Here is a link to [kiali-1194]https://issues.jboss.org/browse/KIALI-1194) so that you can see the UX color chart.

This PR also solves [kiali-1280](https://issues.jboss.org/browse/KIALI-1280), which is the same issue but for edges.  Also, see this [UX design PR](https://github.com/kiali/kiali-design/pull/29) for more on the edges.

### DIFF
--- a/src/components/CytoscapeGraph/graphs/GraphStyles.ts
+++ b/src/components/CytoscapeGraph/graphs/GraphStyles.ts
@@ -1,10 +1,31 @@
 import { PfColors } from '../../../components/Pf/PfColors';
 import { EdgeLabelMode } from '../../../types/GraphFilter';
-import { config } from '../../../config';
-import { FAILURE, DEGRADED } from '../../../types/Health';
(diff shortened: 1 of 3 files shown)


jshaughn:
kiali-1194

jshaughn:
Merging, any further comments can be addressed going forward...

abonas:
@jshaughn the gif is freezing a bit when I view this, so I'm unsure if this is related to it or not - is the hovering covered here (it was part of the UX instructions)? It looks as if you must click on something (or on an empty space to "unselect"), which is not hovering. But again, it could be the gif itself.

jshaughn:
@abonas The hover is covered here.  Sorry about the video, but if you can see the very beginning, that is all hover, starting at hover over ingress and then moving to hover over productpage.  The first selection is the green edge leading from productpage.

PR #586, opened 2018-08-15
KIALI-1131 Use RCUE for productized Patternfly UI

mtho11:
** Describe the change **
Enable selectively building with Patternfly or RCUE(RH productization) based on CLI settings.
This PR dynamically loads the RCUE css stylesheets if specified on the CLI.

By adding a React parameter to the build, we can build using different css libraries (RCUE vs Patternfly).

So to build for RCUE just pass the additional parameter to the build:
```
env REACT_APP_RCUE=true yarn build
```
Or for development:
```
env REACT_APP_RCUE=true yarn start
```

** Issue reference **

https://issues.jboss.org/browse/KIALI-1131

** Backwards compatible? **
Yes

** Screenshot **
Before (Patternfly):
pf-login

After(RCUE):
rcue-login

QE Instructions:
Visually, the above can be seen to tell the difference between Patternfly and RCUE.
However, if you open the JS console you will find a *debug* level message telling you that the RCUE css has loaded:
rcue-message

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -10,6 +10,20 @@ import history from './History';
 import { PersistGate } from 'redux-persist/lib/integration/react';
 import * as Visibility from 'visibilityjs';
 
+/**
+ * Use the Patternfly RCUE productized css styles if set by the environment
+ * @example 'env REACT_APP_RCUE=true yarn start'
+ */
+const loadRcueCssIfNeeded = async (): Promise<void> => {
+  // get the flag from command line if set: env REACT_APP_RCUE=true yarn start
+  const useRcue = process.env.REACT_APP_RCUE;
+  if (useRcue === 'true') {
+    console.debug('REACT_APP_RCUE set to true');
+    Promise.all([require('patternfly/dist/css/rcue.css'), require('patternfly/dist/css/rcue-additions.css')]);
+    console.debug('Loaded RCUE css libraries loaded');
+  }
+};
+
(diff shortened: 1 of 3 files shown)


src/app/App.tsx
```diff
@@ -10,6 +10,20 @@ import history from './History';
 import { PersistGate } from 'redux-persist/lib/integration/react';
 import * as Visibility from 'visibilityjs';
 
+/**
+ * Use the Patternfly RCUE productized css styles if set by the environment
+ * @example 'env REACT_APP_RCUE=true yarn start'
+ */
+const loadRcueCssIfNeeded = async (): Promise<void> => {
+  // get the flag from command line if set: env REACT_APP_RCUE=true yarn start
+  const useRcue = process.env.REACT_APP_RCUE;
```
cfcosta: How exactly does this work? It's replaced at config time? I'd imagine `process.env` not to be available on the browser, no?

mtho11: It's compile time (build time) not runtime. So for your productization build use: 
`env REACT_APP_RCUE=true yarn start` for that build.

cfcosta:
LGTM.

lucasponce:
This PR has some basic style problems, for example:
image
Note the underlying style when navigating on the menu.

Are we sure we want in upstream this style ?

abonas:
@lucasponce I would indeed expect that this PR is not supposed to change anything in upstream style, but only add a flag to switch the style on demand.

PR #566, opened 2018-08-08
KIALI-1218 Service details: Pods and deployments removed + adding workloads

xeviknal:
** Describe the change **

Adapted the modifications performed into ServiceDetails endpoint: https://github.com/kiali/kiali/pull/418
List of pods and deployments are removed. Workloads added.

Also there is a proposal on moving the Istio icon next to service title (as suggested in the comments)

This PR needs to be deployed at the same time as https://github.com/kiali/kiali/pull/418

** Issue reference **

[KIALI-1304](https://issues.jboss.org/browse/KIALI-1304): Update ServiceDetails endpoint.

** Backwards in compatible? **
Yes.

** Screenshot **

If Istio is enabled for the service:
screenshot of kiali console 9

Otherwise:
screenshot of kiali console 10

PR #587, opened 2018-08-16
KIALI-1218 Service details: Pods and deployments removed + adding workloads

lucasponce:
This is a clone of 
https://github.com/kiali/kiali-ui/pull/566
just to resolve locally conflicts with master and pushed to upstream.

IMPORTANT: This PRs cannot be merged as it is, it requires this backend PR first https://github.com/kiali/kiali/pull/418


src/components/Pf/PfInfoCard.tsx
```diff
@@ -1,24 +1,36 @@
...
+  }
+
   render() {
     return (
       <div className="card-pf">
         <div className="card-pf-heading">
           <h2 className="card-pf-title">
             <Icon type={this.props.iconType} name={this.props.iconName} style={{ marginRight: '10px' }} />
+            {this.istoIcon()}
             {this.props.title}
```
aljesusg: Not Blocker but why not { this.props.istio && (Istio sidecar) }

lucasponce: Yeah, probably code could be refactor in that way.
But as @xeviknal was the original author and it's more a style side, I let you comment it when he is back and in a future iteration can be refined.

PR #588, opened 2018-08-16
Release 0.7.0

mwringe:
Prepare for the 0.7.0 release of the Kiali UI

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@kiali/kiali-ui",
-  "version": "0.6.0",
+  "version": "0.7.0",
   "bugs": {
     "url": "https://issues.jboss.org/projects/KIALI/issues/"
   },


mwringe:
@lucasponce can you take a look?

lucasponce:
Sure, let me to test it locally

lucasponce:
@mwringe do you remember the tag which is needed for the intergration with Istio 1.0 GA ?
Not sure if it was 0.7.0.
Perhaps in some email @jmazzitelli commented.
I mean, if the tag doesn't match, Istio 1.0.GA won't pull the right image by default.

lucasponce:
```
<lponce> It sounds to me that v0.6.0 was the expected one, but not sure, to be honest
<cfcosta> it is
<cfcosta> 0.6.0 for both core and ui
<cfcosta> let me find the email, I remember posting about this
```

JStickler:
I think this PR has the information about what the tag should be (Kiali image label "istio-release-1.0".)

https://github.com/istio/istio/pull/7510

mwringe:
The 'istio-release-1.0' is not what we are doing here. We still need the versioned tag like always.

The 'istio-release-1.0' is tag that gets updated and points to the latest version of Kiali that if for Istio 1.0 (so it could be 0.6.0 today, and then 0.6.1 next week if we find a serious enough problem). The 'istio-release-1.0' stuff for now will need to be done manually until we can figure out how to automate it.

mwringe:
Lets use https://github.com/kiali/kiali-ui/pull/589 instead as it has more accurate description and has the proper jira number.

This change is not for this release, but for the next release. The one we are releasing now will be for 0.6.0

PR #592, opened 2018-08-17
KIALI-1217 Add a new Workload details page

aljesusg:
** Describe the change **

Add the page for WorkloadDetails
workload_details

** Issue reference **

[KIALI-12-17](https://issues.jboss.org/browse/KIALI-1217)

### DIFF
--- a/src/types/IstioObjects.ts
+++ b/src/types/IstioObjects.ts
@@ -1,0 +1,82 @@
+import Namespace from './Namespace';
+
+// NamespaceValidations are grouped per 'namespace'
+export type NamespaceValidations = { [key: string]: Validations };
+
+// validations are grouped per 'objectType' first in the first map and 'name' in the inner map
+export type Validations = { [key1: string]: { [key2: string]: ObjectValidation } };
+
+export interface ObjectValidation {
+  name: string;
+  objectType: string;
+  valid: boolean;
+  checks: ObjectCheck[];
+}
+
+export interface ObjectCheck {
+  message: string;
+  severity: string;
+  path: string;
+}
+
+export interface Reference {
+  name: string;
+  kind: string;
+}
+
+export interface ContainerInfo {
+  name: string;
+  image: string;
+}
+
+export interface Port {
+  protocol: string;
+  port: number;
+  name: string;
+}
+
+export interface Autoscaler {
+  name: string;
+  labels?: { [key: string]: string };
+  createdAt: string;
+  minReplicas: number;
+  maxReplicas: number;
+  targetCPUUtilizationPercentage: number;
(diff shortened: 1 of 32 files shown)


aljesusg:
Hooray !!! Updated

lucasponce:
@aljesusg thanks !
One comment, I think we are going to have "Health" at Workload level, similar as it was in Metrics.
So I suggest to create the icon and perhaps use the replica info as part of the tooltip.
We can implement it in a following step.

lucasponce:
Another comment.
In the current breadcrumb style (this will be migrated in the future but meanwhile we should use the current one).
We should have a "Workload Info" like in Services:
image
As we will change add in a next step the "Workload Inbound Metrics" / "Workload Outbound Metrics" tabs.

src/types/IstioObjects.ts
```diff
@@ -0,0 +1,82 @@
+import Namespace from './Namespace';
```
lucasponce: The file IstioObjects.ts is confusing as it is a little bit ambiguous, in theory the pattern for naming was to use the elements close of the pages.
I understand the goal, it was to reuse some common types between workloads/services.
But well, I guess we can refactor in the future, I don't have a better name.
Once all main entities are consolidated perhaps we can pure refactor some interfaces and classes and clean something as I am sure we have some redundant classes under the types folder.
But no any action required, just leaving here a note to not forget.

src/types/Workload.ts
```diff
@@ -1,8 +1,36 @@
...
+  labels: { [key: string]: string[] };
+  createdAt: Date;
+  resourceVersion: string;
+  replicas: Number;
+  availableReplicas: Number;
+  unavailableReplicas: Number;
+  autoscaler: Autoscaler;
+}
+
 export const WorloadLink = (ns: string, name: string) => {
```
lucasponce: remember, methods start with lowercase, I think I saw more in the code.

lucasponce:
Another topic I see:
image
The tooltip with number of warnings is good but this requires that you would enter in some details to show the warning details.
As we don't plan to have it, I would use the same way health did: showing the warning/error details in the tooltip, if not that information is missing and user doesn't know what is wrong in the pod.

Also, it happens as well as workload level, there is no warnings related the label validation, and that information should appear at this level as well.

aljesusg:
Comments addresed @lucasponce . Thanks
Added IstioSidecarIcon if it's enabled
latest

How we are displaying now the errors in pods
pods_warning

aljesusg:
increased the width in the tooltip
pods_warning

aljesusg:
hooray !!! Thanks @lucasponce

PR #598, opened 2018-08-20
KIALI-1366 Add handling for edges showing TCP traffic

israel-hdez:
A new option is added to the "Graph Settings" dropdown to be able to
show TCP traffic in the edges.

Edges with TCP traffic have a cue to be differentiated from those with
HTTP traffic. Coloring for TCP edges is PfBlue-600.

**Backwards in compatible?** It's compatible

image


israel-hdez:
Design issue for reference: https://github.com/kiali/kiali-design/issues/33

PR #604, opened 2018-08-21
KIALI-1323 Filtering homogenization

xeviknal:
** Describe the change **

Bugs fixed:

1. Filter duplication
screen recording

2.  Istio filter is now: Present, Not Present (instead of Deployed/Not Deployed)
screenshot of kiali console 1

screenshot of kiali console 2

3. Let filters be updated or appended: e.g. istio filter is a boolean, so have both filters shouldn't exist. Otherwise, namespace filter can be added several times with different values. (This is done to be consistent with the workload list)
screen recording 1

4. Before this fix, switching between workload/service/istioconfig list wasn't filtering properly with the filters brought from the previous list.
screen recording 2

** Issue reference **
[KIALI-1323](https://issues.jboss.org/browse/KIALI-1323)

Yes.

[x] Is your pull-request introducing changes in behaviour?

### DIFF
--- a/src/components/NamespaceFilter/NamespaceFilter.tsx
+++ b/src/components/NamespaceFilter/NamespaceFilter.tsx
@@ -141,7 +141,11 @@ export class NamespaceFilter extends React.Component<NamespaceFilterProps, Names
   filterValueSelected = (filterValue: FilterValue) => {
     const { currentFilterType, currentValue } = this.state;
 
-    if (filterValue && filterValue.id !== currentValue) {
+    if (
+      filterValue &&
+      filterValue.id !== currentValue &&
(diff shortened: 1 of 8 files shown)


lucasponce:
@xeviknal I wonder if you could rebase from master as there are new App filters that could be affected by this PR and perhaps we could run it in a single step.

lucasponce:
> I merged this after test in local @lucasponce I swa the DNM now....Did you add the label for any reason?

Please, next time, just ping me asking.
Nothing wrong but I wanted to check this too.

aljesusg:
Ok @lucasponce I did it because I saw the review request, after I saw the DNM. Sorry !!!

PR #612, opened 2018-08-23
KIALI-1370 Update release information

mwringe:
Updates the release documentation to let it be know that a line has to be removed from the travis.yaml file when doing a release. Otherwise it will add 'snapshot' to the release which is not what is wanted.

Update the .travis.yml file so that it will be able to push out npm packages on tags. Otherwise it will only allow the packages to be pushed from master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ script:
     yarn test --forceExit --maxWorkers=4
   - ./run_itest.sh
 before_deploy:
-  - yarn set-snapshot-version $TRAVIS_BUILD_NUMBER
+  - [ -z "$TRAVIS_TAG" ] && yarn set-snapshot-version $TRAVIS_BUILD_NUMBER || true
 deploy:
   skip_cleanup: true
   provider: npm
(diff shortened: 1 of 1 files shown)


josejulio:
Yet to test, but i think we can use something like
```
 [ -z "$TRAVIS_TAG" ] && yarn set-snapshot-version $TRAVIS_BUILD_NUMBER;
```
Instead of having to manually update that line.

> TRAVIS_TAG: If the current build is for a git tag, this variable is set to the tag’s name.

https://docs.travis-ci.com/user/environment-variables/

mwringe:
@josejulio Ah right, ok, I have updated the PR. Its now a bit more simple :) Thanks

josejulio:
Actually i was going to say that i can't see travis check for this PR, only the "security/snyk"

israel-hdez:
I tested this on bash, and it worked. I though that was temporary travis failure. But now that this is merged i see this:

image

Can you check @mwringe ?
Probably travis does not like the inline bash test.

PR #615, opened 2018-08-24
KIALI-1350 Avoid displaying 'Invalid Date' in Service Details page

xeviknal:
** Describe the change **

In Service Details page, avoid showing 'Invalid date' as it follows:
bildschirmfoto 2018-08-15 um 12 56 56

and show this:
screenshot of kiali console

The hyphen is shown when time is empty or has invalid format.

** Issue reference **
https://issues.jboss.org/browse/KIALI-1350

Yes.

### DIFF
--- a/src/components/Time/LocalTime.tsx
+++ b/src/components/Time/LocalTime.tsx
@@ -6,6 +6,14 @@ interface TimeProps {
 
 export default class LocalTime extends React.Component<TimeProps> {
   render() {
-    return new Date(this.props.time).toLocaleString();
+    let renderedTime: string;
+
+    if (this.props.time) {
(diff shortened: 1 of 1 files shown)


src/components/Time/LocalTime.tsx
```diff
@@ -6,6 +6,14 @@ interface TimeProps {
 
 export default class LocalTime extends React.Component<TimeProps> {
   render() {
-    return new Date(this.props.time).toLocaleString();
+    let renderedTime: string;
+
+    if (this.props.time) {
```
lucasponce: This also works.
```
return this.props.time ? new Date(this.props.time).toLocaleString() : '-';
```

PR #616, opened 2018-08-24
KIALI-1198: Add error icon and message to service graph

mtho11:
** Describe the change **
Adds the error icon and error message to the service graph page when an error occurs.

** Issue reference **
Jira: https://issues.jboss.org/browse/KIALI-1198

Compatible yes

** Screenshot **
Before:
before

After:
error-msg


israel-hdez:
It makes sense to remove the button. I'll open a new Jira to remove it

PR #617, opened 2018-08-24
KIALI-1421 KIALI-1422 Details page should show reporter="source" metrics

israel-hdez:
reporter=source metrics are the most reliable and complete ones for the metrics tab of the service and workload details pages.

### DIFF
--- a/src/components/Metrics/Metrics.tsx
+++ b/src/components/Metrics/Metrics.tsx
@@ -141,11 +141,11 @@ class Metrics extends React.Component<MetricsProps, MetricsState> {
         const charts = this.getChartsDef();
         Object.keys(charts).forEach(k => {
           const chart = charts[k];
-          const histo = response.data.dest.histograms[k];
+          const histo = response.data.source.histograms[k];
           if (histo) {
             chart.metrics = histo;
           } else {
-            chart.metrics = response.data.dest.metrics[k];
+            chart.metrics = response.data.source.metrics[k];
           }
         });
         this.setState({


lucasponce:
@israel-hdez could you also explain the intention of this change here too ?
I guess "Monday's call" is the scrum call, but we don't participate on that.

Did you follow the discussion in @jshaughn PR when this change was introduced ?

israel-hdez:
@lucasponce I followed the discussion at https://issues.jboss.org/browse/KIALI-1344 and looked at the resulting PRs.

Some background... Previously, the backend had the decision whether to query reporter="source" or reporter="destination". The decision was made based on the chosen tab (inbound vs outbound) and the kind of entity (workload, service, app). In general, reporter="destination" were being queried for inbound metrics and reporter="source" were being queried for outbound.

The current state is that the backend always returns both reporter (source and destination) metrics, leaving the choice to the UI about what to show. Currently, the UI is always displaying metrics reporter="destination" metrics. As a result, TCP metrics for _services_ aren't being shown (this is https://issues.jboss.org/browse/KIALI-1422) because for the _services_ case, TCP is only correctly reported from reporter="source".

For _workloads_ there is also a problem. For TCP-only workloads some HTTP traffic is being reported in the _outbound_ metrics (this is https://issues.jboss.org/browse/KIALI-1421). That HTTP traffic is own Istio infrastructure communication (sidecars -> istio-{policy,telemetry}). This probably happens on all workloads, but is only evident in TCP-only workloads which aren't supposed to have HTTP traffic.

In this PR my idea was to use reporter="source" metrics instead of the destination ones, since they look more reliable. But then, "border" services won't show inbound metrics because traffic from the _unknown_ source is reported only destination side.

Perhaps, I could do in the UI what was previously being done at the backend: show reporter="destination" for _inbound_ and reporter="source" for _outbound. This will solve KIALI-1421, but KIALI-1422 will persist because in the _services_ case, TCP is only correctly reported from reporter="source".

israel-hdez:
@lucasponce That sounds reasonable to me. Much better than trying to guess what're the better metrics to show. WDYT @mwringe?

mwringe:
We might want to bring this up on the list and discuss there.

If understand this correctly, we have two types of metrics:

- "source" -> the metrics the workload itself deals with

- "destination" -> the overall metrics, including the overhead of the Istio system

[I hope I didn't get the names mixed up, and I wish we had a more descriptive name for them. Calling them 'source' and 'destination' isn't intuitive to me]

So we need to figure out how we want to show those metrics. I think letting a user display those might be a good idea, but we don't give users this option on the graph view. So either the graph view is doing things right and we need to use that option on the details page. Or the graph view also needs to include an option to select what should be shown.

@jshaughn any comments on the above? ^

For KIALI-1421, what is causing those extra http calls? Is this just the traffic from the sidecar to the components in istio-system? If so, should we be filtering those out? or also including them?

jmazzitelli:
FWIW: "source" and "destination" metrics are simply the new names Istio gave their telemetry data - in Istio 0.8.0 they used to be named "client" and "server" respectively.

So "source" are the outgoing requests (outgoing bytes) sent out of the workload. "destination" are the incoming requests (incoming bytes) received by the workload.

I actually think client/server made more sense and that source/destination are even more confusing, but it is what it is. These come directly from Istio as seen in Prometheus.

israel-hdez:
@mwringe the extra http calls for KIALI-1421 is, as you say, traffic from the sidecar to istio components. I'm wondering if this traffic is useful for users. If it isn't useful, we should filter out.

My personal preference is that this traffic should, at least, be distinguishable. You can do this by using the "group by" option in the metrics page, but you can't get the "overall traffic minus istio traffic".

jshaughn:
The graph does not show istio nodes unless the user selects the istio namespace.  That is to keep the graph more simple as the istio components add a lot of complexity.  For metrics I think interactions with istio are valid and should be included.  I also think the user should be able to see both source and destination metrics for both inbound and outbound, in the detail charts.  Perhaps with only destination being initially enabled for inbound metrics (this will include requests from unknown and any istio component) and only source for outbound (this will include client failures).  Although it's not completely simple, it is possible to then distinguish the traffic distribution by going to the graph and examing the incoming or outgoing edges.  A workload/app/service-centric graph option would make this easier.  A graph that also showed source (incoming) or destination (outgoing) traffic breakdown could also be nice.

israel-hdez:
I'll proceed to include the dropdown to select source/destination metrics. It's probably the better choice for now.

israel-hdez:
Closing in favor of #622

lucasponce:
Closed as #622 was merged.

PR #620, opened 2018-08-27
KIALI-1345 Add a new Application details page

lucasponce:
Including
KIALI-1393 Add Inbound and Outbound metrics tab on App Details page

image

image


src/pages/AppDetails/AppDetailsPage.tsx
```diff
@@ -0,0 +1,164 @@
...
+      case 'out_metrics':
+        tab = 'Outbound Metrics';
+        break;
+      default:
+        tab = 'Info';
+    }
+    return (
+      <Breadcrumb title={true}>
+        <Breadcrumb.Item componentClass="span">
+          <Link to="/applications">Applications</Link>
```
xeviknal: you might be interested on clear filters `onClick` otherwise you will be filtering apps by filters stored in `NamespaceFilterSelected`

src/pages/AppDetails/AppInfo/AppDescription.tsx
```diff
@@ -0,0 +1,124 @@
...
+        <Link to={this.serviceLink(namespace, service)}>{service}</Link>
+      </div>
+    ));
+  }
+
+  renderWorkloadItem(namespace: string, workload: AppWorkload) {
+    /*
+      Not sure if we need a common icon per Workload instead of an icon per type of Workload
+     */
+    let iconName = WorkloadIcons.Deployment;
```
xeviknal: In my opinion, it is good to have an icon for each type :) It gives more information to the user, it is more explicit. And this should help to detect problems easier.

lucasponce: Yes, inside the "Workloads" page.
I mean, I know that we are under workloads, then a different icon per workload type (deployment/replicaset/etc.).
But in this page, I was not sure, It sounded weird to me to set a different icon per workload without bring the "workload details".
That was my "rationale" for this comment.
I guess I forgot to mention that context was related just to this page, not in general.

lucasponce:
Thx @xeviknal. I have rebased and addressed your comments.

PR #622, opened 2018-08-27
KIALI-1425 Add control to change reporter in the metrics detail pages

israel-hdez:
Showing metrics only for one reporter leads to some gaps in the available data. In order to fill the gaps, a dropdown is added to the metrics page to let the user choose the reporting entity to view the available metrics.

Related to KIALI-1421 KIALI-1422 which are about an issue with an inconsistency in the metrics being shown. This is not really an inconsistency, but a difference in the metrics, depending on the point of view of the reporter. The reporter could be the workload generating the traffic (source) or the workload receiving the traffic (destination).

_By default, inbound metrics are reporter="destination" and outbound are reporter="source"_

image

### DIFF
--- a/src/components/Metrics/Metrics.tsx
+++ b/src/components/Metrics/Metrics.tsx
@@ -29,10 +29,11 @@ type ChartDefinition = {
 };
 
 type MetricsState = {
-  charts: { [key: string]: ChartDefinition };
   alertDetails?: string;
   grafanaLink?: string;
   pollMetrics?: number;
+  metricReporter: string;
+  metricData?: any;
 };
 
 type ObjectId = {
@@ -55,8 +56,14 @@ class Metrics extends React.Component<MetricsProps, MetricsState> {
 
   constructor(props: MetricsProps) {
     super(props);
+
+    let metricReporter = 'source';
+    if (this.props.metricsType === 'inbound') {
+      metricReporter = 'destination';
+    }
+
     this.state = {
-      charts: this.getChartsDef()
+      metricReporter: metricReporter
     };
   }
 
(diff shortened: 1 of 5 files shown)


israel-hdez:
@lucasponce done!
Also changed the title and description.

src/components/Metrics/Metrics.tsx
```diff
@@ -123,6 +143,11 @@ class Metrics extends React.Component<MetricsProps, MetricsState> {
     this.setState({ pollMetrics: newRefInterval });
   };
 
+  onReporterChanged = (reporter: string) => {
+    console.log('New reporter: ', reporter);
```
lucasponce: Do you want this log ? or was just for debugging the feature ?

jshaughn:
I haven't run this, my initial input is more at the design level.  I do think the user should be able to see both the source and destination telemetry.  But I'm not sure a dropdown/toggle is the way to go.  I think I'd rather see two available chart lines, with only the default enabled.  This way the user has access to either or both on the same chart, without having to leave the chart focus.  Also, I'm concerned that "destination" and "source" may not have meaning to the user.  So I feel we may need a "?" icon with some information about what these things mean.  For example, "source telemetry is reported from the workloads originating the requests and includes request failures, network latency, etc. destination telemetry is reported from the workloads processing the requests and excludes network latency, etc"

lucasponce:
@israel-hdez thx, let me review/test it locally and I will merge it.

src/components/Metrics/Metrics.tsx
israel-hdez: It was for debugging. I'll remove it.

lucasponce:
@jshaughn your comment is right.
But I don't mind to have this as it is, or perhaps to unify with the "official" grafana dashboards provided with Istio, which are still using the old "client/server" terminology.

I don't mind to merge this first and iterate more in the wording next.

Both approaches sound good to me.

lucasponce:
@israel-hdez Do you mind to rebase with master ? Then I can test it with the global app metrics and check that nothing is missing.
Thanks !

israel-hdez:
@lucasponce Rebased.

PR #627, opened 2018-08-28
KIALI-1433 Fix edge font-size

jshaughn:
- also, formalize font for node and edge text

@cshinn , @xeviknal ,  This change fixes a regression such that the font-size of edge labels was getting ignored.  The fix involved setting a specific font-family.  The thing is, I can't find any document from UX that specifies a recommended font-family or font-size for node or label text.  So the chosen values are up for review.  See the screenshot below for the values I'm currently using, which are:

font (both node and edge):  Verdana, Arial, Helvetica, sans-serif  // see note
node font-size: 8px
edge font-size: 6px

Note that the font is a list, in order of preference, depending on what is available.  I just chose a common grouping of sans-serif fonts.

screenshot

### DIFF
--- a/src/components/CytoscapeGraph/graphs/GraphStyles.ts
+++ b/src/components/CytoscapeGraph/graphs/GraphStyles.ts
@@ -10,9 +10,11 @@ const EdgeColor = PfColors.Green400;
 const EdgeColorDead = PfColors.Black500;
 const EdgeColorDegraded = PfColors.Orange;
 const EdgeColorFailure = PfColors.Red;
+const EdgeFont = 'Verdana,Arial,Helvetica,sans-serif';
+const EdgeFontSize = '6px';
+const EdgeTextMargin = '6px';
 const EdgeWidth = 1;
 const EdgeWidthSelected = 3;
-const EdgeText = '6px';
 const NodeColorBorder = PfColors.Black400;
 const NodeColorBorderDegraded = PfColors.Orange;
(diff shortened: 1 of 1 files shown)


jmazzitelli:
I know it's not part of this PR, but why are the edge labels cutting through some of the edge labels? Sometimes the edge lines cut directly through the text (see "reviews v2->ratings v1" edge in your screenshot above - the 47% label)

cshinn:
The standard PatternFly font family is “Open Sans”, Helvetica, Arial, sans-serif

jshaughn:
@cshinn Thanks, I'll update the font-family list to replace Verdana with "Open Sans".  The 12px fonts cause much more line intersection and come across sort of busy.  If you are not opposed I'll leave it as is in the PR.

@jmazzitelli The edge label text sometimes crossing the edge line is really annoying.  It must have to do with the "autorotate" math just not being perfect with regards to the layout logic. And the fact that we are only offered x and y margins (we use x).  I feel like this has actually gotten worse more recently and I don't know why.  We could go back to not using autorotate and just let the labels have a standard horizontal layout in the middle of the edge.  I don't really know what is best.

cshinn:
@jshaughn if the smaller font helps reduce line cut-in, I'm all for it.

jshaughn:
@cshinn The smaller font does help the line cut-in issue, at least a little bit.

jshaughn:
@xeviknal 👍 

screenshot

PR #628, opened 2018-08-28
KIALI-1398 Add app detail links from graph side panel

jshaughn:
screenshot


jshaughn:
By the way, if you happen to go to an app that doesn't have detail (perhaps backed by a non-deployment at the moment), you get an OK message:

screenshot

You get a similar message for a missing workload (note that this PR fixes the workload message).

PR #631, opened 2018-08-29
KIALI-1442 Create cytoscape toolbar for zoom/fit and legend

josejulio:
** Describe the change **

Removes the panzoom cytoscape extension and uses the proposed design by UXD
https://github.com/kiali/kiali-design/blob/master/service-graph/design.md#graph-zoom-tool

This doesn't touch the legend window, only removes the legend from the "Graph Settings"

** Screenshot **

kiali-1442

edit:
Moved it a bit to the left and up, also added box shaddows
box-1442


jmazzitelli:
Neat. 
note that legend is going to be totally re-designed... it may end up being just a simple image to display (but its going to be close to 3 times the size of the current legend).

josejulio:
> note that legend is going to be totally re-designed

Right, that's the next step, display the dialog with the legend (one image)

josejulio:
I'll adjust the space, the left part seems to be from the layout, but i don't think i'll affect anything if I make it smaller.

josejulio:
I didn't had to modify the layout, just pulled the box to the left

box-1442

PR #632, opened 2018-08-30
KIALI-1156 Fix behavior of message center; add system errors panel

israel-hdez:
Is your pull-request introducing changes in behaviour? **Yes**

Per some discussion, the notification bell and the "open issues" message
aren't following the pattern suggested by UX. This fixes the behavior of
the messsage center, and the notifications in the main header of Kiali
page.

* Regular kiali errors and warnings (mainly, network errors) will now
show the standard bell with a badge when there are messages in the
message center. If there are errors or warnings, the badge will be red;
other types of messages will show the blue badge (we don't have any of
those kinds as of now). This is the standard PF behavior of the
notification bell and notification drawer.

image

* The warning "open issues" won't replace the notifications bell. For
now, the only issue that raise this kind of warning to show up is the
unsupported Istio version warning (reported by the backend).

image

* An "Open issues" panel is added to the message center. This panel is
visible only when there are system errors (for now, only the Istio version warning).
Messages in this new panel can be marked as read, but cannot be cleared.

image

_Open issues panel is open by default when clicking in the "open issues" message. "Notifications" panel is open by default when clicking the bell_

* Marking system errors as read will clear the notifications count, but won't clear the "open issues" message:

image

This is to remind the user that there is something wrong.

* Backend status is now fetched on browser refresh, if user is already
logged in. This is to fetch and raise the Istio version warning, if
needed. This ensures that the user is aware of the problematic
condition.

### DIFF
--- a/src/components/MessageCenter/MessageCenterTrigger.tsx
+++ b/src/components/MessageCenter/MessageCenterTrigger.tsx
@@ -1,31 +1,55 @@
 import * as React from 'react';
 import * as PfReact from 'patternfly-react';
 
-type StateType = {};
 type PropsType = {
   newMessagesCount: number;
+  systemErrorsCount: number;
   badgeDanger: boolean;
   toggleMessageCenter: () => void;
+  toggleSystemErrorsCenter: () => void;
 };
 
-export default class MessageCenterTrigger extends React.PureComponent<PropsType, StateType> {
+export default class MessageCenterTrigger extends React.PureComponent<PropsType, {}> {
   render() {
-    let icon;
-    if (this.props.newMessagesCount > 0) {
-      icon = (
-        <div>
-          <PfReact.Icon name="warning-triangle-o" type="pf" /> {this.props.newMessagesCount} open issues
-        </div>
-      );
-    } else {
-      icon = <PfReact.Icon name="bell" />;
+    return (
+      <>
+        {this.renderSystemErrorBadge()}
+        {this.renderMessageCenterBadge()}
+      </>
+    );
+  }
+
+  private renderSystemErrorBadge = () => {
+    if (this.props.systemErrorsCount === 0) {
+      return;
     }
+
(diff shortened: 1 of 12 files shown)


mwringe:
Where did 'system issues' come from? I don't remember that being mentioned in the discussions.

If this is something that a user does need to interact with (eg the version being unsupported) then why don't we want to use 'Open Issue' for this case?

israel-hdez:
It wasn't mentioned in the discussion. What I understood is that the suggestion was "the pattern" and the text could be whatever sounded ok for our case. So, I changed it to "system errors", because I think it describes better the situation. But I can revert the text if it's wrong. That's something quick.

src/components/MessageCenter/MessageCenterTrigger.tsx
```diff
@@ -1,31 +1,55 @@
...
+  private renderSystemErrorBadge = () => {
+    if (this.props.systemErrorsCount === 0) {
+      return;
     }
+
+    return (
+      <li className="drawer-pf-trigger">
+        <a className="nav-item-iconic" onClick={this.props.toggleSystemErrorsCenter}>
+          <PfReact.Icon name="warning-triangle-o" type="pf" /> {this.props.systemErrorsCount}
+          {this.props.systemErrorsCount === 1 ? ' system issue' : ' system issues'}
```
josejulio: We probably want to capitalize this

israel-hdez: Capitalized :)

israel-hdez:
@mwringe Reverted to "Open issues".
However, I kept the panel title in the message center with "System errors". Not sure if it would be good to also set the title to "Open issues". (see second image in the description)

israel-hdez:
The "System errors" panel title is now changed to "Open issues" for consistency.

PR #633, opened 2018-08-30
KIALI-1434 Fix Grafana link on App/Workload metric pages

lucasponce:
Grafana link was pointing always to Services dashboard, so, in App/Workload details it was showed a wrong empty dashboard.


lucasponce:
Ops, travis failed, need to update test snapshots.

PR #635, opened 2018-08-30
 KIALI-1078 Adds more info to the graph legend 

josejulio:
** Describe the change **

 - Node shapes
 - Node colors
 - Edges
 - Traffic animations

The legend content is an image, taken from https://github.com/kiali/kiali-design/pull/39
Specifically, used [this]( https://github.com/kiali/kiali-design/raw/e57285a2f1959f5f80c35e2848af06579472b616/service-graph/img/legend-details.png) image and modified with gimp.

This work is on top of https://github.com/kiali/kiali-ui/pull/631 but is not a strong dependency (IIRC) and could rebase if needed.

** Screenshot **

kiali-legend

Edit:

Updated to latest legend 
last-legend


mwringe:
I really don't think we want to use an image here. Its going to be a huge pain if we need to change anything.

Can we get this updated to be done more in html?

jmazzitelli:
I actually thought it would be EASIER to do it in a single image. No muss, no fuss. Its all static, so it should be perfect for an image. 

Think about all the things we have to do to get this to look right in html - we're still going to need images (like the shapes and the lines for edges) - except its going to be a bunch of images, with a bunch of html all around it, except for a single image.

I think we have one image, and we just use GIMP or some other image editor to edit it as we need in the future. Or better yet, get UXD to give us any changes in a new image showing what they want like they did this time (and since they will be the ones to tell us what the changes should look like anyway, this is probably the route we are going to take in the future). But even if UXD isn't involved, GIMP to edit the image (and then just commit the new image) would be much easier than dorking with several little images and html to fit them all together.

mwringe:
Using an SVG would probably help quite a bit. That would make opening and editing the file a lot easier.

I think there might also be problems with things like the fonts not exactly being the same if its an image/svg.

Being an image also prevents us from doing more advanced things. UXD was talking about making this responsive today and adjusting to the width of the screen. Adding in things like experimental features might also mean having multiple svg/image files to load.

I think eventually we will need to move towards doing this in html. But I don't think that needs to be a blocker here.

josejulio:
> Finally, I think it might be better if the legend starts in the lower or upper right as opposed to the lower left, such that less graph is covered.

Here are some combinations
Bottom right
legend-bottom-right

Top right
legend-top-right

Top Left
legend-top-left

Bottom Left
legend-bottom-left

Bottom right without occluding side panel
legend-bottom-right-2

Top right without occluding side panel
legend-top-right-2

jshaughn:
My vote is bottom right.  Try to cover the least amount of graph possible, and still maintain the top of the side panel.

josejulio:
> My vote is bottom right. Try to cover the least amount of graph possible, and still maintain the top of the side panel.

It makes sense there, and you can still move it around if is covering anything you would like to see

josejulio:
Updated the code with some comments about the fact that the image size is the same as the dialog (this is computed) and moved the dialog to the bottom right

josejulio:
Updated with [latest](https://github.com/kiali/kiali-design/pull/39#issuecomment-417433448) legend suggestion from @cshinn 

We have this already working. I could try to port it to HTML in other JIRA/PR and we could compare.

jshaughn:
OK, food improvements, let's get this in and iterate as needed.  But we do need some sort of "source" code for the image so that it can be edited going forward.

PR #636, opened 2018-08-31
KIALI-1436 List pages: Reset page number when filters change

xeviknal:
** Describe the change **

Wrong pagination counts on list pages after using filtering.

** Issue reference **
https://issues.jboss.org/browse/KIALI-1436

yes.

** Is your pull-request introducing changes in behaviour? **
Yes, it resets the page when changing filters.

** Screenshot **
screen recording 5

This behavior is applied also in App list, workload list and istio config list.

### DIFF
--- a/src/pages/AppList/AppListComponent.tsx
+++ b/src/pages/AppList/AppListComponent.tsx
@@ -62,7 +62,7 @@ class AppListComponent extends React.Component<AppListComponentProps, AppListCom
     this.updateApps();
   }
 
-  updateApps = () => {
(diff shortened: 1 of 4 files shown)


lucasponce:
As commented, I see a glitch when cleaning filters as page in the component is not changed, but the page parameter in the URL is gone.

xeviknal:
@lucasponce comments approached. 
It looks like this only happening into Workload list. The other list worked well.
The problem here is connected to code duplicity: I missed to add the resetPagination flag into that scenario.

PR #637, opened 2018-08-31
KIALI-1443 Ensure RCUE is loaded only when demanded

lucasponce:
** Describe the change **

UFO

** Issue reference **

https://issues.jboss.org/browse/KIALI-1443


lucasponce:
No, it is not complete.
Added DNM; it progress in the problem but require() doesn't work dynamically as expected.
Still working on it.

lucasponce:
@aljesusg can you test this approach ?
Before to build the container version, use the yarn remove-rcue, then rcue and rcue-additions will be empty and the container will deploy the right style.
Also, using yarn restore-rcue will return the files and the yarn start will work as usual.

If this works, it can be polished to include the remove-rcue on the build or using some conditional based on the same variable, but at least it is a workaround to create the desired built.

aljesusg:
It's working, maybe we should add this kind of tasks like this maybe and check a environment like KIALi_env, What do you think?
```json
    "start": "if [[ ${KIALI_ENV} == \"production\" ]]; then npm run start:prod; else npm run start:dev; fi",
    "start:kiali": "npm run copy-fonts && npm run copy-img && npm run build-css && npm run prettier && REACT_APP_VERSION=$npm_package_version REACT_APP_NAME=$npm_package_name REACT_APP_GIT_HASH=$(git rev-parse HEAD) react-scripts-ts start",
    "start:dev": "npm run remove-rcue; env REACT_APP_RCUE=false npm run start:kiali",
    "start:prod": "npm run restore-rcue; env REACT_APP_RCUE=true npm run start:kiali",
    "build": "if [[ ${KIALI_ENV} == \"production\" ]]; then npm run build:prod; else npm run build:dev; fi",
    "build:dev": "npm run remove-rcue; env REACT_APP_RCUE=false npm run build:kiali",
    "build:prod": "npm run restore-rcue; env REACT_APP_RCUE=true npm run build:kiali",
    "build:kiali": "npm run copy-fonts && npm run copy-img && npm run build-css && npm run prettier && REACT_APP_VERSION=$npm_package_version REACT_APP_NAME=$npm_package_name REACT_APP_GIT_HASH=$(git rev-parse HEAD) react-scripts-ts build",
    
```

abonas:
@lucasponce what is UFO?

lucasponce:
@aljesusg yes, I thought in something similar as you suggest.
Now that the solution works, I will take a look into the scripts.
Thanks for testing.

lucasponce:
> @lucasponce what is UFO?

An informal way to refer to a very strange issue, apparently trivial, that you can't explain.
(The JIRA has proper details).

lucasponce:
@aljesusg well, just one comment in the scripts:
```
    "build:dev": "npm run remove-rcue; env REACT_APP_RCUE=false npm run build:kiali",
```
Note that the REACT_APP_RCUE is useless on the build but it can be added for consistency.

PR #640, opened 2018-08-31
KIALI-1380 Add links to the website and github organization

israel-hdez:
[ ] Is your pull-request introducing changes in behaviour? **No**

--

Links are added to the 'about' dialog.

Links, icons and text are configurable in the config.ts file. The 'about' section config.ts file can be absent completely, or have only the webpage or project sections. The about box will hide and show what is available in the config file.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,6 +111,12 @@
   ],
   "author": "Red Hat",
   "license": "Apache-2.0",
+  "homepage-comment": [
+    "By default, Create React App produces a build assuming your app is hosted at the server root.",
+    "To override this, specify the homepage in your package.json:",
+    "We don't want this, so don't touch homepage.",
+    "https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#building-for-relative-paths"
+  ],
   "homepage": "/",
   "publishConfig": {
     "access": "public"
--- a/src/components/About/AboutUIModal.tsx
+++ b/src/components/About/AboutUIModal.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
-import { AboutModal } from 'patternfly-react';
+import { AboutModal, Icon } from 'patternfly-react';
 import { Component } from '../../store/Store';
-import { KialiLogo } from '../../config';
+import { config, KialiLogo } from '../../config';
 
 const KIALI_CORE_COMMIT_HASH = 'Kiali core commit hash';
 const KIALI_CORE_VERSION = 'Kiali core version';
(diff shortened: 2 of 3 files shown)


israel-hdez:
Unfortunately, since this is not in backend side, no env-vars nor configmap settings can be used. So, I'm using the already existent `config.ts` file as a place to unify these kind of settings.

israel-hdez:
I'm a little unsure about the icons. Mainly, because of the github icon. There are some guidelines here: https://github.com/logos

Guidelines forbids changing colors and sizes from the provided ones. FontAwesome allows to do both of these. Here, I'm respecting the white color. The size is surely changed. So... I don't know if this is OK.

package.json
```diff
@@ -111,7 +111,7 @@
   ],
   "author": "Red Hat",
   "license": "Apache-2.0",
-  "homepage": "/",
+  "homepage": "http://kiali.io",
```
josejulio: I think this setting is used by react. I'll check

josejulio: > By default, Create React App produces a build assuming your app is hosted at the server root.
To override this, specify the homepage in your package.json, for example:

https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#building-for-relative-paths

josejulio: It's not the first time this param is touched, its worth to add a ""comment"" above it:
```js
"homepage-comment": [
  "By default, Create React App produces a build assuming your app is hosted at the server root.",
  "To override this, specify the homepage in your package.json:",
  "We don't want this, so don't touch homepage."
  "https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#building-for-relative-paths"
]
```

israel-hdez: Yikes!
That's a different usage from what is documented at npm in https://docs.npmjs.com/files/package.json#homepage

>The url to the project homepage.
> Example:
>     "homepage": "https://github.com/owner/project#readme"

I will revert that and place the comment.

josejulio: I know, that's why i was suggesting a new entry in the package.json ("homepage-comment") and using multiple quotes when saying ""comment"" :)

That's a suggested way to comment stuff in the package.json

josejulio:
> Guidelines forbids changing colors and sizes from the provided ones. FontAwesome allows to do both of these. Here, I'm respecting the white color. The size is surely changed. So... I don't know if this is OK.

I think is fine, github seems to be aware of the icons https://github.com/FortAwesome/Font-Awesome/issues/8905

package.json
israel-hdez: Ok; comment added.

PR #648, opened 2018-09-05
KIALI-1235 Graph: lock icon disappear when json indicates that

xeviknal:
** Describe the change **
In graph, lock icon didn't disappear when json stop sending `isMTLS` field. It turns out that there was a missing default value which made the graph to keep the mTLS previous value.

** Issue reference **

https://issues.jboss.org/browse/KIALI-1235
https://issues.jboss.org/browse/KIALI-1238

** Backwards compatible? **
Yes.

Is your pull-request introducing changes in behaviour?
Yes.

Lock icon disappears when istio switch from using security connection to non-security one.

** Screenshot **

screen recording 2

### DIFF
--- a/src/actions/ServiceGraphDataActions.ts
+++ b/src/actions/ServiceGraphDataActions.ts
@@ -22,7 +22,8 @@ const decorateGraphData = (graphData: any) => {
       percentErr: undefined,
       percentRate: undefined,
       latency: undefined,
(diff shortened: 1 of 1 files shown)


hhovsepy:
Verified the PR, secure icon is gone when mTLS is disabled.
screenshot from 2018-09-05 13-10-03

PR #649, opened 2018-09-05
KIALI-1482 Clicking menu items erase applied filters

xeviknal:
** Describe the change **

bug fix: application and workload lists didnt removed the applied filters on clicking to its respective menu item.

** Issue reference **

https://issues.jboss.org/browse/KIALI-1482

** Backwards compatible? **
yes.

** Screenshot **

screen recording 4

screen recording 3

### DIFF
--- a/src/pages/AppList/AppListComponent.tsx
+++ b/src/pages/AppList/AppListComponent.tsx
@@ -260,7 +260,7 @@ class AppListComponent extends React.Component<AppListComponentProps, AppListCom
         isSortAscending: !this.state.isSortAscending,
         appListItems: sorted
       });
-      this.props.onParamChange([{ name: 'direction', value: this.state.isSortAscending ? 'desc' : 'asc' }]);
+      this.props.onParamChange([{ name: 'direction', value: this.state.isSortAscending ? 'asc' : 'desc' }]);
     });
   };
 
--- a/src/pages/AppList/AppListComponent.tsx
+++ b/src/pages/AppList/AppListComponent.tsx
@@ -143,6 +143,58 @@ class AppListComponent extends React.Component<AppListComponentProps, AppListCom
     });
   }
 
+  componentDidUpdate(prevProps: AppListComponentProps, prevState: AppListComponentState, snapshot: any) {
+    if (!this.paramsAreSynced(prevProps)) {
+      this.setState({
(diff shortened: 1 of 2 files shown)


israel-hdez:
(I haven't tested the PR, this is just a comment)

This approach solves the Jira :)  but I was expecting it to be solved doing a "no-op" when clicking the menu item. This feels like having a hidden "clear filters" option. 

Should the user be disoriented -> click again -> filters gone -> :worried: 

That's my $0.02

israel-hdez:
@lucasponce Yes, I agree with you. However, I was thinking in, for example, this flow:
* Click "workloads" in the menu to view the workloads list.
* Apply filters.
* Click (again) "workloads" in the menu.

This is the flow that seems to be shown in the gif. And the OS console does not clear filters in a similar flow. 

I like to think the main nav menu as a kind of "tab" control. So, if I click on an item that is already the active item, (usually) nothing changes. This is the base of my reasoning.

lucasponce:
> I like to think the main nav menu as a kind of "tab" control. So, if I click on an item that is already the active item, (usually) nothing should change. This is the base of my reasoning.

That suggestion is valid.
But question is what to do when I enter in the details page, the same navigation entry will be active.
Using back button will move to the previous URL which should have all parameters in place defining the filtering and pagination options.
Question is if we want to add a similar logic in the navigation item links.
Today those use a static link, and then the behaviour is correct, no parameters in the URL, no parameters in the filter.
Another enhancement could be to remember the previous state and to build the URL in base to that state.
That scenario brought more complexity and we decided to not covered as we have same result with the back button, but it could be implemented in a future iteration.

xeviknal:
@israel-hdez I know what you mean and I can quite agree with you. IMO, I would move this navigation question to UXD team instead of changing it into a bugfix issue. Also, because we should change it in all the pages, including the Graph page (currently it gets user back to a fresh page - no options changed).

israel-hdez:
@xeviknal @lucasponce Yes, I understand. I just wanted to expose my point of view and do a suggestion. I don't want to block this PR, because the proposed behavior certainly solves the Jira. It's just that, personally, I wouldn't expect this behavior and wanted to say what I like better :)
So, merge this as is (after reviews) and pass to UX.

PR #654, opened 2018-09-06
KIALI-1475: right align refresh button on service graph toolbar 

mtho11:
This is to make service graph page more match the other screens with refresh toolbar item on the right instead of left. No changes in behaviour.

** Issue reference **
https://issues.jboss.org/browse/KIALI-1475

** Backwards compatible? **
Yes

** Screenshot **

right-justified


jotak:
You referenced KIALI-1495 instead of KIALI-1475 in commit msg and PR title, which transitioned the wrong jira ticket... can you update please?

PR #655, opened 2018-09-07
KIALI-1491 Metrics pages don't propagate the parameters into the URL

xeviknal:
** Describe the change **
On both service, app, workload metrics page, when changing options they are mapped into param in the URL. They are not creating a new item in the history because navigation it is not implemented yet.

** Issue reference **
https://issues.jboss.org/browse/KIALI-1491

** Backwards compatible? **
yes.

** Screenshot **
screen recording 8

### DIFF
--- a/src/app/History.tsx
+++ b/src/app/History.tsx
@@ -3,3 +3,16 @@ import { createBrowserHistory } from 'history';
 const baseName = '/console';
 const history = createBrowserHistory({ basename: baseName });
 export default history;
+
+export namespace HistoryManager {
+  export const setParam = (name: string, value: string) => {
+    const urlParams = new URLSearchParams(history.location.search);
+    urlParams.set(name, value);
+    history.replace(history.location.pathname + '?' + urlParams.toString());
+  };
+
+  export const getParam = (name: string): string | null => {
+    const urlParams = new URLSearchParams(history.location.search);
+    return urlParams.get(name);
+  };
+}
--- a/src/components/Metrics/Metrics.tsx
+++ b/src/components/Metrics/Metrics.tsx
@@ -59,7 +59,11 @@ class Metrics extends React.Component<MetricsProps, MetricsState> {
     super(props);
 
     let metricReporter = 'source';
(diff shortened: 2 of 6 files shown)


src/app/History.tsx
```diff
@@ -3,3 +3,16 @@ import { createBrowserHistory } from 'history';
 const baseName = '/console';
 const history = createBrowserHistory({ basename: baseName });
 export default history;
+
+export namespace HistoryManager {
```
mtho11: Why a `namespace` and not a class with static methods?

src/components/Metrics/Metrics.tsx
```diff
@@ -59,7 +59,11 @@ class Metrics extends React.Component<MetricsProps, MetricsState> {
     super(props);
 
     let metricReporter = 'source';
-    if (this.props.metricsType === 'inbound') {
+
+    const metricReporterParam = HistoryManager.getParam('reporter');
+    if (metricReporterParam != null) {
```
mtho11: `null` is falsey so you can just use `if(!metricsReporterParam)`

src/app/History.tsx
```diff
@@ -3,3 +3,16 @@ import { createBrowserHistory } from 'history';
...
+  export const setParam = (name: string, value: string) => {
+    const urlParams = new URLSearchParams(history.location.search);
+    urlParams.set(name, value);
+    history.replace(history.location.pathname + '?' + urlParams.toString());
```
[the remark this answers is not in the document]
xeviknal: I think it is not a problem because there is no re-assignation for `urlParams`. It is changing the content of the var, but not re-assigning it.

src/app/History.tsx
mtho11: An object with static methods won't get created everytime.

src/app/History.tsx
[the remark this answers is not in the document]
mtho11: This should be ok. The Object referenced is not changing just properties within are mutated.

PR #690, opened 2018-09-20
KIALI-1580 Increase font size for highlighted graph elements

jshaughn:


jshaughn:
kiali-1580

jshaughn:
@cshinn  What do you think of this idea?  If it gets merged before you can comment we can easily revert.

jmazzitelli:
I like this **a lot**

PR #692, opened 2018-09-20
KIALI-1564 Always run build:dev in travis

josejulio:
** Describe the change **

Forces the dev build in travis to avoid using rcue in upstream builds


cfcosta:
LGTM, but I didn't test the generated images, did you? If not, is there a way to test this before merging it to master?

josejulio:
> LGTM, but I didn't test the generated images, did you? If not, is there a way to test this before merging it to master?

I didn't, but the way to test this would be to run the same commands as travis does and testing the local image.

PR #704, opened 2018-09-24
KIALI-1473 Highlight top of first selected item in nav bar

mwringe:
** Describe the change **

We were not highlighting the top of the first element in the vertical nav bar:

screenshot from 2018-09-24 16-34-14

While any other selected element did have it top highlighted:

screenshot from 2018-09-24 16-34-20

This change will now highlight the top of the first element in the vertical nav bar if its selected:

screenshot from 2018-09-24 16-34-25

Note: the top of the first selected item is a bit thicker, this is due to have the current top border is when things are not selected (see the second screenshot).

See https://issues.jboss.org/browse/KIALI-1473

### DIFF
--- a/src/app/App.scss
+++ b/src/app/App.scss
@@ -107,6 +107,10 @@ a:focus {
 .nav-pf-vertical .list-group-item {
   border-color: #18546b;
 }
-.nav-pf-vertical {
+.nav-pf-vertical .list-group .list-group-item:first-child {
   border-top: 2px solid #18546b;
 }
+.nav-pf-vertical .list-group .list-group-item.active:first-child {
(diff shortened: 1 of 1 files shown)


jshaughn:
@mwringe I'm impressed you found the tweak, I couldn't figure this out when I looked at it.  BUT, did you see my previous comment in the Jira?  This l+f may not be correct, it differs from OS and also the patternfly examples.

mwringe:
@jshaughn I think we need to figure out what is happening and why our L&F is different than that of the out of the box patternfly. At this point I don't know if this is as designed, or just something we did on our own.

PR #710, opened 2018-09-25
minor: add catch to some promises.

jshaughn:
@israel-hdez I think we should add these catch-blocks but not sure.  Please take a look.

### DIFF
--- a/src/actions/GraphDataActions.ts
+++ b/src/actions/GraphDataActions.ts
@@ -94,7 +94,31 @@ export const GraphDataActions = {
         restParams['includeIstio'] = true;
       }
       if (node) {
-        return API.getNodeGraphElements(authentication(), namespace, node, restParams).then(
+        return API.getNodeGraphElements(authentication(), namespace, node, restParams)
+          .then(
+            response => {
+              const responseData: any = response['data'];
+              const graphData = responseData && responseData.elements ? responseData.elements : EMPTY_GRAPH_DATA;
+              const timestamp = responseData && responseData.timestamp ? responseData.timestamp : 0;
+              dispatch(GraphDataActions.getGraphDataSuccess(timestamp, graphData));
+            },
+            error => {
+              let emsg: string;
+              if (error.response && error.response.data && error.response.data.error) {
+                emsg = 'Cannot load the graph: ' + error.response.data.error;
+              } else {
+                emsg = 'Cannot load the graph: ' + error.toString();
+              }
+              dispatch(MessageCenterActions.addMessage(emsg));
+              dispatch(GraphDataActions.getGraphDataFailure(emsg));
+            }
+          )
+          .catch(err => {
+            console.debug(API.getErrorMsg('Cannot load the graph: ', err));
+          });
+      }
+      return API.getGraphElements(authentication(), namespace, restParams)
+        .then(
           response => {
             const responseData: any = response['data'];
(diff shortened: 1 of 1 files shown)


rhqci:
:heavy_check_mark: Jenkins CI: kiali-ui-pr-e2e-test [#114](https://paste.fedoraproject.org/paste/iOiNo-M-74-YH2Ym7hf9Hg)
  * :heavy_check_mark: Istio-and-Kiali-on-Openshift [#[1393]](https://paste.fedoraproject.org/paste/eXinDROk0auU4TZgH8CoCw)
  * :heavy_check_mark: Istio-Upstream [#[268]](https://paste.fedoraproject.org/paste/CG6Bxv-edGBp9OGNH-rcig)
  * :heavy_check_mark: Kiali-Istio-Checker [#[1336, 1337]](https://paste.fedoraproject.org/paste/3suykl8NATiq9hU28skiCQ)
  * :heavy_check_mark: install-bookinfo [#[753]](https://paste.fedoraproject.org/paste/ZLeEu~Eprff~3nvzwdyPUw)
  * :heavy_check_mark: Kiali-on-Openshift [#[763]](https://paste.fedoraproject.org/paste/d82pkKvUnmXsTEf1OGCBuQ)
  * :heavy_check_mark: run-kiali-e2e-tests [#[209]](https://paste.fedoraproject.org/paste/OmoRG~TC1F~YlLKaM2gDhw)


KIALI_NAME=kialiuipr710-istio-system

src/actions/GraphDataActions.ts
```diff
@@ -111,26 +135,10 @@ export const GraphDataActions = {
...
-            emsg = 'Cannot load the graph: ' + error.toString();
-          }
-          dispatch(MessageCenterActions.addMessage(emsg));
-          dispatch(GraphDataActions.getGraphDataFailure(emsg));
-        }
-      );
+        )
+        .catch(err => {
+          console.debug(API.getErrorMsg('Cannot load the graph: ', err));
+        });
```
israel-hdez: Is the goal to log messages to the console?
The `then` block is already "catching" errors because two arguments are being passed. (See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#Methods_2)

You could log to the console in the second argument of the `then` (the `error` arrow fn).

TBH, I prefer to have individual `then` and `catch` blocks because it makes more evident what code is handling the success and what code is catching errors. So, if you want to "split" the `then` I will be happy to merge that change.

src/actions/GraphDataActions.ts
```diff
@@ -94,7 +94,31 @@ export const GraphDataActions = {
...
+                emsg = 'Cannot load the graph: ' + error.response.data.error;
+              } else {
+                emsg = 'Cannot load the graph: ' + error.toString();
+              }
+              dispatch(MessageCenterActions.addMessage(emsg));
+              dispatch(GraphDataActions.getGraphDataFailure(emsg));
+            }
+          )
+          .catch(err => {
+            console.debug(API.getErrorMsg('Cannot load the graph: ', err));
```
mtho11: This is a serious error that user should be notified of via Message Center and `console.error` not `console.debug`

src/actions/GraphDataActions.ts
jshaughn: @israel-hdez No, it's fine.  This is why I asked for review.  I understand now that this would be redundant handling. Thanks, I'm going to close this.

PR #120, opened 2018-03-14
Change title to Kiali from default Patternfly

mwringe:

PR #713, opened 2018-09-26
KIALI-1618 Overview: sort namespaces after refresh

jotak:
Also:
- avoid refresh after changing sort field, it's not needed
- reverse sort order for status: by default, failures first

JIRA: https://issues.jboss.org/browse/KIALI-1618

### DIFF
--- a/src/components/ListPage/ListPage.ts
+++ b/src/components/ListPage/ListPage.ts
@@ -25,6 +25,7 @@ export namespace ListPage {
     setFiltersToURL: (filterTypes: FilterType[], filters: ActiveFilter[]) => ActiveFilter[];
     filtersMatchURL: (filterTypes: FilterType[], filters: ActiveFilter[]) => boolean;
     isCurrentSortAscending: () => boolean;
+    currentSortFieldId: () => string | undefined;
     currentDuration: () => number;
     currentPollInterval: () => number;
   }
@@ -168,6 +169,10 @@ export namespace ListPage {
 
     isCurrentSortAscending(): boolean {
       return (this.getSingleQueryParam('direction') || 'asc') === 'asc';
+    }
+
+    currentSortFieldId(): string | undefined {
+      return this.getSingleQueryParam('sort');
     }
 
     currentDuration() {
--- a/src/pages/Overview/FiltersAndSorts.ts
+++ b/src/pages/Overview/FiltersAndSorts.ts
@@ -60,11 +60,11 @@ export namespace FiltersAndSorts {
       isNumeric: false,
       param: 'h',
       compare: (a: NamespaceInfo, b: NamespaceInfo) => {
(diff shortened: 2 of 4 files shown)


src/components/ListPage/ListPage.ts
```diff
@@ -170,6 +171,10 @@ export namespace ListPage {
       return (this.getSingleQueryParam('direction') || 'asc') === 'asc';
     }
 
+    currentSortFieldId(): string | undefined {
```
jotak: @xeviknal I think we could do something better here but I restrained to do it since you're doing some UI refactoring, that would more likely fall into your domain.
What I thought was:, in `ListPage`, having `currentSortField(): SortField` instead of `currentSortFieldId(): string | undefined`
but that's currently impossible due to `SortField` being different classes depending on the context (one for apps, one for workloads etc.)
I believe with the refactoring we could have a common interface `SortField<T>` for all of them. Do you agree?

rhqci:
:heavy_check_mark: Jenkins CI: kiali-ui-pr-e2e-test [#120](https://paste.fedoraproject.org/paste/XPcHepZfq~GTzSD-DdOzwA)
  * :heavy_check_mark: Istio-and-Kiali-on-Openshift [#[1403]](https://paste.fedoraproject.org/paste/oGHQOoOG6HcfOClEu39n2A)
  * :heavy_check_mark: Istio-Upstream [#[276]](https://paste.fedoraproject.org/paste/8cNMJB8qKG2LaAOea6RMNw)
  * :heavy_check_mark: Kiali-Istio-Checker [#[1358, 1359]](https://paste.fedoraproject.org/paste/SMI-fH2G6qa9nP0tMrV4xg)
  * :heavy_check_mark: install-bookinfo [#[770]](https://paste.fedoraproject.org/paste/BXgl~nhTuj-gHWTpZXvgFA)
  * :heavy_check_mark: Kiali-on-Openshift [#[772]](https://paste.fedoraproject.org/paste/Dk~vLfPrl95kcjdFy0ZGHA)
  * :heavy_check_mark: run-kiali-e2e-tests [#[220]](https://paste.fedoraproject.org/paste/LDoGfwwvYzd1sNVvcfATCw)


KIALI_NAME=kialiuipr713-istio-system

src/components/ListPage/ListPage.ts
xeviknal: @jotak I do agree. I am playing a bit with `SortField<T>` parameterized type. What I have doubts if TS will allow us to return `SortField` when we have `SortField<T>` as a type.

jotak: No it's fine with generics then the signature would be `currentSortField(): SortField<T>` which is really fine (maybe the whole `ListPage` would have to have parameter `<T>` ... I think it can work, let me know / chat if there's some troubles with that)

PR #719, opened 2018-09-28
KIALI-1571 Replace use of parseInt with Number()

burmanm:
Also ban parseInt in the tslint so future use is discouraged


rhqci:
!!! Couldn't read commit file !!!

src/components/GraphFilter/GraphRefresh.tsx
```diff
@@ -23,7 +23,7 @@ const GraphRefresh: React.SFC<GraphRefreshProps> = props => {
 
   const formatRefreshText = (key, isTitle: boolean = false): string => {
     // Ensure that we have an integer (for comparisons).
-    key = parseInt(key, 10);
+    key = Number(key);
```
lucasponce: Isn't skipping the default value ?

lucasponce: Ah, 10 is the base. Sorry.

PR #720, opened 2018-09-28
KIALI-1634 Make yarn use standard POSIX comparators for sh commands

burmanm:
Ubuntu uses dash as a standard for /bin/sh, so the current version of yarn does not work correctly.


rhqci:
!!! Couldn't read commit file !!!

gbaufake:
kialiqeci

PR #732, opened 2018-10-05
KIALI-1695 Performance issues on apps/svc/workload list pages

jotak:
- When rate interval changes, do not trigger an explicit reload as state change will trigger one (avoiding 2 reloads for the same action)
- When fetching backend data, filter it before calling for health (avoiding unnecessary health fetch)
- Remove incorrect health fetching from workloads 'ItemDescription' that triggered additional and incorrect queries

JIRA: https://issues.jboss.org/browse/KIALI-1695

### DIFF
--- a/src/pages/AppList/AppListComponent.tsx
+++ b/src/pages/AppList/AppListComponent.tsx
@@ -62,9 +62,8 @@ class AppListComponent extends ListComponent.Component<AppListComponentProps, Ap
   }
 
   rateIntervalChangedHandler = (key: number) => {
-    this.setState({ rateInterval: key });
     HistoryManager.setParam(URLParams.DURATION, String(key));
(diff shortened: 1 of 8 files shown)


src/pages/AppList/AppListComponent.tsx
```diff
@@ -62,9 +62,8 @@ class AppListComponent extends ListComponent.Component<AppListComponentProps, Ap
   }
 
   rateIntervalChangedHandler = (key: number) => {
-    this.setState({ rateInterval: key });
     HistoryManager.setParam(URLParams.DURATION, String(key));
-    this.updateListItems();
+    this.setState({ rateInterval: key });
```
xeviknal: I guess that here, we avoid the double trigger of the updateListItems()

PR #747, opened 2018-10-12
Customize persist key

clyang82:
Signed-off-by: Chun Lin Yang <clyang@cn.ibm.com>

** Describe the change **

right now, the persist key is hard-coded as `root` in config. it may conflict with other application which is also using `persist:root` as key in localStorage. Since Kaili supports customize base path (fixed in https://github.com/kiali/kiali-ui/pull/569) so that we should get the WEB_ROOT instead of hardcode.

** Backwards compatible? **

Yes

### DIFF
--- a/src/store/ConfigStore.ts
+++ b/src/store/ConfigStore.ts
@@ -16,8 +16,11 @@ import { INITIAL_STATUS_STATE } from '../reducers/HelpDropdownState';
 
 declare const window;
 
+const webRoot = (window as any).WEB_ROOT ? (window as any).WEB_ROOT : undefined;
+const persistKey = webRoot && webRoot !== '/' ? webRoot.substring(1) : 'root';
+
 const persistConfig = {
-  key: 'root',
+  key: persistKey,
   storage: storage,
   whitelist: ['authentication', 'statusState', 'userSettings']
 };
--- a/src/store/ConfigStore.ts
+++ b/src/store/ConfigStore.ts
@@ -17,7 +17,7 @@ import { INITIAL_STATUS_STATE } from '../reducers/HelpDropdownState';
 declare const window;
 
 const webRoot = (window as any).WEB_ROOT ? (window as any).WEB_ROOT : undefined;
-const persistKey = webRoot && webRoot !== '/' ? webRoot.substring(1) : 'root';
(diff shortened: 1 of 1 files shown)


rhqci:

pilhuhn:
@clyang82 Thanks for your contribution, I think that makes sense.
One question though: if you are worried that 'root' is taken and that when you put a context root of e.g. '/monitoring' in, wouldn't there be an equal chance that a persist key of 'monitoring' is taken?

Would it make sense here to do a composite key, that starts with e.g. 'io.kiali' to indicate that whatever you use it points to Kiali?

jmazzitelli:
adding @mtho11 as reviewer

jmazzitelli:
> Would it make sense here to do a composite key, that starts with e.g. 'io.kiali' to indicate that whatever you use it points to Kiali?

+1 I think we should have the root key be more specific to "kiali" regardless of context root (why not just rename "root" to "kiali", for example?")

src/store/ConfigStore.ts
```diff
@@ -16,8 +16,11 @@ import { INITIAL_STATUS_STATE } from '../reducers/HelpDropdownState';
 
 declare const window;
 
+const webRoot = (window as any).WEB_ROOT ? (window as any).WEB_ROOT : undefined;
+const persistKey = webRoot && webRoot !== '/' ? webRoot.substring(1) : 'root';
```
mtho11: This is a great idea! could we also maybe change 'root' to 'kiai-root'

clyang82: > 
> Would it make sense here to do a composite key, that starts with e.g. 'io.kiali' to indicate that whatever you use it points to Kiali?

That is great idea to using composite key to ensure it is unique. I would like to using 'kiali' as a prefix. Thanks.

clyang82:
> 
> Would it make sense here to do a composite key, that starts with e.g. 'io.kiali' to indicate that whatever you use it points to Kiali?

@pilhuhn That is great idea to using composite key to ensure it is unique. I would like to using 'kiali' as a prefix. Thanks.

pilhuhn:
With that change and as @mtho11 approved, I am merging this.

Thanks a lot @clyang82 👍

PR #774, opened 2018-10-26
KIALI-1830 Fix sparkline issue for edge leading from service

jshaughn:
Fix sparkline issue for edge leading from service and including traffic from unknown
- make sure this fix does not break the fix in kiali-1732 for TCP sparkine
  for edges leading from a service.

After the fix we have both sparklines correct, The http sparkline combines  traffic from unknown and ingress:

image

The TCP traffic is shown corrrectly as well:

image

### DIFF
--- a/src/pages/Graph/SummaryPanelEdge.tsx
+++ b/src/pages/Graph/SummaryPanelEdge.tsx
@@ -215,10 +215,11 @@ export default class SummaryPanelEdge extends React.Component<SummaryPanelPropTy
 
     this.metricsPromise.promise
       .then(response => {
+        // HTTP
         let useDest = sourceData.nodeType === NodeType.UNKNOWN;
-        useDest = useDest || this.props.namespace === 'istio-system';
-        const reporter = useDest ? response.data.dest : response.data.source;
-        const metrics = reporter.metrics;
+        useDest = useDest || this.props.namespace === 'istio-system' || sourceData.nodeType === NodeType.SERVICE;
+        let reporter = useDest ? response.data.dest : response.data.source;
+        let metrics = reporter.metrics;
         const histograms = reporter.histograms;
         const reqRates = this.getNodeDataPoints(
           metrics['request_count_in'],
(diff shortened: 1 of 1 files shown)


src/pages/Graph/SummaryPanelEdge.tsx
```diff
@@ -262,6 +263,11 @@ export default class SummaryPanelEdge extends React.Component<SummaryPanelPropTy
           destMetricType,
           sourceData
         );
+        // TCP (uses slightly different reporting)
+        useDest = sourceData.nodeType === NodeType.UNKNOWN;
+        useDest = useDest || this.props.namespace === 'istio-system';
```
jmazzitelli: I'm pretty sure we have a variable or something that gives us the istio-system namesapce so we do not hardcode it. There is a way in kiali that the user can actually configure it for a different name - so we need to make sure we rely on that. See: https://github.com/kiali/kiali/blob/master/config/config.go#L161 - I'm sure the UI is getting this value and storing it somewhere that we can use rather than hardcoding istio-system.

rhqci:
:heavy_check_mark: Jenkins CI: kiali-ui-pr-e2e-test [#276](https://paste.fedoraproject.org/paste/tCgjL4Nnkh5daVwbdWDBDw)
  * :heavy_check_mark: Istio-and-Kiali-on-Openshift [#[1793]](https://paste.fedoraproject.org/paste/P6N6DtvCarF34WlQDIZdPA)
  * :heavy_check_mark: Istio-Upstream [#[637]](https://paste.fedoraproject.org/paste/7Z-ZtwwUGblY3ADw~5LcbQ)
  * :heavy_check_mark: Kiali-Istio-Checker [#[2138, 2139]](https://paste.fedoraproject.org/paste/J9tKKcGRMIXq1p4W5Llnjw)
  * :heavy_check_mark: install-bookinfo [#[1170]](https://paste.fedoraproject.org/paste/5bxz1U9zJS87ANxJ-Fvynw)
  * :heavy_check_mark: Kiali-on-Openshift [#[1168]](https://paste.fedoraproject.org/paste/6sPW799kFP3tAZKo6D1KlQ)
  * :heavy_check_mark: run-kiali-e2e-tests [#[571]](https://paste.fedoraproject.org/paste/mZXWn4CnJ3GfTbwfZCbHKw)


KIALI_NAME=kialiuipr774-istio-system

src/pages/Graph/SummaryPanelEdge.tsx
jshaughn: I agree it's a problem but afaik we don't have that config available yet in the UI.

PR #779, opened 2018-10-30
KIALI-1800 Avoid overlapping of compounds nodes

josejulio:
** Describe the change **

In order to keep compound nodes from being bigger than needed, a pre-process is done on the input to:

1. Remove child nodes and reserve enough size for them.
2. Create synthetic edges to replace edges that go or leave from a compound node.

After the layout is finished, the child nodes are added back and the synthetic edges are removed.

This should work for any discrete layout and any dynamic layout that issues `layoutstop`.

This allows us to keep all the compound nodes as tight as possible, preventing overlap (if the layout prevents that) as the layout positioning only happen on 1st level nodes and we should not have any unwanted surprise when a child node is positioned  (by the layout) on the top left of the screen while the other is positioned on the bottom right of the screen.

Code is mostly done -I have a few things to polish, like setting back values I affected-. I will need to test to ensure is always yielding right results and i'm not missing anything.

Below are some screenshots, but If you have seen a graph with lots of overlaps, please share the json used to generate that.

** Screenshot **

 ### Bookinfo step by step

I'm posting a step by step of what I'm doing to help spot problems. Even though bookinfo doesn't have this problem, I'm going to use it for simplicity along with dagre layout:

 1. Bookinfo with dagre without any change

bookinfo-before

 2. Remove children and nodes

bookinfo-step-1

 3. Reserve enough space for the children (this lets know the layout how big the node needs to be)

bookinfo-step-2

 4. Add synthetic edges; for every node that was connected to any children, it creates a synthetic edge from the node to the compound, and from compound to nodes as well.

bookinfo-step-3

 5. After the layout is finished, the synthetic edges are removed and the children are added back, ordering from top to bottom. We could use other strategies here, row, column, matrix, etc.

bookinfo-step-4

 6. Then we put back the regular edges.

bookinfo-step-5

And here is what it looks like in a gif:

bookinfo-gif

Notice how cola and cose both looks similar, i'm not sure if is because of the removal of the children or if I have a bug and i'm using the same layout for both :)

 ### KIALI-1013 faulty graph

I tried with the oudated json found in [KIALI-1013](https://issues.jboss.org/browse/KIALI-1013) and got the following results
Note that the nodes say "error/unknown" because is an old json that doesn't have the new properties.

 #### Dagre 
 It crashed before, it seems to be working now, maybe is because less nodes are being processed with this approach.
kiali-1013-dagre

 #### Cola
 Freeze my browser before and now it does the same (Eventually it will load, but since it takes ages, is not worth the wait)

edit: It loaded while i was finishing this.

kiali-1013-cola

 #### Cose 

Notice how the before has some overlaps and the after does not (so much).

- Before:
 cose-layout-before
- After:
kiali-1013-cose

~edit: It looks i have something not as good as I want, when adding the children, it appears as if the parent moves.~
edit: It seems that even the cola layout is having problems with only the compound nodes (without children)
kiali-1013-cose-boxes

### DIFF
--- a/src/components/CytoscapeGraph/CytoscapeGraph.tsx
+++ b/src/components/CytoscapeGraph/CytoscapeGraph.tsx
@@ -371,7 +371,12 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
       // overlap, but we need to have them enabled (nodeDimensionsIncludeLabels: true)
       this.turnNodeLabelsTo(cy, true);
       const layoutOptions = LayoutDictionary.getLayout(this.props.graphLayout);
-      cy.layout({ ...layoutOptions, name: 'group-compound-layout', realLayout: this.props.graphLayout.name }).run();
+      if (cy.nodes('$node > node').length > 0) {
+        // if there is any parent node, run the group-compound-layout
+        cy.layout({ ...layoutOptions, name: 'group-compound-layout', realLayout: this.props.graphLayout.name }).run();
+      } else {
+        cy.layout(layoutOptions).run();
+      }
     }
 
     // Create and destroy labels
--- a/src/components/CytoscapeGraph/CytoscapeGraph.tsx
+++ b/src/components/CytoscapeGraph/CytoscapeGraph.tsx
@@ -370,7 +370,8 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
       // Enable labels when doing a relayout, layouts can be told to take into account the labels to avoid
       // overlap, but we need to have them enabled (nodeDimensionsIncludeLabels: true)
       this.turnNodeLabelsTo(cy, true);
-      cy.layout(LayoutDictionary.getLayout(this.props.graphLayout)).run();
+      const layoutOptions = LayoutDictionary.getLayout(this.props.graphLayout);
+      cy.layout({ ...layoutOptions, name: 'group-compound-layout', realLayout: this.props.graphLayout.name }).run();
     }
 
     // Create and destroy labels
--- a/src/components/CytoscapeGraph/Layout/GroupCompoundLayout.ts
+++ b/src/components/CytoscapeGraph/Layout/GroupCompoundLayout.ts
@@ -1,0 +1,142 @@
+const NAMESPACE_KEY = '_group_compound_layout';
+const CHILDREN_KEY = 'children';
+const PADDING_WIDTH = 5;
+const BETWEEN_NODES_PADDING = 3;
+
+const normalizeToParent = element => {
+  return element.isChild() ? element.parent() : element;
+};
+
+class SyntheticEdgeGenerator {
+  private nextId = 0;
+  private generatedMap = {};
+
+  public getEdge(source: any, target: any) {
+    const sourceId = normalizeToParent(source).id();
+    const targetId = normalizeToParent(target).id();
+    const key = `${sourceId}->${targetId}`;
+
+    if (this.generatedMap[key]) {
+      return false;
+    }
+    return (this.generatedMap[key] = {
+      group: 'edges',
+      data: {
+        id: 'synthetic-edge-' + this.nextId++,
+        source: sourceId,
+        target: targetId
+      }
+    });
+  }
+
+  public getGeneratedEdges() {
+    return Object.keys(this.generatedMap).map(key => this.generatedMap[key]);
+  }
+}
+
+export default class GroupCompoundLayout {
+  private options;
+  private cy;
+  private elements;
+  private syntheticEdgeGenerator;
+
+  constructor(options: any) {
+    this.options = { ...options };
+    this.cy = this.options.cy;
+    this.elements = this.options.eles;
+    this.syntheticEdgeGenerator = new SyntheticEdgeGenerator();
+  }
+
+  run() {
+    const { realLayout } = this.options;
+    const parents = this.parents();
+    if (parents.length > 0) {
+      // Prepare parents by assigning a width and height
+      parents.each(parent => {
+        const boundingBox = parent.children().reduce(
+          (accBoundingBox, child) => {
+            const localBoundingBox = child.boundingBox();
+            accBoundingBox.height += localBoundingBox.h;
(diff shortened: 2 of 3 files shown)


josejulio:
Is still WIP but I will be happy if I can have some comments on the method and graph jsons (if possible with overlaps)

//cc: @mtho11 @jshaughn

rhqci:
:heavy_check_mark: Jenkins CI: kiali-ui-pr-e2e-test [#286](https://paste.fedoraproject.org/paste/y66TCklReMqT-Bs-qwJehg)
  * :red_circle: Istio-and-Kiali-on-Openshift [#[1843]](https://paste.fedoraproject.org/paste/qHQ~0uSyzUJye2YG6mhwNw)
  * :heavy_check_mark: Istio-Upstream [#[686]](https://paste.fedoraproject.org/paste/KBBx4fwvA6z8ad43qMwAYg)
  * :heavy_check_mark: Kiali-Istio-Checker [#[2240]](https://paste.fedoraproject.org/paste/YT9rckZuSUyAdPWx75U4Ew)
  * :red_circle: install-bookinfo [#[1227]](https://paste.fedoraproject.org/paste/DKW-6XnsdljmpO0gbMV7YA)


KIALI_NAME=kialiuipr779-istio-system

josejulio:
@jmazzitelli I know you like to break PRs

src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -370,7 +370,8 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
       // Enable labels when doing a relayout, layouts can be told to take into account the labels to avoid
       // overlap, but we need to have them enabled (nodeDimensionsIncludeLabels: true)
       this.turnNodeLabelsTo(cy, true);
-      cy.layout(LayoutDictionary.getLayout(this.props.graphLayout)).run();
+      const layoutOptions = LayoutDictionary.getLayout(this.props.graphLayout);
+      cy.layout({ ...layoutOptions, name: 'group-compound-layout', realLayout: this.props.graphLayout.name }).run();
```
jshaughn: Note that only graphType==versionedApp uses compound/group nodes.  The other types, workload, app, and service don't do any grouping.  So, this special layout could be applied only when necessary.

src/components/CytoscapeGraph/Layout/GroupCompoundLayout.ts
```diff
@@ -0,0 +1,174 @@
+const NAMESPACE_KEY = '_group_compound_layout';
```
jshaughn: This file would benefit from some comments.  It will be hard for anyone looking at this code later to make changes, it's not simple.  Add some high-level approach and some lower level details in tricky spots.  I know what you are tying to do and I'm only partially grokking it.  Maybe explain that this is a new "synthetic" layout, how the bounding box is calculated, etc..  As for the bounding box, I don't really understand, it is creating a single large node to replace the member nodes?  How is that bounding box sized, is it somehow based on the original node sizes?

josejulio:
> Also, is there any noticeable perf impact?

It does not look like there is any, but openshift already puts a lot of stress into my laptop so i might be wrong

src/components/CytoscapeGraph/Layout/GroupCompoundLayout.ts
josejulio: > As for the bounding box, I don't really understand, it is creating a single large node to replace the member nodes? How is that bounding box sized, is it somehow based on the original node sizes?

Is using the parent as a large single node, using the size of the children.

I'll add more comments into the file to make it easier to follow up

josejulio:
@jshaughn I added more information (as comments) of what the code does, I hope this make it easier to understand what's going on there.

rhqci:
:heavy_check_mark: Jenkins CI: kiali-ui-pr-e2e-test [#299](https://paste.fedoraproject.org/paste/qbSUsmDYxZ28KYWjCeTzgw)
  * :red_circle: Istio-and-Kiali-on-Openshift [#[1876]](https://paste.fedoraproject.org/paste/bixl1gWQAXojvlf~e~5dAw)
  * :heavy_check_mark: Istio-Upstream [#[723]](https://paste.fedoraproject.org/paste/7t4MmSoNW-sUPwgvK1jtTA)
  * :heavy_check_mark: Kiali-Istio-Checker [#[2289, 2291]](https://paste.fedoraproject.org/paste/aKE~ghNUdmUGU3UiUN0cnw)
  * :heavy_check_mark: install-bookinfo [#[1261]](https://paste.fedoraproject.org/paste/2R2iTwSyPAv6ng7y00Tdfg)
  * :heavy_check_mark: Kiali-on-Openshift [#[1239]](https://paste.fedoraproject.org/paste/NxPzvix0VowKJU3VcYOFow)
  * :red_circle: run-kiali-e2e-tests [#[631]](https://paste.fedoraproject.org/paste/fTceU665HeA8hIDTOAqJBg)


KIALI_NAME=kialiuipr779-istio-system

src/components/CytoscapeGraph/CytoscapeGraph.tsx
josejulio: I made it more general, this will only run if there are compound nodes.

src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -370,7 +370,13 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
       // Enable labels when doing a relayout, layouts can be told to take into account the labels to avoid
       // overlap, but we need to have them enabled (nodeDimensionsIncludeLabels: true)
       this.turnNodeLabelsTo(cy, true);
-      cy.layout(LayoutDictionary.getLayout(this.props.graphLayout)).run();
+      const layoutOptions = LayoutDictionary.getLayout(this.props.graphLayout);
+      if (cy.nodes('$node > node').length > 0) {
+        // if there is any parent node, run the group-compound-layout
```
jshaughn: OK, this general test is good.  One trivial comment: I think this comment would be more effective above the 'if' (no need to change).

PR #797, opened 2018-11-06
KIALI-1867 Istio Config - Filtering by Type Fails

aljesusg:
[KIALI-1867](https://issues.jboss.org/browse/KIALI-1867)
cc @lucasponce the backend returns a null value when there isn't items
Request /api/namespaces/bookinfo/istio?objects=destinationrules
```json
{"virtualServices":{"permissions":{"update":false,"delete":false},"items":null}}
```

I think that we should return empty array instead of null or set a omitEmpty to avoid this instead of make the change in the front. What do you think?


aljesusg:
Fixed in Backend with the last version and fixed in the front. Closing PR and I'll update the JIRA for @hhovsepy . Thanks @jotak

rhqci:
Jenkins CI: kiali-ui-pr-e2e-test [#324](https://paste.fedoraproject.org/paste/u6lh1Jio9hUCaveILnamQA)
 :red_circle: PRT Failed, Please Contact QE

PR #798, opened 2018-11-06
KIALI-1561 Add status-filtered links to overview page

jotak:
When clicking on a status icon in overview (ex: degraded in namespace X), it will bring you to the Apps list page with namespace=X and health=degraded filters.

JIRA: https://issues.jboss.org/browse/KIALI-1561


jotak:
@xeviknal I agree, I hesitated but think it's still ok for 2 args

PR #799, opened 2018-11-06
KIALI-1872 Support configured version label name in side panels

jshaughn:
image


jmazzitelli:
What about "app"? That is also customizable.

PR #800, opened 2018-11-07
KIALI-1852 User proper lock icon for inaccessible link in side panel

jshaughn:
@mwringe pointed me to the correct lock icon  (pficon-private), so there is no longer a re-use of the fa-lock icon and no an inaccessible node and it's links are handled with the same image.

image

### DIFF
--- a/src/pages/Graph/SummaryLink.tsx
+++ b/src/pages/Graph/SummaryLink.tsx
@@ -66,7 +66,7 @@ export const renderLink = (data: NodeData, nodeType?: NodeType) => {
   return (
     <>
       {link}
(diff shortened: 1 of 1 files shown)


jshaughn:
@jmazzitelli I don't know what to do to solve this snyk thing. It seems unrelated to this PR.

josejulio:
> Occasionally you might see a "Failed to detect issues" commit status which may block your PR. This means that we tried to run a test against your changes but unfortunately something went wrong / we encountered an internal problem. Please reach out if this happens to you and let us know which user, organisation and project you experienced the issue with on support@snyk.io

https://support.snyk.io/webhooks/failed-to-detect-issues

jshaughn:
I rebased to re-trigger the checks and now it is fine.

PR #801, opened 2018-11-07
KIALI-1880 Make Namespace and Duration handling consistent

jshaughn:
The updated duration handling for KIALI-746 caused an issue with the new
activeNamespace handling in KIALI-1782.  Updating the duration handling to
be consistent with the namespace handling solves this issue.
Also, for clarity, rename durationInterval to just duration.


src/pages/Graph/GraphRouteHandler.tsx
```diff
@@ -170,21 +175,16 @@ export class GraphRouteHandler extends React.Component<RouteComponentProps<Graph
       nodeTypeHasChanged;
 
     // update the redux store with the URL information
+    if (durationHasChanged) {
+      store.dispatch(UserSettingsActions.setDuration(nextGraphDuration));
+    }
```
israel-hdez: Not a blocker, but we should find a better way to do this.

The `getDerivedStateFromProps` method is turning somewhat brittle the url and redux managing. If, for any reason, React raises an update/re-render of this `GraphRouteHandler` component after dispatching a setDuration/setActiveNamespace, these lines may override the newly setted values and revert to the old ones present in the URL, before the `GraphPage` component can update the URL with the new values.

src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -552,7 +551,7 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
     if (!cy) {
       return;
     }
-    const duration = this.props.graphDuration.value;
+    const duration = store.getState().userSettings.duration;
```
israel-hdez: Minor, but I'm wondering why the store is being referenced directly.
The `CytoscapeGraph` component is wrapped and "connected" to redux at the bottom of this file. So, we should use `mapStateToProps` to bind to the state and use `this.props`. Same case with `store.dispatch`.

src/components/GraphFilter/GraphFilterToolbar.tsx
```diff
@@ -70,7 +61,7 @@ export default class GraphFilterToolbar extends React.PureComponent<GraphFilterT
         // @ts-ignore
         GraphDataActions.fetchGraphData(
           store.getState().namespaces.activeNamespace,
-          this.props.graphDuration,
+          store.getState().userSettings.duration,
```
israel-hdez: Now that this component has more usages of the `store` we should consider wrapping it in a container to follow the redux patterns, rather than using the store directly.

src/containers/GraphSettingsContainer.tsx
```diff
@@ -87,7 +87,7 @@ class GraphSettings extends React.PureComponent<GraphSettingsProps> {
         // @ts-ignore
         GraphDataActions.fetchGraphData(
           store.getState().namespaces.activeNamespace,
-          this.props.graphDuration,
+          store.getState().userSettings.duration,
```
israel-hdez: Same note here: `GraphSettings` is already wrapped in a container. We should revert back to using `this.props` and use `mapStateToProps` to bind to the state to follow redux patterns.

src/components/GraphFilter/GraphFilter.tsx
```diff
@@ -104,12 +96,11 @@ export default class GraphFilter extends React.PureComponent<GraphFilterPropsRea
...
           />
           <Toolbar.RightContent>
-            <GraphRefreshWithDefaultOptions
+            <GraphRefreshContainerDefaultRefreshIntervals
               id="graph_refresh_container"
               disabled={this.props.disabled}
               handleRefresh={this.handleRefresh}
-              graphDuration={this.props.graphDuration}
-              onUpdateGraphDuration={this.handleDuration}
+              duration={store.getState().userSettings.duration}
```
israel-hdez: I think, you don't need this property, since `GraphRefreshContainerDefaultRefreshIntervals` is overriding this internally (you already changed that component to pass the duration). So, no need to import the store.

jshaughn:
> There is still an issue in the drilled-down view. Sometimes, the dropdown may display "all" instead of the namespace of the drilled-down node. But that's another issue.

@israel-hdez This is actually the correct behavior.  When you drill down we don't want to change the activeNamespace, so that when you go "Back to Namespace" you go back to the namespace you were looking at. 

I think what we should probably do here is remove the namespace dropdown completely.  This will be even more relevant when we have multi-namespace support, because a frozen multi-select dropdown is not useful at all.  We could then change "Back to Namespace" to just "Back" or maybe "Back to Graph" or something like that.

src/pages/Graph/GraphRouteHandler.tsx
jshaughn: We need to be able to change the redux state based on the URL.  In theory this should be getting called when a URL is being processed, either by a new browser entry, an F5, or a programmatic push.  In any of these cases the idea is that the URL should "rule the world" so I'm not sure if there is a problem here, as the new URL would have its values applied.  I may be missing something.  I'm open to ideas for changing things if needed.

jshaughn:
@israel-hdez Ah, OK, that's good.  Although I still think we need to eliminate it because of what is coming in multi-namespace.  So I think I will do that.

src/components/GraphFilter/GraphFilter.tsx
jshaughn: Excellent catch, I  created KIALI-1894 for this, and will have a PR soon.

PR #802, opened 2018-11-07
WIP - KIALI-1724 Allows multiple selections in the namespace filter

josejulio:
** Describe the change **

 - Allows to select multiple namespaces from the namespace selector.
 - Removes the "all" namespace

Still WIP because of the following known issues and because more testing is needed

 - [ ] Namespace needs to be split in the right panel
 - [ ] Charts for multiple namespaces needs to be checked (is possible to show an aggregated chart or should be disabled?)
 - [ ] Empty selections yields the "all" namespace. I think we should probably want to have an error "Please select one namespace" or automatically default to one (the first).
 - [ ] Need to refactor my own code a bit more to remove repetitive LOCs

** Screenshot **

namespace-wip


israel-hdez:
Duh! Your PR only lasted one hour and it got conflicts

src/actions/GraphDataActions.ts
```diff
@@ -104,7 +112,7 @@ export const GraphDataActions = {
       dispatch(GraphDataActions.getGraphDataStart());
       const duration = graphDuration.value;
       let restParams = { duration: duration + 's', graphType: graphType, injectServiceNodes: injectServiceNodes };
-      if (namespace.name === serverConfig().istioNamespace) {
+      if (namespaces.find(namespace => namespace.name === serverConfig().istioNamespace)) {
```
israel-hdez: Not sure, but perhaps this needs special handling?
Istio communicates with all nodes with a sidecar. So, when enabling the "istio-system" namespace, it may not make sense to enable/disable the rest of the namespaces.

josejulio: IIUC, do you think we should disallow to show `istio-system` and other namespaces together? or do yo say that I should only pass the `restParams['includeIstio'] = true;` when `istio-system` is the only namespace?

package.json
```diff
@@ -2,6 +2,7 @@
   "name": "@kiali/kiali-ui",
   "version": "0.10.0",
   "description": "React UI for [Kiali](https://github.com/kiali/kiali).",
+  "proxy": "https://localhost:1080",
```
jshaughn: Don't want this, even in the WIP

src/actions/GraphDataActions.ts
```diff
@@ -91,7 +99,7 @@ export const GraphDataActions = {
 
   // action creator that performs the async request
   fetchGraphData: (
-    namespace: Namespace,
+    namespaces: Namespace[],
```
jshaughn: Perhaps we should have a Namespaces type instead of having to add the array brackets.

src/actions/GraphDataActions.ts
jshaughn: I think we should just "do what the user says".  So, if they explicitly select istio-system as one of the namespaces just include it.  So I think the code is OK as is.

src/actions/GraphDataActions.ts
```diff
@@ -72,6 +73,13 @@ const decorateGraphData = (graphData: any) => {
   return graphData;
 };
 
+let lastCancelableRequest = makeCancelablePromise(Promise.resolve());
+const setCurrentRequest = (promise: Promise<any>) => {
+  lastCancelableRequest.cancel();
+  lastCancelableRequest = makeCancelablePromise(promise);
+  return lastCancelableRequest.promise;
+};
```
jshaughn: I think something like this is already available using the CancelablePromises utility, see https://github.com/kiali/kiali-ui/blob/2caa94b16212adc3d240ed8e160e0d1c5d895a0f/src/utils/CancelablePromises.ts#L37

src/actions/GraphDataActions.ts
josejulio: I'm not sure I understand, `Namespace[]` is already a type (array of Namespaces). Unless you mean we might need to add data (now or in the future) apart of the array of namespaces.

src/actions/NamespaceAction.ts
```diff
@@ -7,7 +7,8 @@ export enum NamespaceActionKeys {
   NAMESPACE_REQUEST_STARTED = 'NAMESPACE_REQUEST_STARTED',
   NAMESPACE_SUCCESS = 'NAMESPACE_SUCCESS',
   NAMESPACE_FAILED = 'NAMESPACE_FAILED',
-  SET_ACTIVE_NAMESPACE = 'SET_ACTIVE_NAMESPACE',
+  TOGGLE_ACTIVE_NAMESPACE = 'TOGGLE_ACTIVE_NAMESPACE',
```
jshaughn: Why do we need this?  Do you want to react to every click?  I'm not sure but perhaps it is better to process changes as a group, and set all namespaces at the same time.  This would prevent firing off requests that will quickly become obsolete.

src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -178,7 +180,7 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
         <ReactResizeDetector handleWidth={true} handleHeight={true} skipOnMount={true} onResize={this.onResize} />
         <EmptyGraphLayout
           elements={this.props.elements}
-          namespace={store.getState().namespaces.activeNamespace.name}
+          namespace={activeNamespacesAsStringSelector(store.getState())}
```
jshaughn: I think the property should go from 'namespace' to 'namespaces' and I think the conversion to a comma-separated-string value should be done only where it's needed (i.e. in EmptyGraphLayout).   Also, on a separate topic, we will probably want to stop accessing the store directly.

src/components/NamespaceDropdown.tsx
```diff
@@ -20,32 +20,52 @@ export class NamespaceDropdown extends React.PureComponent<NamespaceListType, {}
...
+      return this.props.activeNamespaces[0].name;
+    } else {
+      return `${this.props.activeNamespaces.length} selected`;
+    }
+  }
+
   handleToggle = (isOpen: boolean) => isOpen && this.props.refresh();
 
   render() {
-    const disabled = this.props.disabled ? true : false;
```
jshaughn: Did you mean to get rid of disabled?

src/actions/NamespaceAction.ts
josejulio: Reacting to every click is far easier and makes the UI look better (the url updates as i click), but I could set an strategy to wait for some time before actually changing the state.

src/components/Nav/NavUtils.tsx
```diff
@@ -11,11 +11,12 @@ const buildCommonQueryParams = (params: GraphParamsType): string => {
 };
 
 export const makeNamespaceGraphUrlFromParams = (params: GraphParamsType): string => {
```
jshaughn: maybe rename to makeNamespacesGraphUrlFromParams

src/components/Nav/NavUtils.tsx
```diff
@@ -11,11 +11,12 @@ const buildCommonQueryParams = (params: GraphParamsType): string => {
...
-  const namespace = store.getState().namespaces.activeNamespace.name;
+  const namespaces = store
+    .getState()
+    .namespaces.activeNamespaces.map(namespace => namespace.name)
+    .join(',');
```
jshaughn: ooh, nice way of making CSV :-)

src/containers/NamespaceDropdownContainer.tsx
```diff
@@ -20,7 +20,7 @@ const mapDispatchToProps = (dispatch: Dispatch<any>) => {
       dispatch(NamespaceActions.fetchNamespacesIfNeeded());
     },
     onSelect: (namespace: Namespace) => {
-      dispatch(NamespaceActions.setActiveNamespace(namespace));
+      dispatch(NamespaceActions.toggleActiveNamespace(namespace));
```
jshaughn: Yeah, I'm not sure about this...

src/components/NamespaceDropdown.tsx
josejulio: I think i'm going to get rid of it after the rebase

src/components/Nav/NavUtils.tsx
josejulio: +1

src/components/Nav/NavUtils.tsx
josejulio: thanks, probably going to put a separate function to handle that without having to repeat it all over the place.

src/containers/NamespaceDropdownContainer.tsx
josejulio: Ok ok, i'm going to try a timed-based approach, one second and the event is going to fire.

src/actions/GraphDataActions.ts
josejulio: Will check that out

src/pages/Graph/GraphRouteHandler.tsx
```diff
@@ -149,14 +150,15 @@ export default class GraphRouteHandler extends React.Component<
...
-    const currentNamespaces = store.getState().namespaces.activeNamespace;
+    const currentNamespaces = store.getState().namespaces.activeNamespaces;
 
     const durationHasChanged = nextDuration.value !== currentState.graphDuration.value;
     const edgeLabelModeChanged = nextEdgeLabelMode !== currentState.edgeLabelMode;
     const graphTypeChanged = nextGraphType !== currentState.graphType;
     const injectServiceNodesChanged = nextInjectServiceNodes !== currentState.injectServiceNodes;
     const layoutHasChanged = nextLayout.name !== currentState.graphLayout.name;
-    const namespaceHasChanged = !nextNode && nextNamespaces.name !== currentNamespaces.name;
+    const namespaceHasChanged =
```
jshaughn: rename to namespacesHaveChanged

src/pages/Graph/SummaryPanel.tsx
```diff
@@ -36,7 +36,7 @@ export default class SummaryPanel extends React.Component<MainSummaryPanelPropTy
         {this.props.data.summaryType === 'group' ? (
           <SummaryPanelGroup
             data={this.props.data}
-            namespace={this.props.namespace}
+            namespace={this.props.data.summaryTarget.namespace}
```
jshaughn: yes, this was a mistake in the past, good fix. Same for Node.

src/services/Api.ts
```diff
@@ -276,28 +276,29 @@ export const getGraphElements = (auth: AuthToken, params: any) => {
...
-  node: NodeParamsType,
-  params: Partial<GraphParamsType>
-) => {
+export const getNodeGraphElements = (auth: AuthToken, node: NodeParamsType, params: Partial<GraphParamsType>) => {
   switch (node.nodeType) {
     case NodeType.APP:
       return newRequest(
         HTTP_VERBS.GET,
-        urls.appGraphElements(namespace.name, node.app, node.version),
+        urls.appGraphElements(node.namespace.name, node.app, node.version),
```
jshaughn: Yes, another good fix, I think.

src/types/Namespace.ts
```diff
@@ -3,3 +3,11 @@ interface Namespace {
 }
 
 export default Namespace;
+
+export const namespaceFromString = (namespace: string) => ({ name });
```
jshaughn: Just out of curiosity, why do we even have Namespace type as an object with only a name field?  It's sort of a pain, I don't see the value of it over just being a string.

josejulio: Me neither, I think it was just in case we needed to add more data to it?
 @mtho11 do you know?

src/actions/GraphDataActions.ts
jshaughn: Yeah, maybe I was getting confused with Golang, never mind.

src/actions/NamespaceAction.ts
jshaughn: OK, I'll defer to your preference.  Don't worry about the wait, I guess.  I just didn't want to see the graph getting updated and then immediately re-updating and then again, etc.  But if it seems to work well for you then OK.

src/components/CytoscapeGraph/CytoscapeGraph.tsx
jshaughn: Yes, not in this PR.  Actually, I may do some of that today in a separate PR, as a follow-up to Edgar's comments on the 1880 PR yesterday.

src/components/NamespaceDropdown.tsx
jshaughn: Why get rid of it?  Is it because with the cancelablePromises we feel it's OK to change the setting during an update?

src/containers/NamespaceDropdownContainer.tsx
jshaughn: I guess I was thinking we could wait until the dropdown is closed but I'm OK if you think it works well as is, without a timer.  To be honest, I'm not sure how often users will actually add lots of namespaces in one go, so  it may not even matter.

src/types/Namespace.ts
jshaughn: If no one knows then I'm not against changing it to a string and a string[] for namespaces.

src/components/NamespaceDropdown.tsx
josejulio: Exactly, currently it will update when you click a namespace, disabling would block the user if he wants to use multiple namespaces. 

Even if we changed to a "wait before actually changing the state" it would add additional waiting time if the user takes long to decide what namespace or if the auto-refresh triggers while we have this open.

I think the way to go would be to [cancel](https://github.com/axios/axios#cancellation) the current loading and do a new one with the update params, that way we ensure that there is no in-between data.

I think we could even show the GraphLoading screen if we detect we are loading a different namespace set to avoid any confusion on what graph we are looking at.

jshaughn: Sounds good to me.

src/actions/GraphDataActions.ts
josejulio: Did this

josejulio:
@jshaughn please review again

jmazzitelli:
> We need to prevent this from being merged until its discussed a bit more.

@mwringe you can create a review and select "Request changes" - that blocks the PR until you create a new review with "Approved"

jshaughn:
I'd suggest a short demo of the feature to get feedback.

josejulio:
> I'd suggest a short demo of the feature to get feedback.

That's a good idea, i'll organize one.

mtho11:
Looks like this needs rebasing

PR #803, opened 2018-11-08
KIALI-1185 - Implement oAuth support for OCP/OKD/Maistra clients

cfcosta:
** Describe the change **

UI counterpart for kiali/kiali#640.

This PR implements the server-side support for oauth-proxy, enabled only on Openshift. It uses the oauth token (the same that we can use to connect to k8s) and handles most of the use cases.

** What still needs to be done **

* [x] Handle signout gracefully.

** Backwards incompatible? **

No.


cfcosta:
Closed in favour of https://github.com/kiali/kiali/pull/768 and #906.

PR #804, opened 2018-11-08
KIALI-1878 Overview: allow to switch to workloads health

jotak:
- Add switch between app-based health and workload-based health
- In toolbar, some perf enhancements as refresh is not needed in every component change, and prevents unnecessary change checks & state updates (less renders)
- Remove unused function setActiveNamespace

JIRA: https://issues.jboss.org/browse/KIALI-1878

Screenshoft:
overview

### DIFF
--- a/src/app/History.ts
+++ b/src/app/History.ts
@@ -9,6 +9,7 @@ export enum URLParams {
   BY_LABELS = 'bylbl',
   DIRECTION = 'direction',
   DURATION = 'duration',
+  OVERVIEW_TYPE = 'otype',
   PAGE = 'page',
   PER_PAGE = 'perPage',
   POLL_INTERVAL = 'pi',
--- a/src/pages/Overview/FiltersAndSorts.ts
+++ b/src/pages/Overview/FiltersAndSorts.ts
@@ -53,11 +53,11 @@ export namespace FiltersAndSorts {
       isNumeric: false,
       param: 'h',
       compare: (a: NamespaceInfo, b: NamespaceInfo) => {
-        let diff = b.appsInError.length - a.appsInError.length;
+        let diff = b.inError.length - a.inError.length;
         if (diff !== 0) {
           return diff;
         }
-        diff = b.appsInWarning.length - a.appsInWarning.length;
+        diff = b.inWarning.length - a.inWarning.length;
         if (diff !== 0) {
           return diff;
         }
--- a/src/pages/Overview/NamespaceInfo.ts
+++ b/src/pages/Overview/NamespaceInfo.ts
@@ -1,8 +1,8 @@
 export type NamespaceInfo = {
   name: string;
-  appsInError: string[];
-  appsInWarning: string[];
-  appsInSuccess: string[];
+  inError: string[];
+  inWarning: string[];
+  inSuccess: string[];
 };
 
 export default NamespaceInfo;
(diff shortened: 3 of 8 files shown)


jotak:
PS: tell me if you think there's a better label name than "Health Type" to show

jotak:
> > PS: tell me if you think there's a better label name than "Health Type" to show
> 
> "Show health for" would be perhaps more clear.
> Maybe also consider aligning this with the switcher on the graph that also switch between different entity types? only if it will make sense :)

If you're thinking about the "Graph Type" switch, this is what I based on this change already (and it's why I named it "Health Type" as an equivalent of "Graph Type" , but I agree "Show health for" is more meaningful). But it doesn't make sense to show "versionned app" here, it's only Apps or Workloads.

> For future PRs I recommend to split enhancements/removal of unused code to either separate PRs or at least separate commits.

I can only partly agree ... I could have split the PR for the small perf enhancements, I agree. But very little changes (such as removing unused variables in the file that we're editing in) doesn't need extra PR IMO, as it would be a lot of noise (new JIRA, notifications flood etc.) for very little thing - slowing down processes / triage / etc. Which in the end would encourage NOT to fix little things I fear.

abonas:
> > > PS: tell me if you think there's a better label name than "Health Type" to show
> > 
> > 
> > "Show health for" would be perhaps more clear.
> > Maybe also consider aligning this with the switcher on the graph that also switch between different entity types? only if it will make sense :)
> 
> If you're thinking about the "Graph Type" switch, this is what I based on this change already (and it's why I named it "Health Type" as an equivalent of "Graph Type" , but I agree "Show health for" is more meaningful). But it doesn't make sense to show "versionned app" here, it's only Apps or Workloads.

Yes, I'm not suggesting to add versionned apps here, but rather only look at a similar control at the graph page "for inspiration". But if "Show health for" sounds good, go for it :)
> 
> > For future PRs I recommend to split enhancements/removal of unused code to either separate PRs or at least separate commits.
> 
> I can only partly agree ... I could have split the PR for the small perf enhancements, I agree. But very little changes (such as removing unused variables in the file that we're editing in) doesn't need extra PR IMO, as it would be a lot of noise (new JIRA, notifications flood etc.) for very little thing - slowing down processes / triage / etc. Which in the end would encourage NOT to fix little things I fear.

The idea is absolutely *not* to increase the noise and extra work with extra JIRAs. some PRs can hold minor changes without JIRA. But I also mentioned an option of separate commits where those commits are stacked one on another and are part of the same PR. After PR is merged, it has no meaning and no visibility in git history. So when someone looking at the code and using git history, they will gain more clarity per every commit if every commit has a single purpose rather than "why the commit that adds functionality X also does Y? was that change necessary for the feature or just a cleanup?"

lucasponce:
> The idea is absolutely not to increase the noise and extra work with extra JIRAs. some PRs can hold minor changes without JIRA. But I also mentioned an option of separate commits where those commits are stacked one on another and are part of the same PR. After PR is merged, it has no meaning and no visibility in git history. So when someone looking at the code and using git history, they will gain more clarity per every commit if every commit has a single purpose rather than "why the commit that adds functionality X also does Y? was that change necessary for the feature or just a cleanup?"

A comment in the commit message should be enough.
I.e. we force to re-arrange a PR when adding a new library dependencies.

For the problem commented here, just adding a comment is how the project is being doing.

hhovsepy:
The link to Workloads filter by Health is referenced to Applications.
screenshot from 2018-11-12 11-35-40

jotak:
@hhovsepy ah you're right, I've only changed one link, I forgot this one... good catch, thanks!

jotak:
@abonas @hhovsepy fixed:

updated

hhovsepy:
Verified 
kiali-ui 0.11.0 (3ae65df1eccca44cb0ff8ecee8626a212fa5dce3)
kiali v0.11.0-SNAPSHOT (12af09a47f5335f0f06c1a8bbf1f5f99ade962e9)

xeviknal:
Code reviewed and tested in my local env :)

PR #805, opened 2018-11-08
KIALI-1803 Same workload is counted more than once in source workloads tab

xeviknal:
** Describe the change **
In Source Workloads tab within Service details page counts duplicated workloads.

** Issue reference **
https://issues.jboss.org/browse/KIALI-1803

** Screenshot **
screenshot of kiali console 12


hhovsepy:
Verified

src/pages/ServiceDetails/ServiceInfo.tsx
```diff
@@ -74,6 +74,21 @@ class ServiceInfo extends React.Component<ServiceDetails, ServiceInfoState> {
...
+    let differentDependencies: string[] = [];
+    Object.keys(dependencies).forEach(key => {
+      dependencies[key].forEach((dependency: SourceWorkload) => {
+        if (!differentDependencies.includes(dependency.name)) {
+          differentDependencies.push(dependency.name);
+        }
+      });
+    });
+
+    return differentDependencies;
```
abonas: Why does this function return an array and not a number? the return array content is not used later, only the length (aka the number :)) is used.

src/pages/ServiceDetails/ServiceInfo.tsx
```diff
@@ -74,6 +74,21 @@ class ServiceInfo extends React.Component<ServiceDetails, ServiceInfoState> {
     return validationChecks;
   }
 
+  differentSources(): string[] {
+    const dependencies = this.props.serviceDetails.dependencies || {};
+
+    let differentDependencies: string[] = [];
```
abonas: perhaps a Set would also be an appropriate solution here? (if indeed so, then you won't have to check below on line 83 "if not exist" etc.)

xeviknal: You are totally right :) Change approached.

xeviknal:
@theute do you know how to re-run this snyk process again? I guess there was a glitch, because I didn't change a single line of package.json

josejulio:
Closing/Opening to see if that triggers a snyk build.

jshaughn:
@josejulio When this happens to me I usually do a rebase of master and a force push.  That gets it to recheck (and updates the PR)

PR #810, opened 2018-11-09
KIALI-1873 Show inbound/outbound error rate for requests health

israel-hdez:
In order to bring a better idea about where to look for issues, provide the detail about if errors are for incoming connections or for outgoing connections (or both). This is shown in health pop-overs.

Since pop-over is shared across all Kiali, all pop-overs got the details except for service health where only inbound health is available.

https://issues.jboss.org/browse/KIALI-1873

**Dependant PR:** https://github.com/kiali/kiali/pull/647

Example screenshot:

image

### DIFF
--- a/src/types/Health.ts
+++ b/src/types/Health.ts
@@ -33,6 +33,8 @@ export interface WorkloadStatus {
 
 export interface RequestHealth {
   errorRatio: number;
+  inboundErrorRatio: number;
+  outboundErrorRatio: number;
 }
 
 export interface Status {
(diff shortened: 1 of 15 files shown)


src/types/Health.ts
```diff
@@ -121,19 +123,33 @@ const ascendingThresholdCheck = (value: number, thresholds: Thresholds): Thresho
   return { value: value, status: HEALTHY };
 };
 
-export const getRequestErrorsRatio = (rh: RequestHealth): ThresholdStatus => {
-  if (rh.errorRatio < 0) {
+export const getRequestErrorsRatio = (rh: RequestHealth, propName?: string): ThresholdStatus => {
```
jotak: propName could be `keyof RequestHealth` maybe?

jotak:
I am undecided about what's the better algorithm here:

- Either as you did, inbound and outbound error rate status has no impact on global error rate status which is computed apart

- Or make the global error rate status being "worst of [inbound, outbound]" . This is the algorithm applied for other subcomponents of health. It has the advantage to not "dilute" a failing status of a subcomponent with better statuses of other subcomponents when looking at the global icon.

I think I would prefer the latter. Any thought?

jshaughn:
> I am undecided about what's the better algorithm here:

For me inbound and outbound are unrelated.  They represent different things.  To be honest, each inbound and outbound "edge" represent different things.  So I don't think averaging error rates is very meaningful.  If I have 1000 inbound requests at 0% error, and 10 outbound requests at 100% error, I have a bad health situation, despite a very low average error rate.   So, I'd prefer to see the max(errIn,errOut).

src/types/Health.ts
israel-hdez: +1 I'll change

israel-hdez:
@jotak 
> I am undecided about what's the better algorithm

I also prefer to show the worst icon. But I was unsure, since this is how it has been working for a while.

@jshaughn 
> So I don't think averaging error rates is very meaningful.

I think I didn't understand. Are you saying that the "top level" should display the worst error rate? or are  you referring also to the status icon?

jotak:
@israel-hdez indeed there would be some changes to bring to the text too (violations reports).
Also, there's already an existing function to merge statuses.
I've looked at how we could implement a good report text, I opened a PR onto your branch, can you have a look? https://github.com/israel-hdez/swsui/pull/1

israel-hdez:
@jotak I prefer your approach (I merged your PR), since the "global" error rate doesn't make too much sense. The text it's somewhat verbose; perhaps the health indicator should hide the detail, and leave that to the popover. Anyway, that could be left to a follow-up PR.

jotak:
Can you rebase BTW? (maybe related to that snyk failure)

abonas:
this is looking great!

PR #814, opened 2018-11-12
KIALI-1712 IstioConfig sort refactored to follow same patter as others

xeviknal:
** Describe the change **

Refactoring Istio Config sorting method.

** Issue reference **

https://issues.jboss.org/browse/KIALI-1712

### DIFF
--- a/src/pages/IstioConfigList/FiltersAndSorts.ts
+++ b/src/pages/IstioConfigList/FiltersAndSorts.ts
@@ -49,7 +49,7 @@ export namespace IstioConfigListFilters {
           } else if (!a.validation.valid && b.validation.valid) {
             sortValue = 1;
           } else if (a.validation.valid && b.validation.valid) {
-            sortValue = 0;
+            sortValue = a.validation.checks.length - b.validation.checks.length;
           } else if (!a.validation.valid && !b.validation.valid) {
             sortValue = b.validation.checks.length - a.validation.checks.length;
           }
--- a/src/pages/IstioConfigList/FiltersAndSorts.ts
+++ b/src/pages/IstioConfigList/FiltersAndSorts.ts
@@ -10,28 +10,53 @@ export namespace IstioConfigListFilters {
       title: 'Namespace',
       isNumeric: false,
       param: 'ns',
-      compare: (a: IstioConfigItem, b: IstioConfigItem) => 1
+      compare: (a: IstioConfigItem, b: IstioConfigItem) => {
+        return a.namespace.localeCompare(b.namespace) || a.name.localeCompare(b.name);
+      }
     },
     {
       id: 'istiotype',
       title: 'Istio Type',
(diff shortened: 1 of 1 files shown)


hhovsepy:
Did a regression testing, all worked as it was.
Just one question, does sorting by "Config" consider Valid and Warning Configs equal?

screenshot from 2018-11-15 12-30-33

xeviknal:
@hhovsepy could you please provide us an screenshot with the all the columns? To dig into the bug I would use some help on names and bookinfos of each row.

hhovsepy:
screenshot from 2018-11-15 19-34-18
screenshot from 2018-11-15 19-34-29
screenshot from 2018-11-15 19-34-43

src/pages/IstioConfigList/FiltersAndSorts.ts
```diff
@@ -10,28 +10,53 @@ export namespace IstioConfigListFilters {
...
+        } else if (!a.validation && b.validation) {
+          sortValue = 1;
+        } else if (!a.validation && !b.validation) {
+          sortValue = 0;
+        } else if (a.validation && b.validation) {
+          if (a.validation.valid && !b.validation.valid) {
+            sortValue = -1;
+          } else if (!a.validation.valid && b.validation.valid) {
+            sortValue = 1;
+          } else if (a.validation.valid && b.validation.valid) {
```
jotak: If I understand correctly, this means: if both are valid, order with the one that has more checks first. If both are invalid, do the opposite. Correct?

I'm not really against that but I wonder if it really brings anything to have such complex logic. Maybe here, having sortValue=0 (hence fallback to name sorting) would be enough? As you wish, just suggesting.

jotak: Oh actually, I'm just seeing this isn't (entirely) new, it's just moving an existing piece of code - except that in this piece of code, when both are valid (not invalid) we were doing sortValue=0. I guess then we should keep it as before?

hhovsepy:
Verified 
screenshot from 2018-11-19 11-33-58

src/pages/IstioConfigList/FiltersAndSorts.ts
xeviknal: yep, I didn't want to enter into refactoring the logic itself, just adapting that code to the new shape of sortings. However, Hayk found a bug that affects to the existent logic (master branch has it), so I use the JIRA to fix it.
The reason of changing that line is because you can have to valid items with checks in it. So, you'll need to place first valid items without checks, then valid items with checks and then invalid items.

xeviknal: It would probably be easier if we had the 'valid' field as a non-binary attribute (i.e.: valid, warning, error). This way, the comparative would be quite trivial.

PR #820, opened 2018-11-13
KIALI-1670 add support for groupBy options

jshaughn:
This is currently set to DNM, it needs to be demoed and determined to be useful.  It's also depended on some work in progress by @josejulio .

THIS PR REQUIRES SERVER PR: https://github.com/kiali/kiali/pull/659

PR #821, opened 2018-11-14
KIALI-1916 Use dagre for the layout of compound nodes

josejulio:
** Describe the change **

Adds a real layout for the contents of the compounds nodes, we were using a "vertical" layout and now we are using dagre that works better with https://github.com/kiali/kiali-ui/pull/820

** Screenshot **

## With master

Before:

before

After:

after

## In the context of #820 

Before:

without-dagre-for-compounds

After:

dagre-for-compounds

### DIFF
--- a/src/components/CytoscapeGraph/CytoscapeGraph.tsx
+++ b/src/components/CytoscapeGraph/CytoscapeGraph.tsx
@@ -32,6 +32,7 @@ import { NamespaceAppHealth, NamespaceServiceHealth, NamespaceWorkloadHealth } f
 import { makeNodeGraphUrlFromParams } from '../Nav/NavUtils';
 import { NamespaceActions } from '../../actions/NamespaceAction';
 import { DurationInSeconds } from '../../types/Common';
+import { DagreGraph } from './graphs/DagreGraph';
 
 type CytoscapeGraphType = {
   activeNamespace: Namespace;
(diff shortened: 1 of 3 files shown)


src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -355,14 +356,13 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
...
-    // if (updateLayout) {
-    //   cy.remove(cy.elements());
-    // }
+    // KIALI-1291 issue was caused because because some layouts (can't tell if all) do reuse the existing positions.
+    // We got some issues when changing from/to cola/cose, as the nodes started to get far away from each other.
+    // Previously we deleted the nodes prior to a layout update, this was too much and it seems that only reseting the
+    // positions to 0,0 makes the layout more predictable.
+    if (updateLayout) {
+      cy.nodes().positions({ x: 0, y: 0 });
+    }
```
jshaughn: This is an excellent find.

josejulio:
> One question, can you post a screen shot of the "old" app boxes to prove that there is no major change there?

Done, i think it looks better, more spaced and centered.

src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -355,14 +356,13 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
...
 
-    // Note, to resolve some layout issues for KIALI-1291 we introduced a complete wipe of cy elements prior to a layout change. This
-    // caused the issue in KIALI-1649. I'm not sure we still need the wipe, my testing shows good behavior without wiping the existing
-    // elements. So, I'm commenting it out.  If we need to re-instate it then we need to also add a fix for 1649. For that we'd probably
-    // need to gather the elements that have added classes (like Health info) and also capture any selected element. Then re-apply the
-    // settings on the replacement elements, for those elements that still exist.
-    // if (updateLayout) {
-    //   cy.remove(cy.elements());
-    // }
+    // KIALI-1291 issue was caused because because some layouts (can't tell if all) do reuse the existing positions.
```
jmazzitelli: "because because"

josejulio: what what?

jshaughn: mazz is a "dumb dumb"

PR #822, opened 2018-11-14
KIALI-1887: Make sure overview page refresh rate is in sync with other pages

mtho11:
This PR uses redux global state to enable synchronization of the refresh rate with other pages in the application. So refresh rates appear consistent in their state between screen navigation. Adds reduxification of Refresh component.

** Describe the change **
This is the quick low-risk fix for this issue as this just provides common global state between multiple components. The ultimate solution would be to create a new common `Refresh` component that is shared all over the application. However, we still have design and unanswered questions on this so here is the immediate fix.

** Issue reference **
https://issues.jboss.org/browse/KIALI-1887

Next steps:
https://issues.jboss.org/browse/KIALI-1929

** Backwards compatible? **
Yes

** Screenshot **
overview

### DIFF
--- a/src/components/Refresh/Refresh.tsx
+++ b/src/components/Refresh/Refresh.tsx
@@ -4,12 +4,17 @@ import { Button, MenuItem, Icon, DropdownButton } from 'patternfly-react';
 import { config } from '../../config';
 import { PollIntervalInMs } from '../../types/Common';
 
-type Props = {
+type ComponentProps = {
   id: string;
-  pollInterval?: PollIntervalInMs;
   handleRefresh: () => void;
-  onSelect: (selected: PollIntervalInMs) => void;
 };
+
+type ReduxProps = {
+  pollInterval: PollIntervalInMs;
+  setRefreshInterval: (pollInterval: PollIntervalInMs) => void;
+};
+
+type Props = ComponentProps & ReduxProps;
 
 type State = {
   pollInterval?: PollIntervalInMs;
@@ -46,7 +51,7 @@ class Refresh extends React.Component<Props, State> {
       newRefInterval = window.setInterval(this.props.handleRefresh, pollInterval);
     }
     this.setState({ pollerRef: newRefInterval, pollInterval: pollInterval });
-    this.props.onSelect(pollInterval);
+    this.props.setRefreshInterval(pollInterval); // notify redux of the change
   };
 
   render() {
--- a/src/containers/RefreshContainer.tsx
+++ b/src/containers/RefreshContainer.tsx
@@ -1,19 +1,20 @@
 import { connect } from 'react-redux';
-import { bindActionCreators, Dispatch } from 'redux';
+import { Dispatch } from 'redux';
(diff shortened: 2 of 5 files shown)


src/containers/RefreshContainer.tsx
```diff
@@ -13,7 +14,10 @@ const mapStateToProps = (state: KialiAppState) => ({
 const mapDispatchToProps = (dispatch: Dispatch<any>) => {
   return {
     onSelect: bindActionCreators(UserSettingsActions.setRefreshInterval, dispatch),
-    onUpdatePollInterval: bindActionCreators(UserSettingsActions.setRefreshInterval, dispatch)
+    onUpdatePollInterval: bindActionCreators(UserSettingsActions.setRefreshInterval, dispatch),
```
jotak: isn't it weird to have two callbacks for the same action? Shouldn't we have only one?

jotak: Actually, I had a closer look and it seems some logic is broken. From `OverviewToolbar`, the function `this.updatePollInterval` isn't called anymore on refresh rate update, which breaks the URL update.

mtho11: Yes, I agree it should just be one callback. I will change that.

src/components/Refresh/Refresh.tsx
```diff
@@ -4,13 +4,19 @@ import { Button, MenuItem, Icon, DropdownButton } from 'patternfly-react';
...
 
-type Props = {
+type ComponentProps = {
   id: string;
-  pollInterval?: PollIntervalInMs;
   handleRefresh: () => void;
+};
+
+type ReduxProps = {
+  pollInterval: PollIntervalInMs;
```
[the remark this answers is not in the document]
mtho11: It is called `pollInterval` because that is what this `Refresh` component calls it and I didn't want to go renaming others variable names until we have agreed on a standard. This occurs in many places BTW.

src/containers/RefreshContainer.tsx
jotak: It looks like I forgot to ask for changes in my review? Weird, I was pretty sure I did. My bad.
But the second problem I mentioned has been ignored: we now have a regression on URL handling, again.

jshaughn: I expected that to have been fixed when changes were pushed.  My bad for approving and not testing it.  Sorry (again).

mtho11: @jshaughn I did resolve those: https://github.com/kiali/kiali-ui/blob/5e1f3f84e07419929ea6a19eb7e994ffb88d75bc/src/containers/RefreshContainer.tsx#L10

jshaughn: @mtho11 the issue is not the duplication but rather ... "From OverviewToolbar, the function this.updatePollInterval isn't called anymore on refresh rate update, which breaks the URL update."

I'm looking into this already, I created https://issues.jboss.org/browse/KIALI-1960, if I need help I'll let you know.

PR #825, opened 2018-11-14
KIALI-1806 Fixing broken layout for destination rules in Service Details

xeviknal:
** Describe the change **
In destination Rules, within Service Details has a trafficPolicy layout broken in subset column.

** Issue reference **
https://issues.jboss.org/browse/KIALI-1806

** Screenshot **

Before:
screenshot of kiali console 21
After:
screenshot of kiali console 20

### DIFF
--- a/src/pages/ServiceDetails/ServiceInfo/ServiceInfoDestinationRules.tsx
+++ b/src/pages/ServiceDetails/ServiceInfo/ServiceInfoDestinationRules.tsx
@@ -139,17 +139,23 @@ class ServiceInfoDestinationRules extends React.Component<ServiceInfoDestination
   generateSubsets(subsets: Subset[]) {
     let childrenList: any = [];
     subsets.map(subset => {
-      Object.keys(subset.labels).forEach((key, v) =>
-        childrenList.push(
-          <li key={this.generateKey() + '_k' + v}>
-            <span style={{ float: 'left', paddingRight: '10px', paddingTop: '3px' }}>{subset.name}</span>{' '}
-            <Label name={key} value={subset.labels[key]} />
-            <DetailObject name="trafficPolicy" detail={subset.trafficPolicy} />
-          </li>
-        )
+      childrenList.push(
+        <li key={this.generateKey() + '_k' + subset.name} style={{ marginBottom: '13px' }}>
+          <Row>
+            <Col xs={3}>
+              <span style={{ paddingRight: '10px', paddingTop: '3px' }}>{subset.name}</span>{' '}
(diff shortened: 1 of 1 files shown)


hhovsepy:
kialiqeci

rhqci:
Jenkins CI: kiali-ui-pr-UI-test [#329](https://paste.fedoraproject.org/paste/YhmSqGOVb6N1VWral2SxIA)
  * :heavy_check_mark: python-ui-tests-parallel [#[197]](https://paste.fedoraproject.org/paste/drUWinWmpievTQRzq6yDrw)

abonas:
is this dividing the input into 3 columns? what does the traffic policy one on the right refer to - does it have a value?

hhovsepy:
This is how it started showing, before the "Load Balancer" was in Subsets, now it is in Traffic Policy, as it should be:

screenshot from 2018-11-15 11-59-17

abonas:
@hhovsepy great, what confuses me is that your screenshot seems different from @xeviknal "after" screenshot above

hhovsepy:
@abonas right, I was not able to reproduce that 'after' screenshot, @xeviknal can you provide me the yaml file of DR you have created?

hhovsepy:
Reproduced also the fist "After:" screenshot, 
The "TrafficPolicy" section of "SubSets" is correct, as these are the subsets which are under the "version", and the main "TrafficPolicy" does not have value, so the column is empty.

screenshot from 2018-11-15 12-51-18

pilhuhn:
I like (the idea of) that change. I wonder though if we could do a completely different arrangement. The  4 right most columns (and 1st +2nd)  are mostly empty, but the content of the subset (3rd column) expands very much vertically, which eventually forces people to scroll to see the definitions.

xeviknal:
@pilhuhn I do agree with you. Actually, we have [KIALI-1278](https://issues.jboss.org/browse/KIALI-1278) proposing to move this table into smaller one and place the details into another 'page'. This is what we do with Virtual Service.

pilhuhn:
@xeviknal I don't know about "another 'page'" We already force the user to too much clicking before arriving at a point where the user can see what they want to see

aljesusg:
@xeviknal Can you change the commit to KIALI-1806 instead Kiali-1803 ?

xeviknal:
@pilhuhn okay, so let's move the discussion into [KIALI-1278](https://issues.jboss.org/browse/KIALI-1278). Perhaps, we can change the topic to a broader one and raise it to UXD team.

PR #828, opened 2018-11-15
KIALI-1942 grey out showServiceNodes option when graphType=service

jshaughn:
kiali-1942

### DIFF
--- a/src/components/GraphFilter/GraphSettings.tsx
+++ b/src/components/GraphFilter/GraphSettings.tsx
@@ -10,6 +10,8 @@ import { ListPagesHelper } from '../../components/ListPage/ListPagesHelper';
 import { GraphFilterState, KialiAppState } from '../../store/Store';
 import { KialiAppAction } from '../../actions/KialiAppAction';
 import { GraphFilterActions } from '../../actions/GraphFilterActions';
+import { GraphType } from '../../types/Graph';
+import { PfColors } from '../Pf/PfColors';
 
 type ReduxProps = Omit<GraphFilterState, 'showLegend'> & {
   // Dispatch methods
@@ -27,6 +29,7 @@ type GraphSettingsProps = ReduxProps;
 
 interface VisibilityLayersType {
   id: string;
+  disabled?: boolean;
   labelText: string;
   value: boolean;
   onChange: () => void;
@@ -90,6 +93,7 @@ class GraphSettings extends React.PureComponent<GraphSettingsProps> {
       },
       {
         id: 'filterServiceNodes',
+        disabled: this.props.graphType === GraphType.SERVICE,
         labelText: 'Service Nodes',
         value: showServiceNodes,
         onChange: toggleServiceNodes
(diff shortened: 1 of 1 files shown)


josejulio:
I think it the `Services nodes` checkbox should be `checked` (and disabled) if we are selecting the `graphType=service`, but i think that's for UX to decide.

src/containers/GraphSettingsContainer.tsx
```diff
@@ -181,7 +183,7 @@ class GraphSettings extends React.PureComponent<GraphSettingsProps> {
     const displaySettingItems = visibilityLayers.map((item: VisibilityLayersType) => (
       <div id={item.id} key={item.id}>
         <label>
-          <input type="checkbox" checked={item.value} onChange={() => item.onChange()} />
+          <input type="checkbox" checked={item.value} onChange={() => item.onChange()} disabled={item.disabled} />
           <span className={checkboxStyle}>{item.labelText}</span>
```
josejulio: To gray out, you would have to create a new style and apply to this element when the item is disabled:

```tsx
const disabledTextStyle = style({color: 'gray'}); // or PFColors.Gray
// ...
<span className={checkboxStyle item.disabled? disabledTextStyle : ''}>{item.labelText}</span>
```

jshaughn:
@josejulio @abonas I've rebased and updated to get the gray text.  Thanks for the tip Josejulio.  As for ensuring the checkbox is checked.  I've thought more about it and recommend we leave it in whatever state it was in when setting the graph type.  I recommend this because if you change the graphType from service to something else the setting will get applied.  And note that in another PR we are changing the default to 'on', so that means if it is off the user manually set it that way, and I don't think we should then toggle it on.

So, this PR can be merged unless you want me to make the toggle change.

josejulio:
> @josejulio @abonas I've rebased and updated to get the gray text. Thanks for the tip Josejulio. As for ensuring the checkbox is checked. I've thought more about it and recommend we leave it in whatever state it was in when setting the graph type. I recommend this because if you change the graphType from service to something else the setting will get applied. And note that in another PR we are changing the default to 'on', so that means if it is off the user manually set it that way, and I don't think we should then toggle it on.

I didn't really meant to override the "on/off" setting (e.g. moving the redux value). I was saying that we could have it looking like is "on" when the graphType is service (e.g. is on if setting == 'on' or graphType==service)

That said, I don't know if that could confuse the user.

abonas:
> @josejulio @abonas I've rebased and updated to get the gray text. Thanks for the tip Josejulio. As for ensuring the checkbox is checked. I've thought more about it and recommend we leave it in whatever state it was in when setting the graph type. I recommend this because if you change the graphType from service to something else the setting will get applied. And note that in another PR we are changing the default to 'on', so that means if it is off the user manually set it that way, and I don't think we should then toggle it on.
> 
> So, this PR can be merged unless you want me to make the toggle change.

the reasoning here sounds good to me.
Can you just update the most recent screenshot look n feel?

jshaughn:
@abonas , I rebased to fix conflicts and have recorded an updated video below.  It shows the greyed out text and also the approach of maintaining the toggle setting.  I'm going to assume this is good and proceed to merge.  I think it's fine but any issue can be addressed in a follow up jira.

kiali-1942-2

PR #831, opened 2018-11-16
KIALI-1960 use redux for duration and refreshInterval state in Overview page

jshaughn:
I think this fixes the earlier regression while hopefully making it more clear what is managed in redux.

- avoid having the same state in two places, so remove duration
  and refreshInterval (aka pollInterval) from OverviewToolbar state.
  - use 'refreshInterval' where using redux becase this is what it is called
    in the redux store.
- remove OverviewPageContainer as neither duration or refreshInterval are
  actually manipulated in OverviewPage (although, an actual refresh is handled)
- add OverviewToolbarContainer because this is where redux state comes into
  play. We may set redux state based on the URL and we need to re-render on
  redux state change.
- OverviewToolbar
  - use URL to set initial values, keep URL up to date
  - separate props into ReduxProps and standard props
  - remove duration and pollInterval from State
- Refresh
  - remove pollInterval from State and let it be held only in redux

Note, in a follow-up we need to get the refreshInterval into the graph URL


jshaughn:
Oops, sorry @jotak , I fixed up the test failure.

src/pages/Overview/OverviewToolbar.tsx
```diff
@@ -114,24 +145,44 @@ class OverviewToolbar extends React.Component<Props, State> {
           <ToolbarDropdown
             id="overview-duration"
             disabled={false}
-            handleSelect={this.updateDuration}
+            handleSelect={this.props.setDuration}
```
jotak: I think something is missing here, it's going to dispatch the set-duration action but it doesn't trigger a refresh anymore (when the duration is changed, health data should be reloaded immediately)

src/pages/Overview/OverviewToolbar.tsx
```diff
@@ -114,24 +145,44 @@ class OverviewToolbar extends React.Component<Props, State> {
...
+    }
+  };
+};
+
+const OverviewToolbarContainer = connect(
+  mapStateToProps,
+  mapDispatchToProps
+)(OverviewToolbar);
+
+export default OverviewToolbarContainer;
```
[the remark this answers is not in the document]
jshaughn: +1, I have been tying them together.  It's much more clear, I think.

src/pages/Overview/OverviewToolbar.tsx
jshaughn: Does it not get handled here: https://github.com/kiali/kiali-ui/pull/831/files#diff-146a26206017a43c7930bad618f749c6R122 ?

jshaughn: Does it not get handled here? https://github.com/kiali/kiali-ui/pull/831/files#diff-146a26206017a43c7930bad618f749c6R122

jshaughn: Hmmm, in a quick test it doesn't look that the filter change is triggering the callback, I'll look more closely.

jotak: No, `StatefulFilters` is only responsible for the filter dropdown and its associated field, but not the other dropdowns. You might have been confused because this `ToolbarDropdown` is declared as a child of  `StatefulFilters`, but they are actually independent, it's just something necessary for how the DOM is organized, but nothing more.

jshaughn: OK, I see, that is only invoked when adding an actual filter, not changing a dropdown.

jshaughn: I'll make a change.

jotak:
thanks @jshaughn

src/pages/Overview/OverviewToolbar.tsx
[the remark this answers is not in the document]
mtho11: I was out last week so reviewing now. I agree with this and have created a jira for it: https://issues.jboss.org/browse/KIALI-2010

PR #833, opened 2018-11-19
KIALI-1948 Split Istio Rules in Rules/Adapters/Templates

lucasponce:
** Describe the change **

Split Istio Rules in Rules/Adapter/Templates as a previous step to add CRUD operations on them.

** Issue reference **

https://issues.jboss.org/browse/KIALI-1948

And related to PR https://github.com/kiali/kiali/pull/666

** Backwards compatible? **

No.

** Screenshot **

List Rules/Adapters/Templates
image

Details:
image
image


src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
```diff
@@ -288,16 +212,15 @@ class IstioConfigDetailsPage extends React.Component<RouteComponentProps<IstioCo
...
-        )}
+        {this.state.istioObjectDetails && this.state.istioObjectDetails.rule
+          ? this.renderEditor(this.state.istioObjectDetails.rule)
+          : undefined}
+        {this.state.istioObjectDetails && this.state.istioObjectDetails.adapter
+          ? this.renderEditor(this.state.istioObjectDetails.adapter)
+          : undefined}
+        {this.state.istioObjectDetails && this.state.istioObjectDetails.template
+          ? this.renderEditor(this.state.istioObjectDetails.template)
+          : undefined}
```
xeviknal: I do have doubts about rendering more than one ACEeditor at the same details page. Can we have that scenario?

src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
```diff
@@ -55,12 +48,18 @@ class IstioConfigDetailsPage extends React.Component<RouteComponentProps<IstioCo
...
-    );
+    const promiseConfigDetails = props.objectSubtype
+      ? API.getIstioConfigDetailSubtype(
+          authentication(),
+          props.namespace,
+          props.objectType,
+          props.objectSubtype,
+          props.object
+        )
+      : API.getIstioConfigDetail(authentication(), props.namespace, props.objectType, props.object);
```
xeviknal: There is a huge code complexity reduction in that component, that is great. 
However, the idea of handling the behavior differences with `if-else` makes me think that we should explore the idea of composing components. e.g. having one single page that renders a TypeDetails or SubtypeDetails components depending on the base URL. And both components renders one single component that has as props all the basic information to render the details page layout.
Still not really familiar with the react-way though.

src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
```diff
@@ -55,12 +48,18 @@ class IstioConfigDetailsPage extends React.Component<RouteComponentProps<IstioCo
...
+
+    // Note that adapters/templates are not supported yet for validations
+    // This logic will be refactored later on KIALI-
```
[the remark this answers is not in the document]
lucasponce: +1

src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
lucasponce: The idea behind was to not loose the types, but as renderEditor() is expecting an "any" type, I think we can extract the object first and invoke just once to refactor this.

src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
lucasponce: I'm not sure if we can avoid of the if-else part to extract the object, because in any case, we are dealing with different types and I see only two approaches (with potential variations): to miss the types (as it happens) and to render just a pure yaml, or to create a render component per subtype, which probably it is not worth yet, as we only need the yaml representation.

I think I can refactor a little bit to make it more clear, but I wouldn't enter in major changes as in the next JIRA, plan is to refactor this to enable edition.

xeviknal:
@lucasponce I see that details page is throwing errors after the latest refactoring.
> YAMLException: unacceptable kind of an object to dump [object Undefined]

src/pages/IstioConfigList/IstioConfigListComponent.tsx
```diff
@@ -250,6 +250,14 @@ class IstioConfigListComponent extends ListComponent.Component<
       iconName = 'migration';
       iconType = 'pf';
       type = 'Rule';
+    } else if (istioItem.type === 'adapter') {
```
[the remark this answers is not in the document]
lucasponce: No, because type is dynamic is some cases.

lucasponce:
> YAMLException: unacceptable kind of an object to dump [object Undefined]

There is a problem, thanks for the catch, I'm preparing a fix.

src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
lucasponce: Don't believe in all you read :-) specially related code style.

PR #835, opened 2018-11-21
Faster precommit scripts

cfcosta:
This PR changes from using prettier directly on precommit to using
pretty-quick, which will only run on changed files, and it also changes the
lint to only run when there's change on typescript files.

For lint, it requires processing all the files to get more precise type
information (hence the -p flag), so we always run lint, only avoid when there's
no change.

## Differences

1. It does not fail on changed files, just finishes the commit and restage the
   changed files. The default behaviour is to stage them directly, and there
   are no options to have the same behaviour as prettier.

   This is up to discussion, but I like that behaviour, because it only happens
   on fully-staged files, so if you have non-staged stuff still on the files,
   it will just fail as expected.

2. For some reason, I can only call it using yarn. This is weird, but not
   necessarily bad. It is not called on build though, only on commits.

3. The output format is different (but better).

#### Changes on some source files

List of modified files:

Result:


jotak:
Thanks @cfcosta !
+1 to give it a try
I'm also eager to see what's the differences with prettier in output format. I found prettier too intrusive.

Also, if we go with this solution, I guess we would have to run prettification on the whole codebase as an initial commit.

cfcosta:
@jotak it has the same output as prettier, because it uses prettier on the background, it just runs only for staged changes.

By "output" I meant "console output", which might be parsed somewhere (unlikely but possible).

package.json
```diff
@@ -40,12 +40,12 @@
     "build:dev": "npm run remove-rcue; REACT_APP_RCUE=false npm run build:kiali",
     "build:kiali": "npm run lint:prod && npm run copy-fonts && npm run copy-img && npm run build-css && REACT_APP_VERSION=$npm_package_version REACT_APP_NAME=$npm_package_name REACT_APP_GIT_HASH=$(git rev-parse HEAD) react-scripts-ts build",
     "build:prod": "npm run restore-rcue; REACT_APP_RCUE=true npm run build:kiali",
-    "precommit": "(npm run prettier -- --list-different || (echo 'Above file(s) were modified by prettier, check them before commiting' && false) ) && npm run lint:prod",
+    "precommit": "yarn run pretty-quick -- --staged && npm run lint:prod",
```
josejulio: I think you need to add `pretty-quick` as a dev dependency.

yarn.lock
```diff
@@ -5,14 +5,12 @@
 "@babel/code-frame@^7.0.0-beta.35":
   version "7.0.0"
   resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.0.0.tgz#06e2ab19bdb535385559aabb5ba59729482800f8"
-  integrity sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==
```
josejulio: Not sure why your yarn is removing the `integrity` fields, maybe you are using an old version?
```bash
$ yarn --version
1.10.1
```

jotak:
Next time I work on the UI I'm going to pull that PR first, so that I can try in real-world situation.

package.json
mtho11: I still don't see `pretty-quick` in the dev-dependencies: https://github.com/azz/pretty-quick

mtho11: I get this error: 
`error Command "pretty-quick" not found.`

mtho11:
@cfcosta can update the yarn.lock again?

cfcosta:
@mtho11 @josejulio updated my yarn, readded pretty-quick and regenerated yarn.lock, can you take a look?

mtho11:
I confirm this works; but we are waiting until after release to merge.

mtho11:
And we need to standardize on yarn version to make sure we don't have different `yarn.lock` diffs.

josejulio:
> @mtho11 @josejulio updated my yarn, readded pretty-quick and regenerated yarn.lock, can you take a look?

I tested with latest yarn (1.12.3) and it seems that yarn 1.12 won't add the integrity field back (see [this](https://github.com/greenkeeperio/greenkeeper/issues/955#issuecomment-434542711))

On the other hand, if the integrity is present, it will continue using and updating it.

So i think that you need to revert your `yarn.lock`  do a yarn install again.

cfcosta:
@josejulio done, can you see it again? I'm still seeing a weird diff, but everything should be ok.

package.json
```diff
@@ -40,11 +40,11 @@
...
     "copy-fonts": "cp -r node_modules/patternfly/dist/fonts src/",
     "copy-img": "cp -r node_modules/patternfly/dist/img src/",
     "lint": "tslint --project ./tsconfig.json",
-    "lint:prod": "tslint --project ./tsconfig.json --config tslint.prod.json",
+    "lint:prod": "if git diff --name-only HEAD | grep -E '\\.tsx?$'; then tslint --project ./tsconfig.json --config tslint.prod.json; else true; fi",
```
josejulio: I suppose this means that lint:prod will only run if there is any tsx?
Can we move it to a different entrypoint? `lint:prod-precommit` or put this if in the `precommit` ? 
I'm not sure if other parts of the pipeline use the lint:prod even if no file is modified.

package.json
```diff
@@ -115,6 +115,7 @@
     "node-sass-chokidar": "1.3.0",
     "npm-snapshot": "1.0.3",
     "prettier": "1.13.4",
+    "pretty-quick": "^1.8.0",
```
jotak: We're using fixed versions everywhere, so unless we're changing our policy on versions I guess we should have "1.8.0" here.

cfcosta:
@josejulio fixed your concerns, should be ok now.

@jotak It's a similar behaviour, but it does not fail the stage, it just doesn't add the changes to the commit, so you'd have to do another commit (or amend) with the changes. It does not fail the check like prettier does.

On the other hand, without `--no-restage`, it only adds the changes to the files if the whole file has been staged, which to me alleviate all the problems that we would have with it, since those tend to happen only on files that have been partially staged (you stage a line, the formatter goes there and change your patch).

What do you think?

PR #838, opened 2018-11-23
KIALI-1953 - Add Xmas easter egg

josejulio:
** Describe the change **

Adds a "XMas" easter egg  

Disabled by default, to enable you have to type "xmas" (lower case) while the graph area. 
It will be enabled until you navigate to other page or drill-down to a node (not really sure, but didn't spend time checking out why).
 All the node types are show as xmas tree (This could be easily updated to only show trees in certain node types)

** Backwards compatible? **

Yes 

** Screenshot **

Shows the tree in the background unless there is other background to show (`NodeImageOutLocked` or `NodeImageOut`)

xmas-animation-background

This could be easily disabled to show no backgrounds:
xmas-animation-no-background


josejulio:
Is this what you had in mind?

@theute @pilhuhn @abonas

jmazzitelli:
I know I shouldn't have to ask this, but I will JUST to make sure... 
We are NOT merging this, right?

josejulio:
Maybe we will, but not up to me to decide.
This is an Easter egg. Is hidden by default and we can make it more complicated to make it appear.

So please review if you have some time.

lucasponce:
@josejulio I won't approve if you don't put it MUSIC :-)

abonas:
I'm wondering though why so many files need to be changed to make a change in appearance like this - is this because of cytoscape design/architecture?

theute:
The PNG image isn't necessary IMO (and too noisy), just the shape of the tree. (let's remove the image)

In terms of merging or not. I would prefer if it's really easy to use for demos so if it's part of this week release then it would be perfect for demos of December. 
But it shouldn't be merged at the detriment of code complexity (bugs) or performance (in non-easteregg mode) I leave it to developers to comment...
 
PS:We can remove at anytime if needed.

Thanks !

josejulio:
> I'm wondering though why so many files need to be changed to make a change in appearance like this - is this because of cytoscape design/architecture?

I had to move and refactor a couple of files to make it easier to add this.
 - The appearance is split in two parts, the traffic animation and the graph nodes. 
 - To avoid having it on all the time had to add an `EasterEgg` component that enables to trigger some action on certain key press (in this case `xmas`) and configure the current components to allow them to switch from one to the other AND use it from next redraws.

In the future, if adding more "skins" would require really less work. To disable we would only need to remove the [easter egg trigger](https://github.com/kiali/kiali-ui/pull/838/files#diff-04b15b4602d6b9ccf1eb82d15e07535fR196), and the rest could remain there without incurring in any penalty at runtime.

josejulio:
> The PNG image isn't necessary IMO (and too noisy), just the shape of the tree. (let's remove the image)

Just removed it!

josejulio:
> @josejulio I won't approve if you don't put it MUSIC :-)

Do you really want music? :)

josejulio:
> This changes/adds **11** files (ELEVEN!), it adds another dependency, it adds code complexity (as if our code isn't complex enough). Whether enabled or not, the code is all in there.

I had to create some files to have this gag separated from **real** code.
Would it make more sense if I put all the code related to silly stuff in a single file? Maybe it would be easier to revert once the joke is done and would add less overhead on maintaining.

- 2 files are the package.json and yarn.lock
- 3 files are the cytoscapegraph files (CytoscapeGraph code, CytoscapeWrapper and the TrafficAnimation) that had to update to avoid hardcoding something.
- 1 is a file I had to split (types) to share some code.
- The other 5 files are the GraphStyles and traffic animation that could go into a single file.

About the dependency, i guess we could live without that and do something else to trigger this.

Anyway, if not merging, I think we could (maybe?) use npm's snapshot to download and use this version.

jmazzitelli:
If we want something to demo just for December during Christmas, put this in a separate branch, and build it as an npm snapshot and demo using that snapshot somehow.

If some of this refactors code that is useful "for real" then do that refactoring in a separate PR.

theute:
This is with silly things like this that we get the word spread.

As I said if there is an easy way to keep/run a forked version (ie:  we can use the hack script with some parameter) it's all fine to me. But we need this ASAP (ie: by the end of the week) or it will indeed be useless...
(Note: there are 1 Xmas per quarter, 1 halloween per quarter... ;))

pilhuhn:
I think (and I am really eager to demo that asap) it would be best, if @josejulio would build that version locally and then build the container image with that npm version and push this to dockerhub as `kiali/kiali:xmas` or similar. Perhaps a `-rc` now and then with the 0.11 sources tagged a respective `:v0.11-xmas` version.

Using that image then basically only requires editing the Deployment config for Kiali as described in https://medium.com/p/updating-the-kiali-version-in-your-istio-openshift-install-2fa1a618ce3a

theute:
v0.11-xmas image built on @josejulio machine and deployed on docker hub would be perfect for me

lucasponce:
> Do you really want music? :)

Yes, of course, no Xmas easter egg is completed without a proper Carol.
(in 8 bit style would be interesting).

josejulio:
> > Do you really want music? :)
> 
> Yes, of course, no Xmas easter egg is completed without a proper Carol.
> (in 8 bit style would be interesting).

Now using this music https://www.youtube.com/watch?v=lW_0o2VqlcQ

lucasponce:
"Bah! Humbug!" - you missed :-)

josejulio:
Adding DNM label to avoid Scrooge wrath :)

josejulio:
To avoid merging this into master and following @pilhuhn advice, I uploaded a docker image `kiali/kiali:v0.11.0-xmas.0` that contains this easter egg.
This build is based on kiali/kiali: [b570e91f9d53661e11d5421b06799c160e0f866b](https://github.com/kiali/kiali/commit/b570e91f9d53661e11d5421b06799c160e0f866b) and kiali/kiali-ui: [61f9b06e7a5a752e1d02b8caa9057c4e9710f78d](https://github.com/kiali/kiali-ui/commits/61f9b06e7a5a752e1d02b8caa9057c4e9710f78d)

You still need to type "xmas" (lower case) in the graph screen to have this enabled. I could also modify it to use it directly without having to type anything.

You can modify Kiali's Deployment config file as show here: https://medium.com/p/updating-the-kiali-version-in-your-istio-openshift-install-2fa1a618ce3a to have it running.

jotak:
> And now our image bloated with a 3.7MB sound file. So long as this is going in its own special "demo" branch and not master... I will hold my nose and close my eyes.. I see nothing....

Miss the time when our browsers were able to read 4KB midi files ...

jotak:
On the other hand, this file emulates 8bits style with... 320Kbits/s encoding!! This is insane.

theute:
I just tried: 
`sh cluster-openshift.sh --kiali-enabled true --kiali-version v0.11.0-xmas.0 up`
It seemed to work nicely, thanks !

pilhuhn:
> Miss the time when our browsers were able to read 4KB midi files ...

Can they no longer do that? TBH I was not even aware they could in the past.

jotak:
> > Miss the time when our browsers were able to read 4KB midi files ...
> 
> Can they no longer do that? TBH I was not even aware they could in the past.

I don't remember exactly, was it the browser natively, or the awesome RealPlayer plugin that everybody had to have anyway? :)

josejulio:
https://caniuse.com/#search=midi

josejulio:
I pushed `kiali/kiali:v0.12.0-xmas.1` image on dockerhub in case anyone wants it.

Closing as i think I won't have to update it again.

theute:
Thank you, just in time for my demo today :)

cfcosta:
@josejulio this just covers the midi hardware support, things like keyboards, knobs and so on. Unfortunately :(

PR #841, opened 2018-11-26
KIALI-1985 Refactor IstioObjects to map metadata/spec properly

lucasponce:
** Describe the change **

It propagates new model used for Istio in PR
https://github.com/kiali/kiali/pull/674

into the UI.

** Issue reference **

https://issues.jboss.org/browse/KIALI-1985

** Backwards compatible? **

Yes. New types used for Istio objects so all components need to update the new mapping.

VirtualServices component is updated but untouched from user perspective:
image

But now, Yaml representation is updated for next step, which is to allow edit this config from Kiali:
image
image


src/pages/ServiceDetails/__tests__/__snapshots__/ServiceInfo.test.tsx.snap
```diff
@@ -389,13 +389,6 @@ ShallowWrapper {
                       eventKey="destinationrules"
                     >
                       Destination Rules (1)
-                      <span>
```
xeviknal: I see in this snapshot that now validation icons are not shown for destination rules tab. I think it is because of [this](https://github.com/xeviknal/kiali-ui/blob/a624c044ecbd8d71e795d5bba349c36847ce9a15/src/services/__mockData__/getServiceDetail.json) request mock hasn't been updated. The feature is working well though.

lucasponce:
Rebased and solving conflicts with master.

lucasponce:
@xeviknal done

lucasponce:
@xeviknal mock data updated, probably snapshots are now correct, but in a future PR we should add more coverage to find these differences.

PR #843, opened 2018-11-26
KIALI-2000 Prepares redux and graph page for multiple namespaces

josejulio:
- Handle multiple namespaces in the sidebar
 - Add key attribute to some elements
 - Uses PromiseRegistry to ignore requests
 - Avoids loading an "empty" namespace
 - Unselects cy object when the graph is destroyed

** Describe the change **

This PR adds support for multiple namespaces to the redux store and the graph page, but it doesn't do any change to the Namespace Component (is a subset of #802).

I think this will be easier to review/merge and later we can focus only on the dropdown component and its interaction.

** Backwards compatible? **

Yes

** Screenshot **

No UI changes.

### DIFF
--- a/src/pages/Graph/SummaryPanelGraph.tsx
+++ b/src/pages/Graph/SummaryPanelGraph.tsx
@@ -44,7 +44,7 @@ export default class SummaryPanelGraph extends React.Component<SummaryPanelPropT
   }
 
   componentDidMount() {
-    if (this.props.namespace !== 'all') {
+    if (this.props.namespaces.length === 1) {
       this.updateRpsChart(this.props);
     }
   }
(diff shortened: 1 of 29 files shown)


src/components/NamespaceDropdown.tsx
```diff
@@ -35,6 +35,8 @@ export class NamespaceDropdown extends React.PureComponent<NamespaceListType, {}
       return list;
     }, {});
 
+    console.log('DEBUG:', this.props);
```
jshaughn: remove

josejulio:
> @josejulio other than the console log removal I think this is OK. It's going to be a brutal rebase 

Whoops, sorry!

> for me but oh well :) After this goes in I think we need to hold off on more changes in these files until I can get my work merged. Fix the console log and we can merge. Or, should we wait until after the demo tomorrow?

I can't see any reason to wait until the demo (unless this is failing), this shouldn't introduce any visible changes.

src/pages/Graph/SummaryPanelGraph.tsx
```diff
@@ -44,7 +44,7 @@ export default class SummaryPanelGraph extends React.Component<SummaryPanelPropT
   }
 
   componentDidMount() {
-    if (this.props.namespace !== 'all') {
+    if (this.props.namespaces.length === 1) {
```
jshaughn: This is a problem but just leave it as is, I will fix it in my PR.

PR #840, opened 2018-11-26
KIALI-1802 Make DestinationRule in service details clickable and navigate to Istio Config

aljesusg:
JIRA: [KIALI-1802](https://issues.jboss.org/browse/KIALI-1802)

This change create a link for destinationRule to the overview and yaml editor like the virtualServices.

destinationrule view

PR #844, opened 2018-11-27
KIALI-1884 Unify yaml editor component between Istio Config and Services pages

aljesusg:
Jira: [KIALI-1884](https://issues.jboss.org/browse/KIALI-1884)

link

Now we let the user delete the istioObject from this view (overview/yaml), I think that we should remove this dropdown actions and let the user make the actions from the istioObject page. What do you think   @lucasponce @jotak ?

### DIFF
--- a/src/pages/ServiceDetails/IstioObjectDetails.tsx
+++ b/src/pages/ServiceDetails/IstioObjectDetails.tsx
@@ -2,30 +2,11 @@ import * as React from 'react';
 import { DestinationRule, VirtualService } from '../../types/ServiceInfo';
 import { Validations } from '../../types/IstioObjects';
 import { Col, Nav, NavItem, Row, TabContainer, TabContent, TabPane } from 'patternfly-react';
-import { AceValidations, parseAceValidations } from '../../types/AceValidations';
-import AceEditor, { AceOptions } from 'react-ace';
-import 'brace/mode/yaml';
-import 'brace/theme/eclipse';
 import VirtualServiceDetail from './ServiceInfo/IstioObjectDetails/VirtualServiceDetail';
 import DestinationRuleDetail from './ServiceInfo/IstioObjectDetails/DestinationRuleDetail';
 import './ServiceInfo/IstioObjectDetails/IstioObjectDetails.css';
-import { Link } from 'react-router-dom';
 import IstioActionDropdown from '../../components/IstioActions/IstioActionsDropdown';
 import { ResourcePermissions } from '../../types/Permissions';
-
-const yaml = require('js-yaml');
-
-const safeDumpOptions = {
-  styles: {
-    '!!null': 'canonical' // dump null as ~
-  }
-};
-
-const aceOptions: AceOptions = {
-  readOnly: true,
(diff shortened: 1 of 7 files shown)


aljesusg:
I am thinking about remove the yaml tab and create a link inside the overview. What do you think?

aljesusg:
I thought that the idea talked in https://github.com/kiali/kiali-ui/pull/840 was make a link. I'll share code so

jotak:
I think the initial intent was that a yaml **link** (not tab) would be leading to another section. Which would have been correct, I think, navigation through links is ok.
But now that the link has been transformed to a tab, then yes, it makes it weird.

jotak:
Actually, why don't we have the first yaml link (from service details page) taking us directly to IstioConfig page? And replace the tabs view with just overview page, with again that link to yaml? That would allow us to have a single view for yaml, which was the original intent.

lucasponce:
> Actually, why don't we have the first yaml link (from service details page) taking us directly to IstioConfig page? And replace the tabs view with just overview, with again that link to yaml (but not as a tab)? That would allow us to have a single view for yaml, which was the original intent.

+1 IMO that's what I'd expect, to have all yaml view focus in a single location, too, then we can simplify to maintain these two views.

abonas:
> Actually, why don't we have the first yaml link (from service details page) taking us directly to IstioConfig page? And replace the tabs view with just overview, with again that link to yaml (but not as a tab)? That would allow us to have a single view for yaml, which was the original intent.

1. I would like to understand better the proposal of Yaml to become only available in Istio Config - what is the use case? In which scenarios the user would want to click to go there? to view/edit the yaml? Is there an easy get back to the original location? Is Istio Config proposed to become only a place to view yamls - what's the value of this section in addition to that?
2. In any case, this kind of concept change requires a discussion with UX first. AFAIK the OKD UI has overview/yaml tabs and hence Kiali UI was asked to follow for consistency.

lucasponce:
All,
There are several things that I think all participants don't share with the work for KIALI-1802 and KIALI-1884.

KIALI-1802 was described in a confused way "Make DestinationRule in service details clickable and navigate to Istio Config" as under the Service Details, Destination Rules had a link to their YAML representation. As I put in the comment and as it was done in the PR, the suggestion was to unify same component used for VirtualServices, and that is what is being merged.

Original KIALI-1884 was intended to refactor and use a single component to be used in ServiceDetails and IstioDetails, with no changes in the view. But, with the discussion brought in KIALI-1802 and https://github.com/kiali/kiali-ui/pull/840 @aljesusg @jotak and me thought it could be a good idea instead to propose the link @aljesusg is proposing.

This change requires some UXD approval and it's requested changes by @abonas, so, we will move back to the original JIRA.

Please, @aljesusg, work just to unify the component at implementation side without touching the existing navigation, we will discuss separately if we want to change the UX behaviour.

Thanks.

aljesusg:
Close in favor of https://github.com/kiali/kiali-ui/pull/848 in that PR @lucasponce will do KIALI-1884

PR #845, opened 2018-11-27
KIALI-1979 Convert to fully redux-managed state in the graph

jshaughn:
- Update to consistent redux state and URL management for the graph pages
- add node, layout, graphType, edgeLabelMode to redux store
- add pollInterval (pi) to URL
- remove GraphRouteHandler facade in favor of component managed state and url handling
  - each url param should be managed by a single component


src/components/GraphFilter/GraphFilter.tsx
```diff
@@ -65,7 +134,7 @@ export default class GraphFilter extends React.PureComponent<GraphFilterPropsRea
         <Toolbar>
           <FormGroup className={zeroPaddingLeft}>
             {this.props.node ? (
-              <Button className={namespaceStyle} onClick={this.props.onNamespaceReturn}>
+              <Button className={namespaceStyle} onClick={() => this.handleNamespaceReturn()}>
```
josejulio: nitpick, probably cleaner to do: `onClick={this.handleNamespaceReturn}`.

src/pages/Graph/GraphPage.tsx
```diff
@@ -92,76 +108,140 @@ const GraphErrorBoundaryFallback = () => {
...
+    if (GraphPage.isNodeChanged(urlNode, props.node)) {
+      props.setNode(urlNode);
+    }
   }
 
   componentDidMount() {
-    this.scheduleNextPollingInterval(0);
+    // This is a special bookmarking case. If the initial URL is for a node graph then
+    // defer the graph fetch until the first component update, when the node is set.
+    if (!store.getState().graph.node) {
```
josejulio: Could we get the node from the url instead of the store?

src/reducers/GraphDataState.ts
```diff
@@ -60,6 +75,50 @@ const graphDataState = (state: GraphState = INITIAL_GRAPH_STATE, action: KialiAp
...
+      newState.sidePanelInfo = INITIAL_GRAPH_STATE.sidePanelInfo;
+      break;
+    case getType(GraphFilterActions.toggleTrafficAnimation):
+      newState.filterState.showTrafficAnimation = !state.filterState.showTrafficAnimation;
+      break;
+    case getType(GraphFilterActions.toggleUnusedNodes):
+      newState.filterState.showUnusedNodes = !state.filterState.showUnusedNodes;
+      break;
+    // case getType(GraphFilterActions.showGraphFilters):
+    //   return updateState(state, { disableLayers: action.payload });
```
josejulio: Is this left commented on purpose?

src/store/Selectors.ts
```diff
@@ -1,5 +1,6 @@
 import { createSelector } from 'reselect';
 import { KialiAppState } from './Store';
+// import { EdgeLabelMode } from '../types/GraphFilter';
```
josejulio: Can you remove it if no longer used?

src/components/GraphFilter/GraphFilter.tsx
jshaughn: +1, I will fix in future.

src/components/NamespaceDropdown.tsx
```diff
@@ -43,9 +44,36 @@ export class NamespaceDropdown extends React.PureComponent<NamespaceListType, {}
...
+  };
+};
+
+const mapDispatchToProps = (dispatch: Dispatch<any>) => {
+  return {
+    refresh: () => {
+      dispatch(NamespaceThunkActions.fetchNamespacesIfNeeded());
+    },
+    setActiveNamespace: (namespace: string) => {
+      console.log('setActiveNamespace=' + namespace);
```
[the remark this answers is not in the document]
jshaughn: +1, I will fix now (and therefore I'll fix the other thing as well.)

src/pages/Graph/GraphPage.tsx
jshaughn: I'm going to say yes, I think so, if we want to do the URL parsing a second time.  Why do you ask, to avoid direct access of the store?

src/reducers/GraphDataState.ts
jshaughn: It was there when I started this.  I can remove it though, I don't know why it's actually there.

src/store/Selectors.ts
jshaughn: +1 will remove.

src/pages/Graph/GraphPage.tsx
josejulio: Yes, to avoid direct access to the store.

jshaughn: We can make a change in the future if you like.

PR #847, opened 2018-11-27
KIALI-2018 Show services by default

mwringe:
** Describe the change **

Services in the graph are shown by default instead of being hidden by default.

This means the user has a more complete picture about what is happening in their graph in the default view and could help them understand what is happening.

If a user doesn't want services to be displayed they have the option to disable seeing them.

** Issue reference **

https://issues.jboss.org/browse/KIALI-2018

And https://groups.google.com/forum/#!topic/kiali-dev/RsBBUEIWP2Q

** Backwards compatible? **

Yes.

** Screenshot **

Default view before:

screenshot from 2018-11-27 14-12-58

Default view after:

screenshot from 2018-11-27 14-13-04


abonas:
please update the docs/website with the "new default" if this is getting merged :)

mtho11:
@mwringe looks like this needs rebasing

src/reducers/GraphDataState.ts
```diff
@@ -25,7 +25,7 @@ export const INITIAL_GRAPH_STATE: GraphState = {
     showMissingSidecars: true,
     showNodeLabels: true,
     showSecurity: false,
-    showServiceNodes: false,
```
mtho11: We usually stick these defaults in `src/config.ts` and have `GraphDataState.ts` initialize from there. This way, all of the default settings are changed in one place. And in the future, we could also have overriding settings for dev, test, state, prod...

mtho11: Ok separate Jira is probably better for that.

mtho11:
Is there a separate Jira for the Docs update?

mwringe:
@mtho11 The docs update for the feature description has the default image showing services:

https://github.com/kiali/kiali.io/blob/6d7bf3954437e74e6a451cebfd8fd93cb764f8fb/static/images/features/graph-overview.png

Full PR here: https://github.com/kiali/kiali.io/pull/71

PR #848, opened 2018-11-28
KIALI-1655 Support edit YAML configuration

lucasponce:
KIALI-1884 Unify yaml editor component between Istio Config and Services pages

This is now ready for testing.

Addressed design issues discussed in:
https://github.com/kiali/kiali-design/issues/74
https://github.com/kiali/kiali-design/issues/68


abonas:
@lucasponce I understand this is WIP and not ready. 
Was this discussed from UX perspective?

hhovsepy:
ctrl+z allows to revert already saved changes, but as the resource version is also reverted, it can not be saved.

hhovsepy:
Isn't it better instead of having "Delete" button next to "Save", move it to bottom right, not to confuse the edition.

lucasponce:
> @lucasponce I understand this is WIP and not ready.

You are correct, WIP and DNL labels were properly set as this PR is still in active development.

> Was this discussed from UX perspective?
> Is this consistent with how Openshift UI does edit of yaml?
> How errors/warnings are displayed if user inputs an invalid configuration?

The work is consistent with Kiali and also with Openshift UI.
Main difference is that the Actions -> Delete button has been moved into the bottom of the editor in sync with patternfly and Openshift UI similar proposals.

image

image

Note this is not yet finished, there are pending refactoring tasks for this.

hhovsepy:
RFE: To add kind a "Dry Apply" button, to validate the edited config before finally saving that.
Because if VS is broken, the traffic can be sent to 'unknown' destination before it gets fixed.

lucasponce:
> Control + Z to "undo" some changes.

Yes, but CTRL+Z rollback also the resourceVersion changes and that is the message from the backend.
It can be written in an old 'resourceVersion'.

I think that is expected, and it would also happens in a similar editor.

Perhaps one thing that we can improve is to "remove" the resourceVersion before to send them to the backend or filtering on the backend.

> RFE: To add kind a "Dry Apply" button, to validate the edited config before finally saving that.
Because if VS is broken, the traffic can be sent to 'unknown' destination before it gets fixed.

That sounds reasonable, but perhaps we can address it separately of this feature. Just to consolidate this approach and enter in following iterations. Could you also write a story ticket ? Thanks !

xeviknal:
Particular tests:
* remove path permission for Virtual Service and editor doesn't allow you to edit them. Kiali does allow to edit DR, ServiceEntries and so on. :heavy_check_mark: 
* Edit one item, not saving them and clicking the reload button: :heavy_check_mark: 
* Navigation from service details page to Istio Config. History well maintained: :heavy_check_mark: 

Some small **problems** found:

* Click into cancel button when you have pending changes in the editor: it throws two warning windows instead of one (message: You have unsaved changes, are you sure you want to leave?)

* Ctrl-z remove the whole object content: 
screen recording 4

hhovsepy:
A corner case, not easy to reproduce, on FF 63.0.3:
Go to Service details - VS tab, click on VS to open YAMl edit page.
Change the value of 'weight', in my case from "0" to '01'.
Hit the "Refresh" button. Nothing changes, no alert.
Hit the "Cancel" button, nothing changes and no alert as well.
But 'Save' works.
screenshot from 2018-12-21 12-04-01

lucasponce:
@xeviknal @hhovsepy Ok, I've rebased with master (with some issues due the old branch) and address your comments.
Please, let me know if it is good to consolidate.
This branch is getting old and every change starts to force conflicts.

lucasponce:
I've found an issue in my last commit @xeviknal.
So, I need to send a folloup fix.

lucasponce:
@xeviknal weird, I can't reproduce the bug from my previous comment.
I wait to your review to confirm.

PR #854, opened 2018-11-29
KIALI-2031 support new isServiceEntry node metadata

jshaughn:
- this replaces obsolete isEgress node metadata

This requires server PR https://github.com/kiali/kiali/pull/687

image

### DIFF
--- a/src/pages/Graph/SummaryLink.tsx
+++ b/src/pages/Graph/SummaryLink.tsx
@@ -13,7 +13,7 @@ export const nodeTypeToString = (nodeType: string) => {
 };
 
 const isInaccessible = (data: NodeData): boolean => {
-  return data.isInaccessible || data.isEgress;
+  return data.isInaccessible || data.isServiceEntry;
 };
 
 const getLink = (data: NodeData, nodeType?: NodeType) => {
@@ -33,7 +33,7 @@ const getLink = (data: NodeData, nodeType?: NodeType) => {
       displayName = app;
       break;
     case NodeType.SERVICE:
-      if (!data.isEgress) {
+      if (!data.isServiceEntry) {
         link = `/namespaces/${encodeURIComponent(namespace)}/services/${encodeURIComponent(service)}`;
         key = `${namespace}.svc.${service}`;
       }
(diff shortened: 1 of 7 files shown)


jshaughn:
Note that this PR does not provide an option for toggling showing this badge.  I'm not sure if we need one or if it should always be shown.  Also, it does not update the legend.

jmazzitelli:
maybe that icon should be the background (shown inside the triangle) since that would be the same kind of thing we do for the nodes that are other namespaces (but that uses the "star" background)

src/pages/Graph/SummaryLink.tsx
```diff
@@ -99,17 +99,17 @@ export const renderDestServicesLinks = (node: any) => {
...
     isOutsider: data.isOutsider,
     isRoot: data.isRoot,
-    isEgress: data.isEgress,
-    isInaccessible: data.isInaccessible
+    isServiceEntry: data.isServiceEntry,
+    namespace: data.namespace,
+    nodeType: NodeType.SERVICE,
+    service: '',
+    version: '',
+    workload: ''
```
jmazzitelli: phew... alphabetized!

src/pages/Graph/SummaryPanelCommon.tsx
```diff
@@ -11,17 +11,17 @@ import { Metric } from '../../types/Metrics';
...
   isOutsider: boolean;
   isRoot: boolean;
-  isEgress: boolean;
-  isInaccessible: boolean;
+  isServiceEntry: boolean;
+  namespace: string;
+  nodeType: NodeType;
+  service: string;
+  version: string;
+  workload: string;
```
jmazzitelli: alphabetized... noice!

abonas:
The icon is planned to be discussed in https://github.com/kiali/kiali-design/issues/63, the icon used here is not the best option. Also, if the PR does not handle legend update, that would create a confusion for users. So either the legend should be updated at the same time (when we get the relevant icon), or as a follow up tracked in JIRA so it won't be forgotten.

jshaughn:
@abonas +1, I didn't do extra work because this needs feedback.  No need to update legend with the wrong information.  No need to add a 'show' option until required.  The server PR was set to DNM, I just set this one as well, to be more clear.

jmazzitelli:
> Also in the sidebar we could potentially show the target(s).
> If the SE is of mode resolution=STATIC, then the address(es) of he endpoint field, otherwise the hosts: field (but then, this is already the name of this service entry, shown on the node)

I think that can be a separate JIRA

jmazzitelli:
> I think a different icon instead of the badge would be better.

I agree. I would prefer a background image inside the triangle (or even a different shape other than triangle) can denote a service entry.

For now, though, this PR has removed the badge. We can worry about badges/icons later

jmazzitelli:
image

jshaughn:
@abonas , @pilhuhn As mazz commented, the PR has gone back to it's original purpose of just being a compatibility change for server-side changes, with no UI changes.  Service Entry discussion can continue in other channels.  I'm going to merge...

PR #859, opened 2018-11-30
KIALI-1200 Istio sidecar indicator should be present on pods and not only as summary on service

aljesusg:
JIRA: [KIALI-1200](https://issues.jboss.org/browse/KIALI-1200)

The Jira is an old one , now we have not pods in services so after read comments(@lucasponce & @xeviknal ), I did these changes:

- Now we can see in the list of workloads in a service of they have missing Istio Sidecar like @lucasponce proposed. @abonas I set the icon with tooltip (see demo) next to the name of the workload, It's ok in this place under the perspective of UX ?
- Improved component MissingSidecar to let a new attribute "**tooltip**" iff with this we'll have the **text** like a tooltip and not in the right of the icon.
- Added to the types/WorkloadOverview the istioSidecar value.

sample

### DIFF
--- a/src/components/MissingSidecar/MissingSidecar.tsx
+++ b/src/components/MissingSidecar/MissingSidecar.tsx
@@ -1,20 +1,38 @@
 import * as React from 'react';
 import PropTypes from 'prop-types';
-import { Icon } from 'patternfly-react';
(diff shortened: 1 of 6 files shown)


abonas:
great job

aljesusg:
> great job

So it's ok ^^

PR #860, opened 2018-11-30
KIALI-2026 Improve display of namespaces in node labels

jshaughn:
- fix service entry node labels
- fix service entry side panel
- fix namespace display for multi-namespace graphs
- fix unknown handling in graph and side panel
- change "Destination/Source" -> "To/From" in edge side panel to save space and better handle unknown


jshaughn:
@alissa, it's a minor change but just in case YouX had an opinion, I wanted to make clear the change to the Edge summary panel.  Instead of Source/Destination this PR changes it to To/From in the headers.  The reason is mainly to better satisfy the display of unknown but it also seems a little more clear while also saving space:

image

PR #865, opened 2018-12-04
Trivial: Remove 'coveralls' library since we no longer do coverage

mtho11:
** Describe the change **
We no longer do code coverage and there are no coverage scripts in package.json anymore.

** Backwards compatible? **
Yes

No.

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,7 +106,6 @@
     "@types/url-search-params": "0.10.1",
     "axios-mock-adapter": "1.15.0",
     "babel-core": "6.26.3",
-    "coveralls": "3.0.2",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -109,7 +109,6 @@
     "coveralls": "3.0.2",
     "enzyme": "3.3.0",
     "enzyme-adapter-react-16": "1.6.0",
-    "husky": "0.14.3",
```
josejulio: This is not a coveralls tool :)
`husky` is to integrate the git hooks to npm/yarn. The precommit hook will be useless if we remove this.

I think you meant to remove `"coveralls": "3.0.2",`

PR #869, opened 2018-12-04
Trivial: Run in StrictMode to catch any deprecated lifecycle methods

mtho11:
** Describe the change **
Run in StrictMode to catch any deprecated lifecycle methods. Also note: This does nothing when compiled in production so its safe to leave in code. The lifecycle violations will show up as warnings in the console.

** Issue reference **

https://reactjs.org/docs/strict-mode.html

** Backwards compatible? **
Yes

No

** Screenshot **

image

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -85,13 +85,15 @@ class App extends React.Component {
     loadRcueCssIfNeeded();
     const Sidebar = withRouter(NavigationContainer);
     return (
-      <Provider store={store}>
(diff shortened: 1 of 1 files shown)


jotak:
Is it going to send warnings just for our code, or also from dependencies such as patternfly?

mtho11:
@jotak looks like just our code. I would assume after a sprint or two we will have this deprecated lifecycle issues taken care of and it could be removed at that point.

cfcosta:
Did some testing for this here, could not find any issues. Will merge then.

PR #870, opened 2018-12-04
KIALI-2075 Move namespace selector to a secondary masthead

josejulio:
This is only for the graph page, but it prepares for other pages
 to use it.

** Describe the change **

Moves the namespace selector to a secondary masthead in the GraphPage

** Backwards compatible? **

Moving the namespace of place, but still works as before

** Screenshot **

namespace-selector-2-masthead

### DIFF
--- a/src/components/DefaultSecondaryMasthead/DefaultSecondaryMasthead.tsx
+++ b/src/components/DefaultSecondaryMasthead/DefaultSecondaryMasthead.tsx
@@ -1,0 +1,14 @@
+import React from 'react';
+
+import SecondaryMasthead from '../Nav/SecondaryMasthead';
+import NamespaceDropdownContainer from '../../components/NamespaceDropdown';
+
+export default class DefaultSecondaryMasthead extends React.PureComponent {
+  render() {
+    return (
+      <SecondaryMasthead>
+        <NamespaceDropdownContainer disabled={false} />
+      </SecondaryMasthead>
+    );
+  }
+}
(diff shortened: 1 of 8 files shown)


abonas:
super. @serenamarie125 what shall say you?

jshaughn:
I can't say I like the amount of real estate we give up for this.   It's a big empty space after Kiali logo, followed by a big empty space after the namespace dropdown, followed by a big empty space following the word Graph...

josejulio:
> I can't say I like the amount of real estate we give up for this. It's a big empty space after Kiali logo, followed by a big empty space after the namespace dropdown, followed by a big empty space following the word Graph...

Is exactly 53px more that we are losing.

jshaughn:
I ran the code and it works.  But I really don't like the way this looks.  Also, not sure why the dropdown isn't like the other buttons, it has a different look now and doesn't stand out.

serenamarie125:
the issues with visuals is because the initial design we shared had the PF4 L&F!

We can have mary work with a color that makes sense here - question is do you want to merge this PR and have a follow on?  @abonas ?

josejulio:
is there any change needed for UXD?

abonas:
@jshaughn @josejulio @israel-hdez  I agree with the inputs that we have to do something about the "Graph" title, it is strange the way it is now.
I suggest to consider to remove it completely, since on the left side the menu tab for Graph is selected and highlighted so users have an indication that they are in the Graph page. What do you think?
Per the colors, as @serenamarie125 suggests, we can get a different design for the second masthead to make it look better. I'm in favor of that, but also not block this PR on this.

josejulio:
Removing DNM label as seems fine to merge now and do follow-up PRs for fixing the details.

jmazzitelli:
Is there a followup JIRA to fix this? This looks nothing like OKD and the old menu was much better looking than this one. I assume this was an initial implementation with a followup PR coming that fixes the L&F

PR #871, opened 2018-12-05
KIALI-2086 More optimized metrics API

jotak:
Reporter and Direction now have to be provided in the query, to avoid fetching too many unnecessary metrics
As a consequence, response model doesn't include them anymore

No functional change expected.

JIRA: https://issues.jboss.org/browse/KIALI-2086
backend PR: https://github.com/kiali/kiali/pull/710

### DIFF
--- a/src/pages/Graph/SummaryPanelEdge.tsx
+++ b/src/pages/Graph/SummaryPanelEdge.tsx
@@ -130,7 +130,7 @@ export default class SummaryPanelEdge extends React.Component<SummaryPanelPropTy
     );
   }
 
-  private getByLabelsIn = (sourceMetricType: NodeMetricType, destMetricType: NodeMetricType) => {
+  private getByLabels = (sourceMetricType: NodeMetricType, destMetricType: NodeMetricType) => {
     let sourceLabel: string;
     switch (sourceMetricType) {
       case NodeMetricType.APP:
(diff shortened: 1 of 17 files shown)


jotak:
@israel-hdez @jshaughn this gave me some headaches :) Can you please check carefully the changes in various side-panel classes? The changes are basically to move reporter/direction checks from _after_ querying to _before_ querying, and it was sometimes a bit tricky. I think / hope what I've now is correct.

jshaughn:
@jotak To be honest, I think perhaps you should have deferred this work to Saturn team.  This is some of the trickiest code in graph-land.  Looking now...

jotak:
@jshaughn yes, possibly :) If you think it's necessary you can still open a PR over my own PR's branch - but as I was doing breaking changes that affects both metrics page and summary panel, it would have been difficult to do just half of the work (and leave the other half broken)

src/pages/Graph/SummaryPanelGroup.tsx
```diff
@@ -155,26 +156,27 @@ export default class SummaryPanelGroup extends React.Component<SummaryPanelPropT
...
-        const tcpSentOut = metrics['tcp_sent_out'];
-        const tcpReceivedOut = metrics['tcp_received_out'];
+      .then(responses => {
+        const metricsOut = responses[0].data.metrics;
+        const metricsIn = responses[1].data.metrics;
+        const rcOut = metricsOut['request_count'];
+        const ecOut = metricsOut['request_error_count'];
+        const tcpSentOut = metricsOut['tcp_sent'];
+        const tcpReceivedOut = metricsOut['tcp_received'];
         // use dest metrics for incoming
```
jshaughn: remove this comment

src/pages/Graph/SummaryPanelEdge.tsx
```diff
@@ -207,78 +208,108 @@ export default class SummaryPanelEdge extends React.Component<SummaryPanelPropTy
...
+        props,
+        filtersTCP,
+        'inbound',
+        'source',
+        quantiles,
+        byLabelsIn
+      );
+      this.metricsPromise = makeCancelablePromise(mergeMetricsResponses([promiseHTTP, promiseTCP]));
+    } else {
+      const filters = ['request_count', 'request_duration', 'request_error_count', 'tcp_sent', 'tcp_received'];
```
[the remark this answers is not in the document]
jotak: I'm not sure to understand. Basically, what we do above (in the `if` condition) is split HTTP and TCP queries because they're on different reporters, and here in the `else` they're on the same reporter so we get all at once.

jshaughn:
> but as I was doing breaking changes that affects both metrics page and summary panel, it would have been difficult to do just half of the work (and leave the other half broken)

Yes, understood, no problem.  I was saying it more out of empathy for you having to try and deal with that summary panel code :)

src/pages/Graph/SummaryPanelEdge.tsx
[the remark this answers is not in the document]
jshaughn: Hmmm, this comment actually went with a prior comment that is somehow absent from the review.  Let me restate things here:

I think maybe area of the code  should be broken into two blocks, one for the Http metrics and one for the Tcp metrics.  Above, I don't this useDest has migrated correctly (I could be wrong).  Either way, it may be more clear to have a useDestHttp and useDestTcp, and pull the complementary data using distinct filters.  What do you think?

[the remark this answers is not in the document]
jotak: Ok @jshaughn , I've separated the two calls, it's indeed easier to reason about (last commit)

jshaughn:
OK @jotak , I'm going to try running this and see what falls out...

src/pages/Graph/SummaryPanelNode.tsx
```diff
@@ -108,16 +109,40 @@ export default class SummaryPanelNode extends React.Component<SummaryPanelPropTy
...
+    if (!data.isRoot) {
+      // use dest metrics for incoming, except for service nodes which need source metrics to capture source errors
+      const reporter: Reporter =
+        data.nodeType === NodeType.SERVICE && data.namespace !== serverConfig().istioNamespace
+          ? 'destination'
+          : 'source';
+      // For special service dest nodes we want to narrow the data to only TS with 'unknown' workloads (see the related
+      // comparator in getNodeDatapoints).
+      const byLabels = this.isSpecialServiceDest(nodeMetricType) ? ['destination_workload'] : undefined;
+      promiseIn = getNodeMetrics(nodeMetricType, target, props, filters, 'inbound', reporter, undefined, byLabels);
```
israel-hdez: Not sure, but perhaps you are missing the TCP case here.

Unless I'm misunderstanding, this new code is not completely equivalent to the old one. Reading the old code, I understand that here if `reporter` is resolved to `destination`, you will need an additional call to `getNodeMetrics` to correctly fetch TCP data (since the old code is always using reporter=source for TCP).

jotak: Yeah, looks like I missed a case

jotak: Fixed in last commit, splitting TCP and HTTP in that case

PR #873, opened 2018-12-06
KIALI-2100 refresh graph immediately on toggle of showUnused and showSecurity

jshaughn:
- it was only refreshing on toggleOn, but we want the info to disappear immediately when disabled


jshaughn:
@josejulio Can you run this and check to make sure the unused nodes disappear.  For me the refresh is now happening but the nodes don't all go away, even though they are gone from the json.  I am wondering about the recent cy update, or maybe just our own "diff" checking.

josejulio:
> @josejulio Can you run this and check to make sure the unused nodes disappear. For me the refresh is now happening but the nodes don't all go away, even though they are gone from the json. I am wondering about the recent cy update, or maybe just our own "diff" checking.

It might be a bug in the cytoscape (that's the only thing that changed there IIRC), our own "diff" checking only helps decide if we are going to do a re-layout or not, the relayout seems to be happening.

PR #864, opened 2018-12-04
KIALI-2077 React related libraries upgrades

mtho11:
** Describe the change **
Upgrade all of the React related libraries and Patternfly. And this time the redux upgrades because they are minor since we did a major upgrade last sprint for them.

** Issue reference **
https://issues.jboss.org/browse/KIALI-2077

** Backwards compatible? **
Yes

[ ] Is your pull-request introducing changes in behaviour?
No changes in behaviour

PR #875, opened 2018-12-07
KIALI-2109 Snyk install for vulnerability remediation

mtho11:
** Describe the change **
This PR was created as a direct result of https://github.com/kiali/kiali-ui/pull/864 since the library upgrades resulted in Snyk violations that would not allow it to be merged. So splitting the snyk remediation file from the upgrades is what this PR is about. 

**NOTE**: there is now a `snyk` command in `package.json` that can be used by CI (it runs `snyk test`)  if needed to identify vulnerabilities as early as possible. It is probably not necessary in CI but anyway, it can be run without having additional global npm packages installed locally.

_Also_: There is an update to `react-scripts-ts` to get rid of a higher priority vulnerability. That is why the jest snapshots changed.

** Issue reference **

https://issues.jboss.org/browse/KIALI-2109

** Backwards compatible? **
Yes

No

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -1,0 +1,507 @@
+# Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
+version: v1.13.1
+# ignores vulnerabilities until expiry date; change duration by modifying expiry date
+ignore:
+  'npm:bootstrap:20160627':
+    - patternfly > bootstrap:
+        reason: None given
+        expires: '2019-01-06T16:53:37.068Z'
+    - patternfly > eonasdan-bootstrap-datetimepicker > bootstrap:
+        reason: None given
+        expires: '2019-01-06T16:53:37.068Z'
+    - patternfly > patternfly-bootstrap-treeview > bootstrap:
+        reason: None given
+        expires: '2019-01-06T16:53:37.068Z'
+    - patternfly-react > patternfly > bootstrap:
+        reason: None given
+        expires: '2019-01-06T16:53:37.068Z'
+    - patternfly-react > patternfly > eonasdan-bootstrap-datetimepicker > bootstrap:
+        reason: None given
+        expires: '2019-01-06T16:53:37.068Z'
+    - patternfly-react > patternfly > patternfly-bootstrap-treeview > bootstrap:
+        reason: None given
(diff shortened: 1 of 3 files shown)


josejulio:
Travis is failing, please check

mtho11:
Locally, everything is fine. Deleted `node_modules` ran `yarn prettier`and  `yarn test` and `yarn start`. All good. Hmm...

pilhuhn:
Mike, can you help me to better understand the following?

* the PR seems to not only change the Snyk behaviour, but also changes code in .snap files
* Are we adding ignore files now to keep Snyk quiet instead of fixing the vulnerabilities?
* Does the comment about `snyk test` refer to that we do not (no longer?) automatically check for vulnerabilities?

mtho11:
@pilhuhn, here are your answers:

> the PR seems to not only change the Snyk behaviour, but also changes code in .snap files

So I modified this PR to not include any of the fixes but just to introduce the `.snyk` vulnerability remediation file. So no `.snap` files to confuse things.

> Are we adding ignore files now to keep Snyk quiet instead of fixing the vulnerabilities?

We are fixing any vulnerbilities that have fixes and snyk tells us which ones have fixes and even notifies us of fixes when then occur.  The fixes (upgrades are many and affect many files) are in the next PR: https://github.com/kiali/kiali-ui/pull/864 to keep these two things separate.

> Does the comment about snyk test refer to that we do not (no longer?) automatically check for vulnerabilities?

So the `.snyk` file whitelists (for 30 days) the vulnerbilities that are not fixable. We fix everything else. But any new packages are scanned for vulnerbilities and reported.

This is how we got to this point in the first place was that in upgrading the react packages there were vulnerabilities that block the upgrades as non-mergable.

snyk

pilhuhn:
@mtho11 Thanks for your replies.
@josejulio Travis is happy now

mtho11:
Once this is merged then I can rebase and merge: https://github.com/kiali/kiali-ui/pull/864

mtho11:
Also, I purposely put:
`"snyk": "^1.116.2",` in the package.json with the '^' because do want the latest snyk packages in the dev environment (it never even gets compiled into production).

PR #878, opened 2018-12-11
KIALI-2134 Fix links in graph summary page

jshaughn:
- updated links to supports multiple namespace filters

kiali-2134

### DIFF
--- a/src/components/ListPage/ListPageLink.tsx
+++ b/src/components/ListPage/ListPageLink.tsx
@@ -5,6 +5,7 @@ import history from '../../app/History';
 import NamespaceFilter from '../Filters/NamespaceFilter';
 import { ActiveFilter } from 'src/types/Filters';
 import { healthFilter } from '../Filters/CommonFilters';
+import Namespace from '../../types/Namespace';
 
 export enum TargetPage {
   APPLICATIONS = 'applications',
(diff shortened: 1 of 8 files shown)


jshaughn:
@abonas This PR addresses the issue and I think it's fine.  But I did want to point out one possible source of confusion for users.  The graph shows nodes based on the variety of options available.  The number of nodes shown, and listed by type in the summary, will likely not match the number of items listed in the filtered detail lists.  The detail lists filter only on namespace whereas the graph filters dead nodes, unused nodes, service nodes, etc.

I don't know that this is really an area of concern but I thought I'd point it out.  If you think we need to address this in any way I guess another issue would be appropriate.  I don't really have any suggestions but getting the numbers to match is likely not feasible or worth the effort, it would probably need to be some sort of small language change on the link, or something like that.

abonas:
> @abonas This PR addresses the issue and I think it's fine. But I did want to point out one possible source of confusion for users. The graph shows nodes based on the variety of options available. The number of nodes shown, and listed by type in the summary, will likely not match the number of items listed in the filtered detail lists. The detail lists filter only on namespace whereas the graph filters dead nodes, unused nodes, service nodes, etc.
> 
> I don't know that this is really an area of concern but I thought I'd point it out. If you think we need to address this in any way I guess another issue would be appropriate. I don't really have any suggestions but getting the numbers to match is likely not feasible or worth the effort, it would probably need to be some sort of small language change on the link, or something like that.

Hi, thanks for the quick fix first of all :)
Regarding the above, iirc this confusion/mismatch is not related to adding linking capabilities, right?
But it is good that you are bringing it up.
I am not sure I understand though the mismatch explained.
This part "The detail lists filter only on namespace whereas the graph filters dead nodes, unused nodes, service nodes, etc." - the graph does display service nodes, doesn't it?
Perhaps you have a specific example that shows this kind of mismatch? (and better to discuss it not on an already merged PR :))

jshaughn:
> - the graph does display service nodes, doesn't it?

The graph does display service nodes but for example, it may not show a service node without request traffic for the time period.  But the service list will show all services for the namespace. So the number of service nodes indicated on the graph will be less than that listed in the details.

abonas:
> The graph does display service nodes but for example, it may not show a service node without request traffic for the time period. But the service list will show all services for the namespace. So the number of service nodes indicated on the graph will be less than that listed in the details.

Thanks for the clarification. Now I am wondering why is the info shown on the side panel in the first place if it does not match the graph? :) cc @pilhuhn

jshaughn:
The graph side panel categorizes what you see on the current graph.  The numbers for apps, services and workloads reflect the node count for those items: squares, triangles, and circles respectively.

Those numbers may or may not, (probably not) reflect the numbers that exist in the active namespaces, and therefore in the list views.

There is likely value in being able to link to the list views and pre-filter via active namespaces. But these links are misleading (QE already asked me whether there was a bug, so it is clearly confusing).

It is not feasible to generate a list view that matches the graph items without significant work, and frankly it's not worth it.  I'd suggest one of two options:

1) just get rid of the links.  users can already get to the list views (without the pre-filters) by clicking the main LHS menu.  Users can already get to individual detail pages for any selected graph node.  And *IF* the namespace dropdown is to be made global (it is already in the new 2nd masthead on the graph page), then the list views may already have filters set initially.

2) provide new links that better describe what you will get. Maybe something like this:

image

Overall, I don't think it's worth doing too much work on this, removing the links is not terrible.

abonas:
Then I would support removing them. But I think it is important to hear more opinions regarding this. Perhaps ask on kiali dev oe kiali users?

jshaughn:
I will post to the mailing list.

PR #879, opened 2018-12-11
KIALI-2120 Make sure activeNamespaces is always valid

jshaughn:
- Remove from activeNamespaces any namespace that becomes inaccessible


josejulio:
I was going to say that there is still one issue when i noticed is already merged. 
If you have only one namespace selected and the namespace is deleted, when you open the namespace selector it correctly unselects the now-gone namespace, but the graph is still showing the graph.

Clicking refresh doesn't do anything, you have to select and unselect to show the "no namespace is selected" screen.

no-namespace

PR #881, opened 2018-12-12
KIALI-2015 - not crashing when having wrong formatting in service details

xeviknal:
** Describe the change **

** Issue reference **
https://issues.jboss.org/browse/KIALI-2015

** Screenshot **
screenshot of kiali console 8
screenshot of kiali console 9
screenshot of kiali console 10


lucasponce:
@xeviknal I see this is still in DNM, but not sure if I understand the issue.
I guess that the backend sends a VS that doesn't match with the mapping done inside the UI, right ?
Or where is the root issue ?

I was wondering if alternatively to show a better error, if we may want to update the mapping to support this scenario as well.

pilhuhn:
See also 
https://issues.jboss.org/browse/KIALI-2015?focusedCommentId=13674138&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13674138

xeviknal:
@lucasponce yes, the root cause is the one you said. Backend sends an unexpected formatted DR, the UI assign that data in a variable (with a type) and then we try to print something is not correct. Actually, the root cause is that the UI is not throwing any error because of mismatching the type, the UI is complaining about printing an object instead of a basic type (text, number, boolean).

The unexpected format comes here: https://github.com/kiali/kiali-ui/blob/b840650c890ed595bb022116bcd9803411b0df2b/src/types/IstioObjects.ts#L302
DR's shown in the issue, has a subset and the labels in that subset are presented in array of hashmaps instead of a hashmap.
Our UI doesn't complain about this type mismatch but for the fact of trying to print a hashmap instead of a string: https://github.com/kiali/kiali-ui/blob/143f330490f4eea6e8119e75de69885f2df0d633/src/pages/ServiceDetails/ServiceInfo/ServiceInfoDestinationRules.tsx#L181

xeviknal:
What I will propose in this PR is going to be a short-term solution for that problem. There are couple of JIRAs open that will work well to achieve a good solution for this kind of scenarios.

First JIRA is [KIALI-2118](https://issues.jboss.org/browse/KIALI-2118). This aim to have strong typing in the API so all the uncaught format problems by galley are going to be catch in there and not forwarded to the UI.
In this case, the API will try to assign all the DR data to a strong typed DR variable. When checking the label types, it will see that it receives an array of maps instead of a plain map and golang will raise an error. At this point, the API can build a validation and serve it to the UI.

Here it comes the second JIRA, the [KIALI-1671](https://issues.jboss.org/browse/KIALI-1671). Kiali validations will be merged to the data models in a way the frontend will know for each object whether it has format problems or not.

xeviknal:
In case any label in the subset or the name of the subset isn't in the proper format, it will show this:
screenshot of kiali console 13

Again, this should be a short-term solution up until we've got the mentioned jiras.
Alternatively, if there is any exception thrown when rendering any tab, the following message will be shown (instead of breaking the whole page):
screenshot

src/components/ErrorBoundary/WithErrorBoundary.js
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react';
```
lucasponce: I think if you use .js on this file instead of .ts it may skip the typescript compiler.

aljesusg: It should be tsx extension , let js only for logical methods, If you render components or return htmlComponents  like the method withErrorBoundary you should use typescript.

lucasponce:
I have a suggestion.
I understand the work on this, and as commented, this should come from the backend in the future.
But on this situation, I see strange that user would have a green icon on this:
image

So my suggestion is in the context of this PR, if we could add a validation in the UI indicating this.
Same comment "This label has an unexpected format" but also to not show that the object is in green.

Yes, we know that this doesn't come from the backend, but I think user will be confused if don't show it.

Also, as discussed via IRC [1], we need some global indicator that istio/galley is not working.
But that is out of the scope of this PR.



[1] https://issues.jboss.org/browse/KIALI-2160

hhovsepy:
This is the result I saw:
screenshot from 2018-12-14 13-40-53
screenshot from 2018-12-14 13-41-07

xeviknal:
New behavior added. Also generic object for adding ErrorBoundaries to components (thx @aljesusg).
Validations icon goes according the format validation.

screen recording 3
cc/ @hhovsepy @lucasponce

hhovsepy:
screenshot from 2018-12-18 17-07-39

hhovsepy:
screenshot from 2018-12-18 17-17-21

hhovsepy:
Verified the solution.

lucasponce:
One comment. 
This looks correct:
image
This one too:
image
With the discussed comment that we will refactor better the validations.

But on this one:
image
Is missing the validation icon.

I mean in DestinationRuleDetail.tsx#globalStatus we would need to pass that validation to enable the icon that will inform in the overview tab that there is a problem.

src/pages/ServiceDetails/ServiceInfo.tsx
```diff
@@ -181,34 +186,50 @@ class ServiceInfo extends React.Component<ServiceDetails, ServiceInfoState> {
...
                       {(Object.keys(dependencies).length > 0 || this.props.serviceDetails.istioSidecar) && (
                         <ServiceInfoRoutes dependencies={dependencies} />
                       )}
-                    </TabPane>
-                    <TabPane eventKey={'virtualservices'}>
+                    </WithErrorBoundary>
+                    <WithErrorBoundary
+                      component={TabPane}
+                      eventKey={'virtualservices'}
+                      message={this.errorBoundaryMessage('Virtual Services')}
```
lucasponce: "Virtual Services" -> "VirtualServices"
I think it should be one single word as it's a term.

src/pages/ServiceDetails/ServiceInfo/ServiceInfoDestinationRules.tsx
```diff
@@ -138,24 +139,26 @@ class ServiceInfoDestinationRules extends React.Component<ServiceInfoDestination
...
-      subsets:
-        destinationRule.spec.subsets && destinationRule.spec.subsets.length > 0
-          ? this.generateSubsets(destinationRule.spec.subsets)
-          : 'None',
-      host: destinationRule.spec.host ? <DetailObject name="" detail={destinationRule.spec.host} /> : undefined,
-      createdAt: <LocalTime time={destinationRule.metadata.creationTimestamp || ''} />,
-      resourceVersion: destinationRule.metadata.resourceVersion,
-      actions: this.yamlLink(destinationRule)
-    }));
+    return (this.props.destinationRules || []).map((destinationRule, vsIdx) => {
```
lucasponce: This block of code shows changed in github web but I don't see the differences, perhaps it's just a format change ?

src/pages/ServiceDetails/ServiceInfo/types/DestinationRule.ts
```diff
@@ -0,0 +1,70 @@
+import { DestinationRuleSpec, K8sMetadata } from '../../../../types/IstioObjects';
+import Subset from './Subset';
+
+export default class DestinationRule {
```
lucasponce: We are re-reclaring the same DestinationRule in a different place.
So we will have one in IstioObjects.ts and another here to be used for same purpose, I think.

src/pages/ServiceDetails/ServiceInfo/types/Subset.ts
```diff
@@ -0,0 +1,31 @@
...
+  public isValid() {
+    return this.hasValidName() && this.hasValidLabels() && this.hasValidTrafficPolicy();
+  }
+
+  hasValidName() {
+    return typeof this.name === 'string';
+  }
+
+  hasValidLabels() {
+    const valid = Object.keys(this.labels).every((k, i) => new Label(k, this.labels[k]).isValid());
```
lucasponce: I wonder if we can use an aux function instead to create a new "Label" class/abstraction just to perform a validation.
It sounds to me that we need to check a condition between key: k and value: this.labels[k].

src/components/Label/Label.tsx
```diff
@@ -1,25 +1,29 @@
 import * as React from 'react';
+import PropTypes from 'prop-types';
```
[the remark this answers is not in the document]
xeviknal: Didn't follow the discussion about the PropTypes.
cc @aljesusg, I guess we should remove them from the MissingSidecar component too.

src/pages/ServiceDetails/ServiceInfo/ServiceInfoDestinationRules.tsx
xeviknal: you are right. I shouldn't be any change here. I try to remove it from the commit.

src/pages/ServiceDetails/ServiceInfo/types/DestinationRule.ts
xeviknal: I guess that is not the same purpose. The purpose in the IstioObjects is just to have an interface with no logic attached to it. The reason I added a class here is to distribute some logic that are just unique for this components. The validations of format are built depending on which information is shown in the list of this component. It wouldn't be the same in other components (e.g. IstioConfigList/Details).
One more reason I add this class here is not to perform all the logic in the same component. To distribute it between different classes (and try to achieve a one single responsibility class).
Perhaps, instead of having a DestinationRule class, it would be better to have a DestinationRuleValidator.ts where it takes a DR and computes the whole validation.

hhovsepy:
@lucasponce for the validation icon in overview, I had a RFE https://issues.jboss.org/browse/KIALI-2152

xeviknal:
Should we fix the missing status icon into KIALI-2152 instead of here, @lucasponce? In master this is still missing. wdyt?
@lucasponce your comments has been approached, thx for them!

lucasponce:
Comments seems addressed.
I run a smoke test and PR seems safe.

PR #886, opened 2018-12-14
KIALI-1978 Show custom metrics tabs

jotak:
- Show metrics tabs per identified dashboards
- New Dashboard component to display charts
- Make old Metrics page and new Custom metrics use this dashboard

JIRA: https://issues.jboss.org/browse/KIALI-1978
Needs https://github.com/kiali/kiali/pull/735


abonas:
is there a screenshot or this is infrastructure only?

abonas:
this looks like a very big PR. can't it be split to several?
Also, I am wondering about the naming of the tabs. 
right now there is the info tab for general information. And then Inbound metrics and outbound metrics tabs - those are clear. And in the above example there's the vert.x tab which is also about metrics, but the  tab name doesn't suggest so. is there any other non metric information expected in the new tabs?
cc @pilhuhn

lucasponce:
> And in the above example there's the vert.x tab which is also about metrics, but the tab name doesn't suggest so. is there any other non metric information expected in the new tabs?

I also thought about it for me it would be more clear with something like "Vert.x Metrics" or "Runtime Vert.x Metrics" but I won't change the layout of the tabs.

jotak:
I sure, the tab doesn't seem correctly named. Hopefully this is just config (and not part of this PR), as the dashboard title is given by the MonitoringDashboard resource that I pushed as an example for the rest of the work on runtimes epic. We can fix it (on backend), but anyway it's not the definitive dashboard.

jotak:
I've split in two commits, but the first one is still much bigger. Can't do better without too much hurting consistency

lucasponce:
> I've split in two commits, but the first one is still much bigger. Can't do better without too much hurting consistency

IMO is fine to have it in a single PR, all things seem related for this work.

src/components/MetricsOptions/IstioMetricsOptionsBar.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { URLParams, HistoryManager } from '../../app/History';
+import { Reporter, Direction, MetricsOptions } from '../../types/MetricsOptions';
+
+import MetricsOptionsBar from './MetricsOptionsBar';
+
+export interface IstioMetricsOptionsBarProps {
+  direction: Direction;
+}
+
+export class IstioMetricsOptionsBar extends MetricsOptionsBar<IstioMetricsOptionsBarProps> {
```
xeviknal: After some time in the React world, I am start seeing that inheritance aren't one of the most preferred tools. Then, maybe a little late or maybe it could be set as ad-hoc dev-task, do you think this solution could be approached in more react-way? Like using composition of components?
Here a piece of documentation regarding this inheritance thing: https://reactjs.org/docs/composition-vs-inheritance.html#so-what-about-inheritance

Not blocking, but I guess that we all should be aware of this stuff.

jotak: Good point, I tried to avoid inheritance but sometimes I wonder if splitting into a lot of smaller components (to allow composition) really worth a shot. In a first revision I had some inheritance between the Dashboard component and CustomMetrics and IstioMetrics components, but in this case it was quite obvious to change inheritance for composition.
Here with OptionsBar the trade is different and I suppose it would mean a split in many smaller parts to assemble.
But I'll give it another try this week :)

hhovsepy:
Minor issue found on core: https://issues.jboss.org/browse/KIALI-2179

src/components/Metrics/Dashboard.tsx
```diff
@@ -0,0 +1,107 @@
...
+});
+
+const expandedChartBackLinkStyle = style({
+  marginTop: '-1.7em',
+  textAlign: 'right'
+});
+
+type DashboardProps = {
+  dashboard: M.MonitoringDashboard;
+  labelValues: M.AllPromLabelsValues;
```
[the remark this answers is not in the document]
jotak: These are the prometheus labels+values that come along with the metrics, and they're associated with a boolean flag for show/hide on checkbox toggling.
It's briefly documented in types/metrics.ts

jotak:
DNM: removing inheritance

jotak:
@xeviknal I removed inheritance, you can see the diff in last commit. Splitting toolbar content into smaller components made the toolbar components themselves almost useless so I deleted them, moving their logic between Istio/CustomMetric main components and the new smaller components (plus also some shared logic in the Helper class)

src/components/Metrics/CustomMetrics.tsx
```diff
@@ -0,0 +1,139 @@
+import * as React from 'react';
+import { connect } from 'react-redux';
+import { RouteComponentProps, withRouter } from 'react-router';
+import { Toolbar, ToolbarRightContent, FormGroup } from 'patternfly-react';
+
+import RefreshContainer from '../../containers/RefreshContainer';
+import * as API from '../../services/Api';
+import { KialiAppState } from '../../store/Store';
+import { DurationInSeconds } from '../../types/Common';
+import * as M from '../../types/Metrics';
```
lucasponce: Personally, I prefer to not use short acronyms, but no change needed, just more a comment.

jotak:
@lucasponce I've rebased. The issue that @hhovsepy reported is actually a backend (cache) issue, so won't be part of this PR anyway.

PR #887, opened 2018-12-17
KIALI-2110 Updates to cytoscape 3.3.1

josejulio:
Fixes a bug that prevents some nodes from being removed when adding
 and removing the unused nodes


josejulio:
@jshaughn please take a look and see if this fixes the issue

PR #898, opened 2018-12-19
KIALI-2175 Hide namespace selection when rendering node detail graph

jshaughn:
kiali-2175

### DIFF
--- a/src/routes.tsx
+++ b/src/routes.tsx
@@ -133,7 +133,7 @@ const pathRoutes: Path[] = [
 
 const secondaryMastheadRoutes: Path[] = [
   {
-    path: '/graph',
+    path: '/graph/namespaces',
     component: DefaultSecondaryMasthead
   }
 ];


abonas:
can you describe the use case? I don't understand why and when the ns selector needs to be hidden.
It might be confusing to users. For example what will happen if it's hidden and then user navigates to another page? Have you consulted UX on this change? 
cc @pilhuhn

josejulio:
Would it work that changing namespace removes/reset the selected node?

abonas:
> This is how it used to work (BTW: I still like the old namespace drop down location - I really don't like the new position or look - hopefully that gets fixed in the next sprint):

the location shall remain as is, the look was redesigned:
https://github.com/kiali/kiali-design/issues/67

jshaughn:
@josejulio I'm not sure what you mean.  Are you saying that we could keep it and if there is a change then bring the user back to a namespaces graph? Maybe, I don' know.  If UX wants to redefine this whole node detail graph behavior it's fine by me but I think it's fairly clear as-is, and for now I think we need to fix the issue.

jmazzitelli:
> the location shall remain as is, the look was redesigned:

hopefully the look changes - it just doesn't look "right". I guess because it doesn't "look" like a drop down menu (its just all text). It seems out of place the way it is. Is this going to be fixed in the next sprint or is this how the dropdown is going to look going forward?

PR #905, opened 2019-01-04
Kiali-2196 Graph Find

jshaughn:
Add a DSL-based find feature to highlight graph elements of interest.

@jmazzitelli @josejulio @mwringe There is still some console logging but the code can be reviewed. More important is that you run this and see how it works for you.  Try some searches, see if things break or don't work as you expect.  Use the help, see if it makes sense and if the searching is somewhat intuitive.  Basically looking for some user feedback.


abonas:
can u please add a screenshot?

jshaughn:
Made a few changes:
- don't disable while refreshing
- break up node and edge help into two cleaner columns: criteria and keywords
- fix  help dialog height issue

Also, adding some screenshots:
image

image

josejulio:
image

This table is confusing to me, I was trying to understand how  `httpin <op> <number>`  with `[is] appnode`  and why the `keyword` `[is] wlnode | workloadnode` had no `criteria`.

IIUC, the keywords are a kind of boolean criteria, i don't think we should mix them in a table to save space. I would think we should consider using other tabs.

mwringe:
A couple of things I marked from this screenshot:

search

- The help text has 3 different text alignment, should these be aligned differently?

- The app box isn't highlighted for my query. Is this by design?

- "Hint #1" and "Hint #2" are not expressions. Can these be moved out of the expression column? Maybe these just need to be added at the bottom of the table?

- For the expression "rt > 500 (requires response time edge labels)", should the 'requires response time edge labels' part go under the Description column instead of the expression column?

In this screenshot, product page v1 has two outgoing connections that are are '50%' but they are not highlighted (I suspect this is due to a rounding error where we only show '50%' in the graph but the real values are slightly off)

screenshot from 2019-01-04 15-59-50

See also:

screenshot from 2019-01-04 15-59-25

Some weird transparent thing happening with the help text here (see the 'vs | virtual service' line)

screenshot from 2019-01-04 15-58-36

jshaughn:
@mwringe 
that's the standard styling for dialog header, tabs, and tables.  I'll move the tabs over to align with the table columns.  Let me know if you like it better.
I don't think we want to highlight the app boxes.  I tried it and I it looks odd.  Also, when do you do it, only when all members are also highlighted, or just any.  I think we should leave it unhighlighted.
+1 oh hints, I renamed to "Tip" and moved them to the description column.  Another possibility would be to add a new "Tips" tab. Let me know...
+1 on rt, moved note to description
I will look into the numerical =, not sure about that one...
The "transparent help" issue is fixed.

src/components/GraphFilter/GraphSettings.tsx
```diff
@@ -4,16 +4,16 @@ import * as React from 'react';
...
 import { ListPagesHelper } from '../../components/ListPage/ListPagesHelper';
 import { GraphFilterState, KialiAppState } from '../../store/Store';
 import { KialiAppAction } from '../../actions/KialiAppAction';
 import { GraphFilterActions } from '../../actions/GraphFilterActions';
 import { GraphType } from '../../types/Graph';
 import { PfColors } from '../Pf/PfColors';
+import { Omit } from 'lodash';
 
-type ReduxProps = Omit<GraphFilterState, 'showLegend'> & {
+type ReduxProps = Omit<Omit<GraphFilterState, 'showLegend'>, 'showFindHelp'> & {
```
josejulio: I think you can write this as `type ReduxProps = Omit<GraphFilterState, 'showLegend' | 'showFindHelp'> & {`

abonas:
it looks like there is some inconsistency in terminology:
1. for workload there's wl, workload and also wlnode. can only one be picked and used? I also am in favor of clear names, to me workload is the most clear, wl takes time to remember, and wlnode - not sure that the suffix node is necessary.
2. similar to 1 above - svc,service and servicenode, svcnode.
3. for containing I would use the keyword contains instead of *=
4. what is outside/outsider?

src/components/GraphFilter/GraphFind.tsx
```diff
@@ -0,0 +1,372 @@
...
+        const selector = op !== '=' || 0 !== Number(val) ? '[rate ' + op + ' ' + val + ']' : '[^rate]';
+        return { target: 'node', selector: selector };
+      }
+      case 'httpout': {
+        const selector = op !== '=' || 0 !== Number(val) ? '[rateOut ' + op + ' ' + val + ']' : '[^rateOut]';
+        return { target: 'node', selector: selector };
+      }
+      case 'name':
+        if (conjunctive) {
+          MessageCenterUtils.add('Find values do not allow AND expressions with "find by name".');
```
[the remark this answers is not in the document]
jshaughn: @josejulio +1, will ensure all messages include the expression.

jshaughn:
@abonas 

> for workload there's wl, workload and also wlnode. can only one be picked and used? I also am in favor of clear names, to me workload is the most clear, wl takes time to remember, and wlnode - not sure that the suffix node is necessary.

workload | wl are for finding nodes with a matching workload name.
workloadNode | wlnode are for finding nodes of type workloadNode.

It is analogous for app and service.  Note that I have revised the help with better indication of what the variable is meant to be (e.g. \<workloadName\> instead of \<val\>).

> for containing I would use the keyword contains instead of *=

I think some people (like me) prefer the less verbose numerical operators so I'd opt to keep them, I think we need to get some more input, or a formal UX decision, on whether to keep the various aliases for operators and operands, or to remove the potential over-flexibility.

> what is outside/outsider?

An Outsider node is one that is outside the requested namespaces. These are typically indicated with the topology or reverse lock background images, depending on accessibility.

jshaughn:
@jmazzitelli @mwringe @josejulio @abonas I'm looking to remove the DoNotMerge.  If you have any blockers let me know.  I think I have addressed the comments so far and any further code changes will only be to respond to further comments (other than to remove a few console logs).

abonas:
> workload | wl are for finding nodes with a matching workload name.
> workloadNode | wlnode are for finding nodes of type workloadNode.

I reread the above several times, I still don't understand the difference.
There are nodes on the graph representing workloads. Those nodes are workloads. So I am still confused why there are several different terms for them.

jshaughn:
> workload | wl are for finding nodes with a matching workload name.
> workloadNode | wlnode are for finding nodes of type workloadNode.
> I reread the above several times, I still don't understand the difference. There are nodes on the graph representing workloads. Those nodes are workloads. So I am still confused why there are several different terms for them.

@abonas , ah, OK, now I understand your question.    Perhaps it is best answered with screenshots using bookinfo:

Expression "workload contains reviews" :
image

Expression "workload contains reviews and is workloadnode" :
image

Basically, versioned app nodes are backed by workloads and so can be highlighted when searching against workload names.  In general the xxxnode operands limit your results to nodes of the desired type.  Name searches may cross node types in various situations.

jshaughn:
> This illustrates how we need to document better the concepts in the graph. The different node types is one thing. Then when you say "versioned app nodes are backed by workloads" - I don't think people understand what you are saying -- I do :-) but I'm positive most people won't understand it at first reading.

I'm not against doco but I think understanding the graph has a lot to do with understanding Istio in general.  If you click on a versioned app node the summary panel will clearly show you the workload link that "backs" the node.

abonas:
so what is expected to be found/not found for the second example? ""workload contains reviews and is workloadnode" - I'm looking at the screenshot above where it seems nothing is highlighted and I don't understand why.

jshaughn:
> so what is expected to be found/not found for the second example? ""workload contains reviews and is workloadnode" - I'm looking at the screenshot above where it seems nothing is highlighted and I don't understand why.

Nothing is highlighted because there are no workload (circle) nodes in the graph.  The graph only has app (square), service (triangle) and service entry (tag) nodes.

abonas:
> (square), service (triangle) and service entry (tag) nodes.

so in that case why the first query is higlighting something?
if the selected graph is not showing workloads, "workload contains reviews" should not be highlighting anything imo.
this is too complicated IMO

pilhuhn:
@jshaughn Could you make a container image available to play with the current state of affairs?

jmazzitelli:
> Could you make a container image available to play with the current state of affairs?

@pilhuhn see https://cloud.docker.com/repository/docker/jmazzitelli/kiali/tags - I put "pr905" image for people to play with.

jshaughn:
@abonas It seems you are confused between node types and names.  The 'app', 'service' and 'workload' operands are for matching against [string] names and 'appnode', 'servicenode' and 'workloadnode' are for matching against node type.  The other thing to know is that certain node types can be associated with multiple names.  Versioned app nodes have app and workload; service nodes have service and app, etc. 

I think we do need the ability to find using both names and node types.

* Note that what I'm calling 'app' name is actually the app label value.

jshaughn:
@jmazzitelli pushed fix for your issue, and a few others

jmazzitelli:
'splain. httpin > 0.351 but its highlighting things with 0.33 rate.

image

jshaughn:
> 'splain. httpin > 0.351 but its highlighting things with 0.33 rate

I can't really explain it.  I don't think there is an issue with the query parsing or execution.  I think maybe there is an issue with auto-refresh such that the Find results are not showing up on screen as expected.  It seems sometimes to be one refresh behind.

abonas:
> @abonas It seems you are confused between node types and names. The 'app', 'service' and 'workload' operands are for matching against [string] names and 'appnode', 'servicenode' and 'workloadnode' are for matching against node type. The other thing to know is that certain node types can be associated with multiple names. Versioned app nodes have app and workload; service nodes have service and app, etc.
> 
> I think we do need the ability to find using both names and node types.
> 
> * Note that what I'm calling 'app' name is actually the app label value.


then the syntax should be clear this is about type and name. I am not against the functionality, I just think the syntax is not intuitive and is confusing.
why not do something like  'type = workload AND name = X'? IMO it leaves less room for confusion.

jshaughn:
@pilhuhn @abonas @jmazzitelli  OK, So the current operands around find-name and find-nodeTypeseem to be causing confusion.  Let's agree on new operands before making changes.

Please refer to this [google doc](https://docs.google.com/document/d/1FB4h7rEJf2-kCJWIN4n22YRHu-cIeipe5YPDTx39OYo/edit) for the discussion:

src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -371,6 +380,7 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
     };
     cy.scratch(CytoscapeGlobalScratchNamespace, globalScratchData);
 
+    console.log('Start Batch - CyGraph');
```
jmazzitelli: these should be console.debug

src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -347,6 +355,7 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
     if (!cy) {
       return;
     }
+    console.log('Start - CyGraph');
```
jmazzitelli: these should be console.debug

src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -407,6 +417,7 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
     this.turnNodeLabelsTo(cy, this.props.showNodeLabels);
 
     cy.endBatch();
+    console.log('End Batch - CyGraph');
```
jmazzitelli: these should be console.debug

src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -428,6 +439,7 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
     if (this.props.showTrafficAnimation) {
       this.trafficRenderer.start();
     }
+    console.log('End - CyGraph');
```
jmazzitelli: these should be console.debug

src/components/GraphFilter/GraphFind.tsx
```diff
@@ -49,13 +48,8 @@ export class GraphFind extends React.PureComponent<GraphFindProps> {
...
   componentDidUpdate(prevProps: GraphFindProps) {
-    if (!(this.props.graphInfo && this.props.graphInfo.kind === 'graph')) {
-      return;
-    }
-    if (this.cy !== this.props.graphInfo.graphReference) {
-      this.cy = this.props.graphInfo.graphReference;
-    }
-    if (this.findValue.length > 0 && !this.props.isLoading) {
+    if (this.findValue.length > 0 && this.props.cyData.updateTimestamp !== prevProps.cyData.updateTimestamp) {
+      console.log('CDI Find');
```
jmazzitelli: these should be console.debug

src/components/GraphFilter/GraphFind.tsx
```diff
@@ -115,16 +108,21 @@ export class GraphFind extends React.PureComponent<GraphFindProps> {
...
     if (selector) {
-      this.cy.elements(selector).addClass('find');
+      console.log('Start Batch - Find');
+      cy.startBatch();
+      // unhighlight old find-hits
+      cy.elements('*.find').removeClass('find');
+      // add new find-hits
+      cy.elements(selector).addClass('find');
+      cy.endBatch();
+      console.log('End Batch - Find');
```
jmazzitelli: these should be console.debug

src/components/GraphFilter/GraphFind.tsx
```diff
@@ -115,16 +108,21 @@ export class GraphFind extends React.PureComponent<GraphFindProps> {
...
       return;
     }
-    // unhighlight old find
-    this.cy.elements('*.find').removeClass('find');
-
+    const cy = this.props.cyData.cyRef;
     const selector = this.parseFindValue(this.findValue);
     if (selector) {
-      this.cy.elements(selector).addClass('find');
+      console.log('Start Batch - Find');
```
jmazzitelli: these should be console.debug

src/components/CytoscapeGraph/CytoscapeGraph.tsx
jshaughn: yes, I had a comment way above saying that console logging needed to be removed before the DNM label is removed.   I think it's probably fine to remove it now, so I will...

jshaughn:
@jmazzitelli @mwringe @josejulio @abonas I am removing the DoNotMerge label as I believe the feature is stable and the code is in decent shape for merging.  I will now just defer to approvals for a merge.  I would like to see this go in so that we can get some feedback during the next sprint, and make DSL changes as needed.

PR #906, opened 2019-01-06
KIALI-1185 Implement Openshift OAuth support

cfcosta:
This is a big one, and touch many parts of the system, so it should be reviewed with care.

The main idea here is to enable Openshift users to authorize themselves using the OAuth interface provided, and as doing so, also simplifying the task of adding new behaviours and protocols for the user to authorize from.

Following [Kent Beck's](https://twitter.com/kentbeck/status/250733358307500032) advice, to do something hard, you refactor the code until it becomes easy, then do the easy change.

## What changed:

1. All endpoints for authentication (generating tokens, validating tokens) now return the same interface:

```typescript
export interface Session {
  token: string;
  expiresOn: string;
  username: string;
}
```
The usual expiration time calculation was removed, so now the endpoint returns when the token is going to expire, and we just check with that.

2. [https://momentjs.com](MomentJS) is being used to have more robust time calculations and parsing.
3. All login strategies are now defined on `src/services/Login.ts`, with separated classes implementing the same interface:

```typescript
interface LoginStrategy<T = {}> {
  prepare: (info: AuthInfo) => Promise<Result>;
  perform: (request: DispatchRequest<T>) => Promise<LoginResult>;
}
```

The `prepare()` method process some authorization information with the interface:

```typescript
interface AuthInfo {
  strategy: 'login' | 'anonymous' | 'openshift';
  authorizationEndpoint?: string;
}
```
And returns one of those values:

```typescript
enum Result {
  hold = 'hold',
  continue = 'continue',
  failure = 'failure'
}
```

1. `hold` means to stop the authorization flow to wait for some external processing that at some point is going to close the page. Examples: redirecting to an authorization endpoint.
2. `continue` means to continue the authorization flow.
3. `failure` short-circuits the process and force logging out the user.

For `perform()`, it does the meat of the authorization process. The implementation should be straightforward, with an example:

```typescript
class AnonymousLogin implements LoginStrategy {
  public async prepare(info: AuthInfo) {
    return Result.continue;
  }

  public async perform(request: DispatchRequest<{}>): Promise<LoginResult> {
    const session: Session = (await API.login()).data;

    return {
      status: Result.success,
      session: session
    };
  }
}
```

### How to test

1. Update the branches on both repositories (core and UI).
2. Build the UI.
3. Compile with: `AUTH_STRATEGY=<strategy> CONSOLE_VERSION=local CONSOLE_LOCAL_DIR=<ui dir> make build docker-build openshift-deploy`, where `<strategy>` is one of `openshift`, `login`, `anonymous`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,6 @@
   "name": "@kiali/kiali-ui",
   "version": "0.15.0",
   "description": "React UI for [Kiali](https://github.com/kiali/kiali).",
-  "proxy": "https://kiali-istio-system.10.7.0.11.nip.io",
   "keywords": [
     "istio service mesh",
     "kiali",
--- a/package.json
+++ b/package.json
@@ -2,6 +2,7 @@
   "name": "@kiali/kiali-ui",
   "version": "0.15.0",
   "description": "React UI for [Kiali](https://github.com/kiali/kiali).",
+  "proxy": "https://kiali-istio-system.10.7.0.11.nip.io",
   "keywords": [
     "istio service mesh",
     "kiali",
@@ -72,6 +73,7 @@
     "json-beautify": "1.0.1",
     "lodash": "4.17.11",
     "logfmt": "^1.2.1",
+    "moment": "2.23.0",
     "patternfly": "3.59.1",
     "patternfly-react": "2.25.3",
     "patternfly-react-extensions": "2.16.8",
--- a/src/actions/LoginActions.ts
+++ b/src/actions/LoginActions.ts
@@ -1,3 +1,5 @@
+import { RawDate } from '../types/Common';
+
 import { ActionType, createAction } from 'typesafe-actions';
 import { LoginSession, LoginStatus } from '../store/Store';
 
@@ -13,6 +15,7 @@ export interface LoginPayload {
   status: LoginStatus;
   session?: LoginSession;
   error?: any;
+  uiExpiresOn: RawDate;
 }
 
 // synchronous action creators
@@ -29,7 +32,8 @@ export const LoginActions = {
     resolve({
       status: LoginStatus.loggedIn,
       session: session,
-      error: undefined
+      error: undefined,
+      uiExpiresOn: session.expiresOn
     } as LoginPayload)
   ),
   loginFailure: createAction(LoginActionKeys.LOGIN_FAILURE, resolve => (error: any) =>
--- a/src/actions/LoginActions.ts
+++ b/src/actions/LoginActions.ts
@@ -1,5 +1,5 @@
 import { ActionType, createAction } from 'typesafe-actions';
-import { Session, LoginStatus } from '../store/Store';
+import { LoginSession, LoginStatus } from '../store/Store';
 
 enum LoginActionKeys {
   LOGIN_REQUEST = 'LOGIN_REQUEST',
@@ -11,21 +11,21 @@ enum LoginActionKeys {
 
 export interface LoginPayload {
   status: LoginStatus;
-  session?: Session;
+  session?: LoginSession;
   error?: any;
 }
 
 // synchronous action creators
 export const LoginActions = {
   loginRequest: createAction(LoginActionKeys.LOGIN_REQUEST),
-  loginExtend: createAction(LoginActionKeys.LOGIN_EXTEND, resolve => (session: Session) =>
+  loginExtend: createAction(LoginActionKeys.LOGIN_EXTEND, resolve => (session: LoginSession) =>
     resolve({
       status: LoginStatus.loggedIn,
       session: session,
       error: undefined
     } as LoginPayload)
   ),
-  loginSuccess: createAction(LoginActionKeys.LOGIN_SUCCESS, resolve => (session: Session) =>
+  loginSuccess: createAction(LoginActionKeys.LOGIN_SUCCESS, resolve => (session: LoginSession) =>
     resolve({
       status: LoginStatus.loggedIn,
       session: session,
--- a/src/actions/LoginActions.ts
+++ b/src/actions/LoginActions.ts
@@ -21,13 +21,15 @@ export const LoginActions = {
   loginExtend: createAction(LoginActionKeys.LOGIN_EXTEND, resolve => (session: Session) =>
     resolve({
       status: LoginStatus.loggedIn,
-      session: session
+      session: session,
+      error: undefined
     } as LoginPayload)
   ),
   loginSuccess: createAction(LoginActionKeys.LOGIN_SUCCESS, resolve => (session: Session) =>
     resolve({
       status: LoginStatus.loggedIn,
-      session: session
+      session: session,
+      error: undefined
     } as LoginPayload)
   ),
   loginFailure: createAction(LoginActionKeys.LOGIN_FAILURE, resolve => (error: any) =>
@@ -40,7 +42,8 @@ export const LoginActions = {
   logoutSuccess: createAction(LoginActionKeys.LOGOUT_SUCCESS, resolve => () =>
     resolve({
       status: LoginStatus.loggedOut,
-      session: undefined
+      session: undefined,
+      error: undefined
     } as LoginPayload)
   )
 };
--- a/src/actions/LoginActions.ts
+++ b/src/actions/LoginActions.ts
@@ -1,6 +1,5 @@
 import { ActionType, createAction } from 'typesafe-actions';
-import { Token } from '../store/Store';
-import { config } from '../config/config';
+import { Session, LoginStatus } from '../store/Store';
 
 enum LoginActionKeys {
   LOGIN_REQUEST = 'LOGIN_REQUEST',
@@ -11,43 +10,38 @@ enum LoginActionKeys {
 }
 
 export interface LoginPayload {
-  logged?: boolean;
-  sessionTimeOut?: number;
-  token: Token;
-  username: string;
-}
-
-export interface LoginFailurePayload {
-  error: any;
+  status: LoginStatus;
+  session?: Session;
+  error?: any;
 }
 
 // synchronous action creators
 export const LoginActions = {
   loginRequest: createAction(LoginActionKeys.LOGIN_REQUEST),
-  loginExtend: createAction(
-    LoginActionKeys.LOGIN_EXTEND,
-    resolve => (token: Token, username: string, currentTimeOut: number) =>
-      resolve({
-        token: token,
-        username: username,
-        sessionTimeOut: currentTimeOut + config.session.extendedSessionTimeOut
-      } as LoginPayload)
+  loginExtend: createAction(LoginActionKeys.LOGIN_EXTEND, resolve => (session: Session) =>
+    resolve({
+      status: LoginStatus.loggedIn,
+      session: session
+    } as LoginPayload)
   ),
-  loginSuccess: createAction(
-    LoginActionKeys.LOGIN_SUCCESS,
-    resolve => (token: Token, username: string, currentTimeOut?: number) =>
-      resolve({
-        token: token,
-        username: username,
-        logged: true,
-        sessionTimeOut: currentTimeOut || new Date().getTime() + config.session.sessionTimeOut
-      } as LoginPayload)
+  loginSuccess: createAction(LoginActionKeys.LOGIN_SUCCESS, resolve => (session: Session) =>
+    resolve({
+      status: LoginStatus.loggedIn,
+      session: session
+    } as LoginPayload)
   ),
   loginFailure: createAction(LoginActionKeys.LOGIN_FAILURE, resolve => (error: any) =>
-    resolve({ error: error } as LoginFailurePayload)
+    resolve({
+      status: LoginStatus.error,
+      session: undefined,
+      error: error
+    } as LoginPayload)
   ),
   logoutSuccess: createAction(LoginActionKeys.LOGOUT_SUCCESS, resolve => () =>
-    resolve({ logged: false } as LoginPayload)
+    resolve({
+      status: LoginStatus.loggedOut,
+      session: undefined
+    } as LoginPayload)
   )
 };
 
--- a/src/actions/LoginThunkActions.ts
+++ b/src/actions/LoginThunkActions.ts
@@ -45,8 +45,8 @@ const loginSuccess = async (dispatch: KialiDispatch, session: LoginSession) => {
 // The `data` argument is defined as `any` because the dispatchers receive
 // different kinds of data (such as e-mail/password, tokens).
 const performLogin = (dispatch: KialiDispatch, state: KialiAppState, data?: any) => {
-  const bail = (error: Login.LoginResult) =>
-    data ? dispatch(LoginActions.loginFailure(error)) : dispatch(LoginActions.logoutSuccess());
+  const bail = (loginResult: Login.LoginResult) =>
+    data ? dispatch(LoginActions.loginFailure(loginResult.error)) : dispatch(LoginActions.logoutSuccess());
 
   Dispatcher.prepare().then((result: AuthResult) => {
     if (result === AuthResult.CONTINUE) {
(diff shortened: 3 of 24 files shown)


package.json
```diff
@@ -2,6 +2,7 @@
   "name": "@kiali/kiali-ui",
   "version": "0.13.0",
   "description": "React UI for [Kiali](https://github.com/kiali/kiali).",
+  "proxy": "https://kiali-istio-system.10.7.0.11.nip.io",
   "keywords": [
```
aljesusg: Remove proxy

src/actions/LoginActions.ts
```diff
@@ -11,43 +10,41 @@ enum LoginActionKeys {
...
-        username: username,
-        sessionTimeOut: currentTimeOut + config().session.extendedSessionTimeOut
-      } as LoginPayload)
+  loginExtend: createAction(LoginActionKeys.LOGIN_EXTEND, resolve => (session: Session) =>
+    resolve({
+      status: LoginStatus.loggedIn,
+      session: session,
+      error: undefined
+    } as LoginPayload)
   ),
```
aljesusg: The configuration about the session is for Kiali session in the UI not for the client, we shouldn't remove this. The ticket of openshift is 24 hours by default.

cfcosta: What do you mean by that? When we extend the token duration, how do we update it?

aljesusg: With another attribute, session was created to control the time of the user in the UI not in the backend, openshift in this case. This was a requirement by UX , maybe we should ask to @kiali/ux-reviewers and @abonas about this.

Is this working only in openshift or you test this in Kubernetes ?

package.json
```diff
@@ -70,6 +71,7 @@
     "deep-freeze": "0.0.1",
     "js-yaml": "3.12.0",
     "lodash": "4.17.11",
+    "moment": "^2.23.0",
```
mtho11: Let's lock in the version.

src/actions/LoginThunkActions.ts
```diff
@@ -1,103 +1,81 @@
...
+import * as Login from '../services/Login';
+
+type KialiDispatch = ThunkDispatch<KialiAppState, void, KialiAppAction>;
+
+const Dispatcher = new Login.LoginDispatcher();
+
+const shouldRelogin = (state?: LoginState): boolean =>
+  !(state && state.status === LoginStatus.loggedIn && new Date() < moment(state.session!.expiresOn).toDate());
+
+const loginSuccess = async (dispatch: KialiDispatch, session: Session) => {
```
mtho11: I'm loving the `async/await`.  Like we have talked about before everything should migrate to this more concise syntax.

src/actions/LoginThunkActions.ts
```diff
@@ -1,103 +1,81 @@
...
+    const response = await API.getServerConfig(authHeader);
+    setServerConfig(response.data);
+  } catch (error) {
+    if (error.response && error.response.status === HTTP_CODES.UNAUTHORIZED) {
+      dispatch(LoginActions.logoutSuccess());
+    }
+  }
+};
+
+const performLogin = (dispatch: KialiDispatch, getState: () => KialiAppState, data?: any) => {
```
mtho11: Can we tighten the `data?: any` to be more specific?

src/components/Nav/UserDropdown.tsx
```diff
@@ -51,11 +54,13 @@ class UserDropdown extends React.Component<UserProps, UserState> {
   }
 
   timeLeft = (): number => {
-    const nowDate = new Date().getTime();
-    if (this.props.sessionTimeOut - nowDate < 1) {
+    const expiresOn = moment(this.props.session.expiresOn).toDate();
```
mtho11: If we are going to use `moment` there is no need for `new Date()` just use `moment()` and its comparisons and arithmetic it's much more concise. There should be no need for `moment(xxxx).toDate()` anywhere with moment conversions.

src/containers/LoginPageContainer.ts
```diff
@@ -6,11 +6,8 @@ import { KialiAppAction } from '../actions/KialiAppAction';
 import LoginThunkActions from '../actions/LoginThunkActions';
 
 const mapStateToProps = (state: KialiAppState) => ({
-  token: state.authentication.token,
-  username: state.authentication.username,
-  logging: state.authentication.logging,
-  error: state.authentication.error,
-  message: state.authentication.message
+  status: state.authentication.status,
```
mtho11: Please use `reselect` selectors they are memoizable and therefore more efficient and also composable.

https://github.com/reduxjs/reselect

Everything should be using **reselect  selectors** now.

src/services/Api.ts
```diff
@@ -59,24 +60,31 @@ const newRequest = <P>(method: HTTP_VERBS, url: string, queryParams: any, data:
     params: queryParams
   });
 
+interface LoginRequest {
+  username: string;
```
mtho11: `string` type is very generic -- it is just string after all. Let's use `UserName` so we can search the codebase for this and have type safety.

src/services/Api.ts
```diff
@@ -59,24 +60,31 @@ const newRequest = <P>(method: HTTP_VERBS, url: string, queryParams: any, data:
...
+  password: string;
```
mtho11: Same for `password` as `username`. We should be able to search for `PasswordType` use and not just depend on naming a string consistently.

src/services/Login.ts
```diff
@@ -0,0 +1,163 @@
...
+// dependencies on the service.
+type Dispatch = any;
+type AppState = any;
+
+// Stores the result of a computation:
+// hold = stop all computation and wait for a side-effect, such as a redirect
+// continue = continue...
+// success = authentication was a success, session is available
+// failure = authentication failed, session is undefined but error is available
+export enum Result {
```
mtho11: The **Result** is very generic there are many kinds of Results. So `AuthResult` or something like that to help comprehension.

src/services/Login.ts
```diff
@@ -0,0 +1,163 @@
...
+  hold = 'hold',
```
mtho11: On `enums` I think we have been using  uppercase items like `SUCCESS` or `MY_COOL_ENUM_ITEM`

src/services/Login.ts
```diff
@@ -0,0 +1,163 @@
...
+  getState: () => AppState;
+  data: T;
+}
+
+class AnonymousLogin implements LoginStrategy {
+  public async prepare(info: AuthInfo) {
+    return Result.continue;
+  }
+
+  public async perform(request: DispatchRequest<{}>): Promise<LoginResult> {
```
mtho11: In `DispatchRequest<{}>`is the `{}` necessary?

src/services/Login.ts
```diff
@@ -0,0 +1,163 @@
...
+    if (window.location.hash.startsWith('#access_token')) {
+      return Result.continue;
+    } else {
+      window.location.href = info.authorizationEndpoint!;
+
+      return Result.hold;
+    }
+  }
+
+  public async perform(request: DispatchRequest<any>): Promise<LoginResult> {
```
mtho11: Can we replace `any` with a type alias or interface?

src/services/Login.ts
```diff
@@ -0,0 +1,163 @@
...
+    this.strategyMapping.set(AuthStrategy.login, new WebLogin());
+    this.strategyMapping.set(AuthStrategy.openshift, new OpenshiftLogin());
+  }
+
+  public async prepare(): Promise<Result> {
+    const info = await this.getInfo();
+    const strategy = this.strategyMapping.get(info.strategy)!;
+
+    try {
+      const delay = async (ms: number = 3000) => {
```
mtho11: `number` can be replaced with `TimeInMillis`

src/store/Store.ts
```diff
@@ -59,18 +59,23 @@ export interface GraphState {
...
 
-export interface Token {
+export enum LoginStatus {
+  logging,
+  loggedIn,
+  loggedOut,
+  error
+}
+
+export interface Session {
```
mtho11: Should we be more specific than `Session` there are many different sessions; or is this the `KialiSession`?

src/store/Store.ts
```diff
@@ -59,18 +59,23 @@ export interface GraphState {
...
   token: string;
-  expired_at: string;
+  expiresOn: string;
```
mtho11: I assume `expiresOn` is some kind of _date_ (or moment now) type that the component can format instead of passing an already formatted string value.

src/store/Store.ts
```diff
@@ -59,18 +59,23 @@ export interface GraphState {
...
+  username: string;
```
mtho11: Some kind of `UserName` type that we can track instead of a generic string.

src/store/Store.ts
```diff
@@ -59,18 +59,23 @@ export interface GraphState {
...
-export interface Token {
+export enum LoginStatus {
+  logging,
+  loggedIn,
+  loggedOut,
+  error
+}
+
+export interface Session {
   token: string;
```
mtho11: There already is a `Token` type is this different?

src/actions/LoginActions.ts
cfcosta: I did test it on kubernetes but only for `anonymous` and `login` strategies, as kubernetes does not provide oauth. Is that what you mean?

src/actions/LoginThunkActions.ts
cfcosta: Agree, next step is to make the whole api async :D

src/store/Store.ts
cfcosta: It actually is the date response from the server itself. I thought about making it a more high level definition with Moment, but that would require some kind of preprocessing that to be honest I'm not much of a fan.

src/containers/LoginPageContainer.ts
mtho11: Ok, we can resolve this one in a later PR.

src/actions/LoginThunkActions.ts
mtho11: What is `data`? What kind of data? By the variable name, I have no clue because everything is data and there is no type so I still have no clue.

mtho11: Just trying to drive toward self-documenting code that is easily grokked without having to dive into the exact context and implementation.

src/services/Login.ts
```diff
@@ -0,0 +1,154 @@
...
+    return {
+      status: AuthResult.SUCCESS,
+      session: session
+    };
+  }
+}
+
+class OpenshiftLogin implements LoginStrategy<any> {
+  public async prepare(info: AuthInfo) {
+    if (info.authorizationEndpoint === undefined) {
```
mtho11: Can just use falsy of `undefined` to simplify to 
```suggestion
    if (!info.authorizationEndpoint) {
```

src/services/Login.ts
```diff
@@ -0,0 +1,154 @@
+import * as API from './Api';
+
+import { AuthInfo, AuthStrategy, AuthResult } from '../types/Auth';
+import { LoginSession } from '../store/Store';
+import { TimeInMilliseconds } from '../types/Common';
+
+// We just mock the types for state for now, to avoid injecting more
+// dependencies on the service.
+type Dispatch = any;
+type AppState = any;
```
mtho11: Isn't `AppState` supposed to be `KialiAppState`. If so, why the downgrade to `any`?

src/services/Login.ts
```diff
@@ -0,0 +1,154 @@
...
+}
+
+interface LoginStrategy<T = {}> {
+  prepare: (info: AuthInfo) => Promise<AuthResult>;
+  perform: (request: DispatchRequest<T>) => Promise<LoginResult>;
+}
+
+interface DispatchRequest<T> {
+  dispatch: Dispatch;
+  getState: () => AppState;
```
mtho11: Why is this a function and not a state? We want to pass the state, not the method that calls returns the state. So wouldn't this be `KialiAppState` as returned by `mapStateToProps`?

src/services/Login.ts
```diff
@@ -0,0 +1,154 @@
...
+
+    return {
+      status: AuthResult.SUCCESS,
+      session: session
+    };
+  }
+}
+
+interface WebLoginData {
+  username: string;
```
mtho11: Stronger types for username/password than `string` so we can search and easily find everywhere a username or password is used.

src/services/Login.ts
```diff
@@ -0,0 +1,154 @@
...
+      return AuthResult.HOLD;
+    }
+  }
+
+  public async perform(_request: NullDispatch): Promise<LoginResult> {
+    const session = (await API.checkOpenshiftAuth(window.location.hash.substring(1))).data;
+
+    return {
+      status: AuthResult.SUCCESS,
+      session: session
```
mtho11: You can use the ES6 shorthand instead of repeating:
```suggestion
      session
```

mtho11:
Looks like most everything is fixed so I won't hold it up any longer ;)

mtho11:
Looks like we may want to use cookies instead of web storage to store our JWT: 

[Where to Store your JWTs – Cookies vs HTML5 Web Storage](https://stormpath.com/blog/where-to-store-your-jwts-cookies-vs-html5-web-storage)

_Of course, this can be in a different PR._

src/actions/LoginActions.ts
mwringe: @aljesusg Can you please update your comments with what your concern is here? What is changing in the UX aspect here?

aljesusg: The session attribute in the UI was controlled by config().session.sessionTimeOut, it was a timeout for the UI, independent of the backend token, this was a request to kick the user from the UI of they have not activity, so my concern is about this. On the other hand is if we remove this session to have all the control in the openshift token, How is going to work for Kiali in Kubernetes without oauth?

mwringe: We don't support Oauth in anything other than OpenShift. For k8s, we use the username and password login.

aljesusg: Ok but with this we are removing the session of the UI, this is my concern, if we are going to control de UI session from the backend or not, there is a pop-up timeout in the UI that it's working with this.

israel-hdez:
After fixing the build, there is an issue with the recent @jshaughn work. It briefly shows the "Sorry there was a problem.." and it goes away. It's trying to read from the prometheus config when is not yet set.

So, not sure if this should go forward.

mwringe:
> After fixing the build, there is an issue with the recent @jshaughn work. It briefly shows the "Sorry there was a problem.."

@israel-hdez  I have seen this if I was logged in previously. Can you try logging out and then back in again?

jshaughn:
@mwringe @israel-hdez Perhaps we should prioritize the @jotak suggestion and immediately implement https://issues.jboss.org/browse/KIALI-2426.  This would remove serverConfig from redux and eliminate the need to get it via REST.

hhovsepy:
Logged in as "admin/admin", namespace selector does not load values:
screenshot from 2019-02-20 17-45-46

hhovsepy:
Login with wrong credentials, no any error message.
screenshot from 2019-02-20 18-01-52

israel-hdez:
@mwringe @jshaughn I would consider (again) merging after this sprint release. My work about moving the token from redux to cookies will also have an impact in the login flow because the token will no longer be accessible to the UI. 

So, this OAuth work, the Token in cookies work and the prometheus config work all are trying to change the login flow.

mwringe:
Yeah, it looks like the PR has at least a few issues when oauth is disabled. We need to get those fixed before we can merge.

jotak:
@jshaughn Not sure how it's related to this PR, but I've started KIALI-2426, backend PR is there https://github.com/kiali/kiali/pull/863 and frontend soon on its way

mwringe:
I created a new PR on @cfcosta branch that fixes the missing error message on the login screen and the namespace drop down.

israel-hdez:
@mwringe Perhaps can you push directly to cfcosta branch? Else, he will need to merge your changes...

mwringe:
@israel-hdez: Ah, ok, that did the trick. Its now pushed to this branch.

hhovsepy:
Verified, the issues with wrong login and namespaces load is fixed.

PR #910, opened 2019-01-07
KIALI-2206 Fetch grafana information on login

mwringe:
** Describe the change **

Loads the grafana details on login instead of just when we verify the credentials on subsequent requests.

Fix for the v0.11 branch. Master is already doing something similar (https://github.com/kiali/kiali-ui/blob/master/src/actions/LoginThunkActions.ts#L25)

** Issue reference **

https://issues.jboss.org/browse/KIALI-2206

** Backwards compatible? **

Yes, everything else should work as expected. This should only fix the broken links for grafana


hhovsepy:
Verified, solution fixes the link issue.

mwringe:
Closing. We will update the release notes instead.

PR #918, opened 2019-01-09
KIALI-2025 Fix tab icons size to not misalign tab markers

israel-hdez:
Tab icons were slightly taller than the text. So, tab markers (the line that appears below the text of the tab header) was being impacted and appearing at different positions when compared to the sibling tabs.

Fixing by slightly reducing the size of the icons.

BEFORE:

(Look the misalignment of the blue and grey lines of Pod/Services tabs)

image

AFTER:

image


src/pages/ServiceDetails/ServiceInfo.tsx
```diff
@@ -42,6 +43,9 @@ interface ValidationChecks {
 }
 
 const tabName = 'list';
+const tabIconStyle = style({
```
mtho11: Nice to see `typestyle` version.

PR #920, opened 2019-01-10
Make Enzyme tests Typescript(ES6) compatible

mtho11:
** Describe the change **
Make the Enzyme tests align with Typescript(ES6) import statements instead of CommonJS (require) statements.

### DIFF
--- a/src/setupTests.ts
+++ b/src/setupTests.ts
@@ -1,5 +1,0 @@
-import * as Enzyme from 'enzyme';
-require('jest-localstorage-mock');
-const Adapter = require('enzyme-adapter-react-16');
-
-Enzyme.configure({ adapter: new Adapter() });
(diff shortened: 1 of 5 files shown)


src/setupTests.ts
```diff
@@ -1,5 +1,4 @@
-import * as Enzyme from 'enzyme';
-require('jest-localstorage-mock');
-const Adapter = require('enzyme-adapter-react-16');
-
-Enzyme.configure({ adapter: new Adapter() });
+import { configure } from 'enzyme';
+import * as ReactSixteenAdapter from 'enzyme-adapter-react-16';
+const adapter = ReactSixteenAdapter as any;
```
josejulio: Is this really required?

mtho11: I believe it is. I copied this code from the Enzyme setup.

mtho11: @josejulio removed the `as any` and everything still works fine.

josejulio: Travis is complaining about something :(

PR #921, opened 2019-01-10
KIALI-1890 UXD: Standarize fetching and refreshing controls

aljesusg:
** Describe the change **

Standarize the toolbar in Workloads, Services and Apps

- Removed rate interval toolbar (WE can change the rateInterval with queryparam _duration_)
- Created the defaultRateInterval => 10 minutes in configuration 

** Issue reference **

[KIALI-1890](https://issues.jboss.org/browse/KIALI-1890)

** Screenshot **

screenshot from 2019-01-10 10-40-41


src/components/ListPage/ListPagesHelper.ts
```diff
@@ -7,7 +7,7 @@ import * as MessageCenter from '../../utils/MessageCenter';
 
 export namespace ListPagesHelper {
   export const perPageOptions: number[] = [5, 10, 15];
-  const defaultDuration = 600;
+  const defaultDuration = config().toolbar.defaultRateInterval;
```
jotak: In config there's already something called "defaultDuration" but which is set to 1 minute.
This is why I think some default values, such as this one, should not be kept in global config be should be just set per page.

aljesusg: The problem is that the graph use this value in the interval duration and toolbar and I don't know if this impact in something else, maybe we can ask to saturn team

jotak: I know this is why I think it's better to *not* have it as a global config

aljesusg: ok I'll keep the configuration in the type so

jshaughn: I agree that the config needs some attention.  'Toolbar' should contain only setting global to every toolbar (is there anything actually global to every toolbar?).  Otherwise setting should be under qualified names.  Having said that, we should settle on some common naming for things like pollInterval/refreshInterval.  intervalDuration/duration  (what even is an intervalDuration :) ), etc.

jotak:
@aljesusg PR with only deletions are simply the best ones ;-)

PR #924, opened 2019-01-11
KIALI-2221 Remove 0 on graph screen when no namespace is selected

josejulio:
- We need to use booleans when doing conditional rendering to avoid
   passing the raw values to react.

The zero was appearing because `this.props.graphTimestamp` has the value `0`, that value is given to the renderer.


israel-hdez:
Travis is green, but GitHub didn't update.
Merging...

PR #931, opened 2019-01-14
KIALI-2230 Reduce charts height and do not adjust anymore

jotak:
JIRA: https://issues.jboss.org/browse/KIALI-2230

two columns

Note: backend PR is also part of the work to split in two columns, but can be merged independently: https://github.com/kiali/kiali/pull/780

### DIFF
--- a/src/components/Metrics/MetricsChartBase.tsx
+++ b/src/components/Metrics/MetricsChartBase.tsx
@@ -93,11 +93,6 @@ abstract class MetricsChartBase<Props extends MetricsChartBaseProps> extends Rea
     return fmt + this.props.unit;
   };
 
-  adjustHeight(columns: any[]): number {
-    const series = columns.length - 1;
-    return 350 + series * 23;
-  }
-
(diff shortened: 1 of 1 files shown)


jotak:
cc @lucasponce @abonas @pilhuhn @serenamarie125 
This is what is discussed here: https://github.com/kiali/kiali-design/issues/69
At the moment I didn't remove the legend as I hope these changes will be good enough to have a nice display. (IMO, as long as we can keep the legend it's better to do so, for better experience - as with hidden legend, the charts page would be less screenshotable, harder to understand quickly)

lucasponce:
This is good, and LGTM but as I commented in the design issue
https://github.com/kiali/kiali-design/issues/69#issuecomment-453532311

We miss some recommendation from UX to define a "list" or similar where user can see all available charts.

I.e. in this pic I can see al charts
image

But not on this one:
image

So, this is in the good direction and it will be merged but we'll keep the design issue open to see what it can be applied in a next iteration.

lucasponce:
Related to the discussion:
https://github.com/kiali/kiali-ui/pull/931#issuecomment-454756615

PR #932, opened 2019-01-14
KIALI-2237 Rename chart model counterRate -> metric

jotak:
... as it can stand for both raw (gauges/counters) or rate data

JIRA: https://issues.jboss.org/browse/KIALI-2237

Backend PR: https://github.com/kiali/kiali/pull/781


jotak:
Example:
ex

In this screenshot, "Server request count" is the exact same metric as "Server request count rate" (displayed on the left) but displayed with "raw" dataType instead of "rate".

PR #935, opened 2019-01-15
KIALI-2216 Overview performances

jotak:
Debounce async health calls: namespace health loaded 10 by 10
Also, loading can be canceled if another load is triggered or if leaving the page

JIRA: https://issues.jboss.org/browse/KIALI-2216

See also https://github.com/kiali/kiali/issues/775


jotak:
@exiaohao @mathspanda this should improve usability on the Overview page for your scenario: although it's still going to load all namespaces health, cards will be displayed before health is fetched (so that you can browse namespaces and navigate to whichever) and also health fetching will be cancelled if you navigate out of overview, so that it won't affect performances after that.

Please let me know what you think. Also, would you be able to test with your scenario? (if you don't know how to build I can provide a docker image)

mathspanda:
Code looks conflict. Could you give us an image? @jotak

jotak:
@mathspanda image is there: jotak/kiali:issue-775 (https://hub.docker.com/r/jotak/kiali/tags)

src/pages/Overview/OverviewPage.tsx
```diff
@@ -99,76 +88,100 @@ class OverviewPage extends React.Component<OverviewProps, State> {
...
-        () => API.getNamespaceAppHealth(authentication(), namespace, rateInterval),
-        () => API.getNamespaceServiceHealth(authentication(), namespace, rateInterval),
-        () => API.getNamespaceWorkloadHealth(authentication(), namespace, rateInterval)
-      )();
-      return healthPromise.then(r => ({
-        namespace: namespace,
-        health: r
-      }));
+    // debounce async for back-pressure, ten by ten
+    _.chunk(this.state.namespaces, 10).forEach(chunk => {
```
mtho11: Interesting approach.

hhovsepy:
Verified:
1. Health chunks 10 by 10
2. When refresh is hit before second chunk is loaded, it ignores the missing chunks and starts new chunks of health checks
3. Only for displayed namespaces the health requests are sent, when adding more namespaces, it refreshes health for all namespaces

src/pages/Overview/FiltersAndSorts.ts
```diff
@@ -53,13 +53,19 @@ export namespace FiltersAndSorts {
...
+            return diff;
+          }
+          diff = b.status.inWarning.length - a.status.inWarning.length;
+          if (diff !== 0) {
+            return diff;
+          }
+        } else if (a.status) {
+          return 1;
+        } else if (b.status) {
+          return -1;
```
israel-hdez: So... items whose health is not yet fetched always go first.
Here, I would prefer to show, first, items with errors/warnings; then, items whose health is not fetched yet OR with green health (i.e., give same priority and order by name). Just because, as is, you can get this weird (but funny) behavior where cards start moving like a train:

peek 2019-01-24 18-13

jotak: Oh, I *never* know when I have to return positive or negative :-)
I'll do the contrary!

src/pages/Overview/OverviewPage.tsx
```diff
@@ -99,76 +88,100 @@ class OverviewPage extends React.Component<OverviewProps, State> {
...
+              nsStatus.inSuccess.push(item);
             } else {
-              info.notAvailable.push(item);
+              nsStatus.notAvailable.push(item);
             }
           });
-          if (show) {
-            allNamespaces.push(info);
-          }
+          result.nsInfo.status = nsStatus;
```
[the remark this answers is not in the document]
jotak: It is on purpose that the state isn't set at this point, because this is after a promise that is itself wrapped into a cancelable. If we update state here and the wrapping promise is cancelled, the "setstate" would still occur.

src/pages/Overview/FiltersAndSorts.ts
jotak: done

PR #939, opened 2019-01-16
KIALI-2128: Upgrade from Node.js 8 to 10(LTS) 

mtho11:
** Describe the change **
Upgrade to node.js 10 (the current LTS as of Jan 1, 2019) from node.js 8 (now maintenance)
This is just a build tools upgrade.

https://issues.jboss.org/browse/KIALI-2128


mtho11:
Like you said @israel-hdez in the jira, this also needs to be tested in the productized environment.

josejulio:
Testing locally, I think that after merging we might need to send a message to the mailing lists to tell about this change and maybe add some instructions/links to upgrade to node 10

mtho11:
Node 11 is current version and not LTS though.

josejulio:
> Node 11 is current version and not LTS though.

Ok, is what i got (i didn't choose that specifically) when installing. I'll try to specifically install node 10 and retest

mtho11:
I'm using Node version 10.15.0

jmazzitelli:
I was forced to downgrade to 8 when I ran the production builds a few weeks ago. FWIW

israel-hdez:
I'll try to build and share feedback.

mtho11:
@israel-hdez https://issues.jboss.org/browse/KIALI-2258

mtho11:
@jmazzitelli node version 8 is maintenance release now. So no new updates unless you pay $$. Very dangerous given the many vulnerabilities out there.

jmazzitelli:
> @jmazzitelli node version 8 is maintenance release now. So no new updates unless you pay $$. Very dangerous given the many vulnerabilities out there.

I'm not against this - I'm just saying - our production builds will fail when attempting to use node 10. So, we need to make sure our production builds are fixed so it will work with node 10.

mtho11:
@jmazzitelli, @israel-hdez is looking at that now. I'll put a DNM until we have confirmation.

package.json
```diff
@@ -122,8 +122,8 @@
     "@types/react": "16.7.13"
   },
   "engines": {
-    "node": ">=8.0.0 ",
-    "npm": ">=5.0.0 ",
+    "node": ">=10.0.0 ",
+    "npm": ">=6.0.0 ",
```
mwringe: Is there any requirements why a user would have to now use Node 10 instead of Node 8?

The old "node >= 8.0.0" still allows for someone to use Node 10 if they want.

I don't know if we should prevent someone from building with Node 8 just because of its current support status.

josejulio: If trying to install from fedora's package manager, f27 has v8.12, I think f28 is the same ([see](https://rpms.remirepo.net/rpmphp/zoom.php?rpm=nodejs)) and only f29 has the v10

mtho11:
I guess we will discuss this one more at our morning scrum.

package.json
mtho11: I assume that everyone uses package managers such as `n` or `nvm`?

mtho11: @mwringe I guess I can't understand why anyone would want to run on a maintenance version and not a LTS version. I can understand why someone wouldn't want to be on a current version or nightly build.

mwringe: >  I guess I can't understand why anyone would want to run on a maintenance version and not a LTS version

Lots of people continue to use older versions of things, either due to them not wanting to update yet or using some piece of software somewhere that still needs the older version. Node 8 isn't even end-of-life, that happens in December 2019, its just in maintenance mode. So it should still get things like critical security updates.

Lots of people also install software through the package manager on their system, and Node8 is still being used by default for a bunch of them.

mwringe:
For Kiali, its probably good to nudge people to be using the better supported version of Node.

I don't believe there is any hard technical reason why we have to run Node 10 (ie using some features that don't exist in Node8). But we probably don't want to have to deal with any Node8 issues that might come up. And having all the developers use the newer version is probably better as well.

If anyone really really wants to run with Node8, they can just modify that line locally :)

mtho11:
Can we say in here what this is waiting on? @mwringe @israel-hdez 
Does it look as if it is ready to be merged or else we need to add a DNM?

israel-hdez:
AFAIK, CI pipeline is ready. We should be safe to merge.

mtho11:
Ok, let's merge. I'm already on 10.15.0 things are good.

PR #941, opened 2019-01-17
KIALI-1874 Document url bookmarking strategy with Redux shared state

mtho11:
** Describe the change **
This is a documentation task that adds some more rules aound url bookmarking and Redux to the style guide.

** Issue reference **
https://issues.jboss.org/browse/KIALI-1874

No, just documentation

### DIFF
--- a/STYLE_GUIDE.adoc
+++ b/STYLE_GUIDE.adoc
@@ -175,3 +175,16 @@ MyComponent.propTypes = {
   })
 };
 ----
+
+== Redux and URL Consistency
+
+Every application page should strive to have it's state in Redux so that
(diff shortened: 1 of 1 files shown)


josejulio:
Should we have a specific component that takes care of this?

mtho11:
@mwringe I think was waiting on you. We decided that there are some differences in the way urls are handled, but both groups are satisfied with the current solutions, and other solutions would just introduce more components (essentially moving the code around and maybe further away from the actual code use), but that current solution is satisfactory.

So we will just document the things to be aware of with respect to priority of URLs, Redux and React state.

PR #945, opened 2019-01-18
KIALI-2272 Handle updated graph JSON

jshaughn:
- add some backward compatible (forward-looking) gRPC support
- this also fixes (I think) a few issues in GraphStyles by replacing
  tests for undefined with tests for > 0.

### DIFF
--- a/src/components/CytoscapeGraph/CytoscapeGraphUtils.ts
+++ b/src/components/CytoscapeGraph/CytoscapeGraphUtils.ts
@@ -1,4 +1,8 @@
 export const CyEdge = {
+  grpc: 'grpc',
+  grpcErr: 'grpcErr',
+  grpcPercentErr: 'grpcPercentErr',
+  grpcPercentReq: 'grpcPercentReq',
   http: 'http',
   http3XX: 'http3XX',
   http4XX: 'http4XX',
@@ -15,6 +19,9 @@ export const CyEdge = {
 export const CyNode = {
   app: 'app',
   destServices: 'destServices',
+  grpcIn: 'grpcIn',
+  grpcInErr: 'grpcInErr',
+  grpcOut: 'grpcOut',
   hasCB: 'hasCB',
   hasMissingSC: 'hasMissingSC',
   hasVS: 'hasVS',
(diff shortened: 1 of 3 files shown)


src/components/CytoscapeGraph/CytoscapeGraphUtils.ts
```diff
@@ -1,4 +1,8 @@
 export const CyEdge = {
+  grpc: 'grpc',
+  grpcErr: 'grpcErr',
+  grpcPercentErr: 'grpcPercentErr',
+  grpcPercentReq: 'grpcPercentReq',
   http: 'http',
   http3XX: 'http3XX',
   http4XX: 'http4XX',
```
jmazzitelli: In all of the Go code on server-side, these do not use capital-X - they are all lowercase "http3xx, http4xx" - but yet here I see http3XX and http4XX and http5XX.

Even if this still works (which I am not convinced just by reading this code), we should be consistent -- use lowercase "x" everywhere, server and client side.

jmazzitelli:
'splain:

image

I installed Istio 1.0.5 bookinfo demo via our hack script `hack/istio/install-bookinfo-demo.sh -tg --mongo`

I wasn't aware (at least I never seen it before) that the ratings-v2 called back into ratings.

jshaughn:
@jmazzitelli I can't really explain it.  I'd start by looking at the prom data you get using the new version of bookinfo.  When I come back from PTO tomorrow let's take a look.

jmazzitelli:
I deleted bookinfo project, re-installed, and immediately started seeing this - ignore the colors, probably because the pods didn't come up fully yet, but notice the loopback edges everywhere. I have never seen this before:

image

jmazzitelli:
workload graph:

image

jshaughn:
I can't replicate what you are seeing.  We'll need to look at your system.

mwringe:
I have tried the PRs and I don't see this issue. It seems to work well for me

jshaughn:
@jmazzitelli I really can not explain the loopbacks but I also can't see a reason that the changes in the PR would be related to that sort of misbehavior.  I think it is OK to correct any issues going forward,  At least we'll get a larger audience of users...

PR #946, opened 2019-01-19
Add source-map-explorer to visualize relative package sizes

mtho11:
This is an optional tool to visualize the relative package sizes in the project.
source_map_explorer

First step: `yarn build`
Then: `yarn run analyze`


package.json
```diff
@@ -89,6 +90,7 @@
     "redux-persist": "5.10.0",
     "redux-thunk": "2.3.0",
     "reselect": "4.0.0",
+    "source-map-explorer": "^1.6.0",
```
josejulio: Please move this to dev-dependencies?

mtho11: Ah, right!

PR #948, opened 2019-01-22
KIALI-2269 Update runtimes model to dissociate with dashboards

jotak:
JIRA https://issues.jboss.org/browse/KIALI-2269
Backend PR: https://github.com/kiali/kiali/pull/796

### DIFF
--- a/src/pages/AppDetails/AppDetailsPage.tsx
+++ b/src/pages/AppDetails/AppDetailsPage.tsx
@@ -30,7 +30,7 @@ class AppDetails extends React.Component<RouteComponentProps<AppId>, AppDetailsS
         name: '',
         workloads: [],
         serviceNames: [],
-        customDashboards: []
+        runtimes: []
       }
     };
     this.fetchApp();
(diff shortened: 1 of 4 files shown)


src/pages/AppDetails/AppDetailsPage.tsx
```diff
@@ -113,12 +113,14 @@ class AppDetails extends React.Component<RouteComponentProps<AppId>, AppDetailsS
...
-              {this.state.app.customDashboards.map(dashboard => {
-                return (
-                  <NavItem key={dashboard.template} eventKey={dashboard.template}>
-                    <div>{dashboard.title}</div>
-                  </NavItem>
-                );
+              {this.state.app.runtimes.map(runtime => {
+                return runtime.dashboardRefs.map(dashboard => {
+                  return (
+                    <NavItem key={dashboard.template} eventKey={dashboard.template}>
```
lucasponce: Minor, if you have two runtimes with same template i.e. Vert.x - Client, EAP - Client, this could throw a warning.
Perhaps better to add runtime id in the key.

src/pages/AppDetails/AppDetailsPage.tsx
```diff
@@ -148,21 +150,23 @@ class AppDetails extends React.Component<RouteComponentProps<AppId>, AppDetailsS
...
-                      app={this.props.match.params.app}
-                      template={dashboard.template}
-                    />
-                  </TabPane>
-                );
+              {this.state.app.runtimes.map(runtime => {
+                return runtime.dashboardRefs.map(dashboard => {
+                  return (
+                    <TabPane
+                      key={dashboard.template}
```
lucasponce: Same here, perhaps safer to add the runtime id in the key.

src/pages/AppDetails/AppDetailsPage.tsx
jotak: Hmm I don't think, as the template here refers to the dashboard name, so it is unique (at least in a namespace).

PR #949, opened 2019-01-22
KIALI-2256 Wizard for creating traffic rules 

lucasponce:
This PR enables two operations at Service detail level:
- Create A/B Traffic Routing.
- Delete ALL Traffic Routing.

Basically to not need to go into individual configs to delete all Istio linked with a particular service.

image

The Wizard only performs a "Create" operation, and it's enabled when service has not defined any prior configuration (for existing configuration, standard YAML edition is enabled).

The A/B patter is extended in a generic way, meaning that it works for 1..N cases, the wizard will control the weights between the workloads and it will create a basic DestinationRule/VirtualService pair.

image

image

### DIFF
--- a/src/components/IstioWizards/IstioWizard.tsx
+++ b/src/components/IstioWizards/IstioWizard.tsx
@@ -211,6 +211,10 @@ class IstioWizard extends React.Component<Props, State> {
               label={'Traffic Weight'}
               value={workload.traffic}
               onSlide={value => {
+                console.log('onSlide value: ' + value);
+                if (value > 100) {
+                  value = 100;
+                }
                 this.onWeight(workload.name, value as number);
               }}
             />
--- a/src/components/IstioWizards/IstioWizard.tsx
+++ b/src/components/IstioWizards/IstioWizard.tsx
@@ -39,8 +39,8 @@ class IstioWizard extends React.Component<Props, State> {
     if (this.props.workloads.length === 0) {
       return;
     }
-    const wkTraffic = Math.round(100 / this.props.workloads.length);
-    const remainTraffic = 100 % this.props.workloads.length;
+    const wkTraffic = this.props.workloads.length < 100 ? Math.round(100 / this.props.workloads.length) : 0;
+    const remainTraffic = this.props.workloads.length < 100 ? 100 % this.props.workloads.length : 0;
     const workloads: WorkloadTraffic[] = this.props.workloads.map(workload => ({
       name: workload.name,
       traffic: wkTraffic
--- a/src/components/IstioWizards/IstioWizard.tsx
+++ b/src/components/IstioWizards/IstioWizard.tsx
@@ -238,7 +238,7 @@ class IstioWizard extends React.Component<Props, State> {
   render() {
     return (
       <Wizard show={this.state.showWizard} onHide={this.onClose}>
-        <Wizard.Header onClose={this.onClose} title="Create A/B Traffic Routing" />
+        <Wizard.Header onClose={this.onClose} title="Create Traffic Routing" />
         <Wizard.Body>
           <Wizard.Row>
             <Wizard.Main>{this.renderContent()}</Wizard.Main>
--- a/src/components/IstioWizards/IstioWizard.tsx
+++ b/src/components/IstioWizards/IstioWizard.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { Button, ListView, ListViewIcon, ListViewItem, Wizard } from 'patternfly-react';
+import { Button, Label, ListView, ListViewIcon, ListViewItem, Wizard } from 'patternfly-react';
 import { WorkloadOverview } from '../../types/ServiceInfo';
 import Slider from './Slider/Slider';
 import { DestinationRule, VirtualService } from '../../types/IstioObjects';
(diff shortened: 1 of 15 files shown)


lucasponce:
This PR needs https://github.com/kiali/kiali/pull/810 merged

hhovsepy:
The red frame and hover-over message is nice solution in case of negative.
screenshot from 2019-01-25 17-39-25

pilhuhn:
@lucasponce What about not modifying the other values if the user types in a bad value (<0, >100) in one of the text boxes and only showing the red frame that @hhovsepy shows above for this bad value (and of course disable saving if the red frame appears).

lucasponce:
Yeah, the inputtext can be a source of errors, it doesn't have a good coverage in the original component but as I cloned here, I'll add these use cases there as well.

lucasponce:
@hhovsepy @pilhuhn Ok, I added additional validation into the inputtext.
Let me know if there is some case missing.

hhovsepy:
In the case of single destination, when it is less that 100% traffic, no notification to user why it is not possible to create.
screenshot from 2019-01-25 18-48-50

lucasponce:
@hhovsepy good point, I've added a label with the warning.
I guess will only happen on single traffic destination cases.

lucasponce:
Updated "create" permission from changes on https://github.com/kiali/kiali/pull/810

abonas:
Have you seen the patternfly touchpin control? I think the numeric input should be that one (the stacked version probably is nicer to use), not the one currently used in the PR:
https://www.patternfly.org/pattern-library/forms-and-controls/data-input/#bootstrap-touchspin

Also is the slider used is the one from PF? (I'm not sure)
https://www.patternfly.org/pattern-library/forms-and-controls/slider/

abonas:
If this is to be used for cases where there are less or more than 2 services, the title "A/B testing" should be changed. A/B testing is defined as "A/B testing (sometimes called split testing) is comparing two versions of a web page to see which one performs better" so this is not the right definition for !=2 services.

lucasponce:
Moved "Create A/B Traffic Routing" to "Create Traffic Routing" more generic approach with less interpretation.

src/components/IstioWizards/IstioWizard.tsx
```diff
@@ -0,0 +1,265 @@
...
+    if (prev.length !== current.length) {
+      return false;
+    }
+    for (let i = 0; i < prev.length; i++) {
+      if (!current.includes(prev[i])) {
+        return false;
+      }
+    }
+    return true;
+  };
```
aljesusg: Only proprosal not blocker:
What do you think about ?
```typescript
let found = prev.some( item=> current.includes(item))
return found ? false : true
```

src/components/IstioWizards/IstioWizard.tsx
```diff
@@ -0,0 +1,265 @@
...
+      workloads: []
+    };
+  }
+
+  resetState = () => {
+    if (this.props.workloads.length === 0) {
+      return;
+    }
+    const wkTraffic = Math.round(100 / this.props.workloads.length);
+    const remainTraffic = 100 % this.props.workloads.length;
```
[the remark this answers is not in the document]
lucasponce: For this iteration I didn't consider it should be more than 100 workloads versions per service.
Most scenario are <10 so I guess it could be ok for this start point.
But yeah, good point, these maths can also be checked for large scenarios.

src/components/IstioWizards/IstioWizard.tsx
lucasponce: I think it wouldn't test the same.
In the code, the idea is that

```
prev:  [A, B, C]
current: [A, B, C, D]

compareWorkloads(prev, current) = false // D is not in prev

but I think that 
prev.some( item => current.includes(item) )
would return true as soon as detect that condition is true for at least one of them but not all.
```

aljesusg: ohh I see but you can check not include instead includes

src/components/IstioWizards/IstioWizard.tsx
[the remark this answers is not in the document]
lucasponce: Added a check for this cornercase, basically in the scenario where more than 100 workloads happen, then the pre-calculation is 0 and users need to manually setup the weights.
(Instead to have this pre-calculated one).

lucasponce:
@pilhuhn I think your comments are addressed, let me know.

abonas:
Why a PR with DNM label is merged?
Also, iirc this was not reviewed by UX team. If indeed so, please add this to design repo and next meeting review. thank you

lucasponce:
> Why a PR with DNM label is merged?

Because I forgot to remove the label, but the PR was tested.

> Also, iirc this was not reviewed by UX team. If indeed so, please add this to design repo and next meeting review. thank you

There is a long issue opened since Nov https://github.com/kiali/kiali-design/issues/64 and also some pending request without response in the repo and from email.

We have an agreement with @pilhuhn and @theute to progress on this. 
On this way, we will help UX team to understand better the issue and we won't enter in risk to delay a required feature.

Future refinements can be done in following PRs.

Indeed, the first minimal MVP scenario would be cover, I'll prepare a PoC for next one and now we will have more specific scenarios where UX can propose a better design.

For these situations, we have used the minimal implementation using standard Patternfly components, so any change won't be a problem.

PR #952, opened 2019-01-22
KIALI-1774 Graph Hide

jshaughn:
- Use the same DSL as Graph Find for filtering.
- Also
  - fixes issue with finding no traffic
  - adds 'trafficsource' operand to node help
  - removes unnecessary 'traffic' field from cy node and edge elements

image

@abonas @mwringe There is also a container for this: https://hub.docker.com/r/jshaughn/kiali tag filter.  I'm not sure if we need a UX design issue for this or if it OK from a design perspective.

### DIFF
--- a/src/components/GraphFilter/GraphFind.tsx
+++ b/src/components/GraphFilter/GraphFind.tsx
@@ -62,7 +62,6 @@ export class GraphFind extends React.PureComponent<GraphFindProps> {
       this.handleFind();
     }
     if (this.hideValue.length > 0 && this.props.cyData.updateTimestamp !== prevProps.cyData.updateTimestamp) {
-      this.hiddenElements = undefined;
       this.handleHide();
     }
   }
--- a/src/components/GraphFilter/GraphFind.tsx
+++ b/src/components/GraphFilter/GraphFind.tsx
@@ -32,10 +32,10 @@ export class GraphFind extends React.PureComponent<GraphFindProps> {
     router: () => null
   };
 
-  private filterInputRef;
-  private filterInputValue: string;
-  private filterValue: string;
-  private filteredElements: any | undefined;
+  private hideInputRef;
+  private hideInputValue: string;
+  private hideValue: string;
+  private hiddenElements: any | undefined;
   private findInputRef;
   private findInputValue: string;
   private findValue: string;
@@ -47,10 +47,10 @@ export class GraphFind extends React.PureComponent<GraphFindProps> {
       props.toggleFindHelp();
     }
 
-    this.filterInputRef = React.createRef();
-    this.filterInputValue = '';
-    this.filterValue = '';
-    this.filteredElements = undefined;
+    this.hideInputRef = React.createRef();
+    this.hideInputValue = '';
+    this.hideValue = '';
+    this.hiddenElements = undefined;
 
     this.findInputRef = React.createRef();
     this.findInputValue = '';
@@ -61,9 +61,9 @@ export class GraphFind extends React.PureComponent<GraphFindProps> {
     if (this.findValue.length > 0 && this.props.cyData.updateTimestamp !== prevProps.cyData.updateTimestamp) {
       this.handleFind();
     }
-    if (this.filterValue.length > 0 && this.props.cyData.updateTimestamp !== prevProps.cyData.updateTimestamp) {
-      this.filteredElements = undefined;
-      this.handleFilter();
+    if (this.hideValue.length > 0 && this.props.cyData.updateTimestamp !== prevProps.cyData.updateTimestamp) {
+      this.hiddenElements = undefined;
+      this.handleHide();
     }
   }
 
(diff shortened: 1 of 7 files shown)


src/actions/GraphDataActions.ts
```diff
@@ -60,7 +60,9 @@ const decorateGraphData = (graphData: any) => {
       graphData.nodes = graphData.nodes.map(node => {
         const decoratedNode = { ...node };
         if (node.data.traffic) {
-          node.data.traffic.map(protocol => {
+          const traffic = node.data.traffic;
```
josejulio: I moved all of this to a selector in a previous PR. The goal was to keep the same response from the server in our store.

See: https://github.com/kiali/kiali-ui/blob/master/src/store/Selectors/GraphData.ts#L10

jshaughn: OK, I rebased and fixed things up.  Thanks.

src/components/GraphFilter/GraphFind.tsx
```diff
@@ -100,10 +166,25 @@ export class GraphFind extends React.PureComponent<GraphFindProps> {
...
+    cy.startBatch();
+    // restore old filter-hits
+    if (this.filteredElements) {
+      this.filteredElements.restore();
+      this.filteredElements = undefined;
+      runLayout(cy, this.props.layout);
+    }
+    if (selector) {
+      // remove new filter-hits
+      this.filteredElements = cy.remove(selector);
```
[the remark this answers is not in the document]
jshaughn: I'm not sure.  The problem is that after a refresh an element that was removed may no longer be validly removed.  For example, say the filter is 'http > 50'.  Initially the rate may be 55 and it is removed.  After a refresh it may be 45 and so we need it to come back.  I'm not sure I completely understood your suggestion but I can try playing with it.  I used the remove() and restore() because it seemed to be a built-in way to do this sort of thing.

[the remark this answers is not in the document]
jshaughn: @josejulio I played with it but ended up with basically the same code, just replacing the remove/restore with the style suggestions.  It works but I'm not sure the benefit?  If you have code you want to share that does things in a better way please let me know.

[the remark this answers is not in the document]
josejulio: > 'm not sure. The problem is that after a refresh an element that was removed may no longer be validly removed. For example, say the filter is 'http > 50'.

Ok, i see the problem. In that situation there should be no difference between remove and display:none

src/components/GraphFilter/GraphFind.tsx
```diff
@@ -100,10 +166,25 @@ export class GraphFind extends React.PureComponent<GraphFindProps> {
...
+      return;
+    }
+    const cy = this.props.cyData.cyRef;
+    const selector = this.parseFindValue(this.filterValue);
+    cy.startBatch();
+    // restore old filter-hits
+    if (this.filteredElements) {
+      this.filteredElements.restore();
+      this.filteredElements = undefined;
+      runLayout(cy, this.props.layout);
```
josejulio: Mmh, maybe you don't need this anymore if you use `this.filteredElements.style({display: 'element'});`

src/components/GraphFilter/GraphFind.tsx
```diff
@@ -100,10 +164,29 @@ export class GraphFind extends React.PureComponent<GraphFindProps> {
...
+    const cy = this.props.cyData.cyRef;
+    const selector = this.parseFindValue(this.filterValue);
+    cy.startBatch();
+    // this could also be done useing cy remove/restore but we had better results
+    // using visible/hidden (i.e. display: 'element'/'none' ).  The latter worked
+    // better when hiding animation, and also prevents the need for running
+    // layout after each update (at least in my tests so far..)
+    if (this.filteredElements) {
+      // make visible old filter-hits
+      this.filteredElements.style({ display: 'element' });
```
josejulio: One minor change, before recommending to use display: element / none i though that we did the layout to have a layout without the filtered stuff, this is not the case.
Lets change it to:
visibility: visible or visibility: hidden


screenshot_2019-01-25_14-37-19


according to the docs[1]: 
> A display: none element does not take up space.
> A visibility: hidden element does take up space.

We definitely want it to take space, but be invisible

[1] http://js.cytoscape.org/#style/visibility

jshaughn: @josejulio Yes, visibility is better for the styling.  We had to do some more work to get the surrounding edges set as well but it is fine. Thanks for your help!

mwringe:
We need to have QE take a look and be aware of this. @mattmahoneyrh & @gbaufake Can you take a look?

abonas:
@jshaughn Is this adding a new filtering functionality that is somewhat similar to the Find one?
For the Find, the relevant items that respond to the query are highlighted on the graph, however the graph remains intact with all the other items on it unchanged.
What is the expected behavior for Filter? If some items from the graph respond to the query, what happens to all the other items? Are they still visible? is the graph rebuilt / several graphs are presented in case some links do not respond to the filter?
This should be discussed with UX from both the behavior perspective and reviewing the controls themselves. please add a relevant issue to the design repo.

jshaughn:
Adding link to the UX design issue: https://github.com/kiali/kiali-design/issues/89

jshaughn:
@mattmahoneyrh I have updated the PR with UX feedback.  Basically a terminology change from "Filter" to "Hide", which makes more sense as filter implies the data you want to keep but the feature here is about hiding graph elements.  This PR should be ready for QE acceptance testing.

@josejulio There should not be a need for any further review.

mattmahoneyrh:
The following steps result in the filtered objects not returning after the condition (such as a CB) is removed:

Repro Example:
1) Navigate to Graph Bookinfo namespace
2) Apply CB (Ex: https://github.com/kiali/kiali/blob/master/tests/e2e/assets/bookinfo-reviews-all-cb.yaml)
    Note on Graph that CBs badges are present on 'reviews' nodes
3) Hide one node that has CB badge, such as with filter"CB and version=v1"
    Note: V1 Node and Edge are hidden
4) With Hide filter still present, delete the CB
5) Manually refresh the graph, or wait until graph auto-refreshes
   Note: The hidden node and edge are STILL hidden
6) Clear the Hide filter, and then refresh the graph
   Note: The hidden node and edge are STILL hidden

image

Additional notes:
- Changing the graph layout returns the node, but not the edge
- To restore the node and edge, do one of the folowing: 
   - Refreshing the browser
   - Changing the namespace list
   - Moving off of the graph and then back to the graph again

jshaughn:
@mattmahoneyrh I have pushed a fix for the empty box issue.  I'll look into the refresh issue next.

jshaughn:
@mattmahoneyrh I have pushed a fix for the refresh issue.  Please retest and hopefully we'll be good to go.

PR #954, opened 2019-01-23
KIALI-2289 Fix validation lookup path for VS IstioConfig overview page

burmanm:
** Describe the change **

The "Not all checks passed" was not displayed correctly in the Istio Config Overview page with VirtualServices as it is looking for the wrong path (old path) in the JSON.

** Issue reference **

KIALI-2289


hhovsepy:
Verified:

Before:
screenshot from 2019-01-31 11-42-35


After:
screenshot from 2019-01-31 11-37-23

PR #956, opened 2019-01-24
KIALI-2125 Replace "Destination Services" tab with a "Traffic" tab

israel-hdez:
https://issues.jboss.org/browse/KIALI-2125

image


israel-hdez:
@jshaughn I fixed the code. I, indeed, was waiting for your gRPC server side PR to go in, since it had impact on this one. All should be good now.

I'll ping @rubenvp8510, because the slight change in JSON structure also has some impact in his https://github.com/kiali/kiali-ui/pull/971 PR.

src/components/Details/DetailedTrafficList.tsx
```diff
@@ -0,0 +1,228 @@
+import { Icon } from 'patternfly-react';
+import { TableGrid } from 'patternfly-react-extensions';
+import * as React from 'react';
+import { NodeType } from '../../types/Graph';
+import { REQUESTS_THRESHOLDS } from '../../types/Health';
+
+type DetailedTrafficProps = {
+  direction: 'inbound' | 'outbound';
```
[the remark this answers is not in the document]
israel-hdez: Yes! Ok, changed it.

src/components/Details/DetailedTrafficList.tsx
```diff
@@ -0,0 +1,228 @@
...
+export interface TrafficData {
+  http?: {
+    http: string;
+    httpPercentErr?: string;
+  };
+  tcp?: {
+    tcp: string;
+  };
+}
+
```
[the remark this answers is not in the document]
israel-hdez: Support added :)

jotak:
I'm starting code review, but just a couple of remarks while looking at the screenshot :
- is the "health" column really necessary? as said in the PoC comments it seems like the right horizontal bar is sufficient to show how healthy it is, and btw "health" terminology is already used to denote a more complete aggregation of checks (not only error rates).

- what is the small "split" on horizontal bars? is it health threshold?

- I think a refresh button would be necessary at least, similar to the one on "info" tab. And maybe, for a later step & task, a duration dropdown too.

src/components/Details/DetailedTrafficList.tsx
```diff
@@ -0,0 +1,261 @@
...
+export interface ServiceNode {
+  id: string;
+  type: NodeType.SERVICE;
+  namespace: string;
+  name: string;
+}
+
+export interface UnknownNode {
```
jotak: what's "unknown node" about? (I mean, what could it be for instance?)

src/components/Details/DetailedTrafficList.tsx
```diff
@@ -0,0 +1,261 @@
...
+    if (traffic.protocol === 'tcp') {
+      return (
+        <TableGrid.Col {...healthColumnSizes}>
+          <Icon type="pf" name="unknown" />
+        </TableGrid.Col>
+      );
+    } else {
+      let percentError: number;
+      if (traffic.protocol === 'http') {
+        percentError = traffic.rates.httpPercentErr ? Number(traffic.rates.httpPercentErr) : 0;
```
jotak: I would say, when there is no request the error rate / health should be "n/a" instead of 100% / success. This is how it's done in other places.

jotak:
+1 to @jshaughn about linking to other services / workloads

src/components/Metrics/TrafficDetails.tsx
```diff
@@ -0,0 +1,221 @@
...
+
+type AppProps = {
+  itemType: MetricsObjectTypes.APP;
+  namespace: string;
+  appName: string;
+};
+
+type TrafficDetailsProps = {
+  rateInterval: DurationInSeconds;
+  trafficData: any;
```
jotak: Can we have proper typing here, or is it that much of a pain?

src/components/Metrics/TrafficDetails.tsx
```diff
@@ -0,0 +1,221 @@
...
+            node: this.buildTrafficNode('out', targetNode)
+          });
+        }
+      }
+    });
+
+    return { inboundTraffic, outboundTraffic };
+  };
+
+  private processServicesTraffic = (edges: any, nodes: any, myNode: any) => {
```
jotak: that makes a lot of `any`, it makes it difficult to understand how the data is structured :-(

src/components/Details/DetailedTrafficList.tsx
```diff
@@ -0,0 +1,261 @@
...
+  private renderHealthColumn = (traffic: TrafficData) => {
+    if (traffic.protocol === 'tcp') {
+      return (
+        <TableGrid.Col {...healthColumnSizes}>
+          <Icon type="pf" name="unknown" />
+        </TableGrid.Col>
+      );
+    } else {
+      let percentError: number;
+      if (traffic.protocol === 'http') {
```
jotak: When loading the page just after the pods are started, I got an error below because `traffic.rates` is null. Also `traffic.protocol` was an empty string. The error was gone after some time (I guess after some metrics were filled in prometheus)

jotak: Also I have a similar error when I'm activating traffic mirroring. When I fix this issue, another issue occurs: error rates from mirrored traffic isn't displayed. This is a known issue in the graph but IMO also raises the problem of pulling data from graph endpoint, where metrics reporting is only taken from one side, whereas a different algorithm is used in health endpoints to get a combination of both reporters and thus won't miss any data.

src/components/Details/DetailedTrafficList.tsx
jshaughn: The "Unknown" node type is to deal with the special "unknown" node in the graph.  I'm not sure it's actually relevant to the detail pages as it's not a real app/service/workload.

jshaughn:
One comment about using the graph API.  Currently we only generate Cytoscape JSON and cytoscape is default "vendor".  But it is also possible to support a custom JSON by adding a "vendor".

src/components/Metrics/TrafficDetails.tsx
israel-hdez: @rubenvp8510 is working in the typing in #971. I hope his PR can go soon, because I also want to take advantage of those typings. So, let's put some pressure on him :)

(Although, I see he is not adding typing on `API.getNodeGraphElements` in his PR)

src/components/Details/DetailedTrafficList.tsx
israel-hdez: This is to, at least, show it as a source or a destination of traffic (see the "inbound" in the image of the main comment). I can remove this and that row won't be shown. I don't know if it is better to show or not. So... you say.

jshaughn: It should not be a destination, but I guess as a source it is valid.

israel-hdez:
@jshaughn @jotak Most comments resolved. It's only the typing comments which are not resolved. I'll wait for Ruben's PR.

israel-hdez:
@jshaughn 

> One comment about using the graph API. Currently we only generate Cytoscape JSON and cytoscape is default "vendor". But it is also possible to support a custom JSON by adding a "vendor".

I can do that, but I prefer to "transform" the data at client side. Indeed, I prefer to have a "generic" json rather than vendored ones and let the UI do the processing, instead of having multiple cases at server side. But I'll do what you see better for Kiali.

jshaughn:
> I can do that, but I prefer to "transform" the data at client side

I was just mentioning it as an option.  The Cytoscape JSON can also be treated as generic.  The idea of vendor is to support [graph] clients (i.e. cy, vizceral) that already have an expected JSON format.

israel-hdez:
@jshaughn @jotak I've updated the screenshot in the main comment.

israel-hdez:
It's also rebased to fix the conflict.

jotak:
I'm seeing some weird stuff but turning out it's not a problem here, same happens in the graph and actually this is in prometheus metrics: sometimes I can see requests telemetry for a given destination service that point to a different destination workload (a workload that relates to a different service). 

weird stuff

Due to this metric:
istio_requests_total{destination_app="**stadium**",destination_service="**ball.mesh-arena.svc.cluster.local**",destination_service_name="**ball**",destination_service_namespace="mesh-arena",source_app="ai-locals",source_workload="ai-locals-basic",source_workload_namespace="mesh-arena", (etc) }

So anyway, it's not a problem with the PR, but just wondering if some of you have already seen that and might have an explanation?

lucasponce:
@jotak also asked me to review the PR
One first question:
image

The two green columns in the right I'm not sure if user will understand that it represents.
I guess is a similar as the graph of errors, but it's not clear to me.
Perhaps it needs some label just to not miss it.

lucasponce:
Another issue.
Trying to navigate from one workload to another it seems to not work correctly:
image

I clicked in user-regions2 and it moved to productpage-v2.
I haven't tested yet if this is an issue from this PR or perhaps it's already in upstream.

src/components/Details/DetailedTrafficList.tsx
```diff
@@ -0,0 +1,261 @@
...
+      return <TableGrid.Col {...trafficColumnSizes}>N/A</TableGrid.Col>;
+    } else {
+      let rps: number;
+      let percentError: number;
+
+      if (traffic.protocol === 'http') {
+        rps = Number(traffic.rates.http);
+        percentError = traffic.rates.httpPercentErr ? Number(traffic.rates.httpPercentErr) : 0;
+      } else {
+        rps = Number(traffic.rates.grpc);
```
jotak: So, typescript here doesn't show an error due to potentially undefined `traffic.rates`, which is correct, it's able to infer type based on previous conditional branches and matches on `traffic.protocol` ... Nothing to add, I just wanted to say loud how awesome is typescript! :-)

jotak:
@israel-hdez thanks for the changes, I feel much better with the types!

Just found one bug however : when you navigate through the links (ex: click on a destination workload link) then move again the traffic tab, this is still the previous page that is displayed, not the one from the new focused workload.

lucasponce:
If we introduce explicit time:
image

It should also be aligned with the same control defined in the next tab for consistency:
image

Because, now user is aware he is pulling metrics.

jotak:
@lucasponce 
> It should also be aligned with the same control defined in the next tab for consistency:

The same would apply to health then, in info tab. Maybe we can leave that for later task?

lucasponce:
> > It should also be aligned with the same control defined in the next tab for consistency:

> The same would apply to health then, in info tab. Maybe we can leave that for later task?

Yes, if it's not minor it can be done in a next work, but I guess from user perspective it's more close to metrics tab rather than main info health (where I only see a summary).

src/helpers/TrafficDetailsHelper.tsx.ts
```diff
@@ -0,0 +1,27 @@
+import { authentication } from '../utils/Authentication';
```
lucasponce: src/helpers/TrafficDetailsHelper.tsx.ts ??
Why do we need this extra .tsx.ts ?

src/pages/WorkloadDetails/WorkloadDetailsPage.tsx
```diff
@@ -13,12 +13,16 @@ import { MetricsObjectTypes } from '../../types/Metrics';
...
+import { GraphType, NodeParamsType, NodeType } from '../../types/Graph';
+import { fetchTrafficDetails } from '../../helpers/TrafficDetailsHelper.tsx';
+import TrafficDetails from '../../components/Metrics/TrafficDetails';
 
 type WorkloadDetailsState = {
   workload: Workload;
   validations: Validations;
   istioEnabled: boolean;
   health?: WorkloadHealth;
+  trafficData: any;
```
lucasponce: I share with @jotak my concerns about using any to propagate types.
For something small or very specific it might be ok, but it's hard to read as it is now.
I see this is already discussed.

lucasponce:
Another related comment:
I see there is a subtask in https://issues.jboss.org/browse/KIALI-2042 to propagate this change into applications.

I miss also to plan to change it in the service details.
Basically the "Source Workloads" represents same concept here.

But, if we are goint to add the new "Traffic" tab, I wonder if it's better to just to think into the mini-graph at details level, just to avoid to have to many tabs.

This last point perhaps it's good to deal it separately but it's related with all of this effort.

jshaughn:
> I'm seeing some weird stuff but turning out it's not a problem here, same happens in the graph

@jotak , this is likely due to an Istio 1.1 telemetry bug that we have already reported to them: See https://github.com/istio/istio/issues/11151

src/helpers/TrafficDetailsHelper.tsx.ts
israel-hdez: Sorry... My mistake. I fixed the naming.

israel-hdez:
@mattmahoneyrh I added the "not enough traffic" legend.

israel-hdez:
@jotak @lucasponce I'm fixing the the navigation issue in a separate https://github.com/kiali/kiali-ui/pull/983 PR, because it exists in master.

israel-hdez:
@lucasponce What browser are you using? The "two green columns" are meant to represent failed vs successful requests. It's complimentary to the text, for example:

image

There should be a little grey marker. @jotak also asked. It's the health threshold. What I see weird, is that your browser is not rendering the marker, giving the impression of two green bars.

israel-hdez:
> > It should also be aligned with the same control defined in the next tab for consistency:

> The same would apply to health then, in info tab. Maybe we can leave that for later task?

Yes, I would like to leave it for a later task :) I'll create the Jira.
Perhaps, I also could add the tooltip for the green bar in another Jira? This PR is a little big and I don't want to add more stuff. Also, I think the tooltip will need to pass UX, which may imply to put on hold this PR.

israel-hdez:
For the interval/duration http://issues.jboss.org/browse/KIALI-2379

jotak:
> It's the health threshold. What I see weird, is that your browser is not rendering the marker, giving the impression of two green bars.

I think it's not obvious to understand what it is. Once I know it's the threshold I would be even tempted to move it to change the threshold, but it's not possible :)
Maybe it was making sense with health displayed aside, but as we removed the health icon, probably we should remove the health threshold as well, WDYT?

lucasponce:
> @lucasponce What browser are you using? The "two green columns" are meant to represent failed vs successful requests.

Just Chrome on Linux (Fedora 29). I'd say that I didn't get it in a first attempt. The health marker is not obvious to me, I guess, and probably others won't map it to that concept. A single graph with % sounds less info and more interesting. IMO

israel-hdez:
@jotak @lucasponce Ok, I will remove the threshold marker because I agree with you that the marker alone is not intuitive, but we lose that the user won't clearly know what workloads are in _warning_ vs _error_ status.

Remember, there is a design issue: https://github.com/kiali/kiali-design/issues/70. I'll pass this feedback there. They have already suggested that the threshold marker with a tool-tip may make things clearer. But I will check with them if it's worth to show the warning/error status or if it's ok to be omitted from this page. We could, also, do the opposite: restore back the health column and remove the bar (with a tooltip in the health icon to show details).

jotak:
Oh, thanks for the link, I didn't notice the parallel discussion.

israel-hdez:
@jotak @lucasponce All changes done. I rebased, so the fix of the navigation issue should also be applied.

lucasponce:
@israel-hdez thx, I'm fetching it locally to test it on my environment and I'll update the status.

lucasponce:
@israel-hdez I have a semi-blocker comment
image
I think graph info should be fetched only when tab is active, if not is adding additional overhead into the page and it has a "blink" effect that will increase when network is slow.

israel-hdez:
@lucasponce Your feedback is resolved :)

hhovsepy:
In a case of unknown egress traffic, the service "unknown" in Traffic tab's bottom should not be a link. 
In "Destination Services" tab it was shown as a not clickable text.
screenshot from 2019-02-14 07-37-29

As shown in graph:
screenshot from 2019-02-14 07-36-24

israel-hdez:
@hhovsepy I added some code to decide whether to show links, but I don't have the scenario. Could you try?... How can I setup your scenario?

hhovsepy:
My env is recreated and unknown service is gone.
The unknown link existed in graph as well. So you can merge this PR and fix in a scope of other PR, once I see this situation again.

lucasponce:
Ok, final local test and I'll merge in some minutes if everything is ok.

lucasponce:
Thanks @israel-hdez

PR #957, opened 2019-01-25
KIALI-2245 gRPC support

jshaughn:
- no longer use 100% default for %traffic, it doesn't play well with Find
- change edge to a single traffic protocol, multiple protocols doesn't make sense.
- add grpc fields to Find and Find Help
- fix aggregation issues in graph summary
- fix kiali-2290 app box issue
- add 'protocol' field to cy edge data
- add 'requestProtocol' option to metrics queries
- KIALI-2296 Only show %traffic label for edges with traffic

THIS PR REQUIRES SERVER PR https://github.com/kiali/kiali/pull/807


jshaughn:
@israel-hdez Thanks for offering a review.  I know that at the moment there is no change to the legend because I didn't want to change the legend and then find out we were making changes to the edges.  If you want to see GRPC traffic without adding any services, you can look at istio-system.   Please take at all of the various side panels to make sure they are still in good shape for http and tcp, and also incorporate grpc as needed.

src/pages/Graph/GraphHelpFind.tsx
```diff
@@ -322,9 +324,12 @@ export default class GraphHelpFind extends React.Component<GraphHelpFindProps> {
...
                       rows={[
-                        { id: 'ec00', c: 'http <op> <number>', n: 'unit: requests per second' },
-                        { id: 'ec10', c: '%err <op> <number>', n: 'range: [0..100]' },
-                        { id: 'ec20', c: '%traffic <op> <number>', n: 'range: [0..100]' },
+                        { id: 'ec00', c: 'grpc <op> <number>', n: 'unit: requests per second' },
+                        { id: 'ec10', c: '%grpcerr <op> <number>', n: 'range: [0..100]' },
+                        { id: 'ec20', c: '%grpctraffic <op> <number>', n: 'range: [0..100]' },
+                        { id: 'ec23', c: 'http <op> <number>', n: 'unit: requests per second' },
+                        { id: 'ec24', c: '%httperr <op> <number>', n: 'range: [0..100]' },
+                        { id: 'ec25', c: '%httptraffic <op> <number>', n: 'range: [0..100]' },
```
israel-hdez: In the "Examples" a change from `%traffic` -> `%httptraffic` is missing

src/pages/Graph/SummaryPanelGraph.tsx
```diff
@@ -77,15 +77,14 @@ export default class SummaryPanelGraph extends React.Component<SummaryPanelPropT
       return null;
     }
 
-    const baseQuery = cy
-      .nodes()
-      .filter('[!isGroup]')
-      .filter('[!isRoot]');
```
israel-hdez: `!isRoot` filter is removed... I remember this was to not count "unknown". But yes, there are other cases where we want to count roots...

src/pages/Graph/SummaryPanelGraph.tsx
```diff
@@ -77,15 +77,14 @@ export default class SummaryPanelGraph extends React.Component<SummaryPanelPropT
...
-    const numSvc = baseQuery.filter('[nodeType="service"]').size();
-    const numWorkloads = baseQuery.filter('[nodeType="workload"]').size();
-    const numApps = baseQuery.filter('[nodeType="app"]').size();
+    const baseQuery = cy.nodes().filter('[!isGroup]');
+    const numSvc = baseQuery.filter(`[nodeType = "${NodeType.SERVICE}"]`).size();
+    const numWorkloads = baseQuery.filter(`[nodeType = "${NodeType.WORKLOAD}"]`).size();
+    const numApps = baseQuery.filter(`[nodeType = "${NodeType.APP}"]`).size();
     const numEdges = cy.edges().size();
-    const trafficRate = getAccumulatedTrafficRate(cy.edges());
+    const nonServiceEdges = baseQuery.filter(`[nodeType != "${NodeType.SERVICE}"]`).edgesTo('*'); // don't count requests from injected service node
```
israel-hdez: +1
good bugfix when counting traffic.

src/types/Graph.ts
```diff
@@ -13,6 +13,13 @@ export interface SummaryData {
   summaryTarget: any;
 }
 
+export enum Protocol {
+  GRPC = 'grpc',
+  HTTP = 'http',
+  TCP = 'tcp'
+}
+export type ProtocolType = Protocol.GRPC | Protocol.HTTP | Protocol.TCP;
```
israel-hdez: Is `ProtocolType` really necessary?
I think you could use `Protocol` enum to type variables.

abonas:
What kind of visual changes does this PR provide? Can you please add a screenshot?
IIRC grpc on the graph is still in UX discussions in
https://github.com/kiali/kiali-design/issues/84

src/components/SummaryPanel/InOutRateChart.tsx
```diff
@@ -2,7 +2,86 @@ import * as React from 'react';
...
+              values: [0, 25, 50, 75, 100]
+            },
+            padding: {
+              top: 20,
+              bottom: 0
+            }
+          }
+        }}
+        data={{
+          groups: [['OK', '3xx', '4xx', '5xx']],
```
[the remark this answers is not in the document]
jshaughn: Yes, good catch.

src/pages/Graph/GraphHelpFind.tsx
jshaughn: Si, gracias.

src/pages/Graph/SummaryPanelGraph.tsx
jshaughn: Seeing your comment I looked at this code again and simplified it just a bit.  Each count is now just using a dedicated cy selector based on node type.  When dealing with app nodes I explicitly filter out the group nodes.  Please take a look, I think it is more clear this way.

src/pages/Graph/SummaryPanelGraph.tsx
jshaughn: Thanks,

src/pages/Graph/SummaryPanelNode.tsx
```diff
@@ -280,20 +317,41 @@ export default class SummaryPanelNode extends React.Component<SummaryPanelPropTy
               </Link>
             </p>
           )} */}
-          {this.hasHttpTraffic(node) ? this.renderHttpRates(node) : renderNoTraffic('HTTP')}
-          <div>{this.renderSparklines(node)}</div>
+          {this.hasGrpcTraffic(node) && this.renderGrpcRates(node)}
+          {this.hasHttpTraffic(node) && this.renderHttpRates(node)}
+          <div>{this.renderCharts(node)}</div>
+          {!this.hasGrpcTraffic(node) && renderNoTraffic('GRPC')}
+          {!this.hasHttpTraffic(node) && renderNoTraffic('HTTP')}
```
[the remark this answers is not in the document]
jshaughn: I agree, this doesn't look great for unknown. But the basic approach is to present as much telemetry data as possible and then follow it with messages about the protocols with no telemetry data.  I think this is still the best layout.  In this special case for unknown I added a horizontal rule to separate the two messages.  It looks a bit better to me.

src/store/Selectors/GraphData.ts
```diff
@@ -69,9 +74,9 @@ const decorateGraphData = (graphData: any) => {
       graphData.edges = graphData.edges.map(edge => {
         const decoratedEdge = { ...edge };
         if (edge.data.traffic) {
-          edge.data.traffic.map(protocol => {
-            decoratedEdge.data = { ...protocol.rates, ...decoratedEdge.data };
-          });
+          const traffic = edge.data.traffic;
+          edge.data.traffic = undefined;
```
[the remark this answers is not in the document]
jshaughn: I added comments to the code to better explain what is happening here.

src/types/Graph.ts
jshaughn: Oops.  +1, At first I had ProtocolType with the string literals and then I added the enum and basically forgot that I no longer needed the type.

jshaughn:
> What kind of visual changes does this PR provide?

@abonas You can see the UX issue for screenshots: [kiali/kiali-design#84](https://github.com/kiali/kiali-design/issues/84)

jshaughn:
@israel-hdez I have pushed changes for your review.  Please take a look, Thanks!

src/components/GraphFilter/__tests__/GraphFind.test.tsx
```diff
@@ -67,21 +67,33 @@ describe('Parse find value test', () => {
     // @ts-ignore
     expect(instance.parseFindValue('outside')).toEqual('node[isOutside]');
     // @ts-ignore
-    expect(instance.parseFindValue('unused')).toEqual('node[isUnused]');
+    expect(instance.parseFindValue('outside')).toEqual('node[isOutside]');
```
israel-hdez: Minor - I think you can remove this line. I cannot find any difference against the previous line.

src/components/CytoscapeGraph/graphs/GraphStyles.ts
```diff
@@ -103,7 +103,13 @@ export class GraphStyles {
           }
 
           if (rate > 0) {
-            content = pErr > 0 ? rate.toFixed(2) + ', ' + pErr.toFixed(1) + '%' : rate.toFixed(2);
+            if (pErr > 0) {
+              let sErr = pErr.toFixed(1);
+              sErr = `, ${sErr.endsWith('.0') ? pErr.toFixed(0) : sErr}%`;
+              content = rate.toFixed(2) + sErr;
```
israel-hdez: Minor -- I think code will look simpler doing scaling:
```js
let sErr = Math.round(pErr * 10) / 10;
content = `${rate.toFixed(2)}, ${sErr.toString()}%`;
```
BUT... I'm not sure if this could fall in [floating point precision problems](https://stackoverflow.com/questions/588004/is-floating-point-math-broken) and display lots of digits. Probably, your version is safer for the UI, but scaling is easier to read for devs.

So, it's up to you.

src/components/CytoscapeGraph/graphs/GraphStyles.ts
```diff
@@ -115,20 +121,26 @@ export class GraphStyles {
...
           if (ele.data(CyEdge.httpPercentReq) > 0) {
             pReq = Number(ele.data(CyEdge.httpPercentReq));
           } else if (ele.data(CyEdge.grpcPercentReq) > 0) {
             pReq = Number(ele.data(CyEdge.grpcPercentReq));
           }
-
-          content = pReq < 100.0 ? pReq.toFixed(1) + '%' : '100%';
+          if (pReq > 0) {
+            const sReq = pReq.toFixed(1);
+            content = `${sReq.endsWith('.0') ? pReq.toFixed(0) : sReq}%`;
```
israel-hdez: Similar comment here (for scaling)

src/pages/Graph/SummaryPanelNode.tsx
```diff
@@ -360,10 +453,17 @@ export default class SummaryPanelNode extends React.Component<SummaryPanelPropTy
...
+            dataErrors={this.state.httpErrorCountOut}
             hide={isServiceNode}
           />
+          {this.isIstioOutgoingCornerCase(node) && (
+            <>
+              <div>
+                <Icon type="pf" name="info" /> Traffic to istio-system not included. Use edge for details.
+              </div>
+            </>
+          )}
```
israel-hdez: Yikes! I see that the edge also doesn't have sparklines. I guess the edge also collects data from source telemetry.
But this message is OK. The edge, at least, will show in the rates-table the specific details.
This is just a comment :)

src/components/GraphFilter/__tests__/GraphFind.test.tsx
jshaughn: Good find, actually the second line is supposed to test 'outsider', it was missing the 'r',

src/components/CytoscapeGraph/graphs/GraphStyles.ts
jshaughn: It's working, I'm OK as is.

src/components/CytoscapeGraph/graphs/GraphStyles.ts
jshaughn: I'm still OK :)

src/pages/Graph/SummaryPanelNode.tsx
jshaughn: We can chase istio-system corner cases for a long time, instead let's stop and wait for any negative feedback. Good testing.

jshaughn:
@mattmahoneyrh I have pushed changes to the PR.  I apologize that this is sort of becoming an uber-PR.  The Graph Find issue you pointed out above is not really related to this PR but was a known issue (at least to me).  As you said, it was technically correct but not particularly useful.  I made a relatively big change and I think it's now fixed.  To solve the issue I changed default values (from '0' to 'NaN') for our numeric fields in the graph elements.  Please re-check the side panels to make sure they look good and we still see the expected charts and sparklines, and we don't see any NaN values creeping in.

Additionally, please re-check the edge sparkline in the "istio-system destination node" case we talked about.  I think I fixed that issue as well, and now you should see a sparkline chart (as opposed to the "Not enough data..." message).

Finally, I made a couple of additions to the Find/Hide DSL.  You can now identify edges via protocol.  For example, "protocol = grpc".   Also, you can test for no activity by testing for NaN.  For example: "httpout = NaN".  (or any activity with != NaN).

mattmahoneyrh:
The "Not enough traffic..." msg is still seen:

image

jshaughn:
> The "Not enough traffic..." msg is still seen:

OK, after meeting with @mattmahoneyrh we figured out what was happening here.  The edge sparkline issue was resolved but traffic to istio services doesn't happen on the same schedule as regular app traffic.  There can be lulls such that for the "Last Minute" there really isn't any traffic between the envoy proxy and the istio service, even if there is plenty of app traffic being processed.  By expanding to "Last 5 minutes" or greater we saw the sparklines, and they showed the dips in traffic that resulted in the lack of a sparkline for the short 1 minute duration.

PR #959, opened 2019-01-25
KIALI-2069 Persist selected namespace

josejulio:
** Describe the change **

Saves the selected namespaces in the localStorage.

** Screenshot **

peek 2019-01-24 23-55

### DIFF
--- a/src/store/ConfigStore.ts
+++ b/src/store/ConfigStore.ts
@@ -1,6 +1,9 @@
 import { createStore, applyMiddleware, compose } from 'redux';
 import { KialiAppState } from './Store';
 import { persistStore, persistReducer } from 'redux-persist';
+import { persistFilter } from 'redux-persist-transform-filter';
+import { createTransform } from 'redux-persist';
+
 import rootReducer from '../reducers';
 import thunk from 'redux-thunk';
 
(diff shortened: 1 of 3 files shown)


src/store/ConfigStore.ts
```diff
@@ -20,10 +23,25 @@ declare const window;
 const webRoot = (window as any).WEB_ROOT ? (window as any).WEB_ROOT : undefined;
 const persistKey = 'kiali-' + (webRoot && webRoot !== '/' ? webRoot.substring(1) : 'root');
 
+// Needed to be able to whitelist fields but allowing to keep an initialState
+const whitelistInputWithInitialState = (reducerName: string, inboundPaths: string[], initialState: any) =>
+  createTransform(
```
mtho11: I have not heard of this `createTransform` before. I would have just used `localStorage` Api. But, this works. +1

mtho11:
@israel-hdez you can add this to the demo if you want.

src/store/ConfigStore.ts
josejulio: I was going to do that, started to read how to tackle the problem that way. But we were already persisting part of the state to the storage with `redux-persist` so i just extended it and added the ability to partially persist sub-elements. `redux-persist-transform-filter` helped a lot in this :)

PR #961, opened 2019-01-25
KIALI-2300 KIALI-2304 Links from Overview page are not filtering by namespace

aljesusg:
[KIALI-2304](https://issues.jboss.org/browse/KIALI-2304)
[KIALI-2300](https://issues.jboss.org/browse/KIALI-2300)

- [X] Moved paths to configuration to avoid hardcoding.
- [X] Refactor Code and removed ListPageLink


hhovsepy:
Verified the solution.

aljesusg:
Thanks @hhovsepy  !!!

src/components/Filters/NamespaceFilter.ts
```diff
@@ -3,7 +3,7 @@ import { authentication } from '../../utils/Authentication';
 import { FILTER_ACTION_APPEND, FilterType } from '../../types/Filters';
 
 export class NamespaceFilter {
-  static id = 'namespace';
+  static id = 'namespaces';
```
jotak: @aljesusg I think you can remove the whole class & file. It is legacy code now, as far as i can tell it's now only used in `ListPageLink` just to get the `id` and `category` constants, so it doesn't worth it keeping as separate thing.

aljesusg: it is using in the istioconfig list breadcumb too
Let me check the code

src/pages/Graph/SummaryPanelGraph.tsx
```diff
@@ -178,15 +179,15 @@ export default class SummaryPanelGraph extends React.Component<SummaryPanelPropT
 
   private renderTopologySummary = (numSvc: number, numWorkloads: number, numApps: number, numEdges: number) => (
     <div>
-      <ListPageLink key="appsLink" target={TargetPage.APPLICATIONS} namespaces={this.props.namespaces}>
+      <Link key="appsLink" to={`/${Paths.APPLICATIONS}?namespaces=${this.props.namespaces}`}>
         {' applications'}
-      </ListPageLink>
-      <ListPageLink key="servicesLink" target={TargetPage.SERVICES} namespaces={this.props.namespaces}>
+      </Link>
+      <Link key="servicesLink" to={`/${Paths.SERVICES}?namespaces=${this.props.namespaces}`}>
```
jotak: This produces a wrong URL: `http://localhost:3000/console/services?namespaces=[object%20Object]`

I guess you have to map namespaces: `namespace => namespace.name`

aljesusg: mmm I see... it seems is namespaces[0].name like https://github.com/kiali/kiali-ui/pull/961/files/29cd51f06235b0d0df9314af91b80083cbb0000b#diff-3fa7d035edf629a0acba1df605940428R135

```js
const promiseHTTP = API.getNamespaceMetrics(authentication(), props.namespaces[0].name, options);
```

jotak: I don't see any reason to filter only for namespace[0], when several namespaces might be selected and if i'm correct the URL is able accept a list of namespace.
IMO that `getNamespaceMetrics` call is buggy and we shouldn't propagate the bug elsewhere.

aljesusg: you're right I'll change it thanks

jotak: Just searched a little bit, actually the code for side panel metrics is ok because it's fetching these metrics only when there's a single namespace selected.

But for building the links, you may have several namespaces selected.
For example, here's the link URL I have with current master code and multiple namespaces selected:
`http://localhost:3000/console/services?namespace=mesh-arena&namespace=istio-system`

aljesusg:
Changed @jotak Thanks

PR #962, opened 2019-01-25
KIALI-1977 show runtime names in workload details

jotak:
JIRA: https://issues.jboss.org/browse/KIALI-1977

details


abonas:
Can you show an example how does this look if there are several runtimes?

jotak:
Sure, there is nothing fancy, it's coma-separated values:

multiple

PR #963, opened 2019-01-25
KIALI-2148 Add padlock next to Kiali logo when mTLS is gloablly enabled

xeviknal:
** Describe the change **

Add a padloc icon in the masthead when mTLS is enabled globally.
Depends on: https://github.com/kiali/kiali/pull/803

** Issue reference **

https://issues.jboss.org/browse/KIALI-2148

** Backwards compatible? **
yes

** Screenshot **

We have 3 scenarios:

1. Both Destination Rule and MeshPolicy are present and well-configured (mtlswise):
screenshot of kiali console 27

2. One config (Destination Rule or MeshPolicy) is not present nor well-configured (mtlswise):
screenshot of kiali console 28

3. Both DR and MeshPolicy are not present or well-configured (mtlswise)
screenshot of kiali console 30


src/components/Nav/GlobalMTLSStatus.tsx
```diff
@@ -0,0 +1,33 @@
...
+  components: Component[];
+  warningMessages: string[];
+};
+
+const iconStyle = style({
+  position: 'absolute',
+  left: 190,
+  top: 23,
+  color: PfColors.Green
+});
```
abonas: Can this styling be moved to CSS? why is it here?

hhovsepy:
screenshot from 2019-01-28 11-13-07

src/components/Nav/GlobalMTLSStatus.tsx
pilhuhn: About 1): for this entire Istio-Mesh controlled by the Istio control plane we are talking to. And yes, this will include all projects including future ones, that are addressed by this control plane. Currently this means all projects that use Istio, but in the future with soft-multitenancy, there could be more than one Istio installation and each one could treat that differently.
1b: one also would have a matching global DR, that has Istio-mTLS enabled.
If we want to be precise here, we can show a) no lock for no mTLS b) a broken lock if one of { MeshPolicy, global mTLS-DR } is missing, a lock if all is correct.
2) I agree, that should move to the right side of the mast-head

xeviknal:
@abonas I did create an issue into kiali-design: https://github.com/kiali/kiali-design/issues/88
I also answered your questions there as suggested.

lucasponce:
One note.

Some of this questions:

> what does "globally" mean? In all openshift projects? current and future ones?
From reading the Istio docs it sounds like there 2 parts to this functionality:
a. MeshPolicy named default. but this is only half of the work "At this point, only the receiving side is configured to use mutual TLS"
so this is also needed:
b. To configure the client side, you need to set destination rules to use mutual TLS.
So, is the lock displayed "if and only if" both a and b are configured? or one of them is enough for the icon to appear? (I didn't check the server code, it might contain the answer...)

Should be better addressed in the JIRA instead in the PR.

PR should be blocked for specific issues related PR itself, if there is some aspect to clarify in the JIRA that's a different.

From my point of view the PR resolves the JIRA that was created.

I'm +1 with the comment to place the icon at the right, but I once that is fixed and if the feature is implemented as defined by @pilhuhn I would go in with the PR and address future changes in a following PR.

xeviknal:
Moving the lock to the right side:
screenshot of kiali console 23

src/components/Nav/GlobalMTLSStatus.tsx
abonas: > About 1): for this entire Istio-Mesh controlled by the Istio control plane we are talking to. And yes, this will include all projects including future ones, that are addressed by this control plane. Currently this means all projects that use Istio, but in the future with soft-multitenancy, there could be more than one Istio installation and each one could treat that differently.
> 1b: one also would have a matching global DR, that has Istio-mTLS enabled.
> If we want to be precise here, we can show a) no lock for no mTLS b) a broken lock if one of { MeshPolicy, global mTLS-DR } is missing, a lock if all is correct.
> 2) I agree, that should move to the right side of the mast-head

yes, making this more precise would be good.
@lucasponce mentioned that those details should be clarified in the JIRA. I have no objections to that. Butn in order to decide if JIRA is resolved by a PR, the functionality should be detailed and clear.
Otherwise I find it hard to review and it also will be hard for QE to test.

xeviknal:
We have 3 scenarios:

1. Both Destination Rule and MeshPolicy are present and well-configured (mtlswise):
screenshot of kiali console 27

2. One config (Destination Rule or MeshPolicy) is not present nor well-configured (mtlswise):
screenshot of kiali console 28

3. Both DR and MeshPolicy are not present or well-configured (mtlswise)
screenshot of kiali console 30

xeviknal:
@abonas @pilhuhn do you mind reviewing this PR feature-wise?

abonas:
@xeviknal @pilhuhn the icons should be shown to UX team. They can advise if those are good or suggest other ones. Please open an issue in the design repo and either link to screenshots here or post them there. 
The orange one is a bit strange to me. orange is typically a warning, I'm not sure that partially enabled is always a warning. But again - this should be reviewed by UX.

lucasponce:
@abonas 

> the icons should be shown to UX team. They can advise if those are good or suggest other ones. 

Yes, but in parallel to this PR, if the feature proposed and worked by Xavi works, let's do both without blocking the progress.

So, let's continue with this and we will change in a following PR, but better to consolidate.

cc @theute @pilhuhn

lucasponce:
Adding DNM label due in the backend PR we have found a potential blocker scenario that we need to resolve first.

lucasponce:
Merging, but as discussed in https://github.com/kiali/kiali-design/issues/88 icons and style will be updated once we have UX feedback.

PR #970, opened 2019-01-30
KIALI-2247 Reduxify graph-find and graph-hide values

jshaughn:
- maintain find/hide when navigating away and then back to the graph
- note: this intentionally is not in the URL and does not survive an F5

### DIFF
--- a/src/components/GraphFilter/GraphFind.tsx
+++ b/src/components/GraphFilter/GraphFind.tsx
@@ -5,6 +5,8 @@ import { ThunkDispatch } from 'redux-thunk';
 import { bindActionCreators } from 'redux';
 
 import { KialiAppState } from '../../store/Store';
+import { findValueSelector, hideValueSelector } from '../../store/Selectors';
+
 import { GraphFilterActions } from '../../actions/GraphFilterActions';
 
 import { KialiAppAction } from '../../actions/KialiAppAction';
@@ -15,8 +17,12 @@ import { CyData } from '../../types/Graph';
 
 type ReduxProps = {
   cyData: CyData;
+  findValue: string;
+  hideValue: string;
   showFindHelp: boolean;
 
+  setFindValue: (val: string) => void;
+  setHideValue: (val: string) => void;
   toggleFindHelp: () => void;
 };
 
@@ -32,13 +38,11 @@ export class GraphFind extends React.PureComponent<GraphFindProps> {
     router: () => null
   };
 
-  private hideInputRef;
-  private hideInputValue: string;
-  private hideValue: string;
-  private hiddenElements: any | undefined;
   private findInputRef;
   private findInputValue: string;
-  private findValue: string;
+  private hiddenElements: any | undefined;
+  private hideInputRef;
+  private hideInputValue: string;
 
   constructor(props: GraphFindProps) {
     super(props);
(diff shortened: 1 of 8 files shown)


src/components/GraphFilter/GraphFind.tsx
```diff
@@ -47,21 +51,28 @@ export class GraphFind extends React.PureComponent<GraphFindProps> {
...
 
   componentDidUpdate(prevProps: GraphFindProps) {
-    if (this.findValue.length > 0 && this.props.cyData.updateTimestamp !== prevProps.cyData.updateTimestamp) {
+    // make sure the input box reflects the redux value
+    this.findInputRef.value = this.props.findValue;
+    this.hideInputRef.value = this.props.hideValue;
+
+    const findChanged = this.props.findValue !== prevProps.findValue;
+    const hideChanged = this.props.hideValue !== prevProps.hideValue;
+    const graphChanged = prevProps.cyData && this.props.cyData.updateTimestamp !== prevProps.cyData.updateTimestamp;
```
josejulio: I'm curious why you tested from `prevProps.cyData` but not for `this.props.cyData`
Are we sure that we will never hit an update with cyData = null?

jshaughn: Because this is in ComponentDidUpdate, which is called after the first render, I think it should exist.  prevProps doesn't exist until the second time.  If you think it's unsafe I can add a check.

src/components/GraphFilter/GraphFind.tsx
```diff
@@ -47,21 +51,28 @@ export class GraphFind extends React.PureComponent<GraphFindProps> {
...
-    this.hideValue = '';
-    this.hiddenElements = undefined;
-
     this.findInputRef = React.createRef();
     this.findInputValue = '';
-    this.findValue = '';
+
+    this.hiddenElements = undefined;
+
+    this.hideInputRef = React.createRef();
```
josejulio: I see in the code that you are doing:

```
inputRef={ref => {
                  this.findInputRef = ref;
                }}
```

It doesn't make sense to use `React.createRef()`, this will cause confusion because react ref use `theref.current` and this doesn't.

The same applies for the `findInputRef`

src/components/GraphFilter/GraphFind.tsx
```diff
@@ -47,21 +51,28 @@ export class GraphFind extends React.PureComponent<GraphFindProps> {
       props.toggleFindHelp();
     }
 
-    this.hideInputRef = React.createRef();
-    this.hideInputValue = '';
-    this.hideValue = '';
-    this.hiddenElements = undefined;
-
     this.findInputRef = React.createRef();
     this.findInputValue = '';
```
josejulio: Do we really need this? This value should be taken from the prop value `findValue`, the same for the `hideValue` part.

josejulio: Also, i think you should use the prop (`findValue`) to initialize the `FormControl`
See `value` field 
```tsx
<FormControl
                type="text"
                value={this.props.findValue}
                style={{ width: '18em' }}
                inputRef={ref => {
                  this.findInputRef = ref;
                }}
                onChange={this.updateFind}
                onKeyPress={this.checkSubmitFind}
                placeholder="Find..."
              />
```

src/components/GraphFilter/GraphFind.tsx
josejulio: `componentDidUpdate()` is called on the first re-render, prevProps should already exist there. I'm not sure if `cyData` is always present, i asked out of curiosity to see if you knew about it.

jshaughn: I will add the check.

src/components/GraphFilter/GraphFind.tsx
jshaughn: OK, I removed the two React.createRef lines and it does seem to work.

src/components/GraphFilter/GraphFind.tsx
jshaughn: > Do we really need this?

Are you asking whether we need this.findInputValue?   I think we do.  It is the current value being displayed in the input.  Without it the value typed by the user is not displayed, 
 Also, we don't want to update redux on every character typed.  This.props.findValue is different, it holds the last "submitted" value.

As for the line that actually initializes the variable to '', I don't think we actually need that line.

And I don't think the FormControl.value attribute works quite the way your expect.  Trying to use it causes some issues as it's not really treated just like an initial value.  It is better to set the ref value.

josejulio: Oh ok, so `this.findInputValue` is the current value of the input, not the value in effect.

josejulio: We still need for a way to initialize the  findInput based on the received prop.

I see that `componentDidUpdate` takes care of updating the value, but is not called in first render.

> componentDidUpdate() is invoked immediately after updating occurs. **This method is not called for the initial render**.

https://reactjs.org/docs/react-component.html#componentdidupdate

jshaughn: OK, I pushed a change to add ComponentDidMount logic.  This is called after the first render and therefore the ref is set and can be manipulated.  Even if we found a way to use FormControl.value or FormControl.defaultValue to set the initial value, it would not be sufficient.  We would still need to perform the "submit" for the initial value, if set in redux at mount time.  What do you think?  I'm open to alternatives.

PR #971, opened 2019-01-30
KIALI-585 Add Type Interface for Service Graph json

rubenvp8510:
Defined the interface contract. The schema is pretty much the Cytoscape elements schema for nodes/edges.

This PR address  https://issues.jboss.org/browse/KIALI-585

### DIFF
--- a/src/actions/GraphDataActions.ts
+++ b/src/actions/GraphDataActions.ts
@@ -1,4 +1,6 @@
 import { ActionType, createAction } from 'typesafe-actions';
+import { GraphElements } from '../types/Graph';
+import { DurationInSeconds, TimeInSeconds } from '../types/Common';
 
 enum GraphDataActionKeys {
   GET_GRAPH_DATA_START = 'GET_GRAPH_DATA_START',
@@ -12,7 +14,7 @@ export const GraphDataActions = {
   getGraphDataStart: createAction(GraphDataActionKeys.GET_GRAPH_DATA_START),
   getGraphDataSuccess: createAction(
     GraphDataActionKeys.GET_GRAPH_DATA_SUCCESS,
-    resolve => (timestamp: number, graphDuration: number, graphData: any) =>
+    resolve => (timestamp: TimeInSeconds, graphDuration: DurationInSeconds, graphData: GraphElements) =>
       resolve({
         timestamp: timestamp,
         graphDuration: graphDuration,
--- a/src/store/Selectors/GraphData.ts
+++ b/src/store/Selectors/GraphData.ts
@@ -3,11 +3,21 @@
 // Put default values for all fields that are omitted.
 import { KialiAppState } from '../Store';
 import { createSelector } from 'reselect';
+import {
+  DecoratedGraphEdgeData,
+  DecoratedGraphEdgeWrapper,
+  DecoratedGraphElements,
+  DecoratedGraphNodeData,
+  DecoratedGraphNodeWrapper,
+  GraphEdgeWrapper,
+  GraphElements,
+  GraphNodeWrapper
+} from '../../types/Graph';
 
 // When updating the cytoscape graph, the element data expects to have all the changes
 // non provided values are taken as "this didn't change", similar as setState does.
 // Put default values for all fields that are omitted.
-const decorateGraphData = (graphData: any) => {
+export const decorateGraphData = (graphData: GraphElements): DecoratedGraphElements => {
   const elementsDefaults = {
     edges: {
       grpc: 'NaN',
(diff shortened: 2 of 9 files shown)


src/actions/GraphDataActions.ts
```diff
@@ -12,7 +13,7 @@ export const GraphDataActions = {
   getGraphDataStart: createAction(GraphDataActionKeys.GET_GRAPH_DATA_START),
   getGraphDataSuccess: createAction(
     GraphDataActionKeys.GET_GRAPH_DATA_SUCCESS,
-    resolve => (timestamp: number, graphDuration: number, graphData: any) =>
+    resolve => (timestamp: number, graphDuration: number, graphData: GraphElements) =>
```
mtho11: Can we also change `timestamp: TimeInMilliseconds` and `duration: DurationInSeconds`

src/store/Selectors/GraphData.ts
```diff
@@ -3,11 +3,12 @@
 // Put default values for all fields that are omitted.
 import { KialiAppState } from '../Store';
 import { createSelector } from 'reselect';
+import { DecoratedGraphElements, GraphEdgeWrapper, GraphNodeWrapper } from '../../types/Graph';
 
 // When updating the cytoscape graph, the element data expects to have all the changes
 // non provided values are taken as "this didn't change", similar as setState does.
 // Put default values for all fields that are omitted.
-const decorateGraphData = (graphData: any) => {
+export const decorateGraphData = (graphData: any): DecoratedGraphElements => {
```
mtho11: What about the `graphData` type? Probably a subset of `DecoratedGraphElements`

src/types/Graph.ts
```diff
@@ -76,3 +76,94 @@ export interface CytoscapeBaseEvent {
 export interface CytoscapeClickEvent extends CytoscapeBaseEvent {}
 export interface CytoscapeMouseInEvent extends CytoscapeBaseEvent {}
 export interface CytoscapeMouseOutEvent extends CytoscapeBaseEvent {}
+
```
mtho11: This part is great!

mtho11:
@rubenvp8510 you should use the PR template with stuff like links back to the Jira. Otherwise, great!

tslint.json
```diff
@@ -64,8 +64,7 @@
       "check-module",
       "check-operator",
       "check-separator",
-      "check-type",
-      "check-typecast"
```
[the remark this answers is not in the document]
jotak: I would prefer to ignore prettier occasionally rather than removing tslint rules. Tslint is more important than prettier IMO.
To ignore prettier, add `// prettier-ignore` before the line to ignore (with eventually a comment explaining why)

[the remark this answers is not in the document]
mtho11: Let's leave as-is for now. Besides, that should be in its own PR. We can revisit this later.

src/actions/GraphDataActions.ts
```diff
@@ -12,7 +14,7 @@ export const GraphDataActions = {
   getGraphDataStart: createAction(GraphDataActionKeys.GET_GRAPH_DATA_START),
   getGraphDataSuccess: createAction(
     GraphDataActionKeys.GET_GRAPH_DATA_SUCCESS,
-    resolve => (timestamp: number, graphDuration: number, graphData: any) =>
+    resolve => (timestamp: TimeInMilliseconds, graphDuration: DurationInSeconds, graphData: GraphElements) =>
```
jshaughn: I think the timestamp is TimeInSeconds, please verify.

src/store/Selectors/GraphData.ts
```diff
@@ -71,24 +80,26 @@ const decorateGraphData = (graphData: any) => {
...
+      decoratedGraph.edges = graphData.edges.map((edge: GraphEdgeWrapper) => {
         const decoratedEdge = { ...edge };
+        const { traffic, ...edgeData } = edge.data;
         // see comment above about the 'traffic' data handling
-        if (edge.data.traffic) {
-          const traffic = edge.data.traffic;
+        if (traffic) {
           edge.data.traffic = undefined;
-          decoratedEdge.data = { protocol: traffic.protocol, ...traffic.rates, ...decoratedEdge.data };
+          decoratedEdge.data = { ...traffic.rates, ...edgeData };
```
jshaughn: Am I reading this wrong or did you stop setting the protocol field?

src/types/Graph.ts
```diff
@@ -82,3 +82,94 @@ export interface CytoscapeBaseEvent {
...
+  id: string;
+  parent?: string;
+  nodeType: NodeType;
+  namespace: string;
+  workload?: string;
+  app?: string;
+  version?: string;
+  service?: string;
+  destServices?: any;
+  traffic?: ProtocolTraffic[];
```
jshaughn: As far as I know the traffic field is not needed, at least not at the moment.  The approach is that the traffic structure in the JSON is parsed out into the top level fields that you have moved into DecoratedGraphNodeData.

src/types/Graph.ts
```diff
@@ -82,3 +82,94 @@ export interface CytoscapeBaseEvent {
...
+  isRoot?: boolean;
+  isServiceEntry?: string;
+  isUnused?: boolean;
+}
+
+export interface GraphEdgeData {
+  id: string;
+  source: string;
+  target: string;
+  traffic?: ProtocolTraffic;
```
jshaughn: Same comment as above but also not that each edge can have at most one protocol field and it is expected to be set. It seems you removed it but if you look at the UI code you will see it is used in several places.

src/store/Selectors/GraphData.ts
rubenvp8510: Ups... my mistake! I'll fix it.

src/actions/GraphDataActions.ts
rubenvp8510: You're right!!!

https://github.com/kiali/kiali/blob/master/graph/cytoscape/cytoscape.go#L149

src/types/Graph.ts
rubenvp8510: Yes, but this structure represents the "undecorated" node data, so I understand that this is not the way that the structure is used on the code because it uses the decorated version.

src/actions/GraphDataActions.ts
rubenvp8510: Changed!

src/actions/GraphDataActions.ts
rubenvp8510: I changed to TimeInSeconds!

src/store/Selectors/GraphData.ts
rubenvp8510: Fixed!

src/store/Selectors/GraphData.ts
```diff
@@ -55,9 +63,10 @@ const decorateGraphData = (graphData: any) => {
       workload: undefined
     }
   };
+  const decoratedGraph: DecoratedGraphElements = {};
   if (graphData) {
     if (graphData.nodes) {
```
jshaughn: A couple of comments.  I think you need to cast `decoratedNode.data` and then remove the comment line. And we don't want to have a `traffic` field in the decorated node data.  I think this whole if-clause should look like:

```
if (graphData.nodes) {
      decoratedGraph.nodes = graphData.nodes.map((node: GraphNodeWrapper) => {
        const decoratedNode = { ...node };
        // parse out the traffic data into top level fields for the various protocols. This is done
        // to be back compatible with our existing ui code that expects the explicit http and tcp fields.
        // We can then set the 'traffic' field undefined because it is unused in the cy element handling.
        // TODO: refactor the code to use the traffic structure.
        if (decoratedNode.data.traffic) {
          const traffic = decoratedNode.data.traffic;
          decoratedNode.data.traffic = undefined;
          traffic.map(protocol => {
            decoratedNode.data = { ...protocol.rates, ...decoratedNode.data };
          });
        }
        decoratedNode.data = <DecoratedGraphNodeData> { ...elementsDefaults.nodes, ...decoratedNode.data };
        return <DecoratedGraphNodeWrapper> decoratedNode;
      });
    }
```

rubenvp8510:
@jshaughn  @mtho11 I did all requested changes, not sure why "security/snyk" is failing though

mtho11:
That snyk error is strange, it won't even tell us what the issue is.

josejulio:
> That snyk error is strange, it won't even tell us what the issue is.

I have seen that error before, i think it goes away by pushing new code.

@rubenvp8510 could you try to push a small change reverting it back and pushing again? Maybe that would kick another snyk build

rubenvp8510:
> > That snyk error is strange, it won't even tell us what the issue is.
> 
> I have seen that error before, i think it goes away by pushing new code.
> 
> @rubenvp8510 could you try to push a small change reverting it back and pushing again? Maybe that would kick another snyk build

Sure, It worth a try

rubenvp8510:
It works! All in green now

mtho11:
I talked with @jshaughn and he is ok with changes. Merging while green.

PR #972, opened 2019-01-31
KIALI-2322 Click on breadcrumb Istio Object Type - filters doubled

aljesusg:
[KIALI-2322](https://issues.jboss.org/browse/KIALI-2322)

Use Cases 

1. The user go to IstioConfigDetails from IstioList    

- In this case the user set a istiotype = virtualService and when the user click the breadcumb this add this istiotype again in activeFilters. It seems that there is some place where is setting twice. i did this control but still there are 2 filters.

```typescript
updateTypeFilter = () => {
    // When updateTypeFilter is called, selected filters are already updated with namespace. Just push additional type obj
    const activeFilters: ActiveFilter[] = FilterSelected.getSelected();

    const filter: ActiveFilter = {
      category: 'Istio Type',
      value: dicIstioType[this.props.match.params.objectType]
    };
    if (!activeFilters.includes(filter)) {
      activeFilters.push(filter);
    }
    FilterSelected.setSelected(activeFilters);
  };
```

2. The user go to istioConfigDetails from other view how destiantionRules tab of Servicedetails

- The breadcumb must add this filter.

So in the meanwhile I set this in the method setSelected

```typescript
export const setSelected = (activeFilters: ActiveFilter[]) => {
    selectedFilters = _.uniqWith(activeFilters, _.isEqual);
  };
```

With this we avoid repeat the elements in the activeFilters.

demo

I am looking where is the problem...but I suspect in lifecycle events...

cc @lucasponce @hhovsepy


aljesusg:
It could b here

```typescript

 constructor(props: StatefulFiltersProps) {
    super(props);

    let active = FilterSelected.getSelected();
    if (!FilterSelected.isInitialized()) {
      active = ListPagesHelper.getFiltersFromURL(this.props.initialFilters);
      FilterSelected.setSelected(active);
    } else if (!ListPagesHelper.filtersMatchURL(this.props.initialFilters, active)) {
      active = ListPagesHelper.setFiltersToURL(this.props.initialFilters, active);
      FilterSelected.setSelected(active);
    }

    this.state = {
      currentFilterType: this.props.initialFilters[0],
      filterTypes: this.props.initialFilters,
      activeFilters: active,
      currentValue: ''
    };
  }
```

aljesusg:
What do you think about create a component for the breadcrumb @lucasponce @jotak ? This code is working check it please

hhovsepy:
Tested:
KIALI-2322 is fixed.
KIALI-2314 still exist:
Go to Istio config.
Reset namespace.
Click on any istio config.
Click on breadcrumb Istio Config Type.
The namespace is selected.

If initially there were not namespaces selected, it should not select while clicking on breadcrumb.
If there were already selected namespaces, then clicking should keep the initial namespaces.

aljesusg:
> Tested:
> KIALI-2322 is fixed.
> KIALI-2314 still exist:
> Go to Istio config.
> Reset namespace.
> Click on any istio config.
> Click on breadcrumb Istio Config Type.
> The namespace is selected.
> 
> If initially there were not namespaces selected, it should not select while clicking on breadcrumb.
> If there were already selected namespaces, then clicking should keep the initial namespaces.

If you have

Istio Config > Namespace: bookinfo > Istio Object Type: destinationrules > Istio Object: /reviews 

I expect:

1. Click Istio Config  => Go to Istio list without filters
2. Click Namespace: bookinfo => Go to Istio list with namespace filter
3.  Istio Object Type: destinationrules => Go to Istio list with namespace filter and type filter


The use case 3 if I have bookinfo and default ns selected I'll see only bookinfo selected after click, I am cleaning all ns and selecting only the ns of the Istio object. We want filter only by types in all the ns?? 

@lucasponce , @jotak , @xeviknal ?

jotak:
For KIALI-2314 , this is still not clear what we want to do, as commented in the JIRA ( https://issues.jboss.org/browse/KIALI-2314 ). I don't know what's better... someone has to decide :)

jotak:
Actually, I think that when we introduced that global namespace selector, the conceptual change it implies are greater than we thought. Making it global puts it out of the "navigation workflow". It's like namespaces shouldn't be in the breadcrumb anymore, because breadcrumb is for hierarchical drilldown to resources, but with the global multi-namespaces this isn't hierarchical anymore. Not sure if you follow me... But I would be for removing namespace from breadcrumb.

aljesusg:
> Actually, I think that when we introduced that global namespace selector, the conceptual change it implies are greater than we thought. Making it global puts it out of the "navigation workflow". It's like namespaces shouldn't be in the breadcrumb anymore, because breadcrumb is for hierarchical drilldown to resources, but with the global multi-namespaces this isn't hierarchical anymore. Not sure if you follow me... But I would be for removing namespace from breadcrumb.

Agree with this, so in that case  should we clean the namespaces if the user select the type ? ^^ We had the same problem ^^

jotak:
@aljesusg I think no, in that case the breadcrumb becomes completely uncorrelated with the namespaces, so they wouldn't be modified on click.
But I'd like to have other's opinion before to go on.

lucasponce:
I'm testing locally and the behaviour looks good to me.
If I select two namespaces:
image

And I go to the detail:
I have the "Istio Config" > "Namespace: istio-system" > "..."

To me click on "Istio Config" and maintain the previous selection of both namespaces is good.
Click on "Namespace: istio-config" and maintain only the single namespace filter is correct, too.
Click on "Istio Type" and filter by the namespace of the single object and the type, looks good.

So, I haven't reviewed code yet, but the scenarios I'm testing locally looks good.

On these usability scenarios, I'd wait to have more feedback as I see all comments valid here, perhaps when more users can play with this they can tell us what they think.

lucasponce:
One issue:
image

The Istio Object takes the "/" char.

src/components/BreadcrumbView/BreadcrumbView.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+  updateItem = () => {
+    const namespaceRegex = /namespaces\/([a-z0-9\-]+)\/([a-z0-9\-]+)\/([a-z0-9\-]+)(\/([a-z0-9\-]+))?/;
+    const match = this.props.location.pathname.match(namespaceRegex) || [];
+    const ns = match[1];
+    const page = Paths[match[2].toUpperCase()];
+    const istioType = match[3];
+    let itemName = match[3];
+    if (page === 'istio') {
+      itemName = match[4];
```
lucasponce: Take a double look here: this could not be true as istio has two types of urls:
http://localhost:3000/console/namespaces/istio-system/istio/templates/logentries/accesslog
http://localhost:3000/console/namespaces/istio-system/istio/policies/disable-mtls-jaeger-collector

Templates and Adapters have subtype but not others objects.

aljesusg: let me check it

hhovsepy:
Confirming the same behavior @lucasponce described.
Tested for all Object types (Application, Service, Workload).
For instance "Workloads > Namespace: bookinfo > Workload: details-v1 > Workload Info
Click on "Workloads" and it keeps the selected Namespaces.
Click on "Namespace: bookinfo" it selectes only bookinfo namespace.

And "/" in istio config details.

hhovsepy:
Name of App, Service and Workload are missing in breadcrumb:

screenshot from 2019-02-06 10-37-23
screenshot from 2019-02-06 10-37-37
screenshot from 2019-02-06 10-37-50

But Istio Object Name is fine.

aljesusg:
Sorry @hhovsepy for the time I forgot  the other use of case when I was fixing istio routes. Now should be fine.

PR #973, opened 2019-01-31
KIALI-2306 Add detailed traffic view to Applications page

israel-hdez:
https://issues.jboss.org/browse/KIALI-2306

**This is a follow-up of https://github.com/kiali/kiali-ui/pull/956** That PR goes first. ~~This one will also show changes for that PR.~~

image


src/pages/AppDetails/AppDetailsPage.tsx
```diff
@@ -69,16 +89,48 @@ class AppDetails extends React.Component<RouteComponentProps<AppId>, AppDetailsS
...
+      namespace: { name: this.props.match.params.namespace },
+      nodeType: NodeType.APP,
+
+      // unneeded
+      workload: '',
+      service: '',
+      version: ''
+    };
+    const restParams = {
+      duration: '600s',
```
jshaughn: "duration" is the correct term for this, but it is confusingly referred to as "rateInterval" below, in TrafficDetails.   I think you should rename the "rateInterval" property in TrafficDetails to "duration".  Also, I'd suggest not using the 600 constant in multiple places.  Perhaps define a single constant and use it here, and also below.

Note that "rateInterval" should be used when working with the actual rateInterval to be applied in a rate() calculation with Prometheus.

Although it's not a blocker, I wonder if this fetch of data should be done in the TrafficDetails component when the tab is selected, and not proactively here.

These comments also apply to the former Workload details.

src/pages/AppDetails/AppDetailsPage.tsx
```diff
@@ -100,12 +152,22 @@ class AppDetails extends React.Component<RouteComponentProps<AppId>, AppDetailsS
...
-                  onRefresh={this.fetchApp}
+                  onRefresh={this.doRefresh}
                   activeTab={this.activeTab}
                   onSelectTab={this.tabSelectHandler}
                   health={this.state.health}
                 />
               </TabPane>
+              <TabPane eventKey="traffic">
+                <TrafficDetails
+                  rateInterval={600}
```
jshaughn: See the comment above about renaming this property and replacing this 600 with a constant.  I know that we'll probably be changing this to use a duration dropdown but until that time, a constant would be better to ensure the graph query and the displayed fetch duration are the same.

israel-hdez: Ok, I fixed the naming.

src/pages/AppDetails/AppDetailsPage.tsx
israel-hdez: You are right about the constants. I fixed that.

And yes... Initially I wanted to keep the `TrafficDetails` component stateless. But it does have some state. So, I think it's completely valid to move the API calls to `TrafficDetails`. I'll add that to my TODO list.

israel-hdez:
@jshaughn I fixed your comments. Regarding the workload-to-workload comment, I'm sure it won't work. It won't crash, but there will be no row showing that traffic :( I wonder if you prefer the fix to be on this PR or to open another one.

israel-hdez:
@jshaughn I also rebased to get the last design of the traffic component.

src/components/Metrics/TrafficDetails.tsx
```diff
@@ -58,7 +60,7 @@ class TrafficDetails extends React.Component<TrafficDetailsProps, TrafficDetails
   }
 
   render() {
-    const rateIntervalName = getName(this.props.rateInterval).toLowerCase();
+    const rateIntervalName = getName(this.props.duration).toLowerCase();
```
jshaughn: To be consistent, I think this should be durationName or maybe durationLabel.

israel-hdez: Ouch. I missed that one. Fixed!

israel-hdez:
@hhovsepy Do you have the time to check this one? It's quite similar to the traffic workloads PR.

israel-hdez:
Merging. Thanks for your reviews @jshaughn @hhovsepy .

PR #975, opened 2019-02-01
KIALI-2036 fix inaccurate wording and redesign health report

jotak:
Now tooltip content is displayed right below the big health indicator, not as a tooltip anymore.
It replaces the previous "report" text.
For small indicator (in list views) it's unchanged, still as a tooltip

Moving popover & overlay components from HealthDetails to HealthIndicator

screenshot

### DIFF
--- a/src/components/Health/HealthDetails.tsx
+++ b/src/components/Health/HealthDetails.tsx
@@ -1,30 +1,14 @@
 import * as React from 'react';
-import { Icon, OverlayTrigger, Popover } from 'patternfly-react';
+import { Icon } from 'patternfly-react';
 import * as H from '../../types/Health';
 
 interface Props {
-  id: string;
   health: H.Health;
-  headline: string;
-  placement?: string;
 }
 
 export class HealthDetails extends React.PureComponent<Props, {}> {
   constructor(props: Props) {
     super(props);
-  }
-
-  render() {
-    return (
-      <OverlayTrigger
-        placement={this.props.placement || 'right'}
-        overlay={this.renderPopover()}
-        trigger={['hover', 'focus']}
-        rootClose={false}
-      >
-        {this.props.children}
-      </OverlayTrigger>
-    );
   }
 
   renderStatus(status: H.Status) {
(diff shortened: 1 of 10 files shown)


jmazzitelli:
can we get it so the title of the Error Rate line doesn't rate? I know its a small thing, but it triggers me when I see "min:" on a second line under "Error Rate over late 10". Just doesn't "look right".

Perhaps say "Error Rate / last 10 min" or something like that so the line doesn't wrap?

jotak:
@jmazzitelli that's right, I'm adding some space for health :+1: 
more space

jshaughn:
> Last minute change, I've renamed Workloads Status to Pods Status

That makes sense.

src/components/Health/HealthDetails.tsx
```diff
@@ -35,33 +19,29 @@ export class HealthDetails extends React.PureComponent<Props, {}> {
...
-              )}
-            </div>
-          );
-        })}
-      </Popover>
-    );
+    return health.items.map((item, idx) => {
+      return (
+        <div key={idx}>
+          <strong>
```
aljesusg: Do we need the key in this div ?? Can we replace this by <> </> ?

src/components/Health/HealthIndicator.tsx
```diff
@@ -50,43 +43,62 @@ export class HealthIndicator extends React.PureComponent<Props, HealthState> {
...
+      position: 'relative',
+      top: -9,
+      left: 10
+    };
     return (
-      <div style={{ color: this.state.globalStatus.color }}>
-        {this.renderIndicator(health, '35px', '24px', this.state.globalStatus.name)}
+      <div>
+        {this.renderIcon('35px', '24px')}
+        <span style={spanStyle}>{this.state.globalStatus.name}</span>
```
aljesusg: Replace this div by <> </> please.

This is by https://reactjs.org/blog/2017/11/28/react-v16.2.0-fragment-support.html from version 16.2.0  . This avoid us to add extra div to the DOM

aljesusg: Maybe we should add this to the Guides What do you think @jotak , @lucasponce, @xeviknal  ?

src/components/Health/HealthDetails.tsx
jotak: I think the `key` is recommended as long as we're outputing JSX by mapping from an array. Else it raises a warning in JS console.

src/components/Health/HealthIndicator.tsx
jotak: done.

PR #977, opened 2019-02-05
KIALI-2319 Fix Menu item deselected

lucasponce:
Fix a typo in the RegExp constructor as adding extra char broke the pattern.
I guess with this is fixed.
Please @hhovsepy let me know if this is fixed.


aljesusg:
Snyk failed close/open PR

PR #983, opened 2019-02-06
KIALI-2357 Fetch data when a new workload (URL) is requested

israel-hdez:
This fixes an issue when user navigates to another workload from Kiali.
The URL updates and the page should fetch the data of the newly
requested workload.

https://issues.jboss.org/browse/KIALI-2357

Before:

before-2019-02-06 11-51

After:

after-2019-02-06 11-49


israel-hdez:
Thanks for the review. I'll merge so that I can rebase #956.

PR #985, opened 2019-02-07
KIALI-2362 Allow different operators than "sum" for raw metrics

jotak:
Add an operator dropdown for custom dashboards to allow different aggregations across pods.
Possible values: min, max, sum, avg, standard deviation, standard variance.

Backend PR: https://github.com/kiali/kiali/pull/829

sshot

### DIFF
--- a/src/app/History.ts
+++ b/src/app/History.ts
@@ -7,7 +7,7 @@ const history = createBrowserHistory({ basename: baseName });
 export default history;
 
 export enum URLParams {
-  AGG_OPERATOR = 'aggOperator',
+  AGGREGATOR = 'aggregator',
   BY_LABELS = 'bylbl',
   DIRECTION = 'direction',
   DURATION = 'duration',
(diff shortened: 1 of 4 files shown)


src/app/History.ts
```diff
@@ -7,6 +7,7 @@ const history = createBrowserHistory({ basename: baseName });
 export default history;
 
 export enum URLParams {
+  AGG_OPERATOR = 'aggOperator',
```
jshaughn: Perhaps avoid another abbreviation by using 'aggregator'?  And aggregator is basically an aggregation operator, it's something that aggregates, creating an aggregation :)  if you like the word I'd suggest carrying it through the code.

jotak: yes, I didn't think about that one but it makes sense.

jotak:
@jshaughn yes this is only in custom dashboards, because it's essentially useful for gauges and there's no gauge in istio metrics.
Arguably it could also make some sense for counter rates (such as the request volume) but it isn't as obvious as for gauges. I mean, we could want to see the average of request volume per pod instead of the total. But... I would say, unless there is some demand for that, we can skip that for the moment.

hhovsepy:
Verified the functionality.

hhovsepy:
screenshot from 2019-02-07 18-11-08

PR #987, opened 2019-02-07
KIALI-2352 Update prom rate query param calculation given change in scrape interval

jshaughn:
Istio 1.1 has updated the scrape interval for relevant metrics from 5s to
15s, presumably to reduce metric collection intensity.  This affects the
rateInterval/step param calculation we do in the UI, resulting in "single-dot"
sparklines for "Last Minute" durations, and affects other calls as well.

The change increases the minimum step to include 2 data points (making rate()
for short durations basically the same as irate()).  This provides us with at
least a minimal sparkline graph of 3 points.

Additionally, do a little refactoring for this compute method and set things
up for a possible follow-on that adds scrap interval to the server config.

The new 3-dot sparklines for a 1 minute duration:

image

### DIFF
--- a/src/services/Prometheus.ts
+++ b/src/services/Prometheus.ts
@@ -1,26 +1,32 @@
-// Hard-coded 20s minimum. It makes some assumption about prometheus scraping interval, default config in Istio
-// being 5s. We assume 4*scrape_interval is a good minimum.
-// In any way, the rate interval should never go beyond scrape interval.
-//  Eventually TODO: make this minRateInterval configurable or deduced from actual scrape interval.
-const minRateInterval = 20;
-const defaultExpectedDataPoints = 50;
+import { DurationInSeconds } from '../types/Common';
+import { serverConfig } from '../config';
 
-export interface PrometheusQueryOptions {
-  step: number;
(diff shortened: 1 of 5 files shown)


src/services/Prometheus.ts
```diff
@@ -1,26 +1,32 @@
...
-  expectedDataPoints?: number
-): PrometheusQueryOptions => {
-  const expectedDp = expectedDataPoints || defaultExpectedDataPoints;
-  let step = Math.floor(duration / expectedDp);
-  if (step < minRateInterval) {
-    step = minRateInterval;
+export const computePrometheusRateParams = (
+  duration: DurationInSeconds,
+  dataPoints?: number,
+  scrapeInterval?: DurationInSeconds
```
jotak: so that parameter is unused at the moment, you plan to use it at some point?

jshaughn: @jotak You are right, it's not really necessary, But because Prom can have several scrapeIntervals I was just afraid that maybe the single setting would maybe need an override.  I probably should have left it out for now and added it only when needed.

mattmahoneyrh:
Verified fixed.

PR #988, opened 2019-02-07
KIALI-2041 Enable edge releases in jenkins and disable in travis

josejulio:
** Describe the change **

Turns off travis releases and defers this to Jenkins.
Due the amount of "garbage" we have doing edge releases to npm (plus is no longer needed to do it this way) we are removing, jenkins job is now in charge of downloading both and doing a single edge release to dockerhub.

To do:
- [x] ~~A way to notify kiali-core to make a new edge release with the latest code once the latest kiali-ui is in npm.~~


Makefile.jenkins
```diff
@@ -46,7 +56,7 @@ endif
 
 # Determine if an end-of-week snapshot is being released and
 # set the "bumped version" accordingly.
-ifeq ($(findstring snapshot, $(RELEASE_TYPE)),snapshot)
+ifeq ($(findstring $(RELEASE_TYPE), snapshot edge),$(RELEASE_TYPE))
```
israel-hdez: RELEASE_TYPE can be:
* patch
* minor
* major
* snapshot.0
* snapshot.1

And, now, it can also be `edge`... This condition will fail with the snapshot ones :(

(see https://github.com/kiali/kiali/blob/master/deploy/jenkins-ci/Jenkinsfile#L102-L107 and https://github.com/kiali/kiali/blob/master/deploy/jenkins-ci/bin/determine-release-type.sh#L7-L15)

Makefile.jenkins
```diff
@@ -86,6 +95,7 @@ ui-test:
 	CI=true yarn test --maxWorkers=4
 
 ui-npm-publish:
+ifeq ($(PUSH_TO_NPM), y)
```
israel-hdez: Did you try that the condition works as expected?
Sometimes I had problems when using spaces inside the parenthesis, because `make` tries to compare a "trimmed" string with an non-trimmed string.

I see that the NPM_DRY_RUN condition also has the space, so just asking to be sure.

package.json
```diff
@@ -115,7 +114,6 @@
     "jest-canvas-mock": "1.1.0",
     "jest-localstorage-mock": "2.3.0",
     "node-sass-chokidar": "1.3.4",
-    "npm-snapshot": "1.0.3",
```
israel-hdez: I hope this won't go back :)

Makefile.jenkins
josejulio: You are very right there, I'll follow other approach

Makefile.jenkins
josejulio: It works, I tested this :)

Makefile.jenkins
josejulio: Updated

PR #989, opened 2019-02-07
KIALI-2364 mTLS support on Traffic Routing wizard

lucasponce:
Loadbalancing support on wizard
Usability fix: dont adjust slider when there are more than 2 workloads

image

### DIFF
--- a/src/components/IstioWizards/IstioWizard.tsx
+++ b/src/components/IstioWizards/IstioWizard.tsx
@@ -1,5 +1,17 @@
 import * as React from 'react';
-import { Button, Label, ListView, ListViewIcon, ListViewItem, Wizard } from 'patternfly-react';
+import {
+  Button,
+  Col,
+  ControlLabel,
+  DropdownButton,
+  Label,
+  ListView,
+  ListViewIcon,
+  ListViewItem,
+  MenuItem,
+  Row,
+  Wizard
+} from 'patternfly-react';
 import { WorkloadOverview } from '../../types/ServiceInfo';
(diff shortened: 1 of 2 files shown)


src/components/IstioWizards/IstioWizard.tsx
```diff
@@ -24,14 +37,32 @@ type WorkloadTraffic = {
...
+});
+
+const lbStyle = style({
+  marginLeft: 40,
+  marginRight: 20
+});
+
+const loadBalancerSimple: string[] = ['NONE', 'ROUND_ROBIN', 'LEAST_CONN', 'RANDOM', 'PASSTHROUGH'];
+
+const mTLSMode: string[] = ['NONE', 'ISTIO_MUTUAL', 'MUTUAL', 'SIMPLE', 'DISABLE'];
```
xeviknal: What about having an enum? https://www.typescriptlang.org/docs/handbook/enums.html
Same for Line 54.
wdyt?

src/components/IstioWizards/IstioWizard.tsx
```diff
@@ -24,14 +37,32 @@ type WorkloadTraffic = {
...
+
 class IstioWizard extends React.Component<Props, State> {
   constructor(props: Props) {
     super(props);
     this.state = {
       showWizard: false,
-      workloads: []
+      workloads: [],
+      mtlsMode: 'NONE',
+      loadBalancer: 'NONE'
```
xeviknal: In case you liked the idea of enumerators, here it would go the corresponding constant.

PR #990, opened 2019-02-08
KIALI-2359 Utilize the new prom config in ServerConfig

jshaughn:
- use scrapeInterval when computing rate params
- use retention to determine valid "Fetch Duration" dropdown values.
   - GraphRefresh, OverviewToolbar, MetricsDuration
- Replace static/default serverConfig with actual serverConfig
   - fetched at login/checkCred time
- Move serverConfig into persisted redux
   - it needs persistence because checkCreds is async (not sure why this is a thunk actually, it seems it should block) and because we may start rendering prior to the serverConfig being updated, we need something in the store. Note that this is not an issue at login time.
- update serverConfig() to actually get the info from the store.
- make config imports more explicit for clarity and to avoid cyclic imports
   - also, break ServerConfig into its own file (from Config)

@jotak @israel-hdez Sorry, this is a relatively large PR although it's not really bigger than it needs to be, I think.   @mattmahoneyrh, While validating that the dropdowns are properly filtered also try various navigations with logouts/logins and F5s from various pages.   The duration dropdowns are on Overview, Graph, and Metrics (inbound/outbound) pages.  The dropdowns should be filtered based on the prometheus retention (probably 6h given the prom config I've seen for our installs, see the prom container's start command for the actual retention setting).

*** THIS PR REQUIRES SERVER PR: https://github.com/kiali/kiali/pull/836 ***  !!! MERGED, use master !!!

### DIFF
--- a/src/actions/LoginThunkActions.ts
+++ b/src/actions/LoginThunkActions.ts
@@ -1,14 +1,53 @@
 import { ThunkDispatch } from 'redux-thunk';
-import { setServerConfig } from '../config';
 import { HTTP_CODES } from '../types/Common';
-import { KialiAppState } from '../store/Store';
+import { KialiAppState, Token, ServerConfig } from '../store/Store';
 import { KialiAppAction } from './KialiAppAction';
 import HelpDropdownThunkActions from './HelpDropdownThunkActions';
 import GrafanaThunkActions from './GrafanaThunkActions';
 import { LoginActions } from './LoginActions';
 import * as API from '../services/Api';
-import { ServerConfig } from '../config/Config';
+import { ServerConfigActions } from './ServerConfigActions';
 
+const ANONYMOUS: string = 'anonymous';
+
+// completeLogin performs any initialization required prior to declaring the login complete. This
+// prevents early rendering for components waiting on a login.
+const completeLogin = (
+  dispatch: ThunkDispatch<KialiAppState, void, KialiAppAction>,
+  token: Token,
+  username: string,
+  timeout?: number
+) => {
+  const auth = `Bearer ${token['token']}`;
+  API.getServerConfig(auth).then(
+    response => {
+      // set the serverConfig before completing login so that it is available for
+      // anything needing it to render properly.
+      const serverConfig: ServerConfig = {
+        istioNamespace: response.data.istioNamespace,
+        istioLabels: response.data.istioLabels,
+        prometheus: response.data.prometheus
+      };
+      dispatch(ServerConfigActions.setServerConfig(serverConfig));
+
+      // complete the login process
+      dispatch(LoginActions.loginSuccess(token, username, timeout));
+
+      // dispatch requests to be done now but not necessarily requiring immediate completion
+      dispatch(HelpDropdownThunkActions.refresh());
+      dispatch(GrafanaThunkActions.getInfo(auth));
+    },
+    error => {
+      /** Logout user */
+      if (error.response && error.response.status === HTTP_CODES.UNAUTHORIZED) {
+        dispatch(LoginActions.logoutSuccess());
+      }
+    }
+  );
+};
+
+// performLogin performs only the authentication part of login. If successful
+// we call completeLogin to perform any initialization required for the user session.
 const performLogin = (
   dispatch: ThunkDispatch<KialiAppState, void, KialiAppAction>,
   username?: string,
@@ -16,19 +55,15 @@ const performLogin = (
 ) => {
   dispatch(LoginActions.loginRequest());
 
-  const anonymous = username === undefined;
-  const loginUser: string = username === undefined ? 'anonymous' : username;
-  const loginPass: string = password === undefined ? 'anonymous' : password;
+  const loginUser: string = username === undefined ? ANONYMOUS : username;
+  const loginPass: string = password === undefined ? ANONYMOUS : password;
 
   API.login(loginUser, loginPass).then(
     token => {
-      dispatch(LoginActions.loginSuccess(token['data'], loginUser));
-      const auth = `Bearer ${token['data']['token']}`;
-      dispatch(HelpDropdownThunkActions.refresh());
-      dispatch(GrafanaThunkActions.getInfo(auth));
+      completeLogin(dispatch, token['data'], loginUser);
     },
     error => {
-      if (anonymous) {
+      if (loginUser === ANONYMOUS) {
         dispatch(LoginActions.logoutSuccess());
       } else {
         dispatch(LoginActions.loginFailure(error));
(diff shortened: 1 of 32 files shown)


src/actions/LoginThunkActions.ts
```diff
@@ -1,34 +1,69 @@
...
     token => {
-      dispatch(LoginActions.loginSuccess(token['data'], loginUser));
-      const auth = `Bearer ${token['data']['token']}`;
-      dispatch(HelpDropdownThunkActions.refresh());
-      dispatch(GrafanaThunkActions.getInfo(auth));
+      completeLogin(dispatch, token['data'], loginUser);
     },
     error => {
-      if (anonymous) {
+      if (loginUser === ANONYMOUS) {
```
jotak: better to check `username === undefined`, no? If the user is really named "anonymous" ...

src/config/ServerConfig.ts
```diff
@@ -0,0 +1,11 @@
+import deepFreeze from 'deep-freeze';
+import { store } from '../store/ConfigStore';
+import { KialiAppState, ServerConfig } from '../store/Store';
+import { PersistPartial } from 'redux-persist';
+
+// It's not great to access the store directly for convenience but the alternative is
+// a huge code ripple just to access some server config. better to just have this one utility.
+export const serverConfig = (): ServerConfig => {
```
jotak: I'll leave this to the redux gurus if there's a better approach :-)

src/actions/LoginThunkActions.ts
jshaughn: @jotak Thanks for proactively taking a look at this PR.  Above we set loginUser to ANONYMOUS based on `username === undefined` and ANONYMOUS has more semantic meaning here than undefined.  So, I think it's OK.

src/config/ServerConfig.ts
jshaughn: I think the alternate approach would be to have every user of the serverConfig be containerized and I just think that's overkill because in several places we need it in "utility" code.  I actually went down this path and it was getting out of hand.   I containerized the actual classes that need to respond to a change for re-render (MetricsDuration, GraphRefresh, OverviewToolbar).

israel-hdez: What I understand is that the `serverConfig`, once set after logging-in or after validating the stored token (because a browser refresh), the values will never change.

Personally, I would have avoided redux. Since the values never mutate once set, *I think* the server config is not really part of *the state of the app*, but just global values. Thus, redux looks overkill for this use-case.
 
Instead, what I'd have done is to create a separate module to store these values as plain global variables. Something similar how the `history` object is handled in https://github.com/kiali/kiali-ui/blob/929598bd0f324219e33e1a9bdbb2004b3de7abdc/src/app/History.ts. On that separate module you could also have placed all utility functions to retrieve the "valid durations" and any other needed helper functions. Then, you could have just imported that new module where needed and read the global variables leaving redux aside and avoiding the burden of reducers, actions, component containerization, etc. 

Then, your `completeLogin` method, instead of dispatching a `setServerConfig` action, would had simply set the global variables.

But you have already done all that redux burden, is working as is and it's quite OK. So this is only a suggestion about considering keeping redux aside the next time global variables can do the work.

src/reducers/ServerConfigState.ts
```diff
@@ -0,0 +1,21 @@
...
+export const INITIAL_SERVER_CONFIG: ServerConfig | null = null;
+
+// This Reducer allows changes to the 'serverConfig' portion of Redux Store
+const serverConfig = (
+  state: ServerConfig | null = INITIAL_SERVER_CONFIG,
+  action: KialiAppAction
+): ServerConfig | null => {
+  switch (action.type) {
+    case getType(ServerConfigActions.setServerConfig):
+      return action.payload;
```
israel-hdez: I think it's worth to check if the `duration` stored in redux is still valid and update the store if it no longer will be a valid duration. That way, components won't try to set `undefined` values.

src/components/GraphFilter/GraphRefresh.tsx
```diff
@@ -79,8 +86,8 @@ export class GraphRefresh extends React.PureComponent<GraphRefreshProps> {
           disabled={this.props.disabled}
           handleSelect={this.props.setDuration}
           value={this.props.duration}
-          label={String(GraphRefresh.DURATION_LIST[this.props.duration])}
-          options={GraphRefresh.DURATION_LIST}
+          label={String(validDurations[this.props.duration])}
```
israel-hdez: Please see my comment of the reducer/ServerConfigState.ts file. I think this is what may become `undefined`.

src/store/Store.ts
```diff
@@ -107,12 +107,22 @@ export interface GrafanaInfo {
   varWorkload: string;
 }
 
+export interface ServerConfig {
+  istioNamespace: string;
+  istioLabels: { [key: string]: string };
```
jotak: Just a suggestion here, because I was changing it in another PR but I'm going to revert my changes in order to not conflict with yours: could you change the istioLabels type to this:

```
  istioLabels: { [key in IstioLabelKey]: string };
```

And define IstioLabelKey above:
```
type IstioLabelKey = 'appLabelName' | 'versionLabelName';
```

By the way, in some places I can see these labels with first letter upper case or not. What's actually given by the server, low case or upper case? It should be low case (camel case), but do we have to change on the server side as well?

src/config/ServerConfig.ts
jshaughn: > I have minor comments for your consideration, but nothing that should block this.

@israel-hdez You are funny, as this is a major comment :)   But you are also correct, the redux 
approach is not fantastic, and it wasn't my first approach.   You are right that the values will not change between login or a credential check.  I did not start with adding serverConfig to redux but I did it for two reasons.  First, by having it in redux we can better re-establish the state surrounding a problem.  The second reason was more critical.  When navigator dispatches LoginThunkActions.checkCredentials (e.g. on an F5) the logic does not block the rendering, so what happens is the page starts rendering before the serverConfig is set, and that's a major problem as there is nothing set when the page needs to reference serverConfig().

I am actually quite OK with an approach that does not use redux if the call to serverConfig() is guaranteed to return the actual serverConfig (or thrown an error).  This could possibly be done  by somehow making checkCredentials block (pehaps convert it from a thunk?), or if a call to serverConfig() fetched it if needed, setting a global the first time it is called (note that it would need an authorization token to do so).  In the second approach login/checkCredentials would not actually need to set serverConfig, and only checkCredentials would need call it, just as an auth check.

I am OK if you want to submit a PR to this PR with a different approach.

src/components/MetricsOptions/MetricsDuration.tsx
```diff
@@ -46,15 +54,19 @@ export default class MetricsDuration extends React.Component<Props> {
...
     return (
       <ToolbarDropdown
         id={'metrics_filter_interval_duration'}
         disabled={false}
         handleSelect={this.onDurationChanged}
         nameDropdown={'Fetching'}
         initialValue={this.duration}
-        initialLabel={MetricsDuration.Durations[this.duration]}
-        options={MetricsDuration.Durations}
+        initialLabel={validDurations[this.duration]}
```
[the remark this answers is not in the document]
jshaughn: I think it's unlikely but It's true.  Before making a change I'd like to know if you want to submit a PR changing the approach.  if so perhaps you would work the validation into a utility.

src/store/Store.ts
jshaughn: The JSON from the looks like this:

```
{
  "istioNamespace": "istio-system",
  "istioLabels": {
    "AppLabelName": "app",
    "VersionLabelName": "version"
  },
  "prometheus": {
    "globalScrapeInterval": 15,
    "storageTsdbRetention": 21600
  }
}
```

I don't know why AppLabelName and VersionLabelName are uppercase, it's been that way for a while now.  It could be changed for consistency.

Again, I will not make changes until I know whether you intend to submit a PR against this PR.

src/config/ServerConfig.ts
jotak: I'm not sure if it's easy or not to have in our current implementation, but `serverConfig()` could also just return a promise, so that anything that calls it would have to deal with the result asynchronously.

src/store/Store.ts
jotak: Not sure which one will be merged first, I'm changing the backend ( https://github.com/kiali/kiali/pull/839 ) and the frontend ( https://github.com/kiali/kiali-ui/pull/997 ) ; if mine goes first, I'll submit a PR on top of yours for merging.

src/config/ServerConfig.ts
jshaughn: @israel-hdez After talking to you offline the decision is to leave the redux solution in place as it is working and an alternative approach could be investigated in a later refactoring.  @jotak, a promise approach may work, the component would have to defer rendering until the response was received.  Also, we still need to ensure we have an auth token before we can make the call.

src/components/MetricsOptions/MetricsDuration.tsx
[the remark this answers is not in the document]
jshaughn: @israel-hdez I added this protection and also added a couple of utility methods in ServerConfig.  Please review and if it seems OK to you now feel free to approve :)

src/store/Store.ts
jshaughn: @jotak OK, I have added the suggested typing.  But the keys are currently upper-cased because that is what is coming from the server.  I think we'll need a follow-up server and UI PR to convert to lowercase.

jotak: Thanks.
FYI I'm changing it here #997 which comes with small server-side change https://github.com/kiali/kiali/pull/839 , if you want to take a look.

src/actions/LoginThunkActions.ts
jmazzitelli: Did you actually test anonymous access? I wrote that and IIRC that "undefined" was special. I'll test it today, but want to ask if you already did. I have a sneaky suspicion that anonymous access has been broken.

jmazzitelli: quick smoke test appears to show me that, yes, anonymous access has now been broken. We need to fix this. We know of at least one community member using anonymous access.

jmazzitelli: This most likely was broke in another PR where I changed it to support the situation when there is no secret installed. I'll write up a JIRA.

PR #993, opened 2019-02-12
Update Style Guide

jotak:
This is following some previous discussions and trying to have clearer guidelines (although not entirely clear yet). See previous discussions here: https://github.com/kiali/kiali-ui/pull/923

In that discussion, @israel-hdez provided a link to Basarat's guide that I'm taking here as a reference. @mtho11 pointed to microsoft's guide (https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines ), which is actually quite similar, but I'm not taking it as a reference because of the disclaimers it shows.

So, changes are:

- Mainly following Basarat's guide https://github.com/basarat/typescript-book/blob/master/docs/styleguide/styleguide.md
- Some exceptions listed
- Removed the section about Proptypes as we don't use it

ping @jshaughn  @josejulio @israel-hdez @mtho11 @aljesusg @lucasponce @xeviknal ...

### DIFF
--- a/STYLE_GUIDE.adoc
+++ b/STYLE_GUIDE.adoc
@@ -3,28 +3,38 @@
 
 toc::[]
 
-= File naming conventions
+= Default conventions
+
+By default, we try to follow community conventions. link:https://github.com/basarat/typescript-book/blob/master/docs/styleguide/styleguide.md[Basarat's Style Guide] is used as a reference, with a few exceptions that are listed below:
+
+== File names
+
+Most files are named after the main type / class / interface it's exporting. For instance, `ServiceList.ts` is named after the `ServiceList` interface that it's exporting. Which means that, unlike in Basarat's guide, we often use `PascalCase`.
+
+Files with more general purpose, or the ones that don't export types or such, are `camelCase` (ex: `routes.ts`).
+
+== Variables, functions, constants
+
+We mostly follow Basarat's guide, so `camelCase`. Exceptions are:
+
+- Redux actions, which are technically constants, are `PascalCase` (ex: `GraphActions`).
+- We sometimes make a difference between constants of global scopes versus constants of local scopes. You may find the former written in `UPPER_SNAKE_CASE` (ex: `const TIMER_REQUEST_PER_SECOND_MIN = 0;`), whereas Basarat's guide makes no distinction between different kinds of constants, all of them being `camelCase`.
+
+== Enum
+
+Enum names follow Basarat's guide, but not the values: they are `UPPER_SNAKE_CASE`. Example:
 
 [source,typescript]
 ----
-// bad
-import Footer from './Footer/index';
-// good
-import Footer from './Footer';
+enum DisplayMode {
+  LARGE,
+  SMALL
+}
 ----
 
+= Handlers
 
-= Methods
-
-=== Naming
-
-* action types - use constants, [NOUN]_[VERB], eg. USER_FETCH or MATERIAL_REQUEST_UPDATE
-* action creators - [verb][Noun]() eg. fetchCurrentUser(), acceptMaterialRequest()
-* selectors - get[NOUN]() or select[NOUN]() eg. getSelectedCards(), selectActiveRows()
-
-=== Handlers
-
-Use consistent naming for events and event handlers: Name the handler methods after their triggering event
+Use consistent naming for events and event handlers: name the handler methods after their triggering event.
 
 Event handlers should:
 
@@ -37,11 +47,11 @@ Event names:
 * in props should start with on
 * should not clash with builtin/native event names, eg. use onSelect instead of onFocus or onClick
 
-=== Use arrow functions
+= Use arrow functions (fat arrow)
 
 [source,typescript]
 ----
-CreateItem = () => {
+createItem = () => {
     return (
       <ul>
         {props.items.map((item, index) => (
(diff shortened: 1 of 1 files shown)


STYLE_GUIDE.adoc
```diff
@@ -3,28 +3,38 @@
...
+== Variables, functions, constants
 
-* action types - use constants, [NOUN]_[VERB], eg. USER_FETCH or MATERIAL_REQUEST_UPDATE
-* action creators - [verb][Noun]() eg. fetchCurrentUser(), acceptMaterialRequest()
-* selectors - get[NOUN]() or select[NOUN]() eg. getSelectedCards(), selectActiveRows()
+We mostly follow Basarat's guide, so `camelCase`. Exceptions are:
 
-=== Handlers
+- Redux actions, which are technically constants, are `PascalCase` (ex: `GraphActions`).
+- We sometimes make a difference between constants of global scopes versus constants of local scopes. You may find the former written in `UPPER_SNAKE_CASE` (ex: `const TIMER_REQUEST_PER_SECOND_MIN = 0;`), whereas Basarat's guide makes no distinction between different kinds of constants, all of them being `camelCase`.
```
jotak: Note here: we should have a clear guideline here instead of the unclear "sometimes". Which means that we should update all the code not following whatever guideline we choose.

What you folk would you prefer: upper snake case on top-level constants, or no scope distinction hence camel case?

jotak: (Personally I think I prefer to make a distinction and keep upper snake case. It makes it clearer as we read it that it's something top-level defined)

lucasponce:
How good or bad are we on this ?
I'm afraid that following a "changes on code style" we may have a PR with thousands of changes to adjust into the new code style, with the nice side effect to create multiple conflicts to our colleagues :-).

And I'm not a fan of those uber PRs.
@mtho11 owes some beers from last time he did something similar :-)

jotak:
@lucasponce this is more an attempt to formalize what we have _now_ than an attempt to change everything!

STYLE_GUIDE.adoc
```diff
@@ -3,28 +3,38 @@
...
-=== Naming
+== Variables, functions, constants
 
-* action types - use constants, [NOUN]_[VERB], eg. USER_FETCH or MATERIAL_REQUEST_UPDATE
-* action creators - [verb][Noun]() eg. fetchCurrentUser(), acceptMaterialRequest()
-* selectors - get[NOUN]() or select[NOUN]() eg. getSelectedCards(), selectActiveRows()
+We mostly follow Basarat's guide, so `camelCase`. Exceptions are:
 
-=== Handlers
+- Redux actions, which are technically constants, are `PascalCase` (ex: `GraphActions`).
```
israel-hdez: For a follow up: Even if Redux actions are declared with a `const` keyword and, then, in an object; they follow the JS way of namespacing. So, for Redux actions, the namespace convention should apply (which we already comply)

STYLE_GUIDE.adoc
israel-hdez: Hmmm... IMO, we should use upper snake case for the constants that can be "mapped" as constants in other languages (the `TIMER_REQUEST_PER_SECOND_MIN` is a good example).

For all other constants, we should apply the _variables_ convention (or any other that may could be applied given the case). For example, this one https://github.com/kiali/kiali-ui/blob/387e5b21199c83394eae4489f80661b2edf5cc26/src/components/MetricsOptions/MetricsDuration.tsx#L21 declared as constant, but other languages won't even allow to declare constants in that scope... So, it's ok to apply the convention for variables.

jotak: @israel-hdez yes, we could do that way, I tried to formalize that analogy with "constants" as understood in other languages by this notion of "top-level constants" versus "local-scoped constants". But our current code is still very far from adopting this pattern.
Just picking a random example, but there's many other examples:
https://github.com/kiali/kiali-ui/blob/master/src/components/CytoscapeGraph/CytoscapeToolbar.tsx#L34-L43

Where we can see side by side top-level camelCase constants and upper-case constants.
So, at the moment we don't really have any rule about that, it's up to the developer.

STYLE_GUIDE.adoc
jotak: I don't know really about the "JS way of namespacing", basically any variable can contain functions and hence be considered as something like a namespace, no?
Typescriptly speaking it's not a namespace so I think it's good to mention.

israel-hdez: Yes, if we weren't using typescript, we would be creating variables that contains functions to do namespacing.
Check this simplistic TS example: https://bit.ly/2X2q0xq. You can see that, indeed, a "typescript namespace" is transpiled to a JS code that results in creating a simple variable with functions. 

I haven't tried, but probably we could convert Redux actions to code using typescript namespaces and, probably, everything may still would work without changes. If this is true, we could convert the Redux actions code to avoid this kind of exceptions to the coding style.

STYLE_GUIDE.adoc
israel-hdez: Hmmm... true...
Being strict, the `cytoscapeToolbarStyle` variable cannot be mapped to a `const` in other languages, because compilers need to resolve a static value, while the `cytoscapeToolbarStyle` value can only be resolved at runtime (because it's using a function).

But I can see that this can be very confusing. Not sure how we could remove the "sometimes" subjectivity.

jotak: depends what is the "other language" :-)
In c/c++ it's pre-processed so you're right.
In java what 'maps' to constants ("static final"), it's still resolved at runtime, allowing function calls, so it would work.
Here there's no preprocessing so no need to consider differently what can only be expressed at runtime or not.

israel-hdez: Hmm... Yes, I was probably thinking about how C/C++ work... I feel old :'(

PR #996, opened 2019-02-13
KIALI-2032 Show runtime icons

jotak:
JIRA: https://issues.jboss.org/browse/KIALI-2032
Related UX issue: https://github.com/kiali/kiali-design/issues/93

Examples:

Node.js:
nodejs

Thorntail:
Thorntail

Vert.x:
vertx

Mixed with text:
mixed


abonas:
nice :)
the things that probably need some improvement are:
- the logo size does not always fit the line so it looks a bit out of place.
- in the above example with 2 runtimes (vertx and mesh arena) it is not very clear this is a list, the comma is tiny

jotak:
@abonas this is a bit tricky to correctly align logos, because it depends on text alignment within the image itself. For instance in nodejs logo the text is top-aligned, in thorntail it's centered vertically, whereas in vertx it fits the whole image.
Maybe a solution could be to define specific css alignment per logo.

abonas:
> @abonas this is a bit tricky to correctly align logos, because it depends on text alignment within the image itself. For instance in nodejs logo the text is top-aligned, in thorntail it's centered vertically, whereas in vertx it fits the whole image.
> Maybe a solution could be to define specific css alignment per logo.

yes, I suspected so. better to ask the UX team, they might suggest some best practice around that

jotak:
@abonas about the list & coma : maybe having a bigger separator, such as a slash `/`, would make it clearer?

abonas:
> @abonas about the list & coma : maybe having a bigger separator, such as a slash `/`, would make it clearer?

could be, perhaps "|" works better than "/" , but again - UX team would know better, please raise a question :)

pilhuhn:
What is the reason to display the 'mesh-arena' at all? This is not a runtime. Otherwise would you want to list the .ear/.war/... files of an EAP deployment there?

I wonder why we need the row title at all -- at least in case we have a logo.

Last but not least - if we have a logo I'd put it more on top close the the WL icon (just a personal taste).

All this is not a blocker from my side.

jotak:
@abonas Question raised on UX issue. An update with "/":

with slash

jotak:
> What is the reason to display the 'mesh-arena' at all? This is not a runtime. Otherwise would you want to list the .ear/.war/... files of an EAP deployment there?

Not at all, it's just here as an example, maybe I didn't pick the best name to show it, but technically it is possible to have multiple runtimes. This is all user defined, if you label your runtime "xyz" in dashboard it will show "xyz".

> 
> I wonder why we need the row title at all -- at least in case we have a logo.

This was just to be consistent with the other information display, but yes, maybe it can be removed.

> 
> Last but not least - if we have a logo I'd put it more on top close the the WL icon (just a personal taste).

Sure we can try that

> 
> All this is not a blocker from my side.

jotak:
Other tries, without row labels:

- displayed below block
below

- displayed above block
above

pilhuhn:
I like the 1st one without row label.

lucasponce:
> Last but not least - if we have a logo I'd put it more on top close the the WL icon (just a personal taste).

+1

israel-hdez:
> Last but not least - if we have a logo I'd put it more on top close the the WL icon (just a personal taste).

I have another personal preference :D Just by looking at the images, I really think that the runtime icons should "fill" that blank space in the middle of the screen ;)

abonas:
> I have another personal preference :D Just by looking at the images, I really think that the runtime icons should "fill" that blank space in the middle of the screen ;)

we could do that too.
there's an older object details design
https://github.com/kiali/kiali-design/blob/master/object-details/design.md
we can revisit the whole page.
some things can be regrouped and relocated there for sure.

jotak:
So, current winner is the logo displayed below info block without row title.
Let's see how it would be with @israel-hdez 's proposal:

middle

src/config/Logos.ts
```diff
@@ -5,3 +5,10 @@ export const IstioLogo = require('../assets/img/istio-logo.svg');
 
 /** Kiali logo */
 export const KialiLogo = require('../assets/img/logo-alt.svg');
+
+// Runtimes
+export const runtimes = {
+  'Vert.x': 'vertx-logo.png',
+  'Node.js': 'nodejs-logo.png',
+  'Thorntail': 'thorntail-logo.png'
+};
```
aljesusg: I don't know if create the full path here instead that in the code, all the images should be in the same place.

WorkloadDescription.tsx L-28
```ts
      const logo = require('../../../assets/img/' + imageName);
      return <img src={logo} alt={name} />;
```
we could do something like 
```ts
const imgPath = '../assets/img/');
/** Kiali logo */
export const KialiLogo = require(imgPath + 'logo-alt.svg');

// Runtimes
export const runtimes = {
  'Vert.x': imgPath + 'vertx-logo.png',
  'Node.js': imgPath +'nodejs-logo.png',
  'Thorntail': imgPath + 'thorntail-logo.png'
};
```
No blocker . What do you think @jotak ?

aljesusg: mm I think that this not fix that because is different path but you cant add the require and use it like we do with KialiLogo and remove the require of the WorkloadDescription code

jotak: Yes, good point, I changed it. However I can't assign `imgPath` and then build the string by concatenating, I get a "module not found" error, I guess this is because `require` calls are preprocessed by webpack and not executed at runtime (?). Setting full relative path as a single string works fine however.

abonas:
> Note, it isn't exactly centered horizontally because we need to leave some place on the left in case the workload has many or long labels.

putting that without a row title is a bit strange and not looking consistent with the rest of the page.

jotak:
So, I'm merging the current version, which is this one: 
below
(with pipes as separators)

Changes in design can still be discussed here: https://github.com/kiali/kiali-design/issues/93
Or through a more global redesign of the workload details page.

PR #997, opened 2019-02-13
KIALI-2384 Code cleaning & consistent naming in "config" dir

jotak:
Fix some inconsistencies, we currently have exported consts:
- PascalCase (ex: KialiLogo)
- UPCASE (ex: ICONS)
- camelCase (ex: config)

So, moving all to camelCase in this "config" directory. Note, to avoid very large PRs across all the code, this change is focused only on config.

Also fixing:
- IstioLogo isn't used anymore, so removing its definition
- deepFreeze was called for each invocation of the resource (config and/or icons) whereas this is something static. So calling only once.

Backend PR required: https://github.com/kiali/kiali/pull/839


src/config/Config.ts
```diff
@@ -126,20 +126,20 @@ const conf = {
   }
 };
 
-export const config = () => {
-  return deepFreeze(conf) as typeof conf;
-};
+export const config = deepFreeze(conf) as typeof conf;
```
jotak: Note about that: I don't know if there was a good reason to have this lazy `deepFreeze`, repeated for each access. Testing locally, it doesn't seem to break anything to set it once for all. Anyone has an idea about the reason? (@israel-hdez , @mtho11  maybe? )

israel-hdez: Not sure. `conf` is also a constant, so it shouldn't mutate. For me, it is better with your change.

jotak:
@jshaughn they export respectively `serverConfig` (the function, not the interface which is defined elsewhere) and `config`, so in both cases it would be camelCase, seems to me.

jshaughn:
@jotak Ah, OK.  So it's named after the export.  So it makes sense.  As long as the export is properly cased, I'll assume it is :)

jotak:
@jshaughn yes... I hope this rule of "named after the export" will be fine in the long run. It was already "kind of" followed by some of us I think, which is not too surprising as for java developers we're used to having files named after their public class.

PR #998, opened 2019-02-13
KIALI-2013 Extend highlighting for selected graph element

jshaughn:
When hovering highlight the element's neighborhood but when selected
highlight the end-to-end paths involving the element.

For example, a selected edge before:
image

A selected edge after:
image

### DIFF
--- a/src/components/CytoscapeGraph/graphs/GraphHighlighter.ts
+++ b/src/components/CytoscapeGraph/graphs/GraphHighlighter.ts
@@ -89,25 +89,22 @@ export class GraphHighlighter {
       .addClass(DIM_CLASS);
   };
 
-  // Returns the nodes to highlight depending the selected or hovered summaryType
-  // If current selected is 'graph' (e.g. no selection):
-  //  Check if we are hovering on a node or edge and hover those relations
-  // If current selected is something else, highlight that node (group, node or edge)
+  // Returns the nodes to highlight. Highlighting for a hovered element
+  // is limited to its neighborhood.  Highlighting for a selected element
+  // is extended to full incoming and outgoing paths.
   getHighlighted() {
-    if (this.selected.summaryType === 'graph') {
-      return this.getHighlightedByEvent(this.hovered);
-    }
-    return this.getHighlightedByEvent(this.selected);
-  }
-
-  getHighlightedByEvent(event: CytoscapeClickEvent | CytoscapeMouseInEvent | undefined) {
+    const isHover = this.selected.summaryType === 'graph';
+    const event = isHover ? this.hovered : this.selected;
     if (event) {
-      if (event.summaryType === 'node') {
-        return this.getNodeHighlight(event.summaryTarget);
-      } else if (event.summaryType === 'edge') {
-        return this.getEdgeHighlight(event.summaryTarget);
-      } else if (event.summaryType === 'group') {
-        return this.getGroupHighlight(event.summaryTarget);
+      switch (event.summaryType) {
+        case 'node':
+          return this.getNodeHighlight(event.summaryTarget, isHover);
+        case 'edge':
+          return this.getEdgeHighlight(event.summaryTarget, isHover);
+        case 'group':
+          return this.getAppBoxHighlight(event.summaryTarget, isHover);
+        default:
+        // fall through
       }
     }
     return undefined;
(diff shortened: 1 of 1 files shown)


josejulio:
What would happen if you select `kiali-test-reviews`? It would highlight almost all the graph?
Wouldn't this make a useless the selection of the nodes that are at the beginning of the request?
( I havent tested yet, so i might be wrong )

If so I think something like:
 - Select starting node
 - Right click second node and chose: "highlight path to here" or "highlight path from here"

mwringe:
I think this is good. Its gives up a way to highlight how a component could potentially affect other things in the graph.

@gbaufake @mattmahoneyrh Can one of you test out this issue?

@abonas Any comments on this?

mwringe:
> What would happen if you select `kiali-test-reviews`? It would highlight almost all the graph? 

'kiali-test-reviews' is the namespace, what are you asking to be clicked on?

And yes, if you were to click something like the product page, then the whole graph is 'highlighted' (well, its not really highlighted, its just that the connections are not faded to grey).

Which I think is the point of this, if we have a component that can affect the whole graph, then lets not fade away parts of the graph since it can affect everything.

> * Right click second node and chose: "highlight path to here" or "highlight path from here"

If we add in extra steps, it makes things more complex and it becomes more difficult for people to use. I think this change as is is a better default than what we were doing before.

Just a note on this. Its not using tracing data, which means its not as accurate as it could be. A future improvement would be to grab the tracing data from Jaeger and use that to create a more accurate picture.

josejulio:
> 'kiali-test-reviews' is the namespace, what are you asking to be clicked on?

sorry, meant `reviews` :)

israel-hdez:
My personal preference is to be able to see the extended highlight without the need of a click. But also, no problem about extending on selection.

What I don't like is that we lose the differentiation between immediate neighbors vs the extended path. I'd prefer to do the highlighting in a different way. I don't have lots of ideas, but the only so-so idea that I can think of is to highlight the path in a similar way to the "find" feature:

image

but using a different background color. Probably, light blue or light grey... I don't know. Just a suggestion to avoid losing the ability to see the immediate neighbors.

abonas:
@mwringe I am not sure I understand the difference between the extended path , or "end to end" path to neighborhood. How each is defined? neighborhood is only "one level up/down"?
And what is the use case of showing each type?
Also, is this expected that this difference of what to show during hovering vs selecting to be clear enough to users? (I'm having doubts...)

josejulio:
> @mwringe I am not sure I understand the difference between the extended path , or "end to end" path to neighborhood. How each is defined? neighborhood is only "one level up/down"?

`neighborhood` (behavior before this PR) are the nodes that are directly connected to the node and the `end-to-end``(this PR) are the predecessors of a node *and* the successors of a node.

Predecessors defined as the nodes that goes to that node and their predecessors  (the incomers, incomer's incomers, etc)
Successors are the outgoers, outgoer's outgoers, etc

jshaughn:
@josejulio @israel-hdez @mwringe , you all have good points.  I agree with Matt that the point this feature is to highlight basically the entire graph if in fact the selected element influences that many end-to-end paths.  This is what the creator of Kiali-2013 (David Sancho) is asking for, I think.   I agree with Josejulio that perhaps this will not be particularly useful at times, but then again, maybe it will.  Because David has a user's perspective I'm hoping he'll like the feature.  Certainly we can ask him for feedback.  I agree with Edgar that it is still good to see the neighborhood, but the idea there is that if you double-click you get that focused view in the node detail graph.  So at least for a first pass I think that is good and we can defer separate neighborhood highlighting until we have user feedback.

So,I am thumbs up and @mwringe is thumbs up.  But @israel-hdez @josejulio @abonas , are you thumbs up or thumbs down, let's be democratic :)

josejulio:
Looking at the JIRA i though this was about highlighting a path (from Node A to Node B) hence my idea above.

After reading the comments it seems that it was agreed that having this extended connectivity would be useful.

I'm thumbs up :+1:

dsanchor:
From the user's perspective and what I have seen during a my last couple of months, when you deploy +100 services and you want to understand where the request may come from and where it may go afterwards, this feature it is very useful, I would say essential. In my case, depending of the service, the service call chain may include from 4 to 8 services and when you have +100 services in the graph, it gives you a better understanding of which service may be involved in the service execution chain.

abonas:
I understand this functionality better now, thanks for the explanations.
I'm not sure that the suggested way for highlighting/selection will do it justice.
Any reason why this is not raised as an issue to UX?

jshaughn:
@abonas I'm going to merge this as-is.  If you want to expose to UX for feedback that's fine too, I sometimes think we don't have quite enough disparity between dimmed and non-dimmed.  But as-is it's just an extension of the current l+f, so it seems reasonable.

israel-hdez:
Ok, since this has user feedback, I'm :+1:

PR #1002, opened 2019-02-18
KIALI-1469 Do not report "no requests" in health on missing sidecar

jotak:
When pods are not istio-sidecared, just omit requests errors in health
(or in case of ServiceHealth where this is the only item checked, just write n/a - no sidecar)

JIRA: https://issues.jboss.org/browse/KIALI-1469

### DIFF
--- a/src/pages/AppDetails/AppDetailsPage.tsx
+++ b/src/pages/AppDetails/AppDetailsPage.tsx
@@ -48,22 +48,19 @@ class AppDetails extends React.Component<RouteComponentProps<AppId>, AppDetailsS
   }
 
   fetchApp = () => {
-    const promiseDetails = API.getApp(authentication(), this.props.match.params.namespace, this.props.match.params.app);
-
-    const promiseHealth = API.getAppHealth(
-      authentication(),
-      this.props.match.params.namespace,
-      this.props.match.params.app,
-      600
-    );
-
-    Promise.all([promiseDetails, promiseHealth])
-      .then(([resultDetails, resultHealth]) => {
-        this.setState({
-          app: resultDetails.data,
-          health: resultHealth
-        });
+    API.getApp(authentication(), this.props.match.params.namespace, this.props.match.params.app)
+      .then(details => {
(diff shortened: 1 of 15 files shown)


jotak:
WIP/DNM: will add one or two tests before merge

hhovsepy:
screenshot from 2019-02-18 18-05-23
screenshot from 2019-02-18 18-05-29
screenshot from 2019-02-18 18-05-42
screenshot from 2019-02-18 18-05-44
screenshot from 2019-02-18 18-05-58
screenshot from 2019-02-18 18-06-02
screenshot from 2019-02-18 18-06-10
screenshot from 2019-02-18 18-06-16
screenshot from 2019-02-18 18-06-21
screenshot from 2019-02-18 18-06-24

src/pages/AppDetails/AppDetailsPage.tsx
```diff
@@ -48,22 +48,19 @@ class AppDetails extends React.Component<RouteComponentProps<AppId>, AppDetailsS
...
-      600
-    );
-
-    Promise.all([promiseDetails, promiseHealth])
-      .then(([resultDetails, resultHealth]) => {
-        this.setState({
-          app: resultDetails.data,
-          health: resultHealth
-        });
+    API.getApp(authentication(), this.props.match.params.namespace, this.props.match.params.app)
```
lucasponce: I see.
It will improve the scenario where there are no sidecars, but when there is, queries are sequentally instead to be in parallel.
Probably not a big deal, but yeah, in that case we will add more latency.

lucasponce: But it's just an observation, this pattern is needed, as it needs the "hasSidecar" as input.

jotak: @lucasponce that's correct, but on the other hand it's now going to set the state right after the details are fetched (without waiting for health) so it's also an improvement from this perspective. But yes, it's not done in parallel anymore.

PR #1004, opened 2019-02-18
KIALI-2410 Replace bars with status column (traffic tab)

israel-hdez:
https://issues.jboss.org/browse/KIALI-2410

* Remove the green/red bars that appear at the end of each row in the detailed traffic tab.
* Add a "status" column showing colored icons, depending on the thresholds.

This is per UXD recommendation: https://github.com/kiali/kiali-design/issues/70#issuecomment-462904571

image

### DIFF
--- a/src/components/Details/DetailedTrafficList.tsx
+++ b/src/components/Details/DetailedTrafficList.tsx
@@ -4,6 +4,7 @@ import * as React from 'react';
 import { Link } from 'react-router-dom';
 import { NodeType, ProtocolTraffic } from '../../types/Graph';
 import { Direction } from '../../types/MetricsOptions';
+import { REQUESTS_THRESHOLDS } from '../../types/Health';
 
 type DetailedTrafficProps = {
   direction: Direction;
@@ -51,21 +52,17 @@ export interface TrafficItem {
   traffic: ProtocolTraffic;
 }
 
-const minichartColumnSizes = {
-  md: 3,
-  sm: 3,
-  xs: 3
+const statusColumnSizes = {
+  md: 1,
+  sm: 1,
+  xs: 1
 };
 const workloadColumnSizes = {
   md: 3,
   sm: 3,
   xs: 3
 };
-const typeColumnSizes = {
-  md: 1,
-  sm: 1,
-  xs: 1
-};
+const typeColumnSizes = statusColumnSizes;
 const trafficColumnSizes = workloadColumnSizes;
 
 class DetailedTrafficList extends React.Component<DetailedTrafficProps> {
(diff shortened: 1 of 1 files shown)


israel-hdez:
@jotak Not sure if I added any bias to the UXD thread (https://github.com/kiali/kiali-design/issues/70#issuecomment-462904571). I think you are in a more neutral position, it's up to you whether to merge.

Instead of labeling the first column as "Health" it's labeled as "Status". I think this may solve the terminology issue.

For me, this look is much cleaner and more understandable than the bars. But you have the last word.

jotak:
Looks good from the screenshot, I'm going to test locally.
BTW is the "unknown" icon already used somewhere else? It could replace our "N/A" text in health, or vice-versa.

abonas:
I probably missed some discussion on this. But why is there a "tab/space" in destination column for some rows? For example reviews vs reviews-v1. is this supposed to indicate hierarchy?

jotak:
@abonas more than a hierarchy I think it shows the sequence of outgoing requests: first "caught" by a service, then a workload. "Caught" by service is symbolic as services don't run any code, but this is the abstract representation shown, as in the graph.

abonas:
> @abonas more than a hierarchy I think it shows the sequence of outgoing requests: first "caught" by a service, then a workload. "Caught" by service is symbolic as services don't run any code, but this is the abstract representation shown, as in the graph.

@jotak typically a grid/tabular view like this does not visualize the dependency/relation/hierarchy between the rows. Each row is on its own. So the way it looks now, it's looks like a formatting bug, although it's not :)
If there is a need to emphasize the relation between the rows, we'll probably need a different or an additional indication. I'll ask on the design issue.

israel-hdez:
> @abonas But why is there a "tab/space" in destination column for some rows?

What @jotak said is the right explanation :)

> @jotak BTW is the "unknown" icon already used somewhere else?

I don't remember other places using it.

abonas:
it seems that compared to the design, there is a "down" arrow missing from the "parent" rows, and that arrow is what indicates hierarchy.

israel-hdez:
@abonas I deliberately omitted the down arrows because I didn't like how it looks when placed next to the service icons. And, although not too important, also because the TableGrid control doesn't seem to support the expand/collapse feature, and the down arrow may give this idea. If there is no problem with this, I will add the down arrows.

The expand/collapse feature can be added without the TableGrid support, it's not hard, but it's more code. So, I could implement it in another PR.

abonas:
> @abonas I deliberately omitted the down arrows because I didn't like how it looks when placed next to the service icons. And, although not too important, also because the TableGrid control doesn't seem to support the expand/collapse feature, and the down arrow may give this idea. If there is no problem with this, I will add the down arrows.
> 
> The expand/collapse feature can be added without the TableGrid support, it's not hard, but it's more code. So, I could implement it in another PR.

I don't have a strong opinion about this and whether more work needs to be invested into something more sophisticated.
The only thing I would like to see improved is how it currently looks like there's a formatting issue, while it's not

israel-hdez:
I'll merge this PR since this is supposed to address the design about the icons vs the bars.
@abonas can we discuss the row indentation in the weekly UXD call? (although I no longer see the entry of the call in the calendar)

PR #1006, opened 2019-02-20
KIALI-2332 Add MTLS percentage to edge side panel

jshaughn:
mTLS is indicated in side panel on selection, regardless of whether 
Full time period:
image

Percentage of the time period:
image

REQUIRES SERVER PR https://github.com/kiali/kiali/pull/827

### DIFF
--- a/src/components/CytoscapeGraph/graphs/GraphStyles.ts
+++ b/src/components/CytoscapeGraph/graphs/GraphStyles.ts
@@ -13,7 +13,7 @@ const EdgeColor = PfColors.Green400;
 const EdgeColorDead = PfColors.Black500;
 const EdgeColorDegraded = PfColors.Orange;
 const EdgeColorFailure = PfColors.Red;
-const EdgeIconLock = '\ue923'; // lock
+const EdgeIconMTLS = icons.istio.mtls.ascii; // lock
 const EdgeTextOutlineColor = PfColors.White;
 const EdgeTextOutlineWidth = '1px';
 const EdgeTextFont = 'Verdana,Arial,Helvetica,sans-serif,FontAwesome,PatternFlyIcons-webfont';
@@ -142,7 +142,7 @@ export class GraphStyles {
       }
 
       if (cyGlobal.showSecurity && ele.data(CyEdge.isMTLS)) {
-        content = EdgeIconLock + ' ' + content;
+        content = EdgeIconMTLS + ' ' + content;
       }
 
       return content;
--- a/src/config/icons.ts
+++ b/src/config/icons.ts
@@ -2,9 +2,10 @@ import deepFreeze from 'deep-freeze';
 
 const mutIcons = {
   istio: {
+    circuitBreaker: { type: 'fa', name: 'bolt', ascii: '\uf0e7 ' },
     missingSidecar: { type: 'pf', name: 'blueprint', ascii: '\ue915 ', color: 'red' },
-    virtualService: { type: 'fa', name: 'code-fork', ascii: '\uf126 ' },
-    circuitBreaker: { type: 'fa', name: 'bolt', ascii: '\uf0e7 ' }
+    mtls: { type: 'pf', name: 'locked', ascii: '\ue923 ' },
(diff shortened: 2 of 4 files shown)


abonas:
Can you explain what does "50% of graph period" means? I don't understand this sentence. thanks

jshaughn:
> Can you explain what does "50% of graph period" means? I don't understand this sentence. thanks

@abonas , This is an attempt to indicate to the user that mTLS was active for only a portion of the data points gathered for the graph.  So, if the graph period represents T1..T2, this is saying that 50% of the datapoints had mTLS enabled, and 50% did not.  Does that make sense?

abonas:
> @abonas , This is an attempt to indicate to the user that mTLS was active for only a portion of the data points gathered for the graph. So, if the graph period represents T1..T2, this is saying that 50% of the datapoints had mTLS enabled, and 50% did not. Does that make sense?

it does make sense, but the phrasing is not clear.

jshaughn:
@abonas How about one of these:

- "mTLS Enabled [50% of the time]"
- "mTLS Enabled [50% of time period]
- "mTLS Enabled [50% of time range]
- "mTLS Enabled [50% of request traffic]
- "mTLS Enabled [partial: 50% of request traffic]

Something else?

abonas:
> * "mTLS Enabled [50% of the time]"
> * "mTLS Enabled [50% of time period]
> * "mTLS Enabled [50% of time range]
> * "mTLS Enabled [50% of request traffic]
> * "mTLS Enabled [partial: 50% of request traffic]

50% of request traffic is not the same like 50% of time period.
which one is more accurate?

jshaughn:
> 50% of request traffic is not the same like 50% of time period. which one is more accurate?

Good point.  So actually, it's the percent of request traffic.  So the better choices would be:
- "mTLS Enabled [50% of request traffic]
- "mTLS Enabled [partial: 50% of request traffic]

xeviknal:
> * "mTLS Enabled [50% of request traffic]

IMO:
I guess that using partial is redundant when % of the traffic doesn't use mTLS. In addition, we use the word `partial` for the mTLS icons on the masthead. Perhaps, it might lead to misunderstandings.

Good feature @jshaughn ! :dancing_men:

jshaughn:
@abonas Here is the updated messaging:
image

abonas:
@jshaughn it's much better!
One thing I'm not sure about is "request traffic". I mean there is request, and response as well, and both are mtls? or only requests? perhaps only saying "traffic" here is suffice?

jshaughn:
@abonas I made the change.

abonas:
> @abonas I made the change.

Is there an updated screenshot? I might have missed that

jshaughn:
@abonas , here you go:

image

pilhuhn:
Is it really traffic? Or time?  If I create a service at noon without TLS, have 100 req/s and then at 1pm enable mTLS and have 20 req/s and then look at it at 2pm, what would be the outcome?
For the scenario when mTLS is not enabled the whole time I'd expect that we show the fraction of time it was (not) enabled. But maybe I misunderstand this.

jshaughn:
> Is it really traffic? Or time?

@pilhuhn , It's requests as the information is based on the request count rate grouped by mTLS vs not-mTLS, for the time period.  If you want to see something time-based a better representation would likely be a sparkline, or something RLE-ish (like the old avail), and it would require actually querying the datapoints so that we could it ordered by time.

pilhuhn:
Hm. I honestly have no idea what users want here (and/or if that really makes sense in such a way and shouldn't be perhaps encoded in a table with timestamp and state change).

*My* expectation is/was to see something like "2h mTLS out of 7", which could be represented as a bar with different colors. Again: I may be completely be wrong and what you have is what folks want to see.

jshaughn:
@pilhuhn I totally agree.  In fact @mwringe and I already pretty much had this discussion.  I'm flipping a penny here and added what could be done reasonable easily in order to give some visibility if in fact they are interested enough to select the edge.  Going any further would be more work, and I think should be motivated by a real-world RFE.

PR #965, opened 2019-01-28
Kiali 1906 Jaeger Integration

aljesusg:
# Jaeger Integration

Require: https://github.com/kiali/kiali/pull/748
This PR add the feature to get the errorTraces of a service

## Issue reference

[KIALI-1906](https://issues.jboss.org/browse/KIALI-1906)

## Tasks 

- [X] Add logfmt library require for search traces by tags
- [X] Created actions in Redux
- [X] Jaeger Toolbar
  - [X] Namespace Selector
  - [X] Service Selector
  - [X] Tags Control
  - [X] LookBack Control
  - [X] RightToolbar
  - [X] Durations and custom range
- [X] Service detail view Traces tab
  - [X] Remove tab if jaeger is not configured
  - [X] Traces container to show errorTraces by default
- [X] Distributed tracing link now show the jaeger Toolbar with the embedded component
- [x] Tests:
  - [x] Jaeger Toolbar Component
     - [x] Namespace Selector
     - [x] Service Selector
     - [X] Tags Control
     - [x] LookBack Control
     - [x] RightToolbar
     - [x] Durations and custom range
  - [x] Redux
    - [x] Jaeger Actions
    - [x] Jaeger Thunk Actions to create the url to search
    - [x] jaeger Reducers
 

### When jaeger is not configured

without_jaeger_url

PR #1008, opened 2019-02-20
KIALI-1906 Support for security and non security environments

aljesusg:
With this we support the integration of Jaeger with Iframe when:

- kiali http and Jaeger http
- kiali https and Jaeger https

When one of them is with security and the other without we'll have the link to the jaeger website.

What do you think @lucasponce, @jotak ?

With this we only need to work in the JAEGER_URL provided by the backend or Makefile


pilhuhn:
Sounds fair.

aljesusg:
Hey @jotak I did the changes requested in the jaeger integration PR, can you review it?

src/config/paths.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Paths
+ */
+
+export enum paths {
```
jotak: Oops, wrong one, enums are PascalCase ;-)
https://github.com/basarat/typescript-book/blob/master/docs/styleguide/styleguide.md#enum

I think this file was good as it was before, also regarding its name (Path.ts) and the enum values (for which we decided to keep upper case)

src/pages/Overview/OverviewStatus.tsx
```diff
@@ -12,7 +11,7 @@ type Props = {
   namespace: string;
   status: Status;
   items: string[];
-  targetPage: Paths;
+  targetPage: string;
```
jotak: Why remove `Paths`? It's more specific typing, so better.

jotak:
@aljesusg there's just a missing comment from the previous PR, not addressed yet:

https://github.com/kiali/kiali-ui/pull/965/files#r258911600 : we don't use `^` for semver in all other modules

aljesusg:
> @aljesusg there's just a missing comment from the previous PR, not addressed yet:
> 
> https://github.com/kiali/kiali-ui/pull/965/files#r258911600 : we don't use `^` for semver in all other modules

Changed Thanks

jotak:
So I tested these scenarios:
- Kiali on https and Jaeger on http: the tab opens Jaeger UI in a new window
- Kiali and Jaeger both on https: the tab shows Jaeger embedded

So I think this is what's expected.
The only suggestion I have would be to add an "external link" icon in the case it's opened in a new window (fa-external-link), so that there's no surprise for the user.

aljesusg:
Done @jotak 
external_link

My concern is about the link of Distributed Tracing Link that opens jaeger in a new tab too

jotak:
@aljesusg yes there's the same problem with the distributed tracing link. But at least, it doesn't affect the nominal case where both jaeger and kiali are under https. I would say it's good enough.

hhovsepy:
When refreshing the Service Details page, it first appears "Traces" tab with external link, then within an 1/2 sec, switches to "Error Traces" tab.

aljesusg:
This is by the redux state I think let me check it @hhovsepy

aljesusg:
@hhovsepy I'll fix this problems in https://github.com/kiali/kiali-ui/pull/1025 see comment https://github.com/kiali/kiali-ui/pull/1025#issuecomment-467769657 Thanks

PR #1009, opened 2019-02-20
KIALI-2428 Fix issue with updating durations

jshaughn:
- Callers must handle the fact that ToolbarDropdown returns a string when a selection is made
  - add some stronger typing to make this clear


jmazzitelli:
drop down menus seem to work

PR #1013, opened 2019-02-21
KIALI-2329 Support multiple matchings on Wizard

lucasponce:
This PR contains:
- It refactors IstioWizard to add new use cases easily.
- It updates Weighted Routing with UX recommendations related lockings, some icons on the way.
- It adds a new use case "Matching Routing" to cover scenarios with filtering by headers.
- It syncs TLS and prepopulate ISTIO_MUTUAL when mTLS is enabled.

** Issue reference **

https://issues.jboss.org/browse/KIALI-2329

image

image


pilhuhn:
Does it warn when the catch-all rule is not the last in the match-wizard?

hhovsepy:
@pilhuhn does not let to create, and warns:
screenshot from 2019-02-21 14-37-37

hhovsepy:
Moved catch-all to last one:
screenshot from 2019-02-21 14-42-12

src/components/IstioWizards/MatchingRouting.tsx
```diff
@@ -0,0 +1,511 @@
...
+      );
+    }
+    return (
+      <div>
+        <ListView>{ruleItems}</ListView>
+      </div>
+    );
+  };
+
+  render() {
```
xeviknal: This component is pretty huge. Do you think it should be splitted in different ones? Break it down in smaller pieces?

src/components/IstioWizards/IstioWizardDropdown.tsx
```diff
@@ -48,10 +51,10 @@ class IstioWizardDropdown extends React.Component<Props, State> {
   };
 
   onAction = (key: string) => {
-    if (key === 'create_traffic_routing') {
-      this.setState({ showWizard: true });
+    if (key === WIZARD_WEIGHTED_ROUTING || key === WIZARD_MATCHING_ROUTING) {
+      this.setState({ showWizard: true, wizardType: key });
     }
-    if (key === 'delete_traffic_routing') {
+    if (key === DELETE_TRAFFIC_ROUTING) {
```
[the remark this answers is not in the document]
lucasponce: Yes, it can be elseif here.

src/components/IstioWizards/MatchingRouting.tsx
lucasponce: Probably in a next interation.
At the moment I want to maintain a first version with all state together, and then split it in minor pieces.

src/components/IstioWizards/WeightedRouting.tsx
```diff
@@ -0,0 +1,186 @@
...
+                    onLock={locked => this.onLock(workload.name, locked)}
+                  />
+                }
+              />
+            );
+          })}
+        </ListView>
+        {!isValid && <div className={validationStyle}>The sum of all weights must be 100 %</div>}
+      </div>
+    );
```
aljesusg: Replace 
```html
<div></div>
```
 with 
```html
<></>
```
 to avoid depth dom

src/components/IstioWizards/IstioWizard.tsx
```diff
@@ -99,15 +85,44 @@ class IstioWizard extends React.Component<Props, State> {
...
+      .forEach(match => {
+        // match follows format:  headers [<header-name>] <op> <value>
+        const i0 = match.indexOf('[');
+        const j0 = match.indexOf(']');
+        const headerName = match.substring(i0 + 1, j0);
+        const i1 = match.indexOf(' ', j0 + 1);
+        const j1 = match.indexOf(' ', i1 + 1);
+        const op = match.substring(i1 + 1, j1);
+        const value = match.substring(j1 + 1);
+        matchHeaders.headers![headerName] = { [op]: value };
```
aljesusg: My proposal here is get values with regexlike something like this. What do you think?
```ts
const reg = /headers \[([a-z0-9\-]+)\] ([a-z0-9\-]+) ([a-z0-9\-]+)/;
const result =match.match(namespaceRegex) || [];
const headerName = result[1]
const op = result[2]
const value = result[3]
```

src/components/IstioWizards/IstioWizard.tsx
```diff
@@ -99,15 +85,44 @@ class IstioWizard extends React.Component<Props, State> {
...
+    matches
+      .filter(match => !match.startsWith('headers'))
+      .forEach(match => {
+        // match follows format: <name> <op> <value>
+        const i = match.indexOf(' ');
+        const j = match.indexOf(' ', i + 1);
+        const name = match.substring(0, i);
+        const op = match.substring(i, j);
+        const value = match.substring(j);
+        matchRequests.push({
```
aljesusg: The same here than above.

src/components/IstioWizards/IstioWizard.tsx
```diff
@@ -166,18 +171,59 @@ class IstioWizard extends React.Component<Props, State> {
...
+          }
+          httpRoute.route = [destW];
+          if (rule.matches.length > 0) {
+            httpRoute.match = this.buildHTTPMatchRequest(rule.matches);
+          }
+          return httpRoute;
+        })
+      };
+    }
+
```
aljesusg: Could we have more types in the future? The value of hosts are the same and maybe we can change this by a switch case (2º conditional if is not necessary be evaluate if the type is WIZARD_WEIGHTED_ROUTING) What do you think?

src/components/IstioWizards/IstioWizardDropdown.tsx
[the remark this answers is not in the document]
aljesusg: I prefer switch case and avoid else if else if else if....in future if there may be more types

src/components/IstioWizards/IstioWizard.tsx
```diff
@@ -166,18 +171,59 @@ class IstioWizard extends React.Component<Props, State> {
...
       }
-    };
+      if (this.state.loadBalancer !== NONE) {
+        wizardDR.spec.trafficPolicy.loadBalancer = {
+          simple: this.state.loadBalancer
+        };
+      }
+    }
+
     return [wizardDR, wizardVS];
```
aljesusg: This could be

```ts
wizardDR.spec.trafficPolicy = {};
if (this.state.mtlsMode !== NONE) {
  wizardDR.spec.trafficPolicy.tls = {
          mode: this.state.mtlsMode
        };
}
if (this.state.loadBalancer !== NONE) {
  wizardDR.spec.trafficPolicy.loadBalancer = {
          simple: this.state.loadBalancer
        };
}
```

or we can't we have the trafficPolicy empty ? the proposal is avoid in the worst of the cases 3 conditionals.

src/components/IstioWizards/WeightedRouting.tsx
```diff
@@ -0,0 +1,186 @@
...
+        for (let i = 0; i < prevState.workloads.length; i++) {
+          if (prevState.workloads[i].name === workloadName) {
+            prevState.workloads[i].weight = newWeight;
+            maxWeight -= newWeight;
+          } else {
+            // Only adjust those nodes that are not locked
+            if (!prevState.workloads[i].locked) {
+              nodeId.push(i);
+            }
+          }
```
aljesusg:
```ts
} else  if (!prevState.workloads[i].locked) {
              nodeId.push(i);
  }
```

src/components/IstioWizards/IstioWizardDropdown.tsx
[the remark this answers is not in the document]
lucasponce: Yep, will send a folloup commit with this suggestion.

src/components/IstioWizards/WeightedRouting.tsx
lucasponce: +1

src/components/IstioWizards/IstioWizard.tsx
lucasponce: Regex are interesting, but main blocker/pain I see is that value in third param can be also a regex, so, having a regex as parameter could lead in a more complex value.
As empty spaces are fixed and internal to that variable I found more easy to "parse" for this particular use case.

src/components/IstioWizards/IstioWizard.tsx
lucasponce: Yeap. Good suggestion, but as commented, <value> can be also a regex and then it may lead in more complexity.

src/components/IstioWizards/IstioWizard.tsx
lucasponce: Will move to a switch case.

src/components/IstioWizards/IstioWizard.tsx
aljesusg: ok !!! I understand

src/components/IstioWizards/IstioWizard.tsx
lucasponce: Yeah, the logic is to not have the trafficPolicy in the object.
With your suggestion, it simplifies the code but the 'trafficPolicy' will be always present.

src/components/IstioWizards/WeightedRouting.tsx
lucasponce: +1

src/components/IstioWizards/IstioWizard.tsx
aljesusg: ok

lucasponce:
Most of feedback provided by UX and @xeviknal and @aljesusg addressed.

hhovsepy:
100% is hidden behind the arrows:
screenshot from 2019-02-22 11-56-50

lucasponce:
thanks for the feedback @hhovsepy 
Does it happen in Firefox only ?

lucasponce:
Yeap, pretty ugly here as well:
image

lucasponce:
@hhovsepy I'm working on a fix, will update the PR with a solution.

lucasponce:
@hhovsepy This is from Firefox
image

And also we address last UX suggestion that it was not covered on this PR.

hhovsepy:
100 is displayed partly on FF 65
screenshot from 2019-02-22 15-29-19

lucasponce:
> 100 is displayed partly on FF 65

Ouch, it should be an easy change, let me to adjust the CSS for that.

lucasponce:
image

@hhovsepy yeah, I had Firefox 65 with 125% and the width was ok, but yeah, it failed on 100%. It should be fixed now.

hhovsepy:
Verified:
screenshot from 2019-02-22 16-01-39

PR #1017, opened 2019-02-22
KIALI-2437 Show "n/a" when there is an error reading the error tr…

josejulio:
…ace count

** Describe the change **

Shows `n/a` instead of `-1` in the UI when there was an error reading the "error trace" count.

I'm not sure if we want to give more context about the error or if is fine as is for now.

** Screenshot **

peek 25-02-2019 09-54

### DIFF
--- a/src/pages/ServiceDetails/ServiceDetailsPage.tsx
+++ b/src/pages/ServiceDetails/ServiceDetailsPage.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { RouteComponentProps } from 'react-router-dom';
-import { Nav, NavItem, TabContainer, TabContent, TabPane, Icon } from 'patternfly-react';
+import { Nav, NavItem, TabContainer, TabContent, TabPane, Icon, OverlayTrigger, Tooltip } from 'patternfly-react';
 import ServiceId from '../../types/ServiceId';
 import * as API from '../../services/Api';
 import * as MessageCenter from '../../utils/MessageCenter';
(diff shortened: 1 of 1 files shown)


jotak:
"Unknown" is much better than "-1" but I fear it's still not so much meaningful. What about a red failure icon?
The tooltip is great

BTW, in case there is an error, is the tab still accessible?

cc @aljesusg @abonas

pilhuhn:
Yes, unknown is not much better than -1. What about "n/a" ? Would need UX input.

josejulio:
Updated to show `n/a`

@pilhuhn  @jotak @abonas

aljesusg:
No, Please this was something that I forget to change. When the value is -1 the traces should show the normal traces of the services because that means that there is a problem getting traces of jaeger

src/pages/ServiceDetails/ServiceDetailsPage.tsx
```diff
@@ -177,7 +177,23 @@ class ServiceDetails extends React.Component<ServiceDetailsProps, ServiceDetails
...
+                            </Tooltip>
+                          }
+                          placement="top"
+                          trigger={['hover', 'focus']}
+                          rootClose={false}
+                        >
+                          <span>n/a</span>
+                        </OverlayTrigger>
+                      )}
                       {errorTraces > 0 && (
```
aljesusg: Hi @josejulio , I was working in another PR to fix this that I forgot. when the value is -1 the Error traces tab should be a tab with name "Traces" ans show all the traces of this service. -1 means that there is a problem getting traces in Jaeger this could mean a timeout. When there is not JaegerURL configured the traces tab is hide.

josejulio:
Fixed with #1025

PR #1020, opened 2019-02-22
KIALI-2402 Fix link to gateway in another namespace

burmanm:
** Describe the change **

 Parse gateway hostname to FQDN and then generate a correct link to the YAML, without assuming it is in the same namespace.

** Issue reference **

KIALI-2402


burmanm:
parseHost could be in some more generic place (I noticed some service links required parsing also), but I couldn't find a proper place.

hhovsepy:
Verified:
1. Link to the gateway.
2. FQDN in "name.namespace" format.

screenshot from 2019-02-26 16-02-11
screenshot from 2019-02-26 16-02-35

xeviknal:
@burmanm what about adding it into src/utils? There are different files with logic conceptually grouped. For example, CancelablePromises, RateIntervals and so on.

PR #1028, opened 2019-02-27
KIALI-2380 Session handling using cookies

israel-hdez:
https://issues.jboss.org/browse/KIALI-2380

Removes the session token from Redux, letting the browser and the server handle the session through cookies.

The session cookie is not accessible from JS/TS.

Back-end PR: https://github.com/kiali/kiali/pull/877


israel-hdez:
@mwringe Since you took over the OAuth work, I hope you can check this PR (and also the server side PR).

mwringe:
What is the point of the initialisation screen?

israel-hdez:
@mwringe The initialization screen is shown while retrieving the auth strategy and while validating if there is a logged-in user (validating the cookie). 

Probably, it doesn't makes sense if there is no logged in user. But when there is a logged in user, it prevents from showing shortly the login form and suddenly switching to the console.

src/app/InitializingScreen.tsx
```diff
@@ -0,0 +1,51 @@
...
+  LoginPageContainer,
+  LoginPageFooter,
+  LoginPageHeader,
+  Spinner
+} from 'patternfly-react';
+import { isKioskMode } from '../utils/SearchParamUtils';
+
+const kialiTitle = require('../assets/img/logo-login.svg');
+
+const InitializingScreen: React.FC<{ errorMsg?: string }> = (props: { errorMsg?: string }) => {
```
mtho11: It took me a minute to realize what the difference between React.FC<> and React.SFC<>. Looks like SFC is now deprecated in favor of FC: https://twitter.com/dan_abramov/status/1057625147216220162?lang=en

src/components/IstioWizards/IstioWizard.tsx
```diff
@@ -234,18 +233,8 @@ class IstioWizard extends React.Component<Props, State> {
...
-      'destinationrules',
-      JSON.stringify(dr)
-    );
-    const createVS = API.createIstioConfigDetail(
-      authentication(),
-      this.props.namespace,
-      'virtualservices',
-      JSON.stringify(vr)
-    );
+    const createDR = API.createIstioConfigDetail(this.props.namespace, 'destinationrules', JSON.stringify(dr));
```
mtho11: I guess team Jupiter should be aware of the API changes.
CC @lucasponce @aljesusg @xeviknal @jotak

mwringe:
> Probably, it doesn't makes sense if there is no logged in user. But when there is a logged in user, it prevents from showing shortly the login form and suddenly switching to the console.

@israel-hdez Except now I am seeing this screen in addition to the login form :/

israel-hdez:
@mtho11

> I'm curious why the formatting changes to index.html? Prettier didn't change

I'm not sure. I guess because it wasn't being checked since it is supposed that only changed files are verified. Probably the formatting was triggered because I removed the `class` in the `<html>` tag.

src/components/IstioWizards/IstioWizard.tsx
lucasponce: It doesn't change the logic.
I assume that the authentication() is refactored.
@hhovsepy perhaps you could run some regression test on this PR ?
Thanks a lot.

PR #1029, opened 2019-02-28
Trivial: Filename changes to match our rules.

mtho11:
** Describe the change **
A simple change to make the filenames consistent with the rest of the app.


jshaughn:
Can you ensure the naming of certain files is in line with the style guide.  I'm fairly sure that @jotak renamed ServerConfig.ts to serverConfig.ts and this PR changes it back.   I don't want us to flip-flop.

jotak:
Indeed, I've written down some rules about that but it can change if we think it has to.
In most Typescript guidelines (such as Basarat's style guide) it's recommended to use camelCase, unlike what we do.
But in order to be more consistent with our codebase I wrote that rule : when primary purpose of a file is to export something, such as MyReactComponent, the file takes the name of that component, so it can be PascalCase.

So, what do you think? Can we keep this rule, or not? (in which case the style guide has to be amended). Also there's other files that are currently lower case, such as routes.ts, probably some others.

mtho11:
Since most all of our codebase is PascalCase, I think we should run with that. And @jotak has documented that rule so I think we are in good standing with it.
I'll look through and see if anything else needs changing and add it to the PR.

mtho11:
@jotak would you mind reviewing, just cleaning up on old stuff.

PR #1032, opened 2019-03-01
KIALI-2321 Better support smaller screen widths

jshaughn:
@jmazzitelli , for now just let me know if this layout prevents your graph toolbar from wrapping in your Jira environment.  Make sure to test with longest menu values.

### DIFF
--- a/src/components/GraphFilter/GraphFilter.tsx
+++ b/src/components/GraphFilter/GraphFilter.tsx
@@ -23,6 +23,7 @@ import Namespace, { namespacesToString, namespacesFromString } from '../../types
 import { NamespaceActions } from '../../actions/NamespaceAction';
 import { GraphActions } from '../../actions/GraphActions';
 import { KialiAppAction } from '../../actions/KialiAppAction';
+import { AlignRightStyle, ThinStyle } from '../../components/Filters/FilterStyles';
 
 type ReduxProps = {
   activeNamespaces: Namespace[];
@@ -41,19 +42,10 @@ type GraphFilterProps = ReduxProps & {
   onRefresh: () => void;
 };
 
-// align under the Graph breadcrumb
-const breadcrumbStyle = style({
-  marginLeft: '-30px',
-  marginRight: '0px',
-  paddingLeft: '0px',
-  paddingRight: '0px'
+// align with separator start / Graph breadcrumb
+const alignLeftStyle = style({
+  marginLeft: '-30px'
 });
-
-// reduce toolbar padding from 20px to 10px to save space
-const thinGroupStyle = {
-  paddingLeft: '10px',
-  paddingRight: '10px'
-};
 
 export class GraphFilter extends React.PureComponent<GraphFilterProps> {
   /**
@@ -135,7 +127,7 @@ export class GraphFilter extends React.PureComponent<GraphFilterProps> {
     return (
       <>
         <Toolbar>
-          <FormGroup className={breadcrumbStyle} style={{ ...thinGroupStyle }}>
+          <FormGroup className={alignLeftStyle} style={{ ...ThinStyle }}>
             {this.props.node ? (
               <Button onClick={this.handleNamespaceReturn}>Back to full {GraphFilter.GRAPH_TYPES[graphTypeKey]}</Button>
             ) : (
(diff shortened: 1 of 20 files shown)


jmazzitelli:
been 5 days, 2 days since I approved - what's the hold up?

abonas:
if this is blocking other things, then you can merge and then add additional improvements if needed after UX discussion. But certainly, we should consult UX on this.

src/components/GraphFilter/GraphFilter.tsx
```diff
@@ -131,15 +135,19 @@ export class GraphFilter extends React.PureComponent<GraphFilterProps> {
...
-          {this.props.node && (
-            <FormGroup className={zeroPaddingLeft}>
-              <Button className={namespaceStyle} onClick={this.handleNamespaceReturn}>
-                Back to Full Graph...
-              </Button>
-            </FormGroup>
-          )}
-          <FormGroup className={zeroPaddingLeft}>
-            <GraphSettingsContainer edgeLabelMode={this.props.edgeLabelMode} graphType={this.props.graphType} />
+          <FormGroup className={breadcrumbStyle} style={{ ...thinGroupStyle }}>
```
mtho11: Using both the **className** and the **style** is probably not the best way to do this.

src/components/GraphFilter/GraphFind.tsx
```diff
@@ -36,6 +36,16 @@ type ParsedExpression = {
   selector: string;
 };
 
+const inputWidth = {
```
mtho11: These should be **TypeStyle** functions.

src/components/GraphFilter/GraphSettings.tsx
```diff
@@ -175,16 +175,17 @@ class GraphSettings extends React.PureComponent<GraphSettingsProps> {
       </Popover>
     );
 
-    const alignWithGraphHeaderStyle = style({ marginLeft: -40, paddingRight: '0.8em' });
+    // const alignWithGraphHeaderStyle = style({ marginLeft: -60, paddingRight: '0.5em' });
```
mtho11: Probably didn't mean to leave the commented code.

src/components/GraphFilter/GraphSettings.tsx
```diff
@@ -175,16 +175,17 @@ class GraphSettings extends React.PureComponent<GraphSettingsProps> {
...
 
     return (
-      <span className={alignWithGraphHeaderStyle}>
-        <OverlayTrigger overlay={graphSettingsPopover} placement="bottom" trigger={['click']} rootClose={true}>
-          <Button id="graph_settings">
-            Display <Icon name="angle-down" />
-          </Button>
-        </OverlayTrigger>
-      </span>
+      // <span className={alignWithGraphHeaderStyle}>
```
mtho11: More unused code.

src/components/GraphFilter/GraphFilter.tsx
jshaughn: So, I'm not too savvy with this stuff.  But when I tried to combine it into one className style, or all style properties, it just didn't seem to do what I want.  So in general I'm not using styles, but I probably should.  Instead I'm using this "style" objects and a spread operator.  I'd be happy to improve this but I may need help, or a PR to this PR.

jshaughn:
@mattmahoneyrh I think this PR is ready for QE acceptance.  Here is a before/after shot of the graph toolbar, and a list of changes to look for:

Before:
image
After:
image

- Applied to Graph page only:
  - Do not wrap toolbar on 1280x1084 resolution (mazz can verify this)
    - if true for graph it will automatically be true for other main pages as the have more available space
  - Remove the clear icon (the X) for find/hide when the input is empty.
  - Reduce the Find/Hide input box widths
  - Swap the graph type and display buttons for a more logical ordering.
  - Update the Tour 
    - place the arrow closer to the widgets
    - add a tour stop at Find/Hide
- Applied to all of the main pages:
  - Remove margin on left and right.  In other words, left align widget #1 with breadcrumb.  Right align refresh button with toollbar edge.
  - Reduce padding between input groups (the space around the vertical bars) from 20px to 10px
- Applied to all pages using duration/refresh dropdowns
  - Remove the "Display" title for Duration/Refresh dropdowns and add tooltips.
  - Change "min" and "sec" to more standard duration notations in the dropdowns.
    - For example, "Last 1m", "Last 1h", Last "1d", "Every 15s", "Every 5m".

PR #1033, opened 2019-03-01
KIALI-2279 Consistent terminology and location for duration/refresh

jshaughn:
- Place widgets on the right side consistently
- Use "Fetch" terminology consistently.  "Fetch" over "Fetching" or "Displaying" because  it is:
  - clear
  - short (saves width/prevents wrapping)
  - a command rather than an action
- also, move RefreshContainer into Refresh component

Graph:
image
Overview:
image
Metrics:
image


abonas:
I'm not sure that "fetch" is a good term here. Why not "display" ? The control is essentially about users asking to *display* data for X time.

abonas:
@serenamarie125 any thoughts here?

jshaughn:
@abonas. @serenamarie125  I guess I thought "Fetch" was more indicative of what the setting was about but "Display" is fine.  To be honest, I'd prefer to get rid of the label completely and replace it with a tooltip or info icon. This would save space as well as give us a bit more text to describe these controls.

Note that there is another Jira about how the graph toolbar wraps on lower resolutions, so there is some work in progress there to see if we can reduce the width.  I'm playing with that as well.

jshaughn:
@abonas I'll change it to "Display" for this PR although let's consider removing the label as part of Kiali-2321 (the other Jira I mentioned above).

abonas:
LGTM

PR #1034, opened 2019-03-04
KIALI-2492 Fix changing duration in metrics page

jotak:
JIRA: https://issues.jboss.org/browse/KIALI-2492

We should not expect changes to be caught on re-rendering, we can deal with it before (as soon as the change hook is called)


jotak:
cc @pilhuhn this is fixing the bug you've seen

hhovsepy:
Verified

pilhuhn:
@jotak Yes, this works now - thanks for the quick fix.

PR #1036, opened 2019-03-04
KIALI-2484 fix console log error for NA health use case

jshaughn:
The call to get health information may not return health for every node. There
are valid use cases for no health (like no traffic for the health calc period).
Change to log no health at debug level and only log an error if the call to the
backend fails.


jshaughn:
@jotak Looking at the code change I made I think it's a good change either way, as we really shouldn't log an error when the call actually succeeds.  But I did check and it's still an issue.  So I will merge this change.

PR #1037, opened 2019-03-04
KIALI-2379 Add duration control in traffic pages

israel-hdez:
https://issues.jboss.org/browse/KIALI-2379

This is adding a dropdown to the traffic pages:

image

### DIFF
--- a/src/pages/AppDetails/AppDetailsPage.tsx
+++ b/src/pages/AppDetails/AppDetailsPage.tsx
@@ -14,6 +14,7 @@ import BreadcrumbView from '../../components/BreadcrumbView/BreadcrumbView';
 import { GraphDefinition, GraphType, NodeParamsType, NodeType } from '../../types/Graph';
 import { fetchTrafficDetails } from '../../helpers/TrafficDetailsHelper';
 import TrafficDetails from '../../components/Metrics/TrafficDetails';
+import MetricsHelper from '../../components/Metrics/Helper';
 
 type AppDetailsState = {
   app: App;
@@ -98,7 +99,7 @@ class AppDetails extends React.Component<RouteComponentProps<AppId>, AppDetailsS
       version: ''
     };
     const restParams = {
-      duration: `${TrafficDetails.defaultDuration}s`,
+      duration: `${MetricsHelper.initDuration({}).duration!}s`,
       graphType: GraphType.APP,
       injectServiceNodes: true,
       appenders: 'deadNode'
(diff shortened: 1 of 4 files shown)


src/pages/AppDetails/AppDetailsPage.tsx
```diff
@@ -98,7 +99,7 @@ class AppDetails extends React.Component<RouteComponentProps<AppId>, AppDetailsS
       version: ''
     };
     const restParams = {
-      duration: `${TrafficDetails.defaultDuration}s`,
+      duration: `${MetricsHelper.initDuration({}).duration!}s`,
```
jotak: It is a bit weird to use `MetricsHelper.initDuration({}).duration!` here, as this function is really useful to compute some metrics related stuff such as the rate interval that depends on duration, which is of no use here. I think the only piece of code that is interesting to re-use is `MetricsDuration.initialDuration()` that takes duration from URL and/or session storage.

Unless there's something I didn't see, I think it's better to change, to have less coupling with the Metrics-specific stuff.

src/pages/AppDetails/AppDetailsPage.tsx
```diff
@@ -157,11 +158,12 @@ class AppDetails extends React.Component<RouteComponentProps<AppId>, AppDetailsS
               </TabPane>
               <TabPane eventKey="traffic">
                 <TrafficDetails
-                  duration={TrafficDetails.defaultDuration}
+                  duration={MetricsHelper.initDuration({}).duration!}
```
jotak: same here

src/pages/WorkloadDetails/WorkloadDetailsPage.tsx
```diff
@@ -115,7 +116,7 @@ class WorkloadDetails extends React.Component<RouteComponentProps<WorkloadId>, W
       version: ''
     };
     const restParams = {
-      duration: `${TrafficDetails.defaultDuration}s`,
+      duration: `${MetricsHelper.initDuration({}).duration!}s`,
```
jotak: same

src/pages/WorkloadDetails/WorkloadDetailsPage.tsx
```diff
@@ -216,11 +217,12 @@ class WorkloadDetails extends React.Component<RouteComponentProps<WorkloadId>, W
               </TabPane>
               <TabPane eventKey={'traffic'}>
                 <TrafficDetails
-                  duration={TrafficDetails.defaultDuration}
+                  duration={MetricsHelper.initDuration({}).duration!}
```
jotak: same

abonas:
is there a need to specify in the titles the same duration? 
right now there is the dropdown "display last 10 mins" but also "inbound last 10 minutes" - is there a situation where those won't match each other so it has to be specified in the title too?

israel-hdez:
@abonas No real need. I was keeping the original design. But, with the new dropdown, I think it's redundant and the text in the titles can go away

src/pages/AppDetails/AppDetailsPage.tsx
israel-hdez: You are right. I fixed all references.

israel-hdez:
@abonas The text in titles are removed!

abonas:
LGTM

PR #1038, opened 2019-03-05
KIALI-2463 Block Wizard Create button until it has finished

lucasponce:
Minor change.
When onCreate is invoked, disable the "Create" button to avoid multiple clicks.

https://issues.jboss.org/browse/KIALI-2463


hhovsepy:
Can you block the Delete button as well?

lucasponce:
> Can you block the Delete button as well?

Good point. Sure.

lucasponce:
> Can you block the Delete button as well?

I reviewed the component and I guess I need to do a workaround.
I think you mean to disable the confirm delete dialog, isn't it ?

I guess that's not possible without hacking the MessageDialog, but what it can be done is to disable the action "Delete All" entry until the delete operation has finished.

lucasponce:
@hhovsepy @jotak I've rebased and addressed Hayk's comment.
Let me know how it looks like.

PR #1041, opened 2019-03-05
KIALI-2476 Improve error messages for Graph Find/Hide

jshaughn:
- give better indication of problematic expression element
- make more generic (apply better to both find AND hide)
- remove full expression from error messages in anticipation of showing
  messages co-located to hide/find inputs (not in message center)

image

### DIFF
--- a/src/components/GraphFilter/GraphFind.tsx
+++ b/src/components/GraphFilter/GraphFind.tsx
@@ -233,7 +233,7 @@ export class GraphFind extends React.PureComponent<GraphFindProps> {
     const conjunctive = validVal.includes(' AND ');
     const disjunctive = validVal.includes(' OR ');
     if (conjunctive && disjunctive) {
-      MessageCenterUtils.add(`Expression can not contain both 'and' and 'or'`);
+      MessageCenterUtils.add(`Expression can not contain both 'AND' and 'OR'`);
       return undefined;
     }
     const separator = disjunctive ? ',' : '';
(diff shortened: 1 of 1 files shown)


src/components/GraphFilter/GraphFind.tsx
```diff
@@ -233,7 +233,7 @@ export class GraphFind extends React.PureComponent<GraphFindProps> {
     const conjunctive = validVal.includes(' AND ');
     const disjunctive = validVal.includes(' OR ');
     if (conjunctive && disjunctive) {
-      MessageCenterUtils.add(`Find value can not contain both 'and' and 'or'.`);
+      MessageCenterUtils.add(`Expression can not contain both 'and' and 'or'`);
```
abonas: AND and OR should be in capital letters (consistency with other messages below)

src/components/GraphFilter/GraphFind.tsx
```diff
@@ -360,10 +360,10 @@ export class GraphFind extends React.PureComponent<GraphFindProps> {
       case 'name': {
         const isNegation = op.startsWith('!');
         if (disjunctive && isNegation) {
-          MessageCenterUtils.add(`Find values do not allow OR expressions with "not find by name": [${expression}]`);
+          MessageCenterUtils.add(`Can not use 'OR' with 'not find by name'`);
```
abonas: to be honest, I don't understand this message. not the previous version and not the current.

PR #1043, opened 2019-03-05
KIALI-2479 Block console rendering until all configs are retrieved

israel-hdez:
This is adding an AuthenticationController component. This component
becomes responsible on determining the moment when it's OK to render the
console.

Code in charge of retrieving server configs is integrated into the
AuthenticationController component.

https://issues.jboss.org/browse/KIALI-2479
https://issues.jboss.org/browse/KIALI-2502


israel-hdez:
@jshaughn @jotak This is adding an AuthenticationController to "orchestrate" the login flow and block the rendering until all configs are retrieved from the server (prometheus, grafana, status). Once all data is fetched, it allows rendering of the console.

It's not yet finished: error handling is pending, and probably I'm abusing of the "Initialization" screen (the login screen should probably show the messages to avoid too much screen switching). But first, I want to check if you like this approach so I proceed to finish the details (so, not yet ready for a detailed review).

This is not moving away from redux, so the PR is not "big". If  you are OK with this approach, we can either stay with redux or I can incorportate @jotak 's changes from #1012 to move the promConfigs out of redux.

pilhuhn:
Will this only "retrieve" the config from the backend? Or also verify that the backends exist?
E.g. Kiali can work without Grafana being present, so we may fetch its url from the backend, but should not block login on Grafana being available.

israel-hdez:
@pilhuhn It's only retrieving config. So, if server side returns blank configs, it won't block rendering.

But... I'm checking how the server side works and it seems there are some cases where an error is thrown and it should be valid to continue the login. I'll fix that once @jotak and @jshaughn give their :+1: to proceed to finish this work.

Thanks for the observation!

jshaughn:
@israel-hdez , I have no problem with a more centralized authentication mechanism.  I'm for anything that:

1. Blocks rendering until auth is established and serverConfig is retrieved
2. Is graceful about auth failures and connection failures
3. Removes serverConfig from redux
4. Passes the clean-machine test, the F5 test, and the session expired test

I know this is WIP but if you name the file AuthenticationController please also name the class AuthenticationController and the container AuthenticationControllerContainer.

mtho11:
Yes, this seems like a good idea to me.

israel-hdez:
Ok, this is ready for review.

I merged #1012 to integrate all @jotak valuable work. 
Server configs are no longer in redux and are fetched before rendering of the main console.

israel-hdez:
This also will fix https://issues.jboss.org/browse/KIALI-2502

src/app/AuthenticationControllerContainer.tsx
```diff
@@ -0,0 +1,144 @@
...
+interface AuthenticationFlowProps {
+  authenticated: boolean;
+  protectedAreaComponent: React.ReactNode;
+  publicAreaComponent: React.ReactNode;
+  setGrafanaInfo: (grafanaInfo: GrafanaInfo) => void;
+  setServerStatus: (serverStatus: ServerStatus) => void;
+}
+
+interface AuthenticationFlowState {
+  stage: 'login' | 'post-login' | 'logged-in';
```
mtho11: This would have better type-safety and reusability if it were an **enum**.

jotak: out of curiosity: why? I've always thought union types were actually a nice alternative to enums (it can be declared explicitly as a type as well)

src/app/AuthenticationControllerContainer.tsx
```diff
@@ -0,0 +1,144 @@
...
+        .catch(error => {
+          MessageCenter.add(API.getErrorMsg('Error fetching Grafana Info.', error), 'default', MessageType.WARNING);
+        });
+
+      const configs = await Promise.all([API.getServerConfig(), getStatusPromise, getGrafanaInfoPromise]);
+      setServerConfig(configs[0].data);
+
+      this.setState({ stage: 'logged-in' });
+    } catch (err) {
+      this.setState({ isPostLoginError: true });
```
jotak: While testing and opening Kiali the first time, I ran in that situation where `isPostLoginError` was true and it displayed the error message hardcoded above. I'm curious what would have been the caught `err` here. Maybe it should be logged? (console.error)

src/app/AuthenticationControllerContainer.tsx
israel-hdez: I would like to keep it as is for now, since state is kind of private. WDYT? We could convert it to a type if the stage moves to the properties.

israel-hdez:
@jshaughn Ok, I'll check the renamings and the Debug dialog.

About the intermediate page in F5, what I can do is to present a blank page while data is fetched. Perhaps it could be good to have a timeout to show the intermediate page in case any of the endpoints takes too long to reply, but for now I'll only make it show a blank page on F5.

jshaughn:
> what I can do is to present a blank page while data is fetched

Maybe that would be better.  I guess this isn't a blocker, though.  F5 typically does have some sort of flash or load so I guess it'as not too uncommon.  If it's easier, or if you don't like the blank page, we can leave it as is.

src/app/AuthenticationControllerContainer.tsx
mtho11: If it's local I guess it's fine.

pilhuhn:
A blank blank screen does not sound appealing. At least it should say something along "loading stuff"

jshaughn:
> A blank blank screen does not sound appealing

+1, I changed my mind.  After looking at various other sites I think seeing the initialization screen on F5 is fine.

PR #1044, opened 2019-03-05
KIALI-2475 Move graph find/hide errors from message center to inline

jshaughn:


jmazzitelli:
It's "beautiful" 

image

PR #1049, opened 2019-03-07
KIALI-2496 Disabled Slider should not show +/- buttons

lucasponce:
Hide pins when there are only two workloads
Align text into center
Minor changes on style buttons

https://issues.jboss.org/browse/KIALI-2496

image

image

### DIFF
--- a/src/components/IstioWizards/IstioWizard.tsx
+++ b/src/components/IstioWizards/IstioWizard.tsx
@@ -30,8 +30,8 @@ type State = {
   rules: Rule[];
   valid: boolean;
   mtlsMode: string;
+  tlsModified: boolean;
   loadBalancer: string;
-  modified: boolean;
 };
 
 export const WIZARD_WEIGHTED_ROUTING = 'create_weighted_routing';
@@ -53,8 +53,8 @@ class IstioWizard extends React.Component<Props, State> {
       rules: [],
       valid: true,
       mtlsMode: NONE,
-      loadBalancer: NONE,
-      modified: false
+      tlsModified: false,
+      loadBalancer: NONE
     };
   }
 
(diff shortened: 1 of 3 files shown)


lucasponce:
DNM label set.
Found a last minute error on my PR.

lucasponce:
Final fix for 2 workloads usecase:

image

pilhuhn:
Looks great!

abonas:
great work! does this remove the pin also for the single route case?

hhovsepy:
@lucasponce I will suggest to hide "Evenly Distribute Traffic" button when there is only one workload.
Also on a case then Mesh wide mTLS is configured:
1. under the Chrome browser, TLS ISTIO_MUTUAL is selected by default, when I change any percent it is reset, when I click on "Evenly Distribute Traffic" it selects back "ISTIO_MUTUAL". Is it on purpose?
screenshot from 2019-03-07 18-28-35
screenshot from 2019-03-07 18-28-51

2. under FF TLS is always NONE by default

lucasponce:
> great work! does this remove the pin also for the single route case?

Yes, in the first comment, there is a capture with that case in the ratings-v1 workload.
As only workload is elegible, it's by default at 100% and disabled.

lucasponce:
Good finding @hhovsepy.
No, it's not intentional, I guess it's a bug, let me work on it.

lucasponce:
Ok @hhovsepy I think I've fixed it
image
"Evenly..." button is out when only one workload and there was a bug in the state between the TLS component and the Wizard.
I think that Chrome & FF hit same bug, locally it seems fixed, let me know.

hhovsepy:
Verified

PR #1054, opened 2019-03-08
KIALI-2273 Fix SessionTimeout dialog

israel-hdez:
https://issues.jboss.org/browse/KIALI-2273
also: https://issues.jboss.org/browse/KIALI-2518

* SessionTimeout was not appearing because of a time unit mismatch (seconds vs milliseconds)
* Extend session feature was not working. UI cannot extend the session by itself; it needs back-end's help. Now the "extend session" button will make the required request to extend the session successfully. *Note:* If Oauth login is configured, the session cannot be extended.
* "Friendly message" (motivation of KIALI-2273)... Instead of displaying an error in the message center, the user will be redirected to the login page:

image

**This requires https://github.com/kiali/kiali/pull/903**

### DIFF
--- a/src/actions/LoginActions.ts
+++ b/src/actions/LoginActions.ts
@@ -6,7 +6,8 @@ enum LoginActionKeys {
   LOGIN_EXTEND = 'LOGIN_EXTEND',
   LOGIN_SUCCESS = 'LOGIN_SUCCESS',
   LOGIN_FAILURE = 'LOGIN_FAILURE',
-  LOGOUT_SUCCESS = 'LOGOUT_SUCCESS'
+  LOGOUT_SUCCESS = 'LOGOUT_SUCCESS',
+  SESSION_EXPIRED = 'SESSION_EXPIRED'
 }
 
 export interface LoginPayload {
@@ -46,7 +47,8 @@ export const LoginActions = {
       session: undefined,
       error: undefined
     } as LoginPayload)
-  )
+  ),
+  sessionExpired: createAction(LoginActionKeys.SESSION_EXPIRED)
 };
 
 export type LoginAction = ActionType<typeof LoginActions>;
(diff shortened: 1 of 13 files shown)


jmazzitelli:
@israel-hdez How long does a session last, and how can one set the timeout period to something shorter (mainly for testing)?

I would like to be able to set the session timeout to, say, 60 seconds so I can see if this works without having to wait 30 minutes or whatever the default session timeout is.

israel-hdez:
@jmazzitelli By default it is 24hrs. You can change the expiration time by adding an LOGIN_TOKEN_EXPIRATION_SECONDS environment variable to the kiali deployment.

jmazzitelli:
I set my env var to have a session timeout of 20 seconds. It appears to be working, however, look at the session timeout message:

"Would you like to extend your session for another 30 minutes?"

Looks like "30 minutes" is hardcoded. This is incorrect because if a user wants to change that (some security people may want this to be something shorter) they will see this message and think they are getting another 30 minutes of session time when they are really getting it for the length of time configured.

To avoid having the UI make a request to the server just to print out in this dialog what the real session time out is, I suggest we just rename this message to something more generic. Something like "Would you like to extend your session?" or "Would you like to restart your session timer to extend your session?"

session-to-msg

jmazzitelli:
Get the session timeout dialog to popup. Now click anywhere outside of the dialog window. It behaves as if I clicked the "Log out" button, which I did not. I think the behavior should be don't do anything unless I click either the "Log out" or the "Continue Session" button.

In this animated gif, I clicked where you see the mouse pointer rest below the dialog window.
 

session-dialog ogv

israel-hdez:
@jmazzitelli I changed the message to the generic "Would you like to extend your session?".
Also, dialog no longer hides when clicking outside it.

PR #1055, opened 2019-03-09
Add new tslint rule to eliminate the old Object.assign() syntax with ES6+

mtho11:
** Describe the change **
Let's replace this old javascript Object.assign() syntax with the current object spread syntax. It's just more descriptive and therefore easier to grok.

This adds the tslint rules to enforce this.

** Documentation **

- https://palantir.github.io/tslint/rules/prefer-object-spread/
- https://redux.js.org/recipes/using-object-spread-operator
- https://davidwalsh.name/merge-objects


mtho11:
Looks like I had to add a ts-ignore for when the object could be null because the spread operator only works with objects.

PR #1056, opened 2019-03-11
KIALI-2535 Move TLS/LoadBalancer under Show Advanced Options

lucasponce:
** Describe the change **

https://issues.jboss.org/browse/KIALI-2535

image

image

### DIFF
--- a/src/components/IstioWizards/IstioWizard.tsx
+++ b/src/components/IstioWizards/IstioWizard.tsx
@@ -14,6 +14,7 @@ import * as MessageCenter from '../../utils/MessageCenter';
 import MatchingRouting, { ROUTE_TYPE, Rule } from './MatchingRouting';
 import WeightedRouting, { WorkloadWeight } from './WeightedRouting';
 import TrafficPolicyConnected from '../../containers/TrafficPolicyContainer';
+import { DISABLE, ROUND_ROBIN } from './TrafficPolicy';
 
 type Props = {
   show: boolean;
@@ -32,6 +33,7 @@ type State = {
   mtlsMode: string;
   tlsModified: boolean;
   loadBalancer: string;
+  lbModified: boolean;
 };
 
 export const WIZARD_WEIGHTED_ROUTING = 'create_weighted_routing';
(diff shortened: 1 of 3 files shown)


abonas:
while at it, maybe add the lock icon next to TLS? ;)

abonas:
also, did you remove the "NONE" option from the TLS dropdown? it was part of the issues that led to this change

lucasponce:
> while at it, maybe add the lock icon next to TLS? ;)

What pattern here,
TLS <icon> or <icon> TLS, or TLS <dropdown> <icon> ?

lucasponce:
> also, did you remove the "NONE" option from the TLS dropdown? it was part of the issues that led to this change

Ok, that means that we will populate always these values, but I guess that's not bad. I'll use the default values for both cases (TLS and LB).

abonas:
> What pattern here,
> TLS or TLS, or TLS ?

sorry, don't understand the above

abonas:
> Ok, that means that we will populate always these values, but I guess that's not bad. I'll use the default values for both cases (TLS and LB).

only if user selected something in the dropdown, no?

hhovsepy:
Verified.

lucasponce:
I'm going to address in a next commit a couple of comments discussed in the PR.

lucasponce:
image

NONE value removed from Dropdowns.
Lock icon added to TLS label.
TLS value is pre-selected depending if mTLS is defined global wide, otherwise is selected DISABLED.

image

Only tls and loadbalancer are added into final DR object when user modifies or Istio_Mutual is enabled.

abonas:
this is a huge improvement. I have one last suggestion, not sure if it's even possible - to increase the font of the words of "Advanced options" - they look small on their own, and even smaller because right below them there are some words in capital letters

lucasponce:
image

abonas:
EXCELLENT!

lucasponce:
@aljesusg @hhovsepy I've sent a couple of commits after your review; if you'd like to give a final look, to check everything is fine, we can - then - merge this one.
Thx

hhovsepy:
Verified it in a pair with https://github.com/kiali/kiali/pull/909
Screenshot from 2019-03-12 14-27-55

PR #1042, opened 2019-03-05
KIALI-2486 Adding Policy and MeshPolicy validations to IstioConfigList

xeviknal:
** Describe the change **

IstioConfig list now shows validations for:

- MeshPolicy
- Policy
- ClusterRbacConfig
- ServiceRole
- ServiceRoleBinding

There are only validations for MeshPolicy at the moment. However, KIALI-2392 should add validations for the latest 4 config types.

** Issue reference **
https://issues.jboss.org/browse/KIALI-2486

** Backwards compatible? **
yes

** Screenshot **

screenshot of kiali console 61

screenshot of kiali console 60

PR #1057, opened 2019-03-11
KIALI-2392 Enable validations for any backend returned type

burmanm:
** Describe the change **

Instead of manually enabling a type each time (like #1042) this patch makes the backend decide which items are validated. Thus it allows future types (such as those coming from KIALI-2392 backend parts) to display validation results without changes to the UI.

Also, removes code duplication.

** Issue reference **

Part of KIALI-2392 which requires new types.

** Backwards compatible? **

Yes.

** Screenshot **

No UI impact.

### DIFF
--- a/src/types/IstioConfigList.ts
+++ b/src/types/IstioConfigList.ts
@@ -20,6 +20,7 @@ import {
   ClusterRbacConfig
 } from './IstioObjects';
 import { ResourcePermissions } from './Permissions';
+import _ from 'lodash';
 
 export interface IstioConfigItem {
   namespace: string;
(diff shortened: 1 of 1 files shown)


hhovsepy:
Error message is missing when wrong ServiceRole is defined ServiceRoleBinding:
Screenshot from 2019-03-15 17-15-26

Also it does not load any IstioConfig when merged with master.

burmanm:
That's probably because master broke it an hour ago.. ;)

hhovsepy:
master works fine itself,
but with these 2 PRs, "Error loading Istio Config"

xeviknal:
@burmanm  most probably is broken due to one missing error message in here: https://github.com/kiali/kiali/pull/918/files#diff-00df504feeb66372dea6b00a33164c66R153

PR #1058, opened 2019-03-12
KIALI-2533 Fix trim space in Matching Routing Wizard

lucasponce:
** Describe the change **

Matching Routing Wizard introduced some spaces when parsing values.
That created non valid configuration; only when Galley is up un running it can be detected.

** Issue reference **

https://issues.jboss.org/browse/KIALI-2533


src/components/IstioWizards/IstioWizard.tsx
```diff
@@ -109,8 +109,8 @@ class IstioWizard extends React.Component<Props, State> {
         const headerName = match.substring(i0 + 1, j0);
         const i1 = match.indexOf(' ', j0 + 1);
```
aljesusg: Why not add .trim() to headerName too?

lucasponce: Ah, probably I thought it was not necessary, let me check and update it.

hhovsepy:
Verified

PR #1062, opened 2019-03-13
KIALI-470 Add Response Code summary to Edge side panel

jshaughn:
*** REQUIRES Servr PR https://github.com/kiali/kiali/pull/913

### DIFF
--- a/src/components/CytoscapeGraph/CytoscapeGraphUtils.ts
+++ b/src/components/CytoscapeGraph/CytoscapeGraphUtils.ts
@@ -17,6 +17,7 @@ export const CyEdge = {
   id: 'id',
   isMTLS: 'isMTLS',
   protocol: 'protocol',
+  responses: 'responses',
   responseTime: 'responseTime',
   tcp: 'tcp'
 };
--- a/src/components/SummaryPanel/ResponseTable.tsx
+++ b/src/components/SummaryPanel/ResponseTable.tsx
@@ -34,20 +34,12 @@ export class ResponseTable extends React.PureComponent<ResponseTableProps> {
   }
 
   render() {
-    const TableRow = ({ row }) => (
-      <tr>
-        <td>{row.code}</td>
-        <td title={this.getTitle(row.flags)}>{row.flags}</td>
-        <td>{row.val}</td>
-      </tr>
-    );
-
     return (
       <>
         <strong>{this.props.title}</strong>
         <table className="table">
           <thead>
-            <tr>
+            <tr key="table-header">
               <th>Code</th>
               <th>Flags</th>
               <th>% of Requests</th>
@@ -55,7 +47,11 @@ export class ResponseTable extends React.PureComponent<ResponseTableProps> {
           </thead>
           <tbody>
             {this.getRows(this.props.responses).map(row => (
-              <TableRow row={row} />
+              <tr key={row['key']}>
+                <td>{row['code']}</td>
+                <td title={this.getTitle(row['flags'])}>{row['flags']}</td>
+                <td>{row['val']}</td>
+              </tr>
             ))}
           </tbody>
         </table>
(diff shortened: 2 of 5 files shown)


abonas:
I see it's still a draft, it will be helpful to have a screenshot at some point (before merging :))

jshaughn:
@abonas Yes, here is a POC video for evaluation.  Also, @mwringe, what do you think?  The basic idea is that for edges we now collect the full set of responses backing the traffic.  We can then use that to display the % of traffic by code and flags (note: TCP only has flags, no codes).  This new "Responses" view can help users better understand the traffic.  I played around a bit with how to include the info on the side panel without adding too much to the size.  At first I tried sort of a collapsible approach but I liked a tabbed approach better, so that is my proposal.  It looks like this:

kiali-470

abonas:
I think this is a GREAT addition.
The only thing that confuses me are the percentage numbers - for example on the graph it says 0.82 (what does it mean?) and then on the side panel it says 100%.
What does it mean?

jshaughn:
> The only thing that confuses me are the percentage numbers

@abonas Yes, let's see if we can get this to be more clear.  The idea is to show the breakdown of responses for all of the traffic on the edge.  So the "%Traffic" column should always add up to 100, each row showing the percentage of traffic returning that particular responseCode-responseFlags tuple.  So, the edge label may show 0.82 requestsPerSecond (for say the last minute),   That's something like 49 requests, if all of the requests returned 200 responses with no flags then this new table would show 1 row: "200 100  -", meaning 100% of the 49 requests returned 200.

I think percentages is the way to go here, we just need to maybe make it more clear.  Would it be helpful to maybe change some titles and maybe the column ordering?  Something like:

Code      Flags    % of Requests
----------------------------------
200           -            100%

Maybe something other than a table?  A pie graph?  I'm open to ideas.

jmazzitelli:
Using a pie chart will probably not be good because the pie will need to be too big. I remember playing with pie charts and they require alot of real estate to show properly. I think labeling it "% of requests" would be nice. Or maybe a stacked bar chart like we already use? Only the stacked items are for the different flags. Most of those will only have one item (since, from what we've seen, it is rare (though possible) to have more than 1 flag per code).

jmazzitelli:
Note also we need a hover or something to help describe those 2-letter codes. No one is going to know what the 2-letter codes mean.

jshaughn:
@mazz, look closely at the vid, I do have hover for the codes.

jshaughn:
A couple of small changes:
- center the tabs
- change the column order putting percentages at the end
- change the percentages column header from "%traffic" to "% of Requests"
- improve table titles

image

jshaughn:
@abonas, @mwringe, I am moving these PRs out of Draft mode as feedback has been positive and the feature seems stable.  I'm not sure whether this needs UX review.  I'll also add @mattmahoneyrh for QE acceptance.

jshaughn:
> Clicking on an Edge and then selecting Response Codes is not remembered.

@mattmahoneyrh It would take a lot to make this enhancement.  We already maintain the selection when going from edge to edge, it reverts back to the default tab only when navigating from edge to non-edge and back.  That is because each side panel is it's own component, and each is unmounted when changing selection.  The detail pages actually save the tab in the URL,  So you can actually bookmark the page and the tab selection.  The graph URL does not allow bookmarking of a selected component (the whole graph is always the initial selection), so it doesn't make sense for the url to maintain information about side-panel configuration. 

For consistency i suppose we could force a return to the default tab when navigating edge to edge.  I'm not sure if that would be better or worse.  Let me know what you think about that idea because the requested change is not going to be doable.

jmazzitelli:
The requested change should be another JIRA so as not to block this PR. Don't let perfect get in the way of the good :)

jshaughn:
@mattmahoneyrh See my response above regarding your change request.  Assuming you are OK with it please let me know if you have any other change requests or go ahead and approve.  Thanks.

@abonas Do you have any blockers before merge?

mazz:
I’m not a contributor to this project.

jshaughn:
Sorry about that, nick conflict.

mazz:
Np, have a great day!

jshaughn:
@abonas I'm going to merge to avoid conflicts, it's early in the Sprint so plenty of time to respond to any issues or enhancements.

PR #1063, opened 2019-03-13
KIALI-2517 Disable logout for anonymous login strategy

israel-hdez:
https://issues.jboss.org/browse/KIALI-2517

Fixing logout option. It's now disabled when anonymous login is enabled:

image


jmazzitelli:
> However, a browser f5 brings back to the login page, where the user then has to login again.

That is not related to this PR. I see that behavior in master. Probably needs another JIRA for that.

PR #1068, opened 2019-03-14
KIALI-2555 Return to the OpenShift login page on error

mwringe:
** Describe the change **

Adds in some better error handling when dealing with OpenShift OAuth logins.

If we encounter a problem during login it will be bring the user back to the OAuth login page so that the user can start the login process again. Otherwise they can get stuck on the main username/password login screen.

I was not able to exactly recreate the issue mention in KIALI-2555, but spent some time modifying the code to do things like setup cookies with invalid tokens, expired OpenShift tokens, etc. So I can't verify it this fixes the exact problem reported.

** Issue reference **

https://issues.jboss.org/browse/KIALI-2555

### DIFF
--- a/src/services/Login.ts
+++ b/src/services/Login.ts
@@ -83,7 +83,12 @@ class OpenshiftLogin implements LoginStrategy<any> {
   }
 
   public async perform(_request: NullDispatch): Promise<LoginResult> {
+    // get the data from the url that was passed by the OAuth login.
     const session = (await API.checkOpenshiftAuth(window.location.hash.substring(1))).data;
+
+    // remove the data that was passed by the OAuth login. In certain error situations this can cause the
+    // page to enter a refresh loop since it tries to reload the page which then tries to reuse the bad token again.
+    history.replaceState('', document.title, window.location.pathname + window.location.search);
 
     return {
       status: AuthResult.SUCCESS,
--- a/src/actions/LoginThunkActions.ts
+++ b/src/actions/LoginThunkActions.ts
@@ -21,8 +21,13 @@ const loginSuccess = async (dispatch: KialiDispatch, session: LoginSession) => {
 // The `data` argument is defined as `any` because the dispatchers receive
 // different kinds of data (such as e-mail/password, tokens).
 const performLogin = (dispatch: KialiDispatch, state: KialiAppState, data?: any) => {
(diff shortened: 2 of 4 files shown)


src/services/Login.ts
```diff
@@ -83,12 +83,24 @@ class OpenshiftLogin implements LoginStrategy<any> {
...
+    try {
+      const session = (await API.checkOpenshiftAuth(window.location.hash.substring(1))).data;
 
-    return {
-      status: AuthResult.SUCCESS,
-      session: session
-    };
+      // Remove the access_token information from the URL so that if we have a reload on LoginThunkActions.checkCredentials
+      // it will take us back to the OpenShift Login page. We don't want to redirect now, but affect the prepare check above.
+      history.pushState('', document.title, window.location.pathname + window.location.search);
```
israel-hdez: In this case, I think `replaceState` is better than `pushState`

mwringe:
> I have the idea that the error condition is already handled here:

It doesn't handle it properly. We receive the error message but then we don't do anything with it and we get stuck on the Kiali login page. We need special handling for the OAuth login so that we go back to the OpenShift login page (to get a new token).

lucasponce:
I wonder if instead to move back to console, we can maintain the error page with the description and a link to the console login (as it happens in OpenShift console).
In that way user will get info to debug.

(I couldn't test it locally yet).

src/services/Login.ts
mwringe: Done

mwringe:
Yeah, I am trying to see if we can provide a nicer way of doing this with a proper error message.

lucasponce:
> Yeah, I am trying to see if we can provide a nicer way of doing this with a proper error message.

So, this is ready for final review or it is missing some last commit to address last comment ?

mwringe:
I am still trying to get this to work properly with a nice UI message. I will update the jira when that is added.

lucasponce:
> I am still trying to get this to work properly with a nice UI message. I will update the jira when that is added.

I don't see activity in this one, I guess it's not yet ready for final review, isn't it ? @mwringe

mwringe:
Updated. Now provides some UI to display the error message and allow the user to go back to the OpenShift OAuth login page:

Screenshot from 2019-03-27 16-31-53

PR #1069, opened 2019-03-15
KIALI-2563 Add RbacConfigs to IstioConfig section

xeviknal:
** Describe the change **
Adding RbacConfigs to IstioConfig section. 

** Issue reference **
https://issues.jboss.org/browse/KIALI-2563

** Screenshot **

Screenshot-2019-3-15 Kiali Console

### DIFF
--- a/src/types/IstioConfigDetails.ts
+++ b/src/types/IstioConfigDetails.ts
@@ -15,7 +15,7 @@ import {
   ServiceEntry,
   VirtualService,
   ObjectValidation,
-  ClusterRbacConfig,
+  RbacConfig,
   ServiceRole,
   ServiceRoleBinding
 } from './IstioObjects';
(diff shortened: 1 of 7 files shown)


src/types/IstioConfigDetails.ts
```diff
@@ -42,7 +42,8 @@ export interface IstioConfigDetails {
   quotaSpecBinding: QuotaSpecBinding;
   policy: Policy;
   meshPolicy: Policy;
-  clusterRbacConfig: ClusterRbacConfig;
+  clusterRbacConfig: RbacConfig;
```
lucasponce: I'd maintain the ClusterRbacConfig type here, they have same fields, but conceptually they are different entities.
I think it makes sense to separate them logically, if not it can point to some mistake.

abonas: wasn't RBacConfig deprecated? https://github.com/istio/istio/issues/8825
and yes, I'd maintain them separately

xeviknal: @abonas in 1.1.0 both resources are leaving. The final deprecation will be in a subsequent step.
In any case, latest version of Istio now is 1.0.6 with RbacConfig (no ClusterRbacConfig).

burmanm: Aren't we supporting only the newer version (1.1.0) ?

lucasponce: Good question.
There is a JIRA for that https://issues.jboss.org/browse/KIALI-1023
As those types are valid and used for 1.0 and 1.1 we thought that we could do the effort to support both scenarios.
But yes, theorically we should support 1.1 (newer) versions

PR #1072, opened 2019-03-18
[Minor] Fix warning about accessibility issue

jotak:
Warning message was: 

Warning: [react-bootstrap] Specifying only the `"hover"` trigger limits the visibility of the overlay to just mouse users. Consider also including the `"focus"` trigger so that touch and keyboard only users can see the overlay as well.

### DIFF
--- a/src/components/ToolbarDropdown/ToolbarDropdown.tsx
+++ b/src/components/ToolbarDropdown/ToolbarDropdown.tsx
@@ -60,7 +60,7 @@ export class ToolbarDropdown extends React.Component<ToolbarDropdownProps, Toolb
           <OverlayTrigger
             key={'ot-' + this.props.id}
             placement="top"
-            trigger="hover"
+            trigger={['hover', 'focus']}
             delayShow={1000}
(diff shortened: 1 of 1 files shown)


jshaughn:
@jotak I realize the warnings are annoying in the dev (yarn start) builds but I had it set to "hover" on purpose.  The default is actually ["hover", "focus"].  We can add back "focus" if you like but for me this was causing undesirable stickiness of the tooltip.  Did you play with the tooltips when adding it back?  Did it seem OK for you, especially when the dropdown was open?

jotak:
@jshaughn I didn't see any issue, now retrying knowing what you say, with chrome and FF, but it still seems fine.
What do you mean by "undesirable stickiness"?
For me, with mouse over the dropdown, the tooltip shows ; when mouse leaves, tootltip hides ; same happens when the dropdown options are opened ; and same when using the keyboard instead of the mouse.
So is it something different? Or maybe a different browser / version? (I'm with FF 65.0.2 here, and Chrome 72.0.3626.119)

jotak:
Ok, so I'm merging, you'll tell me if the problem comes back

PR #1073, opened 2019-03-18
KIALI-2438 Metrics dashboards to manage non-SI underscaling like bytes

jotak:
This is to prevent some data types (currently bytes and bitrates) to be downscaled as "milli-bytes" / "µbps" which are quite uncommon. At the same time, introducing IEC-based versus SI-based data units.

Now, some unit types are recognized from the dashboards renderer and processed in a specific way.
Other units / non-recognized ones continue to fall under the standard SI scaler

JIRA: https://issues.jboss.org/browse/KIALI-2438

Backend PR: https://github.com/kiali/kiali/pull/925

### DIFF
--- a/src/components/Metrics/MetricsChartBase.tsx
+++ b/src/components/Metrics/MetricsChartBase.tsx
@@ -75,7 +75,45 @@ abstract class MetricsChartBase<Props extends MetricsChartBaseProps> extends Rea
     };
   }
 
-  formatYAxis = (val: any): string => {
+  formatYAxis = (val: number): string => {
+    switch (this.props.unit) {
+      case 'bytes':
+      case 'bytes-si':
+        return this.formatDataSI(val, 'B');
+      case 'bytes-iec':
+        return this.formatDataIEC(val, 'B');
+      case 'bitrate':
+      case 'bitrate-si':
+        return this.formatDataSI(val, 'bit/s');
+      case 'bitrate-iec':
+        return this.formatDataIEC(val, 'bit/s');
+      default:
+        // Fallback to default SI scaler:
+        return this.formatSI(val);
+    }
+  };
+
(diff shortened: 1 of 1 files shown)


pilhuhn:
What about https://de.wikipedia.org/wiki/Mops ?

I think op/s also does not lend to easy scaling (and yes, there are FLOP/s).

jotak:
See the new Y-axis :

new

Whereas it was previously showing things like "500 mB" or "500 mpbs" (mpbs being millibit per second, not to be mixed with Mbps, Megabit per second)

jotak:
@pilhuhn , that dog doesn't scale :-D
More seriously, yes, "ops" could probably part of that too

jotak:
@pilhuhn I changed the default behaviour to not down-scale (so, no milli- , micro- and so on). Given the current dashboards that we have, it seems easier to list explicitly the ones that can downscale, than the opposite.

lucasponce:
One thing I see that probably would need some adjustment in the scale:
image
It's weird to have same left '7.6k' in the metrics.
Also, I expected that the metric in the tooltip could show some details but it's same value:
image

I have another example here:
image



Another comment is if when some metric is empty, if we could should a default 0:
image

Not having a metric sounds like if something is missing.

I've tested and this fix is pretty cool, in other metrics it works as expected.

jotak:
@lucasponce now using d3.format('~r') for rounding on significant digits and removing tail 0s, seems to work much better (https://github.com/d3/d3-format#locale_format)

jotak:
@lucasponce actually, DNM, I'm now seeing different problems :-/

jotak:
@lucasponce should be good now, I've added an additional rounding to remove floating-point issue. Just checked all the dashboards, everything looks fine. I just haven't fixed yet the missing "0" indicator that you pointed out, i can do it in another PR

PR #1074, opened 2019-03-18
KIALI-2537 Add columns to Weighted Routing Wizard

lucasponce:
Align evenly button to the right

https://issues.jboss.org/browse/KIALI-2537

image

### DIFF
--- a/src/components/IstioWizards/Slider/BootstrapSlider.tsx
+++ b/src/components/IstioWizards/Slider/BootstrapSlider.tsx
@@ -16,7 +16,12 @@ type Props = {
   ticks_labels: [];
   locked: boolean;
   min: number;
+  // Note that Slider will use max and maxLimit properties to:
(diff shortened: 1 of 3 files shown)


lucasponce:
I've sent another commit to address KIALI-2578 as it's related to the same component I can't work on it without the previous work.

image

Now -/+ buttons are not hidden but disabled, on this way, aspect/ratio continues similar to other slider items.

Also, introduced a max/maxLimit in the slider to be able to limit the max in relative scenarios.

In that case, when in the first slider is fixed i.e. 70, the other sliders will show max 100, but only a value of 30 should be allowed

lucasponce:
Please @abonas give it a try and let me know if it fixes the issues you have seen.

hhovsepy:
FF:
Screenshot from 2019-03-18 19-13-18

Chrome:
Screenshot from 2019-03-18 19-14-20

PR #1075, opened 2019-03-18
KIALI-2550 Hide weights on single routes

lucasponce:
Also moving some VS/DR details files under IstioConfigDetails page.

https://issues.jboss.org/browse/KIALI-2550

When there are more than 1 destination, weight bullet chart is shown:
image

But when there is only a single route, the chart is hidden.
image


lucasponce:
cc @abonas

hhovsepy:
Verified, the same as in screenshots.

PR #1077, opened 2019-03-18
KIALI-1997 App Box summary panel should provide service link and appropriate badging

jshaughn:
- support service links (not sure if multiple is possible bit it doesn't hurt)
- support VS and CB badging

image

### DIFF
--- a/src/pages/Graph/SummaryPanelGroup.tsx
+++ b/src/pages/Graph/SummaryPanelGroup.tsx
@@ -24,6 +24,7 @@ import { Reporter } from '../../types/MetricsOptions';
 import { CancelablePromise, makeCancelablePromise } from '../../utils/CancelablePromises';
 import { serverConfig } from '../../config/serverConfig';
 import { CyNode } from '../../components/CytoscapeGraph/CytoscapeGraphUtils';
+import { icons } from '../../config';
 
 type SummaryPanelGroupState = {
   loading: boolean;
@@ -96,7 +97,7 @@ export default class SummaryPanelGroup extends React.Component<SummaryPanelPropT
     const group = this.props.data.summaryTarget;
     const data = nodeData(group);
     const { namespace } = data;
-
+    const serviceList = this.renderServiceList(group);
     const workloadList = this.renderWorkloadList(group);
 
     return (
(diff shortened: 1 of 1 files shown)


jshaughn:
@pilhuhn , Is this what you were looking for (see above for screenshot).  Does it make sense to have CB as well?  If not I can remove that badging.

abonas:
@jshaughn I have a more general question regarding CB - afaik we don't have that specific CB indication anywhere else in Kiali. And afaik the configuration for CB is done also in VS (or maybe DR?) so won't it confuse the users they can't see this anywhere else?
We have a very "flat" display of VS/DR in services and in Istio Config, showing CB (which I like as a feature) is great, but perhaps needs to be available outside of the graph too? And if CB has its own display, maybe "inject delay" and "inject abort" need to be handled the same as well?

jshaughn:
@abonas This topic should probably be posted outside of this PR, probably to the mailing list.

I always think of the graph as a tool to quickly identify areas of interest in the topology.  The graph, I think, rarely explains a problem but helps the user know where to dig deeper.  Showing VS in the graph is a very general indicator because basically all traffic routing is configured via VS.  But it still lets the user know that there is something there to investigate.  Note that VS badging is applied only to the Service node it decorates.  CB is a common use of VS and so, I guess, we identify it explicitly.  CB itself has 2 major categories (outlier and connection pool) but we stop short in the graph of getting into these details, deferring to our other views.  Note that unlike the more general VS badging, CB badging is applied to the affected destination nodes.  This is a long-winded way of me agreeing with you, we could provide more badging for common types of VS, like FI (fault injection) or mirroring, but that the lack of that badging does not in any way take away from showing CB.   It's a balancing act, though, each additional badge can add more complexity to the graph and its display options, and possibly slow down generation depending on the work needed to do the discovery.

As for other Kiali pages showing or not showing CB (or other types of VS),  I think being able to investigate the yaml is very valuable.   I suppose we could could indicate the VS types via badges but dedicated views for performing VS CRUD for CB, Mirror, Load balancing, etc, that is likely a decent area for Kiali to go.  We are already doing it for the routing weights, right?

abonas:
> @abonas This topic should probably be posted outside of this PR, probably to the mailing list.
> 
true :)
> I always think of the graph as a tool to quickly identify areas of interest in the topology. The graph, I think, rarely explains a problem but helps the user know where to dig deeper. Showing VS in the graph is a very general indicator because basically all traffic routing is configured via VS. But it still lets the user know that there is something there to investigate. Note that VS badging is applied only to the Service node it decorates. CB is a common use of VS and so, I guess, we identify it explicitly. CB itself has 2 major categories (outlier and connection pool) but we stop short in the graph of getting into these details, deferring to our other views. Note that unlike the more general VS badging, CB badging is applied to the affected destination nodes. 
what do you mean? not sure I understand the "unlike" part - what is the difference? 

This is a long-winded way of me agreeing with you, we could provide more badging for common types of VS, like FI (fault injection) or mirroring, but that the lack of that badging does not in any way take away from showing CB.
>

absolutely, this is not a disadvantage in any way

 It's a balancing act, though, each additional badge can add more complexity to the graph and its display options, and possibly slow down generation depending on the work needed to do the discovery.
> 

won't it be very similar processing to detect FI vs CB?

> As for other Kiali pages showing or not showing CB (or other types of VS), I think being able to investigate the yaml is very valuable. I suppose we could could indicate the VS types via badges but dedicated views for performing VS CRUD for CB, Mirror, Load balancing, etc, that is likely a decent area for Kiali to go. We are already doing it for the routing weights, right?

yes we do.
I think it will be helpful to be able to click on CB on the graph side panel and be navigated to its relevant VS/yaml, but that's indeed a topic for a different place :)

PR #1080, opened 2019-03-19
KIALI-2588 Fix headers without name

lucasponce:
Fix style of matching rule

https://issues.jboss.org/browse/KIALI-2588
https://issues.jboss.org/browse/KIALI-2576

Buttons are disabled in case there is a validation message:
image

Text of matching rule is changed to address KIALI-2576, minor adjustments in style to support large header names.

image

image

### DIFF
--- a/src/components/IstioWizards/MatchingRouting.tsx
+++ b/src/components/IstioWizards/MatchingRouting.tsx
@@ -5,6 +5,7 @@ import {
   DropdownKebab,
   Form,
   FormControl,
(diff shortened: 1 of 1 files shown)


hhovsepy:
All works fine, just on FF it shows some gray fonts:
Screenshot from 2019-03-20 09-51-22

lucasponce:
@hhovsepy it should work now:
image

Sadly I realized about some CSS rendered different in Chrome/FF.

PR #1082, opened 2019-03-20
Kiali-907: Error messages pile up infinitely.

mtho11:
** Describe the change **
This keeps repeating error messages from piling up in the message drawer.
It now works like the Openshift messages with a count of how many times the notification message is repeated and for how long the same notification has been going on.

** Issue reference **
https://issues.jboss.org/browse/KIALI-907

** Backwards compatible? **
yes

no, just changes the way the messages are presented

** Screenshot **

**Old**:

_KIALI-907__Error_Messages_Pile_up_infinitely_-_JBoss_Issue_Tracker

**New**:

error-msg

** Documentation **

### DIFF
--- a/src/reducers/MessageCenter.ts
+++ b/src/reducers/MessageCenter.ts
@@ -1,10 +1,11 @@
-import { MessageType } from '../types/MessageCenter';
+import { MessageType, NotificationMessage } from '../types/MessageCenter';
 import { MessageCenterState } from '../store/Store';
 import { KialiAppAction } from '../actions/KialiAppAction';
 import { getType } from 'typesafe-actions';
 import { MessageCenterActions } from '../actions/MessageCenterActions';
 import { updateState } from '../utils/Reducer';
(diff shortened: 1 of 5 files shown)


src/reducers/MessageCenter.ts
```diff
@@ -29,14 +30,16 @@ export const INITIAL_MESSAGE_CENTER_STATE: MessageCenterState = {
   expandedGroupId: 'default'
 };
 
-const createMessage = (id: number, content: string, type: MessageType) => {
+const createMessage = (id: number, content: string, type: MessageType, count, created, firstTriggered) => {
```
israel-hdez: No types on the new params.
This is private, so I guess we can live without the types. Just commenting to make sure that you are OK.

mtho11: Oh you are right, I should have types there. Thanks.

mtho11:
@mattmahoneyrh do you mind checking this out?

PR #1083, opened 2019-03-20
KIALI-2549 Migrate Source Workloads tab to Traffic in Service Details page

israel-hdez:
image


israel-hdez:
@lucasponce This is the initial approach. Not ready for review yet, but I want some early feedback. I have doubts about if this is the right way to show data for the "services" case.

First, the inbound/outbound terminology sounds wrong because services are just routing traffic. Not sure what terminology could be used. 

However, I think a unified table makes more sense (having just one table with source and destination columns). IMO, it would be more interesting to see the source and destination of the traffic that went through the service. Also, it would resemble more closely the "Source workloads" tab:

image

For apps/workloads the current approach looks good because they are traffic generators and consumers. For the services case, I'm not sure.

I can finish this approach if it is OK for you and prefer consistency across Kiali. But if you agree that a unified table would be more appropriate, I can implement that instead of this.

cc @mwringe

lucasponce:
> This is the initial approach. Not ready for review yet, but I want some early feedback. I have doubts about if this is the right way to show data for the "services" case.

To me it sounds right, at the end we want here to unify same concepts between details pages.
"Source Workloads" was a pretty old PoC tab put it there since beginning of the project, but conceptually wanted to show to the user the incoming/inbound/who invokes the current service.

About terminology, yeah, if we got in such level of details about "pure producer" we can change, but I don't think is needed, as conceptually app and service are similar, I won't change it on this iteration.

> However, I think a unified table makes more sense (having just one table with source and destination columns). IMO, it would be more interesting to see the source and destination of the traffic that went through the service. Also, it would resemble more closely the "Source workloads" tab.

I'm fine with both approaches, personally I guess that user will like to see same format between app/workload/service, if we change it, perhaps visually can create confusion; but I don't know.

To me remove the Source Workloads with the Traffic was the goal, just to have consistency and upgrade that old tab. We can work inside details in future iterations.

israel-hdez:
@lucasponce Ok, then I'll finish this work and improve in future iterations, if needed.

israel-hdez:
@lucasponce This is ready for your review (or anyone from Jupiter).

I created https://issues.jboss.org/browse/KIALI-2610 to remove from the backend's Json the data that is no longer being used. I'll follow up with this Jira.

Then, also follow up with https://issues.jboss.org/browse/KIALI-2355 to add the necessary and missing tests.

PR #1088, opened 2019-03-21
KIALI-2560 Add expanded mode for overview cards

jotak:
See https://github.com/kiali/kiali-design/issues/57

- Fetch metrics in overview when mode is expanded
- Add toggle button to switch expanded/compact
- Horizontal bar chart (for health) and sparkline (for RPS)

JIRA: https://issues.jboss.org/browse/KIALI-2560

### DIFF
--- a/src/pages/Overview/OverviewPage.tsx
+++ b/src/pages/Overview/OverviewPage.tsx
@@ -1,6 +1,5 @@
 import * as React from 'react';
-import { Link } from 'react-router-dom';
-import { Breadcrumb, Card, CardBody, CardGrid, CardTitle, Col, Icon, Row } from 'patternfly-react';
+import { Breadcrumb, Card, CardBody, CardGrid, CardTitle, Col, Row } from 'patternfly-react';
 import { style } from 'typestyle';
 import { AxiosError } from 'axios';
 import _ from 'lodash';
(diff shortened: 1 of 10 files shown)


jotak:
Not ready to merge (there's a couple of things I want to improve ; for instance when no data is available (N/A))

But early feedback is welcome.
cc @serenamarie125 @abonas @lucasponce 

sshot

serenamarie125:
@jotak thanks for sharing!  originally i thought left aligning made sense, but now wondering if the content above and below the bar chart should be horizontally center aligned instead.  What do you think ?  
Screen Shot 2019-03-21 at 3 47 30 PM

jotak:
New screenshot with centered text and adding a title for the sparkline:



(I merged "Traffic" and "last X minutes" on a single line, tell me if it's not ok)

About left-aligned versus centered, I'm fine with both.

hhovsepy:
Tested from functionality pov, Compact/Expand, Time Range, Overview types.

src/pages/Overview/OverviewPage.tsx
```diff
@@ -299,6 +345,47 @@ class OverviewPage extends React.Component<OverviewProps, State> {
       </>
     );
   }
+
+  renderStatuses(ns: NamespaceInfo): JSX.Element {
```
xeviknal: What about having an smaller component with those Statuses? Is it too much fragmentation? I see this component is growing to 400L. I think is a smell that things should be moved. wdyt?

src/pages/Overview/OverviewPage.tsx
```diff
@@ -299,6 +345,47 @@ class OverviewPage extends React.Component<OverviewProps, State> {
...
+          status={ns.status}
+          type={this.state.type}
+          metrics={ns.metrics}
+        />
+      );
+    }
+    return <div style={{ height: 70 }} />;
+  }
+
+  renderCardLinks(ns: NamespaceInfo): JSX.Element {
```
xeviknal: It might work also here. An smaller component rendering only the links. wdyt?

jotak:
@xeviknal I've split Cards Links to another component, and moved a static function to the helper class.
But no change about Statuses (as there's already a component).

PR #1091, opened 2019-03-22
KIALI-2613 Explicit matching selected values

lucasponce:
Matching Wizard doesn't not explicit indicate that an empty Matching will mean "Match any request".

That may lead in some confusion to the user, a first workaround can be to explicit indicate the Matching selected, included the "Any request" when the matching section is empty.

Style can be arranged in a following PR, but at least feature-wise now it will explicit inform user about the matching before he/she can add a new rule.

** Issue reference **

https://issues.jboss.org/browse/KIALI-2613

image

image

### DIFF
--- a/src/components/IstioWizards/MatchingRouting.tsx
+++ b/src/components/IstioWizards/MatchingRouting.tsx
@@ -469,7 +469,9 @@ class MatchingRouting extends React.Component<Props, State> {
         </Label>{' '}
       </span>
     ));
-    return <div className={labelContainerStyle}>{matches}</div>;
+    return (
+      <div className={labelContainerStyle}>Matching selected: {matches.length > 0 ? matches : 'Match any request'}</div>
+    );
   };
 
   renderRules = () => {
--- a/src/components/IstioWizards/MatchingRouting.tsx
+++ b/src/components/IstioWizards/MatchingRouting.tsx
@@ -470,7 +470,9 @@ class MatchingRouting extends React.Component<Props, State> {
       </span>
     ));
     return (
-      <div className={labelContainerStyle}>Matching selected: {matches.length > 0 ? matches : 'Match any request'}</div>
+      <div className={labelContainerStyle}>
+        Matching selected: {matches.length > 0 ? matches : <b>Match any request</b>}
+      </div>
     );
   };
 


lucasponce:
cc @abonas

abonas:
I'm still unsure I understand:
1. in order to get the "any request" match, should user press on "add match" with empty fields? 
2. what are the steps (or where should user click) in order to get for example: header= user alissa go to v1, all other requests route as usual (round robin, to all versions, etc)

lucasponce:
> in order to get the "any request" match, should user press on "add match" with empty fields?

No. As "Matching selected" tells you that is selected the option "Matching any request".

> what are the steps (or where should user click) in order to get for example: header= user alissa go to v1, all other requests route as usual (round robin, to all versions, etc)

Step [1]: if header = user alissa then go to v1
Step [2]: if Any Request then go to v2

There is not a combined v2/v3 as that requires to add weights into the second option as it's out of the first scope of this wizard.

abonas:
OK thanks. I think we need some more UX involvement for the matching part :)

lucasponce:
> all other requests route as usual (round robin, to all versions, etc)

To do that, it needs to add an option, like "Rest of the workloads" when selecting the Workload.
Doable, of course, in that case, instead to send to a single route, it can iterate on the pending workloads with equal weights.

But that can be done in a next PR, it's not part of this one.

lucasponce:
> OK thanks. I think we need some more UX involvement for the matching part :)

But, perhaps it's necessary the scope the feature :) first.
And yes, a mockup from UX is always welcome.

abonas:
> But that can be done in a next PR, it's not part of this one.

that's fine by me. I of course wonder what kind of matchings are expected to be popular, that can determine the ways we should improve this. but it's too early to know...

lucasponce:
About the UX I understand that we have here a combination of:
- A builder to create the match.
- A selection of those matches.

Normally, when there is a selection from one list, the common approach is to have something like:
https://www.patternfly.org/pattern-library/forms-and-controls/dual-list-selector/#design

But on this case, it was not clear, that the way we defined the Matching selected list of values. With the complexity that the empty case is also a valid value.

lucasponce:
> that's fine by me. I of course wonder what kind of matchings are expected to be popular, that can determine the ways we should improve this. but it's too early to know...

This one is something it could be cheap to add. I'll fill a JIRA.

lucasponce:
@abonas https://issues.jboss.org/browse/KIALI-2615

PR #1093, opened 2019-03-25
KIALI-2645 Clean unused mocks / useless test

jotak:
As discussed here: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/kiali-dev/Wq0A11igIX0/Lho5AeQtBQAJ

- Removing Api.test.tsx as it is really only testing the mocks themselves and the mocking function, but no actual prod code
- Among the 10 mocks we have, just 2 of them are really used in tests (GraphElements and ServiceDetails). So keep those two and remove the others
- Refactor the ServiceDetails mock in a similar way to GraphElements, using typescript definition rather than json read-from-file on execution. It allows to have it typed, hence spot easily what to fix when the model changes.
- Thanks to the mock update, the test ServiceInfo.test.tsx is now much simpler (so we don't have anymore to convert from old invalid mock to new valid model)

JIRA: https://issues.jboss.org/browse/KIALI-2645


src/services/__mocks__/Api.ts
```diff
@@ -40,52 +10,6 @@ export const getGraphElements = (params: any) => {
...
-};
-
-export const getErrorMsg = (msg: string, error: AxiosError) => {
-  let errorMessage = msg;
-  if (error && error.response && error.response.data && error.response.data['error']) {
-    errorMessage = `${msg} Error: [ ${error.response.data['error']} ]`;
-  }
-  return errorMessage;
+export const getServiceDetail = (namespace: string, service: string): Promise<ServiceDetailsInfo> => {
+  return Promise.resolve(ServiceData.SERVICE_DETAILS);
```
israel-hdez: A LOT better than doing I/O! :+1:

jotak:
@israel-hdez thanks for reviewing

> Weird to see that the snapshot has a new <ServiceInfoWorkload namespace="istio-system" workloads={Array []} />.

This is because it was previously using the old model (from pre-istio-1.0) where there was no "workloads" but "pods" and "deployments" instead. Now having an empty workload array (instead of undefined) generates this.

mtho11:
Sorry for the late reply, I was out on PTO. These all look like good changes to me. Nice clean things like this up!

PR #1094, opened 2019-03-25
Add multiple destinations on Matching Wizard

lucasponce:
Matching Routing Wizard only allowed to select one destination.

That limited some valid use cases like:

I.e. reviews service with v1, v2, v3 versions:
- Rule [1]: if user == lucas then go to reviews/v1
- Rule [2]: if Any request then go to reviews/v2, reviews/v3

This PRs includes multiple selection of workloads to accept these kind of VS/DR combinations

https://issues.jboss.org/browse/KIALI-2615

image

image

image

image

### DIFF
--- a/src/components/IstioWizards/MatchingRouting/RuleBuilder.tsx
+++ b/src/components/IstioWizards/MatchingRouting/RuleBuilder.tsx
@@ -1,0 +1,104 @@
+import * as React from 'react';
+import { Button, ListView, ListViewItem, TypeAheadSelect } from 'patternfly-react';
+import MatchBuilder from './MatchBuilder';
+import Matches from './Matches';
+import { style } from 'typestyle';
+import { WorkloadOverview } from '../../../types/ServiceInfo';
+import { PfColors } from '../../Pf/PfColors';
+
+type Props = {
+  // MatchBuilder props
+  category: string;
+  operator: string;
+  headerName: string;
+  matchValue: string;
+  isValid: boolean;
+  onSelectCategory: (category: string) => void;
+  onHeaderNameChange: (headerName: string) => void;
+  onSelectOperator: (operator: string) => void;
+  onMatchValueChange: (matchValue: string) => void;
+  onAddMatch: () => void;
+
+  // Matches props
+  matches: string[];
+  onRemoveMatch: (match: string) => void;
+
+  workloads: WorkloadOverview[];
+  routes: string[];
+  onSelectRoutes: (routes: string[]) => void;
+
+  // RuleBuilder
+  validationMsg: string;
+  onAddRule: () => void;
+};
+
+const routeStyle = style({
+  marginTop: 15,
+  // Yes, you are right, this is a CSS trick to adjust style on combined components
+  $nest: {
+    ['.rbt-token .rbt-token-remove-button']: {
+      right: 5
+    }
+  }
+});
+
+const validationStyle = style({
+  marginTop: 15,
+  color: PfColors.Red100
+});
+
+const createStyle = style({
+  marginTop: 90,
+  marginLeft: 20
+});
+
+class RuleBuilder extends React.Component<Props> {
+  constructor(props: Props) {
+    super(props);
+  }
+
+  render() {
+    return (
+      <ListView>
+        <ListViewItem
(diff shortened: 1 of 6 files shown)


lucasponce:
Please @abonas, this is one of the scenarios that were not yet updated, with this PR we could also model these type of rules.

lucasponce:
I put the DNM label, I forgot to add a validation. It shouldn't allow to add a rule on empty route.

hhovsepy:
When typing value only for header name and not header value, it allows to add a rule, but ignores the header. But should say that both header name and value are required.

hhovsepy:
On FF, Add Match button is on the next line:
Screenshot from 2019-03-25 18-31-12

lucasponce:
@hhovsepy thanks for the feedback. I refactored the component and it seems I introduced some regressions.

lucasponce:
@hhovsepy comments addressed, let me know if validations are now correct and style in FF is fixed.
(Chrome and FF are different in 100% view and some time I missed it).

hhovsepy:
thanks @lucasponce,
the Add Rule button is fixed now.
But the header value validation is missing when name is added.

The Routes now by default selects the first Route, I think this can be confusing for users.
What if it selects all Routes by default? Or selects only one, but shows the rest of available ones?

lucasponce:
> But the header value validation is missing when name is added.

Ok, missed.

> The Routes now by default selects the first Route, I think this can be confusing for users.
What if it selects all Routes by default? Or selects only one, but shows the rest of available ones?

Yeah, I was unsure about what option is better, perhaps just leave it empty with the placeholder "Select Workloads" and let user always fill the field.

lucasponce:
@hhovsepy @aljesusg comments addressed.
Please, let me know how it looks like now.

src/components/IstioWizards/MatchingRouting/RuleBuilder.tsx
```diff
@@ -0,0 +1,109 @@
...
+                      defaultSelected={this.props.routes}
+                      options={this.props.workloads.map(wk => wk.name)}
+                      onChange={(r: string[]) => this.props.onSelectRoutes(r)}
+                    />
+                  </FormGroup>
+                </Form>
+              </div>
+              {!this.props.isValid && <div className={validationStyle}>{this.props.validationMsg}</div>}
+            </div>
+          }
```
aljesusg: Please replace empty divs by <> </>

src/components/IstioWizards/MatchingRouting/Rules.tsx
```diff
@@ -0,0 +1,144 @@
...
+          {this.props.rules.length > 1 && (
+            <DropdownKebab key={'move-rule-actions-' + index} id={'move-rule-actions-' + index} pullRight={true}>
+              {index > 0 && <MenuItem onClick={() => this.props.onMoveRule(index, MOVE_TYPE.UP)}>Move Up</MenuItem>}
+              {index + 1 < this.props.rules.length && (
+                <MenuItem onClick={() => this.props.onMoveRule(index, MOVE_TYPE.DOWN)}>Move Down</MenuItem>
+              )}
+            </DropdownKebab>
+          )}
+        </div>
+      );
```
aljesusg: Replace empty divs by <></>

src/components/IstioWizards/MatchingRouting/Rules.tsx
```diff
@@ -0,0 +1,144 @@
...
+              ))}
+              {!isValid && (
+                <div className={validationStyle}>
+                  Match 'Any request' is defined in a previous rule.
+                  <br />
+                  This rule is not accessible.
+                </div>
+              )}
+            </div>
+          }
```
aljesusg: Replace empty divs by <></>

lucasponce:
@aljesusg done
I wonder if there is some tlint rule for this. It sounds pretty automatic check/fix I think that perhaps can be done at format phase.

aljesusg:
> @aljesusg done
> I wonder if there is some tlint rule for this. It sounds pretty automatic check/fix I think that perhaps can be done at format phase.
They are working https://github.com/palantir/tslint-react/issues/147

PR #1095, opened 2019-03-25
KIALI-2556 Fix links to unknown destination service

israel-hdez:
https://issues.jboss.org/browse/KIALI-2556

This prevents creating links for the "unknown" node when it is a destination of traffic.

This fix is for the side panels of the graph and, also for the traffic tabs of the details pages.

### DIFF
--- a/src/components/Details/DetailedTrafficList.tsx
+++ b/src/components/Details/DetailedTrafficList.tsx
@@ -145,7 +145,7 @@ class DetailedTrafficList extends React.Component<DetailedTrafficProps> {
       }
     } else if (NodeType.SERVICE === node.type) {
       icon = <Icon type="pf" name="service" style={style} />;
-      if (!node.isServiceEntry || !node.isInaccessible) {
+      if (!node.isServiceEntry && !node.isInaccessible) {
         name = (
           <Link to={`/namespaces/${encodeURIComponent(node.namespace)}/services/${encodeURIComponent(node.name)}`}>
             {node.name}
--- a/src/pages/Graph/SummaryPanelEdge.tsx
+++ b/src/pages/Graph/SummaryPanelEdge.tsx
@@ -251,8 +251,14 @@ export default class SummaryPanelEdge extends React.Component<SummaryPanelPropTy
       this.metricsPromise = undefined;
     }
 
-    // Just return if the metric types are unset, there is no data, or charts are unsupported
-    if (!destMetricType || !sourceMetricType || !this.hasSupportedCharts(edge) || (!isGrpc && !isHttp && !isTcp)) {
+    // Just return if the metric types are unset, there is no data, destination node is "unknown" or charts are unsupported
+    if (
+      !destMetricType ||
+      !sourceMetricType ||
+      !this.hasSupportedCharts(edge) ||
+      (!isGrpc && !isHttp && !isTcp) ||
+      destData.isInaccessible
+    ) {
       this.setState({
         loading: false
       });
@@ -407,6 +413,15 @@ export default class SummaryPanelEdge extends React.Component<SummaryPanelPropTy
         <>
           <Icon type="pf" name="info" /> Service graphs do not support service-to-service aggregate sparklines. Use the
           workload graph type to observe individual workload-to-service edge sparklines.
+        </>
+      );
+    }
+
+    const data = nodeData(edge.target());
+    if (data.isInaccessible) {
+      return (
+        <>
+          <Icon type="pf" name="info" /> Sparkline charts cannot be shown because the destination is inaccessible.
         </>
       );
     }
--- a/src/pages/Graph/SummaryPanelNode.tsx
+++ b/src/pages/Graph/SummaryPanelNode.tsx
@@ -126,6 +126,12 @@ export default class SummaryPanelNode extends React.Component<SummaryPanelPropTy
       return;
     }
 
+    // If destination node is inaccessible, we cannot query the data.
+    if (data.isInaccessible) {
+      this.setState({ loading: false });
+      return;
+    }
+
     let promiseOut: Promise<Response<Metrics>> = Promise.resolve({ data: { metrics: {}, histograms: {} } });
     let promiseIn: Promise<Response<Metrics>> = Promise.resolve({ data: { metrics: {}, histograms: {} } });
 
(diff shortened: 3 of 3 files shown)


israel-hdez:
@jshaughn I updated this. Your review wasn't dropped. Do you want to re-check?

src/components/Details/DetailedTrafficList.tsx
```diff
@@ -145,7 +145,7 @@ class DetailedTrafficList extends React.Component<DetailedTrafficProps> {
       }
     } else if (NodeType.SERVICE === node.type) {
       icon = <Icon type="pf" name="service" style={style} />;
-      if (!node.isServiceEntry || !node.isInaccessible) {
+      if (!node.isServiceEntry && !node.isInaccessible) {
```
jshaughn: This doesn't look right to me.  Are you sure about this?

src/pages/Graph/SummaryPanelEdge.tsx
```diff
@@ -251,8 +251,14 @@ export default class SummaryPanelEdge extends React.Component<SummaryPanelPropTy
...
 
-    // Just return if the metric types are unset, there is no data, or charts are unsupported
-    if (!destMetricType || !sourceMetricType || !this.hasSupportedCharts(edge) || (!isGrpc && !isHttp && !isTcp)) {
+    // Just return if the metric types are unset, there is no data, destination node is "unknown" or charts are unsupported
+    if (
+      !destMetricType ||
+      !sourceMetricType ||
+      !this.hasSupportedCharts(edge) ||
+      (!isGrpc && !isHttp && !isTcp) ||
+      destData.isInaccessible
```
jshaughn: I don't think this is right.  You can have traffic to an inaccessible node.  Inaccessible just means that the user can't navigate to or see it.  You may be able to just get rid of this condition. If that causes a problem rendering the edge chart then you may need to update getNodeMetricType() such that for nodeType SERVICE the metricType is WORKLOAD when the service is unknown.

src/pages/Graph/SummaryPanelEdge.tsx
```diff
@@ -411,6 +417,15 @@ export default class SummaryPanelEdge extends React.Component<SummaryPanelPropTy
...
 
+    const data = nodeData(edge.target());
+    if (data.isInaccessible) {
+      return (
+        <>
+          <Icon type="pf" name="info" /> Sparkline charts cannot be shown because the destination is inaccessible.
+        </>
+      );
+    }
+
```
jshaughn: Again, I'm not sure this is true.  Maybe I am wrong, but I would expect to be able to see the edge traffic to an inaccessible node.

hhovsepy:
Verified:
Screenshot from 2019-03-27 13-53-00

src/pages/Graph/SummaryPanelEdge.tsx
israel-hdez: Well... Tried this with a more "real" scenario. What I did is to deploy a traffic generator to a side "myproject" namespace and make a request to productpage in bookinfo namespace. Then, configured Kiali with reduced permissions giving access only to "myproject". So, I get this graph "myproject" namespace:

image

So, _productpage_ is inaccessible. What I see is that when I select the traffic-generator -> productpage edge, the request that goes to the backend is:

`http://localhost:3000/api/namespaces/bookinfo/services/productpage/metrics?queryTime=1553705660&duration=60&step=30&rateInterval=30s&filters[]=request_count&filters[]=request_duration&filters[]=request_error_count&quantiles[]=0.5&quantiles[]=0.95&quantiles[]=0.99&byLabels[]=source_workload&direction=inbound&reporter=source&requestProtocol=http`

You can note that the request asks for data in the `bookinfo` namespace. But that one is inaccessible. So, the back-end correctly replies with an error:

```js
{
  "error":
     "Cannot access namespace data: projects.project.openshift.io \"bookinfo\" is forbidden: User \"system:serviceaccount:istio-system:kiali-service-account\" cannot get projects.project.openshift.io in the namespace \"bookinfo\": no RBAC policy matched"}
```

So, I don't see this wrong since we cannot query the data for the sparklines. Though, the other data of the side panel is available:

image

src/components/Details/DetailedTrafficList.tsx
israel-hdez: I think it's right. The `SummaryLink.isInaccessible` is:
```js
const isInaccessible = (data: NodeData): boolean => {
  return data.isInaccessible || data.isServiceEntry !== undefined;
};
```

It is the same condition, but negated - because I want the links if accessible.

src/pages/Graph/SummaryPanelEdge.tsx
israel-hdez: Well... Tried this with a more "real" scenario. What I did is to deploy a traffic generator to a side "myproject" namespace and make a request to productpage in bookinfo namespace. Then, configured Kiali with reduced permissions giving access only to "myproject". So, I get this graph "myproject" namespace:

image

So, _productpage_ is inaccessible. What I see is that when I select the traffic-generator -> productpage edge, the request that goes to the backend is:

`http://localhost:3000/api/namespaces/bookinfo/services/productpage/metrics?queryTime=1553705660&duration=60&step=30&rateInterval=30s&filters[]=request_count&filters[]=request_duration&filters[]=request_error_count&quantiles[]=0.5&quantiles[]=0.95&quantiles[]=0.99&byLabels[]=source_workload&direction=inbound&reporter=source&requestProtocol=http`

You can note that the request asks for data in the `bookinfo` namespace. But that one is inaccessible. So, the back-end correctly replies with an error:

```js
{
  "error":
     "Cannot access namespace data: projects.project.openshift.io \"bookinfo\" is forbidden: User \"system:serviceaccount:istio-system:kiali-service-account\" cannot get projects.project.openshift.io in the namespace \"bookinfo\": no RBAC policy matched"}
```

So, I don't see this wrong since we cannot query the data for the sparklines. Though, the other data of the side panel is available:

image

src/pages/Graph/SummaryPanelEdge.tsx
israel-hdez: See the other reply... I think the same applies to this and is probably correct to have the `isInaccessible` condition to avoid sending a request that will return with error.

PR #1096, opened 2019-03-26
KIALI-2580 Add filtering and sorting by mTLS in overview

xeviknal:
** Describe the change **
Adding sorting and filtering by mTLS status to overview page.

** Issue reference **
https://issues.jboss.org/browse/KIALI-2580

** Screenshot **
Screen recording (11)

### DIFF
--- a/src/pages/Overview/FiltersAndSorts.ts
+++ b/src/pages/Overview/FiltersAndSorts.ts
@@ -70,6 +70,24 @@ export namespace FiltersAndSorts {
         // default comparison fallback
         return a.name.localeCompare(b.name);
       }
+    },
+    {
+      id: 'mtls',
+      title: 'mTLS',
+      isNumeric: false,
+      param: 'm',
+      compare: (a: NamespaceInfo, b: NamespaceInfo) => {
+        if (a.tlsStatus && b.tlsStatus) {
+          return a.tlsStatus.status.localeCompare(b.tlsStatus.status);
+        } else if (a.tlsStatus) {
+          return -1;
+        } else if (b.tlsStatus) {
+          return 1;
+        }
+
+        // default comparison fallback
+        return a.name.localeCompare(b.name);
+      }
     }
   ];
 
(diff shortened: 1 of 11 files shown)


src/pages/Overview/FiltersAndSorts.ts
```diff
@@ -1,16 +1,56 @@
...
+      id: 'enabled',
+      title: 'Enabled'
+    },
+    {
+      id: 'partiallyEnabled',
+      title: 'Partially Enabled'
+    },
+    {
+      id: 'notEnabled',
+      title: 'Not Enabled'
```
jotak: Should it be rather "Disabled" or "Not Enabled" ? (I guess we have similar labels in other places, no?)

src/pages/Overview/FiltersAndSorts.ts
```diff
@@ -1,16 +1,56 @@
...
+  export const mtlsFilter: FilterTypeWithFilter<NamespaceInfo> = {
+    id: 'mtls',
+    title: 'mTLS status',
+    placeholder: 'Filter by mTLS status',
+    filterType: 'select',
+    action: FILTER_ACTION_APPEND,
+    filterValues: mtlsValues,
+    filter: (namespaces: NamespaceInfo[], filters: ActiveFilter[]) => {
+      return namespaces.filter(
+        ns => ns.tlsStatus && filters.some(f => statusMap.get(ns.tlsStatus!.status) === f.value)
```
jotak: Because of `ns.tlsStatus!.status`, could it be a problem if the filtering function is called while not all asynchronous data have been fetched? (in which case `ns.tlsStatus` would be undefined i guess)

xeviknal: Good point. I think that I won't be a problem. Having `ns.tlsStatus` at the beginning of the statement will prevent the execution of the `some` expression, so I guess that non undefined access error will be rose.

src/pages/Overview/FiltersAndSorts.ts
xeviknal: yep, I think it is more clear to have Disabled.

jotak: Well I see several lines below that "Not Enabled" and "Disabled" are actually two different things?

xeviknal:
@jotak comments approached!

src/pages/Overview/FiltersAndSorts.ts
xeviknal: Technically yes. It means that the NS is explicitly disabled (adding one DR and Policy). And not enabled, means that for that specific NS the user haven't changed the mTLS settings (not enabled). The difference here is made because we need to merge NS-wide mtls status and Mesh-wide mtls status. 
I think that at the end, we want to tell the user whether the ns is using mTLS or not.

hhovsepy:
Verified.

PR #1097, opened 2019-03-27
KIALI-2624 Prevent UI crash even if validation is missing

burmanm:
** Describe the change **

These modifications check if validations is empty from the backend result and then goes to avoid displaying those items that require validation. All the pages should work, but the validation result is missing or text is displayed indicating that the validations that were expected could not be found.

Related to issue 936 in kiali repository. I could not repeat the issue at all and as such, one will need a modified kiali backend to test this. I have posted a docker image to ``docker.io/burmanm/kiali:kiali-2624`` that returns empty validations.

** Issue reference **

KIALI-2624 (I hope, JIRA is not working)

### DIFF
--- a/src/pages/ServiceDetails/ServiceInfo/ServiceInfoVirtualServices.tsx
+++ b/src/pages/ServiceDetails/ServiceInfo/ServiceInfoVirtualServices.tsx
@@ -85,6 +85,14 @@ class ServiceInfoVirtualServices extends React.Component<ServiceInfoVirtualServi
     };
   }
 
+  hasValidations(virtualService: VirtualService): boolean {
+    // This is insane, but doing return to the clause inside the if will cause compiler failure
+    if (this.props.validations && this.props.validations[virtualService.metadata.name]) {
+      return true;
+    }
+    return false;
+  }
+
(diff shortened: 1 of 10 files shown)


src/pages/ServiceDetails/ServiceInfo/ServiceInfoVirtualServices.tsx
```diff
@@ -85,6 +85,14 @@ class ServiceInfoVirtualServices extends React.Component<ServiceInfoVirtualServi
     };
   }
 
+  hasValidations(virtualService: VirtualService): boolean {
+    // This is insane, but doing return to the clause inside the if will cause compiler failure
+    if (this.props.validations && this.props.validations[virtualService.metadata.name]) {
```
xeviknal: yep, I guess this is because `this.props.validations['vs-name']` returns a non-boolean value.
If you wanted to have it in one line, I guess it would be enough with addint the `!!` operator that converts whatever variable to a boolean. So this should work:
`return this.props.validations && !!this.props.validations[virtualService.metadata.name];`
Same thing would apply in the file right above: `ServiceInfoDestinationRules.tsx`

PR #1098, opened 2019-03-29
KIALI-2488 Add dashboards logos

jotak:
JVM :
jvm

MicroProfile:
mp


hhovsepy:
cc @skondkar

abonas:
@jotak is there a tooltip with the runtime name when standing on the logo? the location of the logo seems a bit unnatural, were UX consulted?

jotak:
@abonas yes, there's a tooltip.
The position of the logo hasn't changed here, but it was indeed discussed previously: https://github.com/kiali/kiali-ui/pull/996 where I ended up with the current proposal (without "Runtimes" row title) and leaving the field open for a global redesign of the workloads details page (to better fill all the blank space)

PR #1085, opened 2019-03-21
Add new tslint rule: 'no-literal-string' to clean up type-safe issues…

mtho11:
…. Disallows obj["property"] (which has no type safety) in favor of obj.property. We are totally bypassing Typescript's type-safety here.

WDYT? This open to the team to comment.

PR #1101, opened 2019-04-01
Kiali 577

jshaughn:
Add pod log display via the workload detail pod's display.

*** REQUIRES SERVER PR https://github.com/kiali/kiali/pull/963

kiali-577

### DIFF
--- a/src/pages/WorkloadDetails/WorkloadInfo.tsx
+++ b/src/pages/WorkloadDetails/WorkloadInfo.tsx
@@ -116,7 +116,13 @@ class WorkloadInfo extends React.Component<WorkloadInfoProps, WorkloadInfoState>
                   </Nav>
                   <TabContent>
                     <TabPane eventKey={'pods'}>
-                      {pods.length > 0 && <WorkloadPods pods={pods} validations={this.props.validations!.pod} />}
+                      {pods.length > 0 && (
+                        <WorkloadPods
+                          namespace={this.props.namespace}
+                          pods={pods}
+                          validations={this.props.validations!.pod}
+                        />
+                      )}
                     </TabPane>
                     <TabPane eventKey={'services'}>
                       {services.length > 0 && <WorkloadServices services={services} namespace={this.props.namespace} />}
--- a/src/pages/WorkloadDetails/WorkloadInfo/WorkloadPodLogs.tsx
+++ b/src/pages/WorkloadDetails/WorkloadInfo/WorkloadPodLogs.tsx
@@ -4,6 +4,7 @@ import { Pod, PodLogs } from '../../../types/IstioObjects';
 import { getPodLogs, Response } from '../../../services/Api';
 import { CancelablePromise, makeCancelablePromise } from '../../../utils/CancelablePromises';
 import { ToolbarDropdown } from '../../../components/ToolbarDropdown/ToolbarDropdown';
+import { DurationInSeconds } from '../../../types/Common';
 
 export interface WorkloadPodLogsProps {
   namespace: string;
@@ -189,7 +190,7 @@ export default class WorkloadPodLogs extends React.Component<WorkloadPodLogsProp
     return { container: containers[0], containerOptions: options };
   };
 
-  private fetchLogs = (namespace: string, podName: string, container: string, duration: number) => {
+  private fetchLogs = (namespace: string, podName: string, container: string, duration: DurationInSeconds) => {
     const sinceTime = Math.floor(Date.now() / 1000) - duration;
     const promise: Promise<Response<PodLogs>> = getPodLogs(namespace, podName, container, sinceTime);
     this.loadPodLogsPromise = makeCancelablePromise(Promise.all([promise]));
--- a/src/pages/WorkloadDetails/WorkloadInfo/WorkloadPodLogs.tsx
+++ b/src/pages/WorkloadDetails/WorkloadInfo/WorkloadPodLogs.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
-import { Button, Icon, Table, Toolbar } from 'patternfly-react';
+import { Button, Icon, Toolbar } from 'patternfly-react';
 import { Pod, PodLogs } from '../../../types/IstioObjects';
-import { getPod, getPodLogs, Response } from '../../../services/Api';
+import { getPodLogs, Response } from '../../../services/Api';
 import { CancelablePromise, makeCancelablePromise } from '../../../utils/CancelablePromises';
 import { ToolbarDropdown } from '../../../components/ToolbarDropdown/ToolbarDropdown';
 
@@ -10,19 +10,22 @@ export interface WorkloadPodLogsProps {
   pods: Pod[];
 }
 
+interface ContainerInfo {
+  container: string;
+  containerOptions: object;
+}
+
 interface WorkloadPodLogsState {
-  container?: string;
-  containers?: Object;
+  containerInfo?: ContainerInfo;
   duration: string; // DurationInSeconds
-  loadingPod: boolean;
-  loadingPodError?: string;
   loadingPodLogs: boolean;
-  pod?: Pod;
+  loadingPodLogsError?: string;
+  podValue?: number;
   podLogs?: PodLogs;
 }
 
 const DurationDefault = '300';
-const Durations = {
+const DurationOptions = {
   '60': 'Last 1m',
   '300': 'Last 5m',
   '600': 'Last 10m',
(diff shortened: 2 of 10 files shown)


jshaughn:
@pilhuhn @abonas , FYI: Check out the video above or build the prs and let me know if you have any feedback.

lucasponce:
I think a "Log" tab will be more consistent in how this page is organizing info.
In the Log tab there can be a doble dropdown to select first pod and second container.

In that way, traffic, metrics and log and defined at same level.
Also, I'd also add it to the services/app level too, as we can get pod names from workloads, and all pages should get that info.

Also, note that a pod row may group all pods for same replica, so it would fail into the same scenario.

aljesusg:
I agree with @lponce I think that we should avoid popups, have this log viewer in a tab like a component will be great and reusable to see Kiali logs for example in a future admin tab ^^

lucasponce:
@aljesusg yeah, I think this feature would be better to be shown also under services, where user can also check metrics/traces/logs just changing the tab with the same time interval.

In that way, it's easy to start to correlate, rather than going to an specific popup.

Also, next step is to process and aggregate info, so the Tab sounds better than a popup.

(We use popups for wizard, but I guess that's a different use case than this, as in that case we use it just for the user to introduce some info).

src/services/Api.ts
```diff
@@ -406,7 +422,7 @@ export const getErrorMsg = (msg: string, error: AxiosError) => {
     } else if (error.response.statusText) {
       errorMessage = `${msg}, Error: [ ${error.response.statusText} ]`;
       if (error.response.status === 401) {
-        errorMessage += ' Have your session expired? Try logging again.';
+        errorMessage += ' Has your session expired? Try logging again.';
```
jotak: Good catch, I'm going to fix in typo PR to not have to wait for this one

src/pages/WorkloadDetails/WorkloadInfo/WorkloadPodLogs.tsx
```diff
@@ -0,0 +1,238 @@
...
+  };
+
+  private fetchPod = (namespace: string, podName: string) => {
+    const promise: Promise<Response<Pod>> = getPod(namespace, podName);
+    this.loadPodPromise = makeCancelablePromise(Promise.all([promise]));
+    this.loadPodPromise.promise
+      .then(response => {
+        const pod = response[0].data;
+        const containers: Object = {};
+        if (pod.containers) {
```
jotak: i think code could be a little bit more concise if `pod.containers` and `pod.istioContainers` were not optional (empty arrays instead) - which means to initialize them explicitly from backend rather than relying on golang's default value for slices.

jshaughn:
@lucasponce, @aljesusg, I will investigate a Logs tab.   We may still need a dialog if we decide to to show logs from the graph, as there isn't really an alternate option.   But I will try and move the "meat" to a component that could be wrapped in different ways.  Thanks for the feedback,

lucasponce:
> We may still need a dialog if we decide to to show logs from the graph, as there isn't really an alternate option.

That deserves another discussion. Metrics/Traces/Logs are consolidated at detail level I think.

So, I'd really find strange to have some info available and not others.

To me the use case is to go to the graph to detect the problem, and next step of the user would be to go to the details for more investigation to see the details on metrics, traces, runtime dashboard, and now logs. Sharing the same timeline between those tabs, and just moving from one context to another in one click.

So, that's why I think this feature should be presented as a tab under app/service/workload detail level.

jshaughn:
> That deserves another discussion. Metrics/Traces/Logs are consolidated at detail level I think.

@lucasponce Right, I agree, that is why I am now working on a tab.  I was only saying that if some powers want to be able to spy logs from the graph that a modal would likely be best.

src/pages/WorkloadDetails/WorkloadInfo.tsx
```diff
@@ -116,7 +116,13 @@ class WorkloadInfo extends React.Component<WorkloadInfoProps, WorkloadInfoState>
                   </Nav>
                   <TabContent>
                     <TabPane eventKey={'pods'}>
-                      {pods.length > 0 && <WorkloadPods pods={pods} validations={this.props.validations!['pod']} />}
+                      {pods.length > 0 && (
+                        <WorkloadPods
+                          namespace={this.props.namespace}
+                          pods={pods}
+                          validations={this.props.validations!['pod']}
```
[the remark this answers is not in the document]
jshaughn: @mtho11 It's outside of this PR but I changed it as well as a few other instances in this file.  You can recheck.  But I stopped short of replacing all of the literals here as really all this PR needed was to add a property.

src/pages/WorkloadDetails/WorkloadInfo/WorkloadPodLogs.tsx
```diff
@@ -0,0 +1,238 @@
...
+export interface WorkloadPodLogsProps {
+  className?: string;
+  namespace: string;
+  podName: string;
+  onClose: () => void;
+}
+
+interface WorkloadPodLogsState {
+  container?: string;
+  containers?: Object;
```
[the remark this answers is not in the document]
jshaughn: @mtho11 That was basically a typo, I've changed it now from `Object` to `object`.

jshaughn:
@lucasponce, @aljesusg Here is the latest in Workload detail. You can see it is tab-based and has pod and container selection:

image

Before trying to add a `Logs` tab to Service and App details I just wanted to make sure you guys think it's really necessary.  To me it seems natural to associate logs with a workload as the workload is basically backed by its pods.  Service and apps can be associated with workloads but they are more abstract.  For me it seems OK in the service or app view to drill into any workload of interest to see its pod logs.

From a coding perspective there is a bit of work to do to add a Logs tab to the Service or App details. Neither carries the complete workloads information (no pod info), only the necessary boilerplate info to satisfy their workload display needs.

If we do want to add a logs tab to these detail displays then we should agree on how to do it.  I think there are two main options.  Option one would be to fetch the full workload info in the parent component (perhaps only if they select the logs tab) and supply the resulting set of pods to the logs component (this is what the logs component expects today, a list of pods).  Option two would be to enhance the logs component to optionally take a list of workload names or a list of pods.  If workload names were supplied we'd then present the user with a 3-level selection:

workload [dropdown]   pod [dropdown]   container [dropdown]

The user would first need to select a workload, and then we could populate the pod and container dropdowns.

What are your opinions?  If we provide logs only from the workload details then the existing PRs may be close to providing what we need.

src/pages/WorkloadDetails/WorkloadInfo/WorkloadPodLogs.tsx
jshaughn: I started making this change but the implications to existing UI code became a bit overwhelming as there is quite a lot of code already expecting that these arrays may be unset.  So I'll suffer with the two additional guards here...

src/pages/WorkloadDetails/WorkloadInfo.tsx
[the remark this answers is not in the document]
mtho11: @jshaughn ok that's cool. You don't need to replace all because that is what this PR is for: https://github.com/kiali/kiali-ui/pull/1085

src/pages/WorkloadDetails/WorkloadInfo/WorkloadPodLogs.tsx
```diff
@@ -0,0 +1,238 @@
...
+      container: undefined,
+      containers: undefined,
+      loadingPod: true,
+      loadingPodError: undefined,
+      pod: undefined,
+      podLogs: undefined
+    });
+  };
+
+  private fetchLogs = (namespace: string, podName: string, container: string, duration: number) => {
```
mtho11:
```suggestion
  private fetchLogs = (namespace: string, podName: string, container: string, duration: DurationInSeconds) => {
```

jshaughn: yup, I will make the change.

lucasponce:
> Service and apps can be associated with workloads but they are more abstract. For me it seems OK in the service or app view to drill into any workload of interest to see its pod logs.

To me, the key is to offer to the user the "correlation" key. The tab is perfect to me. But as a user I'd like to see at least at service, app level traces.

That can be do in a following PR, no need to increase this one.

But I guess we need to move user to "correlation" scenarios, also, traces tab only happens at service level but I also wonder we would need also to the app/workloads (with the related services).

I think for traces is more simple, as at the end is just to populate which is the service(s) you can choose.

But as a user, I will dig into the tabs to get more info, so I'd like to see that info close.

jshaughn:
> But I guess we need to move user to "correlation" scenarios, also, traces tab only happens at service level but I also wonder we would need also to the app/workloads (with the related services).

@lucasponce I agree that this feature is not really what we may imagine as logs observability.  It is intended mainly as a first step to prove that we can fetch and display logs in some capacity.  The main benefit here is only to consolidate information to help the user not have to access logs from a separate console.

I agree that correlation would be a much more interesting view, where perhaps we have metric and log and maybe trace information in a combined view.  Also, log metrics, search, or "hot spots".  These additions should certainly be a priority in the next 3-6 month time frame.

jotak:
To me, the Workloads Logs is the most obvious and logical first step. Logs are produced by runtimes, and runtimes are in Workloads. So it makes sense.

For more correlation, maybe we need to extend our current service view - not just for logs, but for everything. Service metrics are also strictly scoped on the *service* aka inbound interface, there is no direct correlation with the underlying workloads. Currently we need to navigate to workloads to see more.

lucasponce:
> For more correlation, maybe we need to extend our current service view - not just for logs, but for everything. Service metrics are also strictly scoped on the service aka inbound interface, there is no direct correlation with the underlying workloads. Currently we need to navigate to workloads to see more. On the other hand, this could also be seen as the role for App: an aggregated view that permits more correlation.

Exactly. 

I think in the same way, I guess we have (more or less) progressed into the initial phase to bring the elements that participate in the scenario and now it's time to put it together.

My point in all of this: I'd prefer that user understands from first step that Kiali is about observability and correlation, and any overlapping with other generic k8s console is circumstantial. 

For example, do we want to operate workloads ? But I guess not at least at the point we are.

lucasponce:
> I agree that correlation would be a much more interesting view, where perhaps we have metric and log and maybe trace information in a combined view.

+1
I was also thinking about this. Perhaps a more rich chart built with metrics + traces (+ logs).

mtho11:
Having metrics charts, logs, health information is all valuable -- but the real win is in the **correlation** of all this information. Having a single screen with all this information available. All visible in the same timeframe.  This is what is going to help someone understand what happened during an outage or degradation quickly.

src/pages/WorkloadDetails/WorkloadInfo/WorkloadPodLogs.tsx
```diff
@@ -0,0 +1,224 @@
...
+
+  private handleRefresh = () => {
+    const pod = this.props.pods[this.state.podValue!];
+    this.fetchLogs(this.props.namespace, pod.name, this.state.containerInfo!.container, Number(this.state.duration));
+  };
+
+  private getContainerInfo = (pod: Pod): ContainerInfo => {
+    const containers: string[] = [];
+    if (pod.containers) {
+      pod.containers.forEach(c => {
```
mtho11: The `forEach` and `push` is an anti-pattern (mutable and not as direct as `const`... `map` immutable structure)

src/pages/WorkloadDetails/WorkloadInfo/WorkloadPodLogs.tsx
```diff
@@ -0,0 +1,224 @@
...
+                  tooltip={'Time range for graph data'}
+                />
+                <span style={{ paddingLeft: '0.5em' }}>
+                  <Button id={'wpl_refresh'} disabled={!this.state.podLogs} onClick={() => this.handleRefresh()}>
+                    <Icon name="refresh" />
+                  </Button>
+                </span>
+              </Toolbar.RightContent>
+            </Toolbar>
+            <textarea
```
mtho11: Please move all of the inline styles to type-style.

src/pages/WorkloadDetails/WorkloadInfo/WorkloadPodLogsDraggable.tsx
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  render() {
+    const className = this.props.className ? this.props.className : '';
+
+    return (
+      <Draggable handle="#wpl_header">
+        <div
+          className={`modal-content ${className}`}
+          style={{
```
mtho11: Again move inline style to typestyle

src/services/Api.ts
```diff
@@ -398,6 +399,21 @@ export const getWorkload = (namespace: string, name: string) => {
   return newRequest<Workload>(HTTP_VERBS.GET, urls.workload(namespace, name), {}, {});
 };
 
+export const getPod = (namespace: string, name: string) => {
+  return newRequest<Pod>(HTTP_VERBS.GET, urls.pod(namespace, name), {}, {});
+};
+
+export const getPodLogs = (namespace: string, name: string, container?: string, sinceTime?: number) => {
+  const params: Object = {};
```
mtho11: `Object`?

src/types/IstioObjects.ts
```diff
@@ -80,13 +80,18 @@ export interface Pod {
...
+  containers?: ContainerInfo[];
   istioContainers?: ContainerInfo[];
   istioInitContainers?: ContainerInfo[];
   status: string;
   appLabel: boolean;
   versionLabel: boolean;
 }
 
+export interface PodLogs {
+  logs?: string;
```
mtho11: Since this is a really large string and it is more interesting than just a string how about assigning it its own type alias like `PodLogString`?

src/pages/WorkloadDetails/WorkloadInfo/WorkloadPodLogs.tsx
jshaughn: Ok, I’ll move to type style

src/pages/WorkloadDetails/WorkloadInfo/WorkloadPodLogs.tsx
jshaughn: Ok, Thanks for the tip, will change .

src/services/Api.ts
jshaughn: Good catch

src/types/IstioObjects.ts
jshaughn: Good idea.

pilhuhn:
I missed all the discussion, but am totally happy with the outcome (via screenshots).

Great work!!

PR #1104, opened 2019-04-03
KIALI-2666 Show open lock whenever traffic is under 100%

xeviknal:
** Describe the change **
In the scenario of having mTLS enabled mesh-wide, the open locks on edges should be visible whenever there is traffic not using mTLS (from 0% to 99% of traffic that uses mTLS).

** Issue reference **
https://issues.jboss.org/browse/KIALI-2666

** Backwards compatible? **
yes

** Screenshot **
mTLS not enabled globally. Graph shows locks whenever there is traffic with using mtls (from 1% to 100% of the traffic using mTLS).
Screen recording (18)

mTLS enabled for service mesh. Graph should show open locks whenever there is traffic **not** using mTLS (from 0% to 99% of traffic using mTLS)
Screen recording (17)


hhovsepy:
Verified both scenarios

PR #1105, opened 2019-04-04
KIALI-2497 Show message to user at first load if secret is missing

israel-hdez:
https://issues.jboss.org/browse/KIALI-2497

These changes will show the _secret missing_ message to the user without the need to try to login. Taking advantage that a request is made to the /api/auth/info endpoint as initialization to check if secret configuration is pending.

Related backend PR: https://github.com/kiali/kiali/pull/979

### DIFF
--- a/src/pages/Login/LoginPage.tsx
+++ b/src/pages/Login/LoginPage.tsx
@@ -79,7 +79,15 @@ export default class LoginPage extends React.Component<LoginProps, LoginState> {
                 <Col sm={10} smOffset={1} md={8} mdOffset={2} lg={8} lgOffset={2}>
                   <div className={'card-pf'}>
                     <header className={'login-pf-header'} />
-                    {this.props.status === LoginStatus.error && <Alert>{this.props.message}</Alert>}
+                    {authenticationConfig.secretMissing && (
+                      <Alert>
+                        The Kiali secret is missing. Users are prohibited from accessing Kiali until an administrator
+                        creates a valid secret. Please refer to the Kiali documentation for more details.
+                      </Alert>
+                    )}
+                    {!authenticationConfig.secretMissing && this.props.status === LoginStatus.error && (
+                      <Alert>{this.props.message}</Alert>
+                    )}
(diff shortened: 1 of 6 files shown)


src/pages/Login/LoginPage.tsx
```diff
@@ -79,7 +79,16 @@ export default class LoginPage extends React.Component<LoginProps, LoginState> {
                 <Col sm={10} smOffset={1} md={8} mdOffset={2} lg={8} lgOffset={2}>
                   <div className={'card-pf'}>
                     <header className={'login-pf-header'} />
-                    {this.props.status === LoginStatus.error && <Alert>{this.props.message}</Alert>}
+                    {authenticationConfig.secretMissing && (
+                      <Alert>
+                        The Kiali secret is missing. Users are prohibited from accessing Kiali until an administrator
+                        creates a valid secret and restarts Kiali. Please refer to the Kiali documentation for more
```
jmazzitelli: while we are here, we can change this message slightly. We no longer need to restart Kiali. Once you create the secret, Kiali will eventually autodetect it and use the credentials it finds.

So - remove the words "and restarts Kiali" here.

```suggestion
                        creates a valid secret. Please refer to the Kiali documentation for more
```

israel-hdez: Agreed.
Change is applied.

jmazzitelli:
I blame myself for not actually trying this, but this caused a regression and should be fixed. See [my JIRA comment here](https://issues.jboss.org/browse/KIALI-2497?focusedCommentId=13719467&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13719467).

src/pages/Login/LoginPage.tsx
jmazzitelli: I say "we no longer need to restart Kiali" but this PR actually broke that feature and causes us to have to restart Kiali (at least that is what I am seeing). This is a regression. We need to fix that.

PR #1109, opened 2019-04-08
KIALI-2423 Add links from traffic tab to metrics

israel-hdez:
https://issues.jboss.org/browse/KIALI-2423

This will let the user navigate to the detailed metrics. Filters are preset depending on the clicked row.

image

### DIFF
--- a/src/components/Details/DetailedTrafficList.tsx
+++ b/src/components/Details/DetailedTrafficList.tsx
@@ -5,6 +5,7 @@ import { Link } from 'react-router-dom';
 import { NodeType, ProtocolTraffic } from '../../types/Graph';
 import { Direction } from '../../types/MetricsOptions';
 import { REQUESTS_THRESHOLDS } from '../../types/Health';
+import history, { URLParam } from '../../app/History';
 
 type DetailedTrafficProps = {
   direction: Direction;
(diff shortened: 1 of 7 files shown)


israel-hdez:
@jotak Well... this first try is a naive approach.

Once I started working on this, I realized that not all rows can have a "View metrics" link. In the detailed metrics tabs, there are no filters for _workloads_, so I cannot build a link to view that kind of metrics. Also, there are filters for _services_ only in the outbound metrics tab... Apps are the only entities that can be linked both for inbound and outbound.

Well... looks like we would need Jiras to add the required filters to fully implement KIALI-2423... or perhaps there is an alternative way to achieve the same result with the existent filters?

Any thoughts?

lucasponce:
@israel-hdez +1 from my side to add links to metrics.
About workloads, not sure if understand the question, the existing filter as
```
https://kiali-istio-system.127.0.0.1.nip.io/console/namespaces/bookinfo/workloads/productpage-v1?tab=in_metrics
```
Doesn't work here to build the link ?

israel-hdez:
@lucasponce I've updated this PR. I used your suggestion for the link.

What I was trying, is to avoid a context switch. For example, if you were in the "app1" details page, what I wanted was to just switch to the Inbound or Outbound tab of the "app1" details page with filters applied so that you can see metrics for that particular connection. I understood this was the idea of the Jira., but this is only possible for some cases because we currently don't have enough filters in the metrics pages.

When that is not possible I'm building the link the way you suggested - but, metrics shown will be overall inbound/outbound traffic instead of traffic for the particular connection.

PR is ready, please check if this satisfies the requirement... Although, I think it could be a little confusing that in some cases you get traffic metrics for the connection, and other cases you get overall traffic metrics...

PR #1111, opened 2019-04-10
KIALI-2454 Add service validations information to the UI

burmanm:
** Describe the change **

Adds validation status for service validations in the service list. Waits for further information from kiali-design #118

** Issue reference **

KIALI-2454

** Screenshot **

In the kiali-design ticket.

### DIFF
--- a/src/pages/ServiceList/ItemDescription.tsx
+++ b/src/pages/ServiceList/ItemDescription.tsx
@@ -5,12 +5,15 @@ import { ServiceHealth } from '../../types/Health';
 import { DisplayMode, HealthIndicator } from '../../components/Health/HealthIndicator';
 import MissingSidecar from '../../components/MissingSidecar/MissingSidecar';
 import { PromisesRegistry } from '../../utils/CancelablePromises';
+import { ObjectValidation } from '../../types/IstioObjects';
+import { ConfigIndicator } from '../../components/ConfigValidation/ConfigIndicator';
 
 interface Props {
   item: ServiceListItem;
 }
 interface State {
   health?: ServiceHealth;
+  validation: ObjectValidation;
 }
 
 export default class ItemDescription extends React.PureComponent<Props, State> {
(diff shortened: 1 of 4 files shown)


hhovsepy:
@burmanm  does it show the validation now, or still WIP?

burmanm:
It shows it in the "service list", not in the details.

src/pages/ServiceList/ItemDescription.tsx
```diff
@@ -5,20 +5,23 @@ import { ServiceHealth } from '../../types/Health';
...
+  validation: ObjectValidation;
 }
 
 export default class ItemDescription extends React.PureComponent<Props, State> {
   private promises = new PromisesRegistry();
 
   constructor(props: Props) {
     super(props);
-    this.state = { health: undefined };
+    this.state = { health: undefined, validation: {} as ObjectValidation };
```
xeviknal: Perhaps I am not fully following this. I see that the validation object is defined in the state interface and there isn't any instruction writting any value different thatn {}. Who is the one adding the actual Validations here?

src/pages/ServiceList/ItemDescription.tsx
```diff
@@ -57,6 +60,14 @@ export default class ItemDescription extends React.PureComponent<Props, State> {
         <Col xs={12} sm={12} md={4} lg={4}>
           {!this.props.item.istioSidecar && <MissingSidecar />}
         </Col>
+        <Col xs={12} sm={12} md={4} lg={4}>
+          <strong>Config: </strong>{' '}
+          <ConfigIndicator
+            id={this.props.item.name + '-config-validation'}
+            validations={[this.state.validation]}
```
xeviknal: Won't we always have {} values? I guess that having the following would be enough to fix this:
```typescript
validations={[this.props.item.validation]}
```
And the state could be out of the state.
Does this make sense?

burmanm: Yeah, it was supposed to be that this.props.item.validation and no state at all, that was just the UI testing code.

burmanm:
Rebased.

hhovsepy:
Error config icon is in service list:
Screenshot from 2019-04-17 14-58-04

PR #1113, opened 2019-04-12
KIALI-2663 Prevent duplicate calls to this.updateListItems()

burmanm:
** Describe the change **

componentDidMount() was calling this.updateListItems() which caused the React to call componentDidUpdate() which called again this.updateListItems() and that caused the backend calls to be duplicated.

Replaced with this.setState({}); in this componentDidMount() which causes only the call to componentDidUpdate() and as such there's only a single call to the backend.

** Issue reference **

KIALI-2663

### DIFF
--- a/src/pages/AppList/AppListComponent.tsx
+++ b/src/pages/AppList/AppListComponent.tsx
@@ -37,7 +37,7 @@ class AppListComponent extends ListComponent.Component<AppListComponentProps, Ap
   }
 
   componentDidMount() {
-    this.updateListItems();
+    this.setState({});
   }
 
   componentDidUpdate(prevProps: AppListComponentProps, prevState: AppListComponentState, snapshot: any) {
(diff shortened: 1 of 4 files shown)


burmanm:
Yeah, I tried to find some "established way of doing this", but I couldn't. Some proposals were for state tricking, some to keep alive certain booleans and all of them looked kludge. So I just picked one. Maybe the shouldComponentUpdate is something to use, but what would it compare to? The props are clearly different in the mount & first update.

jshaughn:
@burmanm, yeah, I understand the difficulty.  I guess what I don't understand is why "paramsAreSynced" returns false on the first fetch.  Maybe the constructor needs to set better defaults for pagination.  Anyway, I'm not going to beat this horse.  I'm OK with it.

burmanm:
``prevProps.rateInterval === this.props.rateInterval`` returns false (different objects?)

burmanm:
Yeah, that's the real reason.. so both have same namespaces if comparing the names, but different objects. Maybe I should fix that compare instead.

PR #1116, opened 2019-04-15
KIALI-2712 Share duration with other tabs

jshaughn:
- introduce tail lines limit
- add disabled and tooltip options to MetricsDuration

By default Tail 500 lines to ensure that we don't ask for an overwhelming number of lines based on duration.  In the screenshot below show 10 lines.  Note that the duration is disabled when using a line limit:
image
When asking for all lines the time limit is enabled:
image

** Question: Perhaps it would be better to keep the duration dropdown enabled even though it's value is trumped when there is a line limit.  Otherwise the user can not alter the duration for the first fetch after selecting "All Lines".  Thoughts?

*** SERVER PR REQUIRED: https://github.com/kiali/kiali/pull/1009

### DIFF
--- a/src/pages/WorkloadDetails/WorkloadInfo/WorkloadPodLogs.tsx
+++ b/src/pages/WorkloadDetails/WorkloadInfo/WorkloadPodLogs.tsx
@@ -146,11 +146,7 @@ export default class WorkloadPodLogs extends React.Component<WorkloadPodLogsProp
                   tooltip={'Show up to last N log lines'}
                 />
                 {'   '}
-                <MetricsDurationContainer
-                  disabled={this.state.tailLines > 0}
-                  tooltip="Time range for log messages"
-                  onChanged={this.setDuration}
-                />
+                <MetricsDurationContainer tooltip="Time range for log messages" onChanged={this.setDuration} />
                 {'  '}
                 <Button id={'wpl_refresh'} disabled={!this.state.podLogs} onClick={() => this.handleRefresh()}>
                   <Icon name="refresh" />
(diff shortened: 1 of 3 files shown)


src/services/Api.ts
```diff
@@ -403,14 +403,23 @@ export const getPod = (namespace: string, name: string) => {
...
   const params: any = {};
   if (container) {
     params.container = container;
   }
   if (sinceTime) {
     params.sinceTime = sinceTime;
   }
+  if (tailLines && tailLines > 0) {
+    params.tailLines = tailLines;
+  }
```
israel-hdez: My only comment here is to exclude `params.sinceTime` if tailLines is not "all". I'm getting this filters combined by time and also number of lines. So, if you select "10 lines" but had previously selected "Last 1m", I sometimes get no logs.

src/pages/WorkloadDetails/WorkloadInfo/WorkloadPodLogs.tsx
```diff
@@ -19,25 +21,24 @@ interface ContainerInfo {
...
-  '1800': 'Last 30m',
-  '3600': 'Last 1h',
-  '10800': 'Last 3h',
-  '21600': 'Last 6h',
-  '43200': 'Last 12h',
-  '86400': 'Last 1d',
-  '604800': 'Last 7d'
+const TailLinesDefault = 500;
+const TailLinesOptions = {
+  '-1': 'All lines',
```
israel-hdez: "All lines" could be "By time"... And then the toolbar would be more human readable like "Tail: By time; Last X min"??
Well... I'm not sure about my suggestion; up to you ;)

src/services/Api.ts
jshaughn: You're right, it isn't working well.  But after thinking about this a bit more I think I prefer both filters to be active.  I'm not going to disable the duration dropdown.  This way the user if free to play with either setting and by default will limit to 500 lines for the current duration set for the tabs.

src/pages/WorkloadDetails/WorkloadInfo/WorkloadPodLogs.tsx
jshaughn: This is not really an option now that both filters active.

israel-hdez: Yes, with both active, not an option.

mtho11:
Oh looks like you just merged it 🍰

PR #1118, opened 2019-04-17
KIALI-2598 protect against negative countdown

jshaughn:
This doesn't really do much but maybe it's enough to solve this minor issue.  I'm not 100% sure this countdown mechanism really works as expected.  We may want to consider changing to a more simple approach that just pops up a dialog when the session expires and then gives the user a minute or so to extend before logging out.

### DIFF
--- a/src/components/SessionTimeout/SessionTimeout.tsx
+++ b/src/components/SessionTimeout/SessionTimeout.tsx
@@ -34,9 +34,7 @@ export class SessionTimeout extends React.Component<SessionTimeoutProps, {}> {
               <Icon name="warning-triangle-o" type="pf" style={{ fontSize: '48px' }} />
             </Col>
             <Col xs={12} sm={10} md={10} lg={10}>
-              {authenticationConfig.strategy === AuthStrategy.login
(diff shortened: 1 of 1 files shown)


src/components/SessionTimeout/SessionTimeout.tsx
```diff
@@ -70,19 +70,27 @@ export class SessionTimeout extends React.Component<SessionTimeoutProps, {}> {
...
-        Your session will timeout in {this.props.timeOutCountDown.toFixed()} seconds.
+        {message}
         <br />
         Would you like to extend your session?
       </p>
     );
   };
 
   private textForOpenshiftStrategy = () => {
+    const message =
```
mtho11: Looks like some duplication of code here. Perhaps the `const message` can be moved out of the `textForLoginStrategy` and `textForOpenshiftStrategy` and put above these so it can be shared?

jshaughn: OK, I changed things around in the hopes that his sire will find it aesthetically pleasing.

PR #1122, opened 2019-04-17
KIALI-2689 Support OAuth logout in OpenShift 4

mwringe:
** Describe the change **

If running on OpenShift 4 we will now redirect to the OpenShift logout page so that the user is properly logged out.

** Issue reference **

https://issues.jboss.org/browse/KIALI-2689

** Backwards compatible? **

This is backwards compatible with OpenShift 3.

### DIFF
--- a/src/components/Nav/Masthead/UserDropdown.tsx
+++ b/src/components/Nav/Masthead/UserDropdown.tsx
@@ -5,7 +5,8 @@ import { config } from '../../../config';
 import { MILLISECONDS } from '../../../types/Common';
 import Timer = NodeJS.Timer;
 import { KialiAppState, LoginSession } from 'src/store/Store';
-
+import authenticationConfig from '../../../config/AuthenticationConfig';
+import { AuthStrategy } from '../../../types/Auth';
 import moment from 'moment';
 import { ThunkDispatch } from 'redux-thunk';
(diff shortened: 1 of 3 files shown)


src/components/Nav/Masthead/UserDropdown.tsx
```diff
@@ -122,6 +127,11 @@ class UserDropdownConnected extends React.Component<UserProps, UserState> {
           toggle={<DropdownToggle onToggle={this.onDropdownToggle}>{this.props.session.username}</DropdownToggle>}
           dropdownItems={[userDropdownItems]}
         />
+        {authenticationConfig.strategy === AuthStrategy.openshift && authenticationConfig.logoutEndpoint && (
+          <form id="openshiftlogout" action={authenticationConfig.logoutEndpoint} method="post">
+            <input type="hidden" name="then" value={authenticationConfig.logoutRedirect} />
+          </form>
+        )}
       </>
```
aljesusg: Can we use a PF4 component here? If not please can you add a JIRA to the PF4 story and I'll Migrate this in another PR.

cfcosta:
Not a very clean solution, but to be honest all of the other ones are going to be exactly like this, just hidden in some form.

Didn't do functional testing, but @mattmahoneyrh did and I trust him, so I'm giving the +1 on this :)

PR #1123, opened 2019-04-17
Fix metrics settings when filters are first checked

israel-hdez:
When working on links from Traffic tab, needed a way to preselect filters. I succeeded to preselect filters, but also broke the normal behavior and sub-filters are not being checked by default when user selects a filter. This seems to fix the issue.


israel-hdez:
@jshaughn This is a "hot fix" :)

PR #1124, opened 2019-04-17
Related KIALI-2721 add a missing catch block

jshaughn:
I don't think this solves the Jira, it's just a change based on the
console log message captured with the problem.  This may not even be the
source of the uncaught promise, but it's the only instance I found when
looking in the graph-related code.


jshaughn:
@israel-hdez You are right, I don't see how this can really be a problem or the cause of the "uncaught" console logging.  I'm going to close it,

PR #1127, opened 2019-04-22
KIALI-2734 Refactor logout openshift to avoid html elements

aljesusg:
[KIALI-2734](https://issues.jboss.org/browse/KIALI-2734)

Tested. With this we avoid have more html elements, we can avoid adding this feature to the action instead to the component. What do you think @mwringe ?

Can you review that it's working correctly @mwringe ?

### DIFF
--- a/src/actions/LoginThunkActions.ts
+++ b/src/actions/LoginThunkActions.ts
@@ -7,6 +7,7 @@ import { AuthResult, AuthStrategy } from '../types/Auth';
 import { KialiDispatch } from '../types/Redux';
 import { MessageCenterActions } from './MessageCenterActions';
 import authenticationConfig from '../config/AuthenticationConfig';
+import axios from 'axios';
 
 const Dispatcher = new Login.LoginDispatcher();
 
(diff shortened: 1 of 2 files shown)


src/actions/LoginThunkActions.ts
```diff
@@ -73,14 +74,27 @@ const LoginThunkActions = {
...
   logout: () => {
     return async (dispatch: KialiDispatch) => {
-      try {
-        const response = await API.logout();
+      if (authenticationConfig.strategy === AuthStrategy.openshift && authenticationConfig.logoutEndpoint) {
+        axios
+          .post(authenticationConfig.logoutEndpoint)
+          .then(() => {
+            if (authenticationConfig.logoutRedirect) {
+              window.location.assign(authenticationConfig.logoutRedirect);
```
mwringe: I don't think this will work. The logout endpoint will only accept a request coming in from a POST, it wont accept any other method. This means we cannot use xmlhttprequests/ajax and we cannot use window.location.assign (which will result in a GET).

The logout endpoint will only work if you access it via a POST like what a form submit does.

aljesusg: Mmm but this is a axios POST, I'll check it tomorrow the status code.

PR #1129, opened 2019-04-22
KIALI-2737 Fix links in the About box

israel-hdez:
The `component="a"` attribute was missing.
I had to add @ts-ignore comments, because typescript complains, but the app works correctly.

Also, swapping icons of Homepage and Repository links, since they look inverted.

### DIFF
--- a/src/components/About/AboutUIModal.tsx
+++ b/src/components/About/AboutUIModal.tsx
@@ -92,7 +92,8 @@ class AboutUIModal extends React.Component<AboutUIModalProps, AboutUIModalState>
     if (config.about && config.about.website) {
       const Icon = icons[config.about.website.icon];
       return (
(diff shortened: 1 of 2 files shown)


israel-hdez:
@aljesusg Perhaps you could check if, for you, this builds correctly without the `@ts-ignore` comment?
Not sure if I have any issue with my env.

aljesusg:
I was thinking about set some property or link inside but I think that this is cleaner, about the ts-lint keep this comment rule, I'll open a PR in PF to fix this , I had the same problem with another component, after that we'll remove this.

aljesusg:
I opened the PR https://github.com/patternfly/patternfly-react/pull/1825 with the issue https://github.com/patternfly/patternfly-react/issues/1824 @israel-hdez

PR #1134, opened 2019-04-23
KIALI-2676 Show ExternalName field for ExternalName Services

xeviknal:
** Describe the change **
Knative introduced a new kind of services, the ExternalName services.
I think it would be a good idea to show which is that ExternalName.

Depending on: https://github.com/kiali/kiali/pull/1030

** Issue reference **

https://issues.jboss.org/browse/KIALI-2676

** Backwards compatible? **
yes

** Screenshot **
ExternalName Service doesn't show any IP but shows the actual ExternalName.
Screenshot of Kiali Console (10)

ClusterIP doesn't show the externalName since it doesn't make sense for this type of services:
Screenshot of Kiali Console (11)

### DIFF
--- a/src/pages/ServiceDetails/ServiceInfo/ServiceInfoDescription.tsx
+++ b/src/pages/ServiceDetails/ServiceInfo/ServiceInfoDescription.tsx
@@ -19,6 +19,7 @@ interface ServiceInfoDescriptionProps {
   labels?: { [key: string]: string };
   type?: string;
   ip?: any;
+  externalName?: string;
   ports?: Port[];
   endpoints?: Endpoints[];
   health?: ServiceHealth;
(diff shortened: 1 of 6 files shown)


hhovsepy:
Verified:
Screenshot from 2019-04-25 16-00-21
Screenshot from 2019-04-25 16-00-39

src/pages/ServiceDetails/ServiceInfo/ServiceInfoDescription.tsx
```diff
@@ -57,9 +58,15 @@ class ServiceInfoDescription extends React.Component<ServiceInfoDescriptionProps
               <div>
                 <strong>Type</strong> {this.props.type ? this.props.type : ''}
               </div>
-              <div>
-                <strong>IP</strong> {this.props.ip ? this.props.ip : ''}
-              </div>
+              {this.props.type !== 'ExternalName' ? (
```
jshaughn: Not a blocker but there is an effort underway to remove string literals. 'type' would be a good candidate for stonger typing.

xeviknal: Totally agree @jshaughn! Already changed that :)
Thanks for the review!

PR #1138, opened 2019-04-25
KIALI-2768 UX updates to compound node labels

mwringe:
**Describe the change**

Change to the compound node labels based on UX feedback. https://github.com/kiali/kiali-design/issues/121

**Issue reference**

https://issues.jboss.org/browse/KIALI-2768

~~**DO NOT MERGE**~~

~~There is a bug here which has my a bit stumped. When you first load the graph everything works fine, if you switch the layout then the labels are not aligned properly.~~

~~I suspect its in how we are doing things in the GroupCompoundLayout with removing elements and adding things back in, I think it might be taking the size with the temporary elements and not the final size.~~

~~If we explicitly set the label margin in GroupCompoundLayout, then it works when you change the layout, but is broken when you first load the page (and it also doesn't work when you hover over things).~~

~~I have submitted the PR to see if we can get feedback from someone who might know more about this than me.~~

~~@josejulio Any ideas here?~~

Thanks to @josejulio this is now working and is ready for review

### DIFF
--- a/src/components/CytoscapeGraph/Layout/GroupCompoundLayout.ts
+++ b/src/components/CytoscapeGraph/Layout/GroupCompoundLayout.ts
@@ -222,6 +222,13 @@ export default class GroupCompoundLayout {
         // Discard the saved values
         parent.removeScratch(CHILDREN_KEY);
         parent.removeScratch(STYLES_KEY);
+
+        // This will fix the label width issue when the layout changes, but it
+        // doesn't work well when the node is selected or hovered. It also prevents
+        // the label from being properly aligned when the graph is first loaded
+        /* parent.style('text-margin-x', (ele: any) => {
+                return '-' + (+ele.width() + +ele.style('padding').slice(0,-2) * 2 )+ "px";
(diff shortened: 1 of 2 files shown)


mwringe:
Yeah, I tested the update and its working nicely. It also fixes that weird shift thing that would happen when I first load the page, which is very nice (I always found that annoying).

I don't know how heavy it is, but it actually seems faster to me due to that shift being gone.

Personally I like the left aligned label better than the centre one. I think it might also have less potential overlap with edges. Such as in these situations:

Screenshot from 2019-05-03 10-31-06

josejulio:
> @mwringe I tested the update and it is working. Although I don't love that the fix is to reset all of the elements, which is a bit heavy iirc (is that true @josejulio ?) .

I'm not really sure, the elements are the same, it shouldn't trigger any remove/add in the elements but it surely triggers a redraw which is what we need to have it working. 
I'll ask to cytoscape maintainers if there is other way to do that and how heavy is calling json(elements) with the same elements.

PR #1139, opened 2019-04-26
Fix Menu Collapsed when mobile screen

aljesusg:
##  Before
before

## After
autocollapse


lucasponce:
Tested locally and it works as described.

PR #1140, opened 2019-04-26
KIALI-2729 PF4 update allows 'logout' when using anonymous login

aljesusg:
JIRA: [KIALI-2729](https://issues.jboss.org/browse/KIALI-2729)

### DIFF
--- a/src/components/Nav/Masthead/UserDropdown.tsx
+++ b/src/components/Nav/Masthead/UserDropdown.tsx
@@ -105,8 +105,10 @@ class UserDropdownConnected extends React.Component<UserProps, UserState> {
 
   render() {
     const { isDropdownOpen } = this.state;
+    const isAnonymous = authenticationConfig.strategy === AuthStrategy.anonymous;
+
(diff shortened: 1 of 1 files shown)


hhovsepy:
"Logout" menu item is disabled.
Screenshot from 2019-04-26 18-14-50

But from customer point of view it looks confusing, as option exists without any info, it is not grayed out or removed at all.

aljesusg:
It was disabled before upgrade from me side I agree with @hhovsepy

@mwringe @jshaughn ?

mwringe:
I think this was the design. From the discussions with UXD its more clear to disable something (with a hover over explanation) than to just remove it.

Does this have a hover over explaining why its disabled?

jshaughn:
There is no hover.   It is disabled (it is grayed out) and not removed.  Disabling is typically the preference from UX perspective when an option exists but is not currently relevant.  Personally, given that the username says 'anonymous' and that there was no login screen, I don't think a hover is necessary.

PR #1141, opened 2019-04-26
KIALI-2772 Support new gateway annotation format for VS gateway parsing

burmanm:
** Describe the change **

Istio 1.1 added new way to write the gateway string in the VirtualService definition. This one allows correct parsing to happen in the UI if the resource is in another namespace (current parsing only works for the same namespace).

** Issue reference **

KIALI-2772

** Backwards compatible? **

Yes

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "name": "@kiali/kiali-ui",
   "version": "0.19.0",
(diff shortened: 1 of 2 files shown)


lucasponce:
@burmanm I forgot to review this one. Do you mind to rebase it ?
I cc @hhovsepy if we can check it.

PR #1143, opened 2019-04-26
KIALI-2104 Shows link to go from workload/service/app page to graph

josejulio:
- This focus on the desired element, centering it on the resulting graph

** Describe the change **

It adds a link on the detail page of app/workload/service to "show on graph" the element.
It does that by going to the graph page and focusing (currently only centers the view on the element(s))

Probably a visual cue is needed aside of just centering the graph on the nodes.

** Screenshot **

show-in-graph-2

//cc: @pilhuhn

### DIFF
--- a/src/actions/GraphFilterActions.ts
+++ b/src/actions/GraphFilterActions.ts
@@ -8,6 +8,7 @@ enum GraphFilterActionKeys {
   SET_FIND_VALUE = 'SET_FIND_VALUE',
   SET_GRAPH_TYPE = 'SET_GRAPH_TYPE',
   SET_HIDE_VALUE = 'SET_HIDE_VALUE',
+  SET_SHOW_UNUSED_NODES = 'SET_SHOW_UNUSED_NODES',
   // Toggle Actions
   TOGGLE_GRAPH_NODE_LABEL = 'TOGGLE_GRAPH_NODE_LABEL',
   TOGGLE_GRAPH_CIRCUIT_BREAKERS = 'TOGGLE_GRAPH_CIRCUIT_BREAKERS',
@@ -28,6 +29,7 @@ export const GraphFilterActions = {
   setFindValue: createStandardAction(GraphFilterActionKeys.SET_FIND_VALUE)<string>(),
   setGraphType: createStandardAction(GraphFilterActionKeys.SET_GRAPH_TYPE)<GraphType>(),
   setHideValue: createStandardAction(GraphFilterActionKeys.SET_HIDE_VALUE)<string>(),
+  setShowUnusedNodes: createStandardAction(GraphFilterActionKeys.SET_SHOW_UNUSED_NODES)<boolean>(),
   // Toggle actions
   showGraphFilters: createStandardAction(GraphFilterActionKeys.ENABLE_GRAPH_FILTERS)<boolean>(),
   toggleGraphNodeLabel: createAction(GraphFilterActionKeys.TOGGLE_GRAPH_NODE_LABEL),
--- a/src/pages/AppDetails/AppInfo/AppDescription.tsx
+++ b/src/pages/AppDetails/AppInfo/AppDescription.tsx
@@ -31,7 +31,9 @@ class AppDescription extends React.Component<AppDescriptionProps, AppDescription
   }
 
   showOnGraphLink(application: string) {
-    return `/graph/namespaces?focusSelector=${encodeURI(new CytoscapeGraphSelectorBuilder().app(application).build())}`;
+    return `/graph/namespaces?graphType=workload&injectServiceNodes=true&unusedNodes=true&focusSelector=${encodeURI(
+      new CytoscapeGraphSelectorBuilder().app(application).build()
+    )}`;
   }
 
   serviceLink(namespace: string, service: string) {
@@ -124,15 +126,16 @@ class AppDescription extends React.Component<AppDescriptionProps, AppDescription
 
   render() {
     const app = this.props.app;
+    const istioSidecar = this.istioSidecar();
     return app ? (
       <PfInfoCard
         iconType="pf"
         iconName="applications"
         title={app.name}
-        istio={this.istioSidecar()}
+        istio={istioSidecar}
         items={
           <>
-            {app.name && (
+            {app && istioSidecar && (
               <Row>
                 <Col xs={12} sm={6} md={4} lg={4}>
                   <Link to={this.showOnGraphLink(app.name)}>Show on graph</Link>
--- a/src/pages/AppDetails/AppInfo/AppDescription.tsx
+++ b/src/pages/AppDetails/AppInfo/AppDescription.tsx
@@ -9,6 +9,7 @@ import { WorkloadIcon } from '../../../types/Workload';
 import { Link } from 'react-router-dom';
 import { CytoscapeGraphSelectorBuilder } from '../../../components/CytoscapeGraph/CytoscapeGraphSelector';
 import { NodeType } from '../../../types/Graph';
+import Namespace from '../../../types/Namespace';
 
 type AppDescriptionProps = {
   app: App;
@@ -31,12 +32,13 @@ class AppDescription extends React.Component<AppDescriptionProps, AppDescription
     return istioSidecar;
   }
 
-  showOnGraphLink(application: string) {
+  showOnGraphLink(application: string, namespace: Namespace) {
     return `/graph/namespaces?graphType=app&injectServiceNodes=true&unusedNodes=true&focusSelector=${encodeURI(
       new CytoscapeGraphSelectorBuilder()
         .app(application)
         .nodeType(NodeType.APP)
         .isGroup(null)
+        .namespace(namespace.name)
         .build()
     )}`;
   }
@@ -138,7 +140,7 @@ class AppDescription extends React.Component<AppDescriptionProps, AppDescription
         iconName="applications"
         title={app.name}
         istio={istioSidecar}
-        showOnGraphLink={this.showOnGraphLink(app.name)}
+        showOnGraphLink={this.showOnGraphLink(app.name, app.namespace)}
         items={
           <>
             <Row>
--- a/src/pages/AppDetails/AppInfo/AppDescription.tsx
+++ b/src/pages/AppDetails/AppInfo/AppDescription.tsx
@@ -134,7 +134,9 @@ class AppDescription extends React.Component<AppDescriptionProps, AppDescription
           <>
             {app.name && (
               <Row>
-                <Link to={this.showOnGraphLink(app.name)}>Show on graph</Link>
+                <Col xs={12} sm={6} md={4} lg={4}>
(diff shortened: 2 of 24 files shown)


pilhuhn:
I like that.
Could we use the highlighting feature? @jshaughn I guess that would be problematic if find is already on - or that that get cleared when the user leaves the graph? In that case we would be on the safe side.

jshaughn:
@josejulio, @pilhuhn, Perhaps don't go to the namespace graph but rather a node detail graph for the element.  Or, you you could use the graph find option and wire it so that the find expression is pre-set to the element of interest.  As it is right now the graph find expression does not survive navigations away from the graph, so it would not be a conflict.

pilhuhn:
Yes, graph find I what I meant.

josejulio:
It actually survives navigation away from the graph.

find

It does not survive page refresh (e.g. f5)

josejulio:
> Perhaps don't go to the namespace graph but rather a node detail graph for the element.

This is how that approach looks:

show-in-graph

But unsure on what to focus if we click "show on graph" from the app detail.

josejulio:
I added a new animation, please check:

focus-animation

josejulio:
> As it is right now the graph find expression does not survive navigations away from the graph, so it would not be a conflict.

Anyway we could replace the current search, the user wants to focus on the requested node(s) I think.
We would have to clear the find and the hide.

pilhuhn:
I like the animation. 
Going to the zoom-in page looks not optimal to me, as it still lacks the overall context of the graph (there may be better connected nodes in the graph, where the zoom-in view is better, as it shows more context).
Resetting the find/hide and populating with the new query to highlight seems ok to me.

jshaughn:
OK, so if using the full [namespace] graph there is a benefit to pre-selecting the node in question.  It will then give you the side-panel info as well emphasizing the end-to-end paths for which the node is a part.

josejulio:
> OK, so if using the full [namespace] graph there is a benefit to pre-selecting the node in question. It will then give you the side-panel info as well emphasizing the end-to-end paths for which the node is a part.

OK, but when showing the "app" in the graph there could be multiple nodes. 
Not going to select anything there if there is more than one result.

josejulio:
Ok, now i select the node to be able to see more info from the side-panel.
I didn't touch the filter/hide because i didn't think it was still relevant.

focus-animation-2

israel-hdez:
Minor comment: I see the node pre-selected, but looks like the "click" callback of the graph is not being called. This is my impression since I don't see the "path" highlighted. This is what I get when using the "Show in graph" link:

image

Small resolution image, but I gues you can see that the graph state is as if no node were selected. Clicking (re-selecting) manually the focused leads to the right path highlighted:

image

israel-hdez:
Also minor: Perhaps in the Application details page, the "Show on graph" legend should have margin. It is too close to that gray edge looking as a mistake:

image

The other pages look good.

src/pages/AppDetails/AppInfo/AppDescription.tsx
```diff
@@ -126,26 +131,33 @@ class AppDescription extends React.Component<AppDescriptionProps, AppDescription
...
-                health={this.props.health}
-                mode={DisplayMode.LARGE}
-                tooltipPlacement="left"
-              />
-            </Col>
-          </Row>
+          <>
+            {app.name && (
+              <Row>
+                <Link to={this.showOnGraphLink(app.name)}>Show on graph</Link>
```
israel-hdez: Probably, the `<Col>` may fix i ndentation/margin.

abonas:
@josejulio this is really cool. I wonder what kind of graph settings does it use when navigate there this way? I mean - which graph type (versioned app/workload/service...)?
If for example user selected previously a workload graph (with no services), and then the user went to a service details page and clicked "show on graph" - which kind of graph will it display now?

josejulio:
> @josejulio this is really cool. I wonder what kind of graph settings does it use when navigate there this way? I mean - which graph type (versioned app/workload/service...)?
> If for example user selected previously a workload graph (with no services), and then the user went to a service details page and clicked "show on graph" - which kind of graph will it display now?

As is it doesn't change anything. 
I think I need to do some checking to decide if the current graph is able to show the requested node (as in your example).

src/pages/AppDetails/AppInfo/AppDescription.tsx
josejulio: Most likely, will try that

josejulio:
> My only so-so important comment is that, probably, the "Show on graph" should be hidden for items marked with "Missing sidecar". These items will be hidden with the default graph options and, when "Show unused nodes" is enabled, these will be displayed aside. So, I think this feature is not relevant for these items.

I wonder if we should leave a message indicating that to the user OR if we are going to update the graph params we could enable "Show unused nodes" on these nodes.

I'm not sure, but, wouldn't the same apply for a node that hasn't received any traffic?

jshaughn:
- I agree with Edgar about disabling when the sidecar is missing
- I think enabling "display unused nodes" is a good idea to avoid the node not appearing in the graph
- I think the graph type could be based on the detail we're looking at:
  - service ->service
  - app -> versioned app
  - workload -> workload.
  - If you prefer the same type every time, then a workload graph with display services option enabled.

josejulio:
An alternative to try to change the current visualization of the user is to display an info message explaining that the node wasn't found on the current view, if the user updates the view and the node appears it will be highlighted.

See this: (the message doesn't appear yet)

focus-animation-node-not-found

but not sure if this is going to be confusing.

WDYT @abonas  @israel-hdez  ?

jshaughn:
@josejulio I don't love the idea of extra messaging.  Actually, I think this feature is already good and we should move to get it in sooner than later.

josejulio:
> @josejulio I don't love the idea of extra messaging. Actually, I think this feature is already good and we should move to get it in sooner than later.

I'll add your last comments here and will be ready to merge

src/pages/AppDetails/AppInfo/AppDescription.tsx
josejulio: Fix this

josejulio:
> Minor comment: I see the node pre-selected, but looks like the "click" callback of the graph is not being called.

Fixed, I wasn't calling the graphhighlighter.

josejulio:
Ok, I think this is ready to go:

focus-animation-4


focus-animation-5

src/actions/GraphFilterActions.ts
```diff
@@ -8,6 +8,7 @@ enum GraphFilterActionKeys {
   SET_FIND_VALUE = 'SET_FIND_VALUE',
   SET_GRAPH_TYPE = 'SET_GRAPH_TYPE',
   SET_HIDE_VALUE = 'SET_HIDE_VALUE',
+  SET_SHOW_UNUSED_NODES = 'SET_SHOW_UNUSED_NODES',
```
jshaughn: This is OK but going forward we may just want to have one "display" query param that can handle all of the display options.  Having bookmarkability for just one option is a little weak.

josejulio:
> * I'd prefer the link location to be fixed.  Currently it's included at the end of the labels, which seems odd.  to me. I think it would be better raised up to the right of, or below, the entity name (maybe in parens).  Not critical but may also make sense as an entry in the Actions dropdown.

Ah yes, will move that. I recall I just put the link somewhere to test, but never paid much attention to it.

> * Now that I see it, I think the graph type should be based on the entity type and not always a workload graph.  One major issue with workload graph is that focusing on an app is basically impossible.  By using the corresponding graph type there will be a single node on which to focus.

Will make the change so we can test that.

> * I really think we should abandon the center/zoom and show the full graph.  This provides much more context.  The animated focus and node selection should be enough for the user to see where in the topology the node of interest is located.

Sounds good, center/zoom was the initial way of focusing and seems like got superseded by the focus animation.

josejulio:
Consider this,
When doing the focus, I'm searching for nodes with "app=productpage" (in this example)
This could match nodes that aren't nodeType=app and group nodes (isGroup!=undefined).

Should i add the nodeType and isGroup to the query (see 2) or is OK as is?

1)
focus-animation-8

2)
focus-animation-7

josejulio:
Other gif showing how the different graphTypes react with (1)
focus-animation-9

jshaughn:
> When doing the focus, I'm searching for nodes with "app=productpage" (in this example)
This could match nodes that aren't nodeType=app and group nodes (isGroup!=undefined).

OK, so thinking with more detail. Workloads should be straightforward, a workload graph type with nodeType='workload' and name=wlName.  Services also should be straightforward, a service graph with nodeType='service' and service=svcName.   App is the tricky one.  We could use app graph or versioned app graph.  In general an app box represents and app but there are times (like an app that has not seen traffic) that an app box will not exist.  To be safe (and as we discussed together offline)  you would need to use an app graph with nodeType='app',app=appname,^isGroup.

Also, after doing the initial focus animation the focus should be initialized.  We don't want to re-focus or try to maintain a selection when, say, changing graph types.

josejulio:
I did all the suggested changes.
This is ready for another review.

src/pages/AppDetails/AppInfo/AppDescription.tsx
```diff
@@ -29,6 +30,16 @@ class AppDescription extends React.Component<AppDescriptionProps, AppDescription
     return istioSidecar;
   }
 
+  showOnGraphLink(application: string) {
+    return `/graph/namespaces?graphType=app&injectServiceNodes=true&unusedNodes=true&focusSelector=${encodeURI(
+      new CytoscapeGraphSelectorBuilder()
+        .app(application)
+        .nodeType('app')
```
jshaughn: should use NodeType.APP

josejulio: done

hhovsepy:
In the case when there are bookinfo and bookinfo2 namespaces, it highlights both apps on the graph when clocking on one app's link:
services_traffic_routing

src/pages/AppDetails/AppInfo/AppDescription.tsx
```diff
@@ -29,6 +31,16 @@ class AppDescription extends React.Component<AppDescriptionProps, AppDescription
     return istioSidecar;
   }
 
+  showOnGraphLink(application: string) {
+    return `/graph/namespaces?graphType=app&injectServiceNodes=true&unusedNodes=true&focusSelector=${encodeURI(
+      new CytoscapeGraphSelectorBuilder()
```
jshaughn: Needs namespace here and in the other selector builds.

josejulio:
> In the case when there are bookinfo and bookinfo2 namespaces, it highlights both apps on the graph when clocking on one app's link:

Good catch! 
I already added the namespace to the selector. It should be working now to correctly focus on the correct app/service/workload

hhovsepy:
Verified.

jshaughn:
@josejulio I think this can be merged if we can get a +1 from one of @abonas, @pilhuhn , or @lucasponce.

PR #1146, opened 2019-04-29
KIALI-2138 Show type in confirmation when deleting Istio Objects

israel-hdez:
https://issues.jboss.org/browse/KIALI-2138

All types that show up in the "Istio Config" list page seems to have the
"kind" property. Using this property to show the type in the confirm
dialog. Falling back to "Isito object" just in case.

image

### DIFF
--- a/src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
+++ b/src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
@@ -194,12 +194,8 @@ class IstioConfigDetailsPage extends React.Component<RouteComponentProps<IstioCo
     }
   };
 
-  fetchYaml = () => {
+  getIstioObject = () => {
     let istioObject: IstioObject | undefined;
-    if (this.state.isModified) {
-      return this.state.yamlModified;
-    }
-
     if (this.state.istioObjectDetails) {
       if (this.state.istioObjectDetails.gateway) {
         istioObject = this.state.istioObjectDetails.gateway;
@@ -233,6 +229,14 @@ class IstioConfigDetailsPage extends React.Component<RouteComponentProps<IstioCo
         istioObject = this.state.istioObjectDetails.serviceRoleBinding;
       }
     }
+    return istioObject;
+  };
+
+  fetchYaml = () => {
+    if (this.state.isModified) {
+      return this.state.yamlModified;
+    }
+    const istioObject = this.getIstioObject();
     return istioObject ? jsYaml.safeDump(istioObject, safeDumpOptions) : '';
   };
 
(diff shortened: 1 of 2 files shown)


hhovsepy:
Verified.

src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
```diff
@@ -298,9 +304,12 @@ class IstioConfigDetailsPage extends React.Component<RouteComponentProps<IstioCo
 
   renderRightToolbar = () => {
     const canDelete = this.state.istioObjectDetails !== undefined && this.state.istioObjectDetails.permissions.delete;
+    const istioObject = this.getIstioObject();
+
     return (
       <span style={{ float: 'right' }}>
         <IstioActionDropdown
+          objectKind={istioObject && 'kind' in istioObject ? istioObject.kind : undefined}
```
xeviknal: After reviewing this line, I realized that all the Istio Objects should follow the interface IstioObject. Subsequently, all of them should have the kind field. As a result, I saw that we had two deprecated Istio Objects that didn't follow that inheritance: DestinationPolicies and Route Rules.

I created the following PR for removing them: https://github.com/kiali/kiali-ui/pull/1148
Right after this is merged, we will have that all the Istio Objects will be type of IstioObject and then, this check could be delegated to typescript in form of strong type detection.

src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
```diff
@@ -194,48 +194,54 @@ class IstioConfigDetailsPage extends React.Component<RouteComponentProps<IstioCo
...
       } else if (this.state.istioObjectDetails.serviceRole) {
-        istioObject = this.state.istioObjectDetails.serviceRole;
+        return this.state.istioObjectDetails.serviceRole;
       } else if (this.state.istioObjectDetails.serviceRoleBinding) {
-        istioObject = this.state.istioObjectDetails.serviceRoleBinding;
+        return this.state.istioObjectDetails.serviceRoleBinding;
       }
     }
+
+    return undefined;
```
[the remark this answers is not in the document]
israel-hdez: The reasoning behind the change is that declaring solely as `let istoObject` typescript types the variable as `any`, so the method was being typed as `any`.
Using "return" instructions, the method is typed with a union type and we have type checking enabled.

But with your changes in #1148, we can use the IstioObject base type. The union type is not really needed for now. I rebased and removed the returns.

src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
israel-hdez: Rebased this PR to use the latest changes.

israel-hdez:
@xeviknal PR is updated!

PR #1149, opened 2019-04-30
KIALI-2784 Improve handling of summary collapse/expand

jshaughn:
It is important to re-render the graph page, not just the summary panel, for "fit" to have the correct canvas dimensions.

The idea here is to maximize graph space when collapsing or expanding the side panel.  But, do not resize on expand/collapse if the user has selected a graph element.  In that case it is assumed the user prefers any drag or zoom customization.

### DIFF
--- a/src/pages/Graph/GraphPage.tsx
+++ b/src/pages/Graph/GraphPage.tsx
@@ -12,6 +12,7 @@ import { computePrometheusRateParams } from '../../services/Prometheus';
 import { CancelablePromise, makeCancelablePromise } from '../../utils/CancelablePromises';
 import * as MessageCenterUtils from '../../utils/MessageCenter';
 import CytoscapeGraphContainer from '../../components/CytoscapeGraph/CytoscapeGraph';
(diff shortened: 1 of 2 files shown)


jshaughn:
@mwringe Do you think we could merge this and take further action if UX asks for it?

mwringe:
I would wait for UX to provide feedback here. I would find it confusing to expand or collapse a side panel and then have the graph change its pan and zoom to fit.

mtho11:
@mwringe seems like this should be merged because it does fix overlaying of the graph by the side panel.
The whole zoom/pan/fit is another separate design issue that may take a while to iron out. Especially when summit is upon us.

jshaughn:
@mwringe Shall we continue to wait for UX here? This agenda item was skipped by UX during the last UX meeting.  Or shall we take the incremental improvement?

jshaughn:
After demoing this to Saturn we decided the desired behavior is really subjective and the vote was about 50/50 between the current behavior and this PR.  So, closing in favor of future UX an/or user feedback.

PR #1150, opened 2019-04-30
KIALI-2746 Reduce size of graph sidepanel and main nav menu

israel-hdez:
https://issues.jboss.org/browse/KIALI-2746
Also: https://issues.jboss.org/browse/KIALI-2765

Reducing sizes as a workaround until we have a new design available.

* Font size of side panel is reduced. This produces a more adequate
size to not distract the user from the graph.
* Width of main navigation menu is reduced to make better use of
available screen space.

Before:
image

After:
image

### DIFF
--- a/src/pages/Graph/SummaryPanel.tsx
+++ b/src/pages/Graph/SummaryPanel.tsx
@@ -16,11 +16,13 @@ type MainSummaryPanelPropType = SummaryPanelPropType & {
 };
 
 const expandedStyle = style({
+  fontSize: '74%', // TODO: Remove
   paddingTop: '1em',
   position: 'relative'
 });
 
 const collapsedStyle = style({
+  fontSize: '74%', // TODO: Remove
   paddingTop: '1em',
   position: 'relative',
   $nest: {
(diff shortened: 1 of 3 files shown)


mtho11:
Can we get some before/after screenshots?

israel-hdez:
@mtho11 There you go... I was preparing them :)

mtho11:
LOL! 
That looks much better to me.

src/pages/Graph/SummaryPanel.tsx
```diff
@@ -16,11 +16,13 @@ type MainSummaryPanelPropType = SummaryPanelPropType & {
 };
 
 const expandedStyle = style({
+  fontSize: '74%', // TODO: Remove
```
jshaughn: @israel-hdez This was not removed prior to merge. (also below).

israel-hdez: Yes, it is intentional. It's a note to remove this line once we have a new design... Perhaps my note it's too short and needed more explanation :/

jshaughn: Yes, too short :)

PR #1151, opened 2019-05-01
KIALI-2786 Set bounds on graph legend and find help modals

jshaughn:
@mtho11, by adding bounds on the Draggables for me it seems reasonable if not great.  It does generate vertical scrollbars as needed and I never get into a situation where I lose the handle or can't get rid of the modal.  It does not introduce or generate horizontal scrollbars.  I would sort of expected that to happen but it doesn't and I don't know how to make it happen or if it's even possible.

Give it a try and see whether it's enough to get by or if we need a guru to get this in better shape.

### DIFF
--- a/src/components/GraphFilter/GraphLegend.tsx
+++ b/src/components/GraphFilter/GraphLegend.tsx
@@ -61,7 +61,7 @@ export default class GraphLegend extends React.Component<GraphLegendProps, Graph
     }
     const className = this.props.className ? this.props.className : '';
     return (
-      <Draggable>
+      <Draggable bounds="#root">
         <div className={`modal-content ${className}`}>
           <div className="modal-header">
             <Button className="close" bsClass="" onClick={this.props.closeLegend}>
--- a/src/components/GraphFilter/GraphLegend.tsx
+++ b/src/components/GraphFilter/GraphLegend.tsx
@@ -37,44 +37,49 @@ export default class GraphLegend extends React.Component<GraphLegendProps, Graph
     image.src = this.getLegendImage();
   }
 
-  getLegendImageStyle() {
-    return style({
-      backgroundImage: `url(${this.getLegendImage()})`,
-      margin: '5px 10px',
-      padding: 0
-    });
-  }
-
-  getLegendImage() {
-    let image = graphLegendImage;
-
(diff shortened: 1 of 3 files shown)


abonas:
@jshaughn can you please add a screenshot?

jshaughn:
@abonas @mtho11 Here is a video that basically shows how the legend and help are now bounded by the root and remain located such that they can be closed. Previously it was possible to drag such that closing became impossible (at least for help, not sure about legend).  You can see the vertical scroll kick in when needed.  As mentioned above, horizontal scroll never shows up, I'm not sure why or if it's even possible.

kiali-2786

mtho11:
It seems better than what we have. But someone might argue that we should never see the vertical scrollbar. I think UxD needs to look at this one.

jshaughn:
@mtho11 Yeah, we can see what UX says, I'm just not sure we have much technical flexibility to realize any UX vision/preference.  I just threw some gum at the wall.

abonas:
did someone raise this to UX by mentioning here/opening a design issue? Adding a label is good, but it does not trigger any notification to anyone.

jshaughn:
@mtho11 several changes to limit widths and use horizontal scrolling for both Find/Hide help and the Legend.  Please give it a whirl when you get the chance...

jshaughn:
@mtho11 Here is an updated video.  I'm taking UX label off of this PR, if there are changes to the actual legend it should be as part of a complete overhaul around l+f for smaller devices.

kiali-2786

src/components/GraphFilter/GraphLegend.tsx
```diff
@@ -37,44 +37,49 @@ export default class GraphLegend extends React.Component<GraphLegendProps, Graph
...
+      width: this.state.width,
+      height: this.state.height
+    });
+
     return (
-      <Draggable>
-        <div className={`modal-content ${className}`}>
-          <div className="modal-header">
+      <Draggable bounds="#root">
+        <div className={`modal-content ${parentClassName} ${contentStyle}`}>
```
mtho11: Note: can also use the typestyle `classes` function for style composition (even conditionally): https://typestyle.github.io/#/core

PR #1153, opened 2019-05-01
KIALI-2389 Reduce edge label congenstion by moving percentErr under rate

jshaughn:
This isn't the best example of crowding, see the Jira/design issue, but it should be easy enough to see that the stacked label is less wide, and I think cleaner without the comma-separator.

Before:
image

After:
image


mtho11:
@jshaughn I'm not sure this is necessary with being able to right-click on an edge and show whatever information that you want: https://github.com/kiali/kiali-ui/pull/1014. 
But, I'm also cool with this approach. Whatever UxD wants.

mtho11:
It's much better than what we have right now so I would say merge this.

PR #1155, opened 2019-05-02
KIALI-2799 Adapt logs pane to the height of the browser window

israel-hdez:
https://issues.jboss.org/browse/KIALI-2799

Not a great approach, but I don't see other option rather than making
calculations using the size of the controls at the top of the pane.

As a plus, this also fixes:
* Sizing issues with the main content `div`. The issue can be seen
  in any of the app/workloads/services list pages. The scroll bar
  will run beyond the screen.
* Scroll bar issue in the metrics tab of the service details page.
* Applies the same fix of the logs tab to the YAML editor to fill the available page space.

### DIFF
--- a/src/components/Nav/Navigation.tsx
+++ b/src/components/Nav/Navigation.tsx
@@ -53,12 +53,7 @@ class Navigation extends React.Component<PropsType, NavigationState> {
   }
 
   isContentScrollable = () => {
-    const urlParams = new URLSearchParams(this.props.location.search);
-    let isMetricTab = false;
-    if (urlParams.has('tab')) {
-      isMetricTab = urlParams.get('tab') === 'metrics';
-    }
-    return !this.props.location.pathname.startsWith('/graph') && !isMetricTab;
+    return !this.props.location.pathname.startsWith('/graph');
   };
 
   onNavToggleDesktop = () => {
@@ -85,7 +80,7 @@ class Navigation extends React.Component<PropsType, NavigationState> {
 
     const Header = (
       <PageHeader
-        logo={<Brand src={kialiLogo} alt="Patternfly Logo" />}
+        logo={<Brand src={kialiLogo} alt="Kiali Logo" />}
         toolbar={<Masthead />}
         showNavToggle={true}
         onNavToggle={isMobileView ? this.onNavToggleMobile : this.onNavToggleDesktop}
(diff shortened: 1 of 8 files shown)


jotak:
> Still, I'm impressed because I didn't have any hope that this could be improved.

When beers are in play everything is possible

israel-hdez:
@jshaughn 

> I see horizontal scroll bars immediately when I select the logs tab. Is that expected?

I'm curious, how does it look on your screen? I see this:
image

I only see the scrollbar of the logs pane. Same both in Chrome & Firefox. Perhaps browser renders differently in Windows? I can try in that OS if this is the case.

I'm curious, because I only changed vertical sizes, so it is weird that you are getting horizontal scrollbars

> Also, I don't quite understand what benefit this introduced to the list pages, to me they behave the same way, and fwiw, neither the old code of this PR make it work well with a smaller screen.

See this screenshot, this is how it looked before:
image

Look at the bottom right corner. Perhaps, can you see that the scroll bar does not have its down arrow?
This means that the view is running beyond the viewport of the browser. Also, In my env, I have the page scrolled all the way down. You can see that the pagination bar cannot be viewed entirely. Perhaps, can you check if you also have this issue with the old code? This is what I tried to fix.

israel-hdez:
>> Still, I'm impressed because I didn't have any hope that this could be improved.

> When beers are in play everything is possible

@jotak  LOL!!! with a big ending "L"... :)

jshaughn:
@israel-hdez , sorry, I didn't mean to say horizontal, it's vertical scroll bars.  OK, now I owe tres cervezas.  But the vertical scroll (on kiali, not just the log windos) appeared for seemingly no reason, as soon as I opened the logs tab, as if the logs window were sized a little too large.

As for the list pages, I see no difference between the PR and master.

israel-hdez:
@jshaughn Ok, I updated and applied the fix to the YAML editor.

Not sure if it could be confusing using CSS variables. I did it only to reuse calculations between the logs tab and the YAML editor. If you see this confusing, I can go back to doing the full calc in place.

Remember to re-start yarn ;)

jshaughn:
> Remember to re-start yarn ;)

Yes, my yarn start was stale.  Things are looking good!  Now need to apply to the YAML tabs and add a little bit of commenting so that people understand what you are doing.  I think the change is sensitive to possible changes in the height of the surrounding components.

israel-hdez:
@jshaughn This is ready for another review round.

PR #1159, opened 2019-05-07
KIALI-2798 Ensure find/hide clear buttons survive navigation

jshaughn:
Also:
- add tootips to find/hide clear buttons to help distinguish inputs when both are filled
- add name and autocomplete attributes to find/hide inputs to [maybe] help browsers present previous expressions

### DIFF
--- a/src/components/GraphFilter/GraphFind.tsx
+++ b/src/components/GraphFilter/GraphFind.tsx
@@ -59,7 +59,9 @@ export class GraphFind extends React.PureComponent<GraphFindProps, GraphFindStat
 
   constructor(props: GraphFindProps) {
     super(props);
-    this.state = { errorMessage: '', findInputValue: '', hideInputValue: '' };
+    const findValue = props.findValue ? props.findValue : '';
+    const hideValue = props.hideValue ? props.hideValue : '';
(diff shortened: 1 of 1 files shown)


abonas:
screenshots please ?

jshaughn:
@abonas A video is better for this because you have to show navigating away and then back. When showing that the clear buttons survive you can see the tooltips I added as a nod to kiali-2805...

kiali-2798

src/components/GraphFilter/GraphFind.tsx
```diff
@@ -99,47 +101,67 @@ export class GraphFind extends React.PureComponent<GraphFindProps, GraphFindStat
...
                 placeholder="Find..."
               />
               {this.state.findInputValue && (
-                <InputGroup.Button>
-                  <Button onClick={this.clearFind}>
-                    <Icon name="close" type="fa" />
-                  </Button>
-                </InputGroup.Button>
+                <OverlayTrigger
+                  key="ot-clear_find"
```
jmazzitelli: completely trivial - but it is odd that you use "-" and "_" in the key value here. I'm assuming there is a reason. I'm not requesting a change, but I just bring it up because it looks weird.

PR #1160, opened 2019-05-07
KIALI-2817 preload icon webfonts

jshaughn:
- allow clean/incognito browsers to properly show badges on first render

This is currently a draft until we figure out whether it's safe to preload from static/media, I have a feeling those filenames will change if there is an update to the font.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -13,6 +13,12 @@
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
     <link rel="shortcut icon" href="%PUBLIC_URL%/kiali_icon_lightbkg_16px.png" />
     <!--
+      Preload the icon webfonts used for badging in graph labels. Cytoscape can render prior to the font
+      bring loaed. (Kiali-2817)
+    -->
+    <link rel="preload" href="static/media/fontawesome-webfont.af7ae505.woff2" as="font" type="font/woff2" crossorigin />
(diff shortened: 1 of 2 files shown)


josejulio:
> those filenames will change if there is an update to the font

Yes, they will change if that happens

You can require the resource in typescript and just add it to the link. 

Like here: https://github.com/josejulio/swsui/blob/bb0c3fdc2c7f82bcdccfb44e0510eb88d5b3d6e6/src/components/CytoscapeGraph/CytoscapeGraph.tsx#L94

and use like this:
https://github.com/josejulio/swsui/blob/bb0c3fdc2c7f82bcdccfb44e0510eb88d5b3d6e6/src/components/CytoscapeGraph/CytoscapeGraph.tsx#L210

jshaughn:
@josejulio Thanks for the comment, I figured the approach was not safe but created a PR just to see if it would work for @mtho11 .  It did work, so if we can find a better way to do the pre-load then it should be ok.  I will follow your links.

jshaughn:
@mtho11, @josejulio, OK, I promoted this from Draft to regular PR.  Big thanks to Josejulio for helping figure out how to get the preload links into place.  Give it a try and see if it's OK for you.  Inspecting the html head via F12-Elements I can see the elements in place,  I don't get any load errors in the console, and I'm not seeing the bad characters using the incognito repro steps in the Jira.

PR #1161, opened 2019-05-08
Axios CVE

mtho11:
** Describe the change **

Update snyk for Axios CVE.

** Issue reference **
https://issues.jboss.org/browse/KIALI-2823

** Backwards compatible? **
yes

No new behaviour

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,8 +101,7 @@
     "ssri": "6.0.1",
     "typesafe-actions": "3.2.1",
     "typestyle": "2.0.1",
-    "visibilityjs": "2.0.2",
-    "snyk": "^1.161.1"
+    "visibilityjs": "2.0.2"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -99,7 +101,8 @@
     "ssri": "6.0.1",
     "typesafe-actions": "3.2.1",
     "typestyle": "2.0.1",
-    "visibilityjs": "2.0.2"
+    "visibilityjs": "2.0.2",
+    "snyk": "^1.161.1"
```
josejulio: Isn't this a dev-dependency?

mtho11: So `snyk` removed it from devDependency and put it in dependency.

mtho11: But I agree, I don't see why it is needed as a dependency and it should be a dev-dependency instead.

mtho11: Changes pushed

PR #1162, opened 2019-05-08
KIALI-2641 Add Wizard for 3scale adapter config

lucasponce:
This PRs refers to JIRA
https://issues.jboss.org/browse/KIALI-2641
and it needs backend PR https://github.com/kiali/kiali/pull/1050 with the API services.

Summary
----
Maistra will include a new Istio 3Scale adapter [1] that will allow to autenticate Services deployed in Istio with a 3Scale API. 

The configuration of this integration is somehow complex as a user needs to create Istio objects like rules, handlers and template instances.

The goal of this PR is to provide a high level integration from the UI to make this setup easier.

For this first step, we have defined a new Wizard at Service Details level. 
This Wizard only will be visible when the 3scale adapter is present in the cluster, otherwise, this PR will be transparent to the rest of the users.

image

When the 3scale adapter is available user will be able to "Add a 3Scale" rule for a service or to "Delete" an existing one. Note that this operation is 1to1, so, Kiali will only permit one 3Scale Rule per Service.

This will means that a specific service will query 3Scale for authentication, following the described policies defined in [1].

If cluster has not defined any 3Scale Handler, first action a user will do is to create one. To perform this operation, user needs to supply some parameters provided by the 3Scale Admin site.

image

Once this handler is created; the handler is selected for the service that will be linked.

image

Note, that user can create multiple handlers, but it will be typically only one per cluster, that will be used for multiple services. But multiple handlers can be created using the "Show Advanced Options".

image

Basic delete/update operations on handler(s) are supported from the Wizard:

image

Once the 3Scale Rule is created, a particular service is "linked" with 3Scale API using and indicated into the Service Description details:

image

Note that in the future, there can be more complex operations, like to perform a Rule on mulitple Services in a single operation. 

That is out of the scope of this PR.

How to test
---

There are several preliminary steps before you can test this feature.

First is to use install Maistra enabling the 3Scale adapter. To do that, you need to use a special custom installation yaml like this one:

```
apiVersion: "istio.openshift.com/v1alpha1"
kind: "Installation"
metadata:
  name: "istio-installation"
  namespace: istio-operator
spec:
  threeScale:
    enabled: true
```
That will deploy the 3scale adapter under the istio-system namespace:
image

Also, 3Scale adapter needs to enable policies in Istio, if not, Rules won't take any effect without showing any warning/error on logs (yeap, I hope this would save you some time):

```
echo "Check policies...."
oc -n istio-system get cm istio -o jsonpath="{@.data.mesh}" | grep disablePolicyChecks
cd $ISTIO_HOME
echo "Updating policies..."
helm template install/kubernetes/helm/istio --namespace=istio-system -x templates/configmap.yaml --set global.disablePolicyChecks=false | oc -n istio-system replace -f -
```

For these tests I've created a special API in 3Scale called Kiali, under the API I prepared some configuration to be used with a special demo app called travel-agency:

https://github.com/lucasponce/travel-comparison-demo

So, first you need to deploy the travel-agency app using

```
https://github.com/lucasponce/travel-comparison-demo/blob/master/deploy-travel-agency.sh

Note: no need to install the full example, just the travel-agency is enough for this 3scale testing
```

Next, you would need to expose the travels service with a gateway

```
https://github.com/lucasponce/travel-comparison-demo/blob/master/travel_agency_gateway.yaml
```

And you can invoke the travel service through the gateway using something like this script

```
#!/bin/bash

export INGRESS_HOST=$(oc get po -l istio=ingressgateway -n istio-system -o 'jsonpath={.items[0].status.hostIP}')
export INGRESS_PORT=$(oc -n istio-system get service istio-ingressgateway -o jsonpath='{.spec.ports[?(@.name=="http2")].nodePort}')

TRAVEL_AGENCY_HOST="Host:www.travel-agency.com"
TRAVELS_SERVICE="http://${INGRESS_HOST}:${INGRESS_PORT}"

while true
do
    for city in paris rome london
    do
        for user in vip normal
        do
            RESULT=$(curl -H "${TRAVEL_AGENCY_HOST}" -H "user:${user}" -o /dev/null -s -w %{http_code} ${TRAVELS_SERVICE}/travels/${city}?user_key=4406b917c719df23055919c134651ee4)
            echo "Travel quota for ${city} from ${user} => $RESULT"
            sleep 1
        done
        RESULT=$(curl -H "${TRAVEL_AGENCY_HOST}" -H "user:${user}" -o /dev/null -s -w %{http_code} ${TRAVELS_SERVICE}/travels/${city})
        echo "Travel quota for ${city} from anonymous => $RESULT"
        sleep 1
    done
done
```

Note that this script will add a special 3scale query parameters "user_key" to propagate the user key of the demo user connected with 3Scale (you need to use this value, as it's a valid code).

Then, you can use the following 3scale parameters to create handlers/rule in the Wizard:

```
{
  "name": "my-test-handler",
  "serviceId": "2555417792901",
  "systemUrl": "https://kiali-admin.3scale.net",
  "accessToken": "15586d0814cf49c5fd8fef46940e774063476c3845a368d218ac6edc6befb054"
}

Note: "name" can be anything, but serviceId, systemUrl and accessToken are valid
```

How to validate things are working as expected
---

Once that you have created a 3Scale Rule under the "travels" service, only the calls with the valid user_key are responding a 200 Ok, the others should have a 403 Error.
When no rule is created, all calls should return a 200 Ok.

```
[...]
Travel quota for london from vip => 200
Travel quota for london from normal => 200
Travel quota for london from anonymous => 403
[...]
```

Also, in Istio Config, we can check that the Wizard generates a pair of Hanlder/Instances resources per "3Scale Handler" and a specific "Istio Rule" per link between "Service and 3Scale".

[1] https://github.com/3scale/3scale-istio-adapter

### DIFF
--- a/src/pages/ServiceDetails/ServiceInfo/ServiceInfoDescription.tsx
+++ b/src/pages/ServiceDetails/ServiceInfo/ServiceInfoDescription.tsx
@@ -11,6 +11,7 @@ import { style } from 'typestyle';
 import './ServiceInfoDescription.css';
 import Labels from '../../../components/Label/Labels';
 import { CytoscapeGraphSelectorBuilder } from '../../../components/CytoscapeGraph/CytoscapeGraphSelector';
+import { ThreeScaleServiceRule } from '../../../types/ThreeScale';
 
 interface ServiceInfoDescriptionProps {
   name: string;
@@ -25,6 +26,7 @@ interface ServiceInfoDescriptionProps {
   ports?: Port[];
   endpoints?: Endpoints[];
   health?: ServiceHealth;
+  threeScaleServiceRule?: ThreeScaleServiceRule;
 }
 
 const listStyle = style({
@@ -83,6 +85,9 @@ class ServiceInfoDescription extends React.Component<ServiceInfoDescriptionProps
               <div>
                 <strong>Resource Version</strong> {this.props.resourceVersion}
               </div>
+              {this.props.threeScaleServiceRule && this.props.threeScaleServiceRule.threeScaleHandlerName !== '' && (
+                <div>Service linked with 3Scale API</div>
+              )}
             </Col>
             <Col xs={12} sm={4} md={2} lg={2}>
               <div className="progress-description">
--- a/src/pages/ServiceDetails/ServiceInfo/ServiceInfoDescription.tsx
+++ b/src/pages/ServiceDetails/ServiceInfo/ServiceInfoDescription.tsx
@@ -86,7 +86,7 @@ class ServiceInfoDescription extends React.Component<ServiceInfoDescriptionProps
                 <strong>Resource Version</strong> {this.props.resourceVersion}
               </div>
               {this.props.threeScaleServiceRule && this.props.threeScaleServiceRule.threeScaleHandlerName !== '' && (
-                <div>Service linked with 3Scale API</div>
+                <div>Service linked with 3scale API</div>
               )}
             </Col>
             <Col xs={12} sm={4} md={2} lg={2}>
--- a/src/pages/ServiceDetails/ServiceInfo/ServiceInfoDescription.tsx
+++ b/src/pages/ServiceDetails/ServiceInfo/ServiceInfoDescription.tsx
@@ -85,9 +85,9 @@ class ServiceInfoDescription extends React.Component<ServiceInfoDescriptionProps
               <div>
                 <strong>Resource Version</strong> {this.props.resourceVersion}
               </div>
-              {this.props.threeScaleServiceRule && this.props.threeScaleServiceRule.threeScaleHandlerName !== '' && (
-                <div>Service linked with 3scale API</div>
-              )}
+              {this.props.threeScaleServiceRule &&
+                this.props.threeScaleServiceRule.threeScaleHandlerName !== '' &&
+                'Service linked with 3scale API'}
             </Col>
             <Col xs={12} sm={4} md={2} lg={2}>
               <div className="progress-description">
--- a/src/pages/ServiceDetails/ServiceInfo/ServiceInfoDescription.tsx
+++ b/src/pages/ServiceDetails/ServiceInfo/ServiceInfoDescription.tsx
@@ -85,9 +85,11 @@ class ServiceInfoDescription extends React.Component<ServiceInfoDescriptionProps
               <div>
                 <strong>Resource Version</strong> {this.props.resourceVersion}
               </div>
-              {this.props.threeScaleServiceRule &&
-                this.props.threeScaleServiceRule.threeScaleHandlerName !== '' &&
-                'Service linked with 3scale API'}
+              {this.props.threeScaleServiceRule && this.props.threeScaleServiceRule.threeScaleHandlerName !== '' && (
+                <span>
+                  Service linked with 3scale API Handler <i>{this.props.threeScaleServiceRule.threeScaleHandlerName}</i>
+                </span>
+              )}
             </Col>
             <Col xs={12} sm={4} md={2} lg={2}>
               <div className="progress-description">
--- a/src/components/IstioWizards/IstioWizard.tsx
+++ b/src/components/IstioWizards/IstioWizard.tsx
@@ -18,6 +18,7 @@ import {
   getInitWeights,
   WIZARD_MATCHING_ROUTING,
   WIZARD_SUSPEND_TRAFFIC,
+  WIZARD_THREESCALE_INTEGRATION,
   WIZARD_TITLES,
   WIZARD_UPDATE_TITLES,
   WIZARD_WEIGHTED_ROUTING,
@@ -26,6 +27,8 @@ import {
 } from './IstioWizardActions';
 import { Response } from '../../services/Api';
 import { MessageType } from '../../types/MessageCenter';
+import ThreeScaleIntegration from './ThreeScaleIntegration';
+import { ThreeScaleServiceRule } from '../../types/ThreeScale';
 
 class IstioWizard extends React.Component<WizardProps, WizardState> {
   constructor(props: WizardProps) {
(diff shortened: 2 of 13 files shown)


lucasponce:
@hhovsepy I know this needs a rebase with master, but I guess the feature is more or less complete to start testing it.

Also I will cc @unleashed, @PhilipGough and @vramosp if they want to provide feedback and/or provide feedback.

unleashed:
Hi @lucasponce we are changing a bit how we will be attaching credentials to these templates to minimise the number of objects needed and allow for a more streamlined management of 3scale accounts where multiple services use the same credentials - we are still working on it and experimenting, but if it all goes well we will have a description of the changes soon, which hopefully won't impact this work too much.

Other than that this looks like good progress! :+1:

Just a minor nit, use "3scale" rather than "3Scale". Thanks!

lucasponce:
> multiple services use the same credentials - we are still working on it and experimenting, but if it all goes well we will have a description of the changes soon, which hopefully won't impact this work too much.

@unleashed Also note that Kiali won't update the Pod labels, for the wizard it will create a rule per service using existing data (also to avoid an extra operation to delete/recreate Pods).

In the testing, it looks good as well.

hhovsepy:
Should there be all fields mandatory to fill, then enable the add button?
Screenshot from 2019-05-10 16-35-28

unleashed:
@hhovsepy I think so, yes. The 3scale adapter requires all of that information, so if anything is missing it can't possibly work.

hhovsepy:
[RFE] Add a button and a status on handler that authentication via Access Token was passed on 3scale.

abonas:
1. Can some info/tooltips be added for where the users should be taking the information for all the fields? serviceID for example - what kind of ID is this? systemURL - which system URL?
2. How the handlers are managed? edit some details of a handler for example?
3. A clear indication for mandatory fields (I think there's already a comment about this above)
4. Testing the connection to the handler whether it's OK - how the user can tell if the entered details are working OK?
5. How an access token is passed and stored? Is it secure?
6. If there are several handlers, is there a possible overlap/override that can cause some unpredictable behavior? If rule A uses handler 1 and rule B uses handler 2, what is the expected behavior?

lucasponce:
> [RFE] Add a button and a status on handler that authentication via Access Token was passed on 3scale.
+1

Yes, it shouldn't permit to create a handler without all fields filled.

lucasponce:
> Can some info/tooltips be added for where the users should be taking the information for all the fields? serviceID for example - what kind of ID is this? systemURL - which system URL?

Some tooltip can be added, but this is specific 3scale context info.

> How the handlers are managed? edit some details of a handler for example?

Described in the first comment.

> A clear indication for mandatory fields (I think there's already a comment about this above)

Yes. This is something it can be done in this PR.

> Testing the connection to the handler whether it's OK - how the user can tell if the entered details are working OK?

This is beyond of the scope of this PR. It needs additional support from 3scale adapter too. Once that logic is well defined it can be added to kiali.

> If there are several handlers, is there a possible overlap/override that can cause some unpredictable behavior? If rule A uses handler 1 and rule B uses handler 2, what is the expected behavior?

There is a 1to1 between a rule and a service. So, in your example, if rule/service A uses handler 1, it will autenticate with the 3scale endpoint defined (it can be a different 3scale App or API), same for rule/service B, it will authenticate against different endpoints.

It's a cornercase as I think more normal scenario is to have a single handler scenario, but valid.

src/pages/ServiceDetails/ServiceInfo/ServiceInfoDescription.tsx
```diff
@@ -83,6 +85,9 @@ class ServiceInfoDescription extends React.Component<ServiceInfoDescriptionProps
               <div>
                 <strong>Resource Version</strong> {this.props.resourceVersion}
               </div>
+              {this.props.threeScaleServiceRule && this.props.threeScaleServiceRule.threeScaleHandlerName !== '' && (
+                <div>Service linked with 3scale API</div>
+              )}
             </Col>
```
aljesusg: Remove div please
```typescript
{this.props.threeScaleServiceRule && this.props.threeScaleServiceRule.threeScaleHandlerName !== '' && (
               'Service linked with 3scale API'
              )}
```

lucasponce:
@aljesusg done, thanks for the review.
Also, there is a backend PR related for this feature.

vramosp:
>> Note, that user can create multiple handlers, but it will be typically only one per cluster, that will be used for multiple services. But multiple handlers can be created using the "Show Advanced Options".

At the moment I believe we use one handler per service, although we're trying to improve that part so only one per cluster (or SM instance) is needed.

lucasponce:
> At the moment I believe we use one handler per service, although we're trying to improve that part so only one per cluster (or SM instance) is needed.

Here handler is the triple serviceId/systemUrl/access_token right ?

So, assuming we are calling same API, where the serviceId and systemUrl will be similar, this means that the access_token would be different per service ?

vramosp:
@lucasponce Not the access_token, but the serviceId

lucasponce:
@vramosp ok, then I guess I flipped a term here

3scale "serviceId" is the API ID:

image

And here in 3scale terminology, with that "serviceId" we are refering to all services "mapped" under that API, right ?

lucasponce:
@vramosp a couple of adjustments:
image

And 
image

Let me know how it looks like, I'd like to consolidate this feature to not reach the code freeze deadline.

Also, please let me know if there is any change in the 3scale adapter that can impact this logic.

Thanks

lucasponce:
Thanks @aljesusg, also this PR needs the backend one https://github.com/kiali/kiali/pull/1050 which also needs a review :-)

hhovsepy:
Tested Create/Update/Delete of handlers and Management Rules.
Just a minor request: Add the regexp validation into Handler name field while creating it in wizard, to void errors from OC side:
Screenshot from 2019-05-13 19-18-18

vramosp:
>> different combination of serviceId/accesstoken/systemUrl, user need to create use a new handler.

Correct.

lucasponce:
Good catch @hhovsepy that validation should be in the form.

lucasponce:
@hhovsepy added requested validation
image

Also, some minor refactoring that could affect, would you mind to re-test ? thanks

abonas:
@lucasponce - does the wizard currently contain only an advanced options section?

lucasponce:
> @lucasponce - does the wizard currently contain only an advanced options section?

Yes

PR #1165, opened 2019-05-09
KIALI-2832 Clean errors if there is no namespace selected

josejulio:
** Describe the change **

Clean errors when requesting graph data without namespaces
The issue happens because the graph data without namespaces just returns and doesn't clean previous errors (if any)

The error will stop appearing when Kiali notices that we don't have that namespace anymore (e.g. when opening the namespace selector) 

** Screenshot **

no-namespace-shows-error

### DIFF
--- a/src/actions/GraphDataActions.ts
+++ b/src/actions/GraphDataActions.ts
@@ -6,7 +6,8 @@ enum GraphDataActionKeys {
   GET_GRAPH_DATA_START = 'GET_GRAPH_DATA_START',
   GET_GRAPH_DATA_SUCCESS = 'GET_GRAPH_DATA_SUCCESS',
   GET_GRAPH_DATA_FAILURE = 'GET_GRAPH_DATA_FAILURE',
-  HANDLE_LEGEND = 'HANDLE_LEGEND'
+  HANDLE_LEGEND = 'HANDLE_LEGEND',
+  GET_GRAPH_DATA_WITHOUT_NAMESPACES = 'GET_GRAPH_DATA_WITHOUT_NAMESPACES'
 }
 
 // synchronous action creators
@@ -24,6 +25,7 @@ export const GraphDataActions = {
   getGraphDataFailure: createAction(GraphDataActionKeys.GET_GRAPH_DATA_FAILURE, resolve => (error: any) =>
     resolve({ error: error })
   ),
+  getGraphDataWithoutNamespaces: createAction(GraphDataActionKeys.GET_GRAPH_DATA_WITHOUT_NAMESPACES),
   handleLegend: createAction(GraphDataActionKeys.HANDLE_LEGEND)
 };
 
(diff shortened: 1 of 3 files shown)


israel-hdez:
Hmmm I think this doesn't really fixes the Jira... I think @mattmahoneyrh should check if this is what he expects.

josejulio:
> Hmmm I think this doesn't really fixes the Jira... I think @mattmahoneyrh should check if this is what he expects.

Yes, but please check what happened before, even when we knew the namespace didn't exist (when loading the namespaces) the error persisted there.

mwringe:
@josejulio Could you explain a bit what this is meant to change?

I see if I return to the graph and the namespace I was currently viewing is no longer available that I get an error message and that it automatically clears the namespace list.

I also see that if I was viewing 2 namespaces before and return where only one is deleted then I don't get any error message (but the deleted one is removed from the list).

And I see if I stay on the graph and I delete the namespace I am viewing then I get an error message and it doesn't clear the namespace list.

Its a bit inconsistent in how this is working, could we also make it work better across these scenarios?

josejulio:
> @josejulio Could you explain a bit what this is meant to change?

The way it worked **before** this PR is that when we get a new namespace list (when we fetch it) it compares that every "active namespace" is on that list, the ones that are missing are removed.
One way to fetch the namespace list is to open the namespace dropdown.
In the context of this bug, before this PR:

1. Add a namespace "bookinfo2"
2. Mark "bookinfo2" namespace as active by selecting it on the dropdown list
3. Delete namespace "bookinfo2" (without reloading the list by clicking the namespace dropdown)
4. Wait for the autorefresh (or refresh directly from the app, not the browser refresh)
5. Verify that you see the error about the non existent namespace "bookinfo2"
6. Open the namespace dropdown
7. Verify that the error does not go away and still complains about "bookinfo2"

Current code does the same up to 6, but on 7:

7. Verify that "bookinfo2" got unselected and no error appears

The current fix clears any error on the graph when there is no namespace selected.

> I also see that if I was viewing 2 namespaces before and return where only one is deleted then I don't get any error message (but the deleted one is removed from the list).

The same holds true if you select more namespaces, I suspect you opened the namespace list and it "cleared" the deleted namespace before getting the error.

After talking with @mattmahoneyrh I'm going to try to detect the error and refresh the namespace list on that moment, as soon as we get that the namespace does not exist it should clear it from the "actives" and refresh the graph.

jshaughn:
I guess i don't totally understand the problem here.  From what I am reading the user is presented with an error message like, "Requested namespace [bookinfo] is not accessible.".  If he then opens the namespace dropdown it will get cleared and he can proceed without error, right?   So the "bug" is that the error message repeats is auto-refresh is on and the user just stares at the screen?  Maybe I don't get it but if that is the case I don't think any more work is warranted.

If we absolutely must do this then perhaps on any 403 from the graph API we could refresh the namespaces.

mwringe:
@mattmahoneyrh Can you take a quick look? at least to make sure it doesn't cause any issues in your testing?

josejulio:
Going to merge this before it gets old, if we want to do something else in this area please create other jira and I'll follow up, thanks!

PR #1166, opened 2019-05-10
KIALI-2826 App Box label should increase in size when highlighted

jshaughn:
To get this to work the label had to be centered, the offset calculation
seemed confused when the app-box and/or font size would change.  Perhaps
it needed more refreshing somewhere.


jshaughn:
I created [design issue 126](https://github.com/kiali/kiali-design/issues/126) to propose centered labeling for app boxes.  If that proposal is approved then this PR can be moved forward.

jshaughn:
@mwringe @josejulio @mattmahoneyrh , So this PR does two things: the app box label has now moved to bottom center as approved in the UX design issue, and the label should properly resize when the app box is highlighted (via hover, selecting the app box, or selecting something else that includes the app box in it's end-to-end paths).  Highlight and dimming should all behave as expected.

PR #1168, opened 2019-05-10
KIALI-2831 Add unary 'traffic' operator for edges

jshaughn:
Can now easily find edges with any traffic for any protocol, or negated can
find edges with no traffic for all protocols.

Also, slightly improve unary patterns to now match in a truthy way.

kiali-2831


jshaughn:
@jotak Any thoughts/review about this approach?

PR #1169, opened 2019-05-10
KIALI-2829 Don't hide graph-find clear buttons prematurely

jshaughn:
The buttons can be hidden only when the input is clear *and* submitted. Otherwise
editing is in progress.

also: fix some static keys and ids to be consistently formed.

@xeviknal This video shows the clear button staying in place until the edit is submitted.  Is this sufficient or do you believe we should auto-submit if the user edits to an empty value?  I feel like that is a non-standard thing to do but let me know.

kiali-2829

### DIFF
--- a/src/components/GraphFilter/GraphFind.tsx
+++ b/src/components/GraphFilter/GraphFind.tsx
@@ -113,13 +113,13 @@ export class GraphFind extends React.PureComponent<GraphFindProps, GraphFindStat
                 onKeyPress={this.checkSubmitFind}
                 placeholder="Find..."
               />
-              {this.state.findInputValue && (
+              {this.props.findValue && (
                 <OverlayTrigger
-                  key="ot-clear_find"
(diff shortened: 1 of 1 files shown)


josejulio:
> @xeviknal This video shows the clear button staying in place until the edit is submitted. Is this sufficient or do you believe we should auto-submit if the user edits to an empty value? I feel like that is a non-standard thing to do but let me know.

Maybe we could somehow indicate that the current query is not applied?

jshaughn:
> Maybe we could somehow indicate that the current query is not applied?

I'm not sure I like that, again, it seems non-standard.  If we had to add an indicator I'd probably prefer to just submit whenever the box is cleared.

jshaughn:
@xeviknal, @abonas, I have updated the PR to behave as you prefer.  Editing to an empty input for find/hide is now equivalent to clicking the clear ('x') button.

PR #1174, opened 2019-05-14
KIALI-2636 Adds DurationDropdown to detail pages

josejulio:
** Describe the change **

Adds a way to configure the time of the health in the details page

** Screenshot **

KIALI-2636

@abonas Is this what you meant? If so i can continue to add it to the workload and app page.

### DIFF
--- a/src/components/DurationDropdown/DurationDropdown.tsx
+++ b/src/components/DurationDropdown/DurationDropdown.tsx
@@ -1,0 +1,61 @@
+import ToolbarDropdown from '../ToolbarDropdown/ToolbarDropdown';
+import { Durations, serverConfig } from '../../config/ServerConfig';
+import * as React from 'react';
+import { DurationInSeconds } from '../../types/Common';
+import { KialiAppState } from '../../store/Store';
+import { durationSelector } from '../../store/Selectors';
+import { ThunkDispatch } from 'redux-thunk';
+import { KialiAppAction } from '../../actions/KialiAppAction';
+import { bindActionCreators } from 'redux';
+import { UserSettingsActions } from '../../actions/UserSettingsActions';
+import { connect } from 'react-redux';
+
+type BasicDurationDropdownProps = {
+  duration: DurationInSeconds;
+  disabled?: boolean;
+  id?: string;
+  setDuration: (duration: DurationInSeconds) => void;
+  tooltip?: string;
+};
+
+// These are taken from the serverConfig
+type ExtendedDurationDropdownProps = BasicDurationDropdownProps & {
+  durations: Durations;
+};
+
+export class DurationDropdown extends React.Component<ExtendedDurationDropdownProps> {
+  render() {
+    return (
+      <ToolbarDropdown
+        id={this.props.id}
+        disabled={this.props.disabled}
+        handleSelect={key => this.props.setDuration(Number(key))}
+        value={this.props.duration}
+        label={this.props.durations[this.props.duration]}
+        options={this.props.durations}
+        tooltip={this.props.tooltip}
+      />
+    );
+  }
+}
+
+export const DurationIntervalWithDurations = (props: BasicDurationDropdownProps) => {
+  return <DurationDropdown durations={serverConfig.durations} {...props} />;
+};
+
+const mapStateToProps = (state: KialiAppState) => ({
+  duration: durationSelector(state)
+});
+
+const mapDispatchToProps = (dispatch: ThunkDispatch<KialiAppState, void, KialiAppAction>) => {
+  return {
+    setDuration: bindActionCreators(UserSettingsActions.setDuration, dispatch)
+  };
+};
+
+const DurationDropdownContainer = connect(
+  mapStateToProps,
+  mapDispatchToProps
+)(DurationIntervalWithDurations);
+
+export default DurationDropdownContainer;
--- a/src/components/DurationDropdown/DurationDropdown.tsx
+++ b/src/components/DurationDropdown/DurationDropdown.tsx
@@ -9,21 +9,26 @@ import { KialiAppAction } from '../../actions/KialiAppAction';
 import { bindActionCreators } from 'redux';
 import { UserSettingsActions } from '../../actions/UserSettingsActions';
 import { connect } from 'react-redux';
+import { HistoryManager, URLParam } from '../../app/History';
+import history from '../../app/History';
 
-type BasicDurationDropdownProps = {
+type ReduxProps = {
   duration: DurationInSeconds;
+  setDuration: (duration: DurationInSeconds) => void;
+};
+
+type BasicDurationDropdownProps = ReduxProps & {
   disabled?: boolean;
   id?: string;
-  setDuration: (duration: DurationInSeconds) => void;
   tooltip?: string;
 };
 
 // These are taken from the serverConfig
-type ExtendedDurationDropdownProps = BasicDurationDropdownProps & {
+type DurationDropdownProps = BasicDurationDropdownProps & {
   durations: Durations;
 };
 
-export class DurationDropdown extends React.Component<ExtendedDurationDropdownProps> {
+export class DurationDropdown extends React.Component<DurationDropdownProps> {
   render() {
     return (
       <ToolbarDropdown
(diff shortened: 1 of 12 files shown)


josejulio:
Rebased and updated the gif, it now coves the workload, service and app details.

jotak:
@josejulio I was out of office last week, i'll take care about it today / tomorrow

abonas:
this is looking great!

josejulio:
> Do you have the chance to sync the duration dropdowns? Just to put some order to this disaster :)

Ah yes, it was something that was bothering me but forgot about it

> Now, most (or all?) tabs have the duration dropdown. When changing the duration in one tab, the setting is sometimes propagated to the other tabs, but sometimes not. Even more, there are cases where the setting is propagated only to some of the other tabs. This is a little confusing, since I would expect consistency across the tabs.

It seems that before the "Traffic" tab was the one to go out ot sync, If you change it in `Inbound/Outbound Metrics`,  `Traffic` won't reflect the change, but the other way around works.

Now `Info` Is added to the mix :)

josejulio:
The easy way to fix is to add `mountOnEnter={true} unmountOnExit={true}` to every `TabPane` that uses this MetricDuration component. 
That way we avoid any complication.

The more robust fix (I think) would be to fetch the duration from redux and let redux help with the syncing.

Easy fix is OK for this PR IMO.

jshaughn:
Logs/Inbound/Outbound are synced afaik.  I don't think Traffic is synced.  If we are going to have the dropdown across the tabs I think Info and Traffic should likely also be synced.

josejulio:
> Logs/Inbound/Outbound are synced afaik. I don't think Traffic is synced. If we are going to have the dropdown across the tabs I think Info and Traffic should likely also be synced.

I just "synced" them in the last commit

jotak:
Wait wait wait... It's not without a reason that not everything is synced. Duration in Graph doesn't have the same meaning than, let's say, in metrics. Graph show aggregated-in-time data whereas metrics show not-aggregated-in-time timeseries. This is a big difference.

IMO, what should be synced together are:
- Graph
- Pages with health-related queries (details, lists, overview)

Metrics page must not have synced duration.
For logs, well, I'm not sure. Maybe yes, as we're going to check logs mostly when health is red, so it must coincide with health.

src/pages/AppDetails/AppDetailsPage.tsx
```diff
@@ -117,6 +125,10 @@ class AppDetails extends React.Component<RouteComponentProps<AppId>, AppDetailsS
     return istioSidecar;
   }
 
+  updateRateInterval = (rateInterval: DurationInSeconds) => {
+    this.setState({ rateInterval }, () => this.doRefresh());
```
jotak: This seems to work, but I'm curious... how?
I mean how `this.setState({ rateInterval })` can work where I would expect the state object be more like `{ rateInterval: rateInterval }`? Some dark magic happening here?

jshaughn:
> IMO, what should be synced together are: Graph, Pages with health-related queries (details, lists, overview)

The Overview page and Graph do have synced durations in master.  The duration set there is not applied to the list or detail pages afaik.   I can see an argument for also applying that same setting to the list pages, as they also consider health across many elements.   When moving to detail the user defaults to the Overview tab (aka Info tab).  The health reflected there should be the same as the entry on the list page, so it to should share the same duration.    The other tabs, Logs, Inbound, Outbound, Traffic, I think they should be synced with each other for consistency across the tabs but not with Overview/Graph/Lists/Overview tab.  The value there, as pointed out by Joel, is often going to be a larger duration to be useful.

src/pages/AppDetails/AppDetailsPage.tsx
aljesusg: @jotak you can do this if the name of the variable is the same that the property in the state, typescript sometimes rocks ^^

jshaughn: One man's rocks are another man's mud :)  Personally, it's this sort of optimization that just makes code confusing.

jotak: @aljesusg thanks, good to know! But too much fancy syntactic sugar for me :-)

jotak: @jshaughn this is where we can see we're getting old

israel-hdez: @jotak @jshaughn Haha... I think it is not related to oldness... Otherwise, I'd be feeling "old" about 3-5 years after writing my first line of code :P

BTW, that syntax is "new" in ES6. Read this little section:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#Property_definitions

josejulio: Yes, is dark magic :)

josejulio:
> The Overview page and Graph do have synced durations in master. The duration set there is not applied to the list or detail pages afaik. I can see an argument for also applying that same setting to the list pages, as they also consider health across many elements. When moving to detail the user defaults to the Overview tab (aka Info tab). The health reflected there should be the same as the entry on the list page, so it to should share the same duration. The other tabs, Logs, Inbound, Outbound, Traffic, I think they should be synced with each other for consistency across the tabs but not with Overview/Graph/Lists/Overview tab. The value there, as pointed out by Joel, is often going to be a larger duration to be useful.

Currently the "Overview" (or info) tab is synced with the other  tabs. I think I can fix this in my other PR.

josejulio:
Updated so that Overview (info) tab is now in sync with the graph and the overview main page.
Did some small refactoring to have a new DurationDropdown component / container

src/pages/AppDetails/AppDetailsPage.tsx
```diff
@@ -15,13 +15,21 @@ import { fetchTrafficDetails } from '../../helpers/TrafficDetailsHelper';
...
+import { connect } from 'react-redux';
 
 type AppDetailsState = {
   app: App;
   health?: AppHealth;
   trafficData: GraphDefinition | null;
 };
 
+type AppDetailsProps = RouteComponentProps<AppId> & {
+  duration: DurationInSeconds;
```
jshaughn: Same recommendation for ReduxProps convention.

src/components/DurationDropdown/DurationDropdown.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { DurationInSeconds } from '../../types/Common';
+import { KialiAppState } from '../../store/Store';
+import { durationSelector } from '../../store/Selectors';
+import { ThunkDispatch } from 'redux-thunk';
+import { KialiAppAction } from '../../actions/KialiAppAction';
+import { bindActionCreators } from 'redux';
+import { UserSettingsActions } from '../../actions/UserSettingsActions';
+import { connect } from 'react-redux';
+
+type BasicDurationDropdownProps = {
```
jshaughn: Thanks for adding this overdue component.  The convention is to declare ReduxProps for anything being supplied/managed by the container.  So I'd suggest removing duration and setDuration from BasicDurationDropDownProps and declaring them in ReduxProps.  Also, I think drop the 'Extended' prefix, Something like:


```
type ReduxProps = {
  duration: DurationInSeconds;
  setDuration: (duration: DurationInSeconds) => void;
}

type BasicDurationDropdownProps = {
...
}

type DurationDropdownProps = ReduxProps & BasicDurationDropdownProps & {
  durations: Durations;
}

export class DurationDropdown extends React.Component<DurationDropdownProps>

```

src/pages/Overview/OverviewToolbar.tsx
```diff
@@ -83,14 +83,14 @@ export class OverviewToolbar extends React.Component<Props, State> {
...
-  componentDidUpdate() {
+  componentDidUpdate(prevProps: Props) {
     // ensure redux state and URL are aligned
     HistoryManager.setParam(URLParam.DURATION, String(this.props.duration));
     HistoryManager.setParam(URLParam.POLL_INTERVAL, String(this.props.refreshInterval));
 
     const urlSortField = ListPagesHelper.currentSortField(Sorts.sortFields);
     const urlIsSortAscending = ListPagesHelper.isCurrentSortAscending();
-    if (!this.paramsAreSynced(urlSortField, urlIsSortAscending)) {
+    if (!this.paramsAreSynced(urlSortField, urlIsSortAscending) || this.props.duration !== prevProps.duration) {
```
jshaughn: Is this to replace the fact that onRefresh is no longer called when a duration is selected in the dropdown?

src/pages/AppDetails/AppDetailsPage.tsx
```diff
@@ -30,8 +38,8 @@ const emptyApp = {
   runtimes: []
 };
 
-class AppDetails extends React.Component<RouteComponentProps<AppId>, AppDetailsState> {
-  constructor(props: RouteComponentProps<AppId>) {
+class AppDetails extends React.Component<AppDetailsProps, AppDetailsState> {
+  constructor(props: AppDetailsProps) {
```
jshaughn: I think now that this is sharing the duration setting in redux, it must also respect and manage the setting in the URL.  See OverviewToolbar, for example.  We typically have logic in the constructor and also componentDidUpdate,  This comment would apply to the other detail pages as well.

src/pages/Overview/OverviewToolbar.tsx
josejulio: Yes, instead of connecting that callback, we "listen" for changes on `duration`

josejulio:
Just pushed an update to make the DurationDropdownContainer aware of the URL.

PR #1177, opened 2019-05-15
KIALI-2872 Adjust focus animation

josejulio:
** Describe the change **

Adjust focus animation.

It takes 2 seconds and the first circle in takes 1s, the out half a second and the second "in" the other half second.

I could do something different, please comment what you have in mind.

** Screenshot **

Before:
focus-animation-17

After:

a)
focus-animation-16

b)
focus-animation-15

c) same as (a) but with color PfColors.Blue300

focus-animation-18

### DIFF
--- a/src/components/CytoscapeGraph/FocusAnimation.ts
+++ b/src/components/CytoscapeGraph/FocusAnimation.ts
@@ -98,7 +98,7 @@ export default class FocusAnimation {
 
   private render(element: any, radio: number) {
     const { x, y } = this.getCenter(element);
-    this.context.strokeStyle = PfColors.Black;
+    this.context.strokeStyle = PfColors.Blue300;
     this.context.lineWidth = LINE_WIDTH;
     this.context.beginPath();
     this.context.arc(x, y, radio, 0, 2 * Math.PI, true);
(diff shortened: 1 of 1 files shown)


mtho11:
Can you also add the before screenshot so we can compare?

josejulio:
> Can you also add the before screenshot so we can compare?

I added some, but the quality of my recording is not really good.

jmazzitelli:
I like (a) better. The red obviously won't help color blind people - red is also the color of "ERROR! Something is wrong here!!".

josejulio:
Current code uses option (a).

mtho11:
Personally, I kinda like option b.

jshaughn:
@josejulio option a or b are both fine to me, although in both cases I think I'd prefer the same color as is used for a selected node.  The problem there is that the color differs depending on health.  If we had to settle on one color I'd go with the non-degraded select color (PfColors.Blue300), or just black.

josejulio:
Health is fetched dynamically, so i don't think we can use the health color for this.

josejulio:
> PfColors.Blue300

With that color it looks like this:
focus-animation-18

abonas:
not with red please.
cc @serenamarie125 fyi

josejulio:
Current code has the PfColors.Blue300
I think we can iterate further if UXD proposes something else

PR #1179, opened 2019-05-16
KIALI-2863 Increase font of Delete Dialogs

lucasponce:
Increase font on Delete Dialogs.


skondkar:
Verified with delete confirmation dialog box. The font size is increased and readable.

Screenshot-Delete-Confirmation-Font

PR #1181, opened 2019-05-16
kiali-2827 Improve handling of missing sidecar icon on detail pages

jshaughn:
Because Info tab is the default it renders multiple times, first without all
of the item detail, and then again after the detail is available/fetched. By
assuming a missing sidecar the icon is rendered initially.
- Don't assume missing sidecar by default
  - Only show icon when explicitly set false for optional prop
- Don't render "Show on Graph" link or MS icon until title is available
- Consistently position "Show on Graph" link on the right (i.e. after MS icon)

kiali-2827

### DIFF
--- a/src/components/Pf/PfInfoCard.tsx
+++ b/src/components/Pf/PfInfoCard.tsx
@@ -24,17 +24,15 @@ class PfInfoCard extends React.Component<CardProps> {
           <h2 className="card-pf-title">
             <Icon type={this.props.iconType} name={this.props.iconName} style={{ marginRight: '10px' }} />
             {this.props.title}
-            {this.props.istio ? (
-              this.props.showOnGraphLink && (
-                <>
-                  {' '}
-                  (<Link to={this.props.showOnGraphLink}>Show on graph</Link>)
-                </>
-              )
-            ) : (
+            {this.props.title && this.props.istio !== undefined && !this.props.istio && (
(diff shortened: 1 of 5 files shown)


src/components/Pf/PfInfoCard.tsx
```diff
@@ -24,18 +24,16 @@ class PfInfoCard extends React.Component<CardProps> {
...
-                  (<Link to={this.props.showOnGraphLink}>Show on graph</Link>)
-                </>
-              )
-            ) : (
+            {this.props.title && this.props.istio !== undefined && !this.props.istio && (
               <span style={{ marginLeft: '10px' }}>
                 <MissingSidecar />
               </span>
             )}
+            {this.props.title && this.props.showOnGraphLink && (
```
israel-hdez: Something changed, not sure why.
Previous behavior was to hide "Show on graph" if sidecar is missing. Now, it is always present.

jshaughn: Yes, I changed that on purpose because I didn't see a reason to not show it on the graph if it was lacking a sidecar.  The graph also shows nodes lacking a sidecar,  especially when the "show unused" option is selected, which is true when this link is generated, right?  Am I missing something?

israel-hdez: If it was intentional, it is OK. I have no strong opinion.

PR #1182, opened 2019-05-16
KIALI-2394 Don't update the graph if it is still empty

mwringe:
** Describe the change **

Previously if the graph was empty it would still switch to the 'loading' graph page when fetching new 
data. The loading graph page should only show up on the initial load and not in between data fetches (like how it works when the graph has some data).

** Issue reference **

https://issues.jboss.org/browse/KIALI-2394


mwringe:
Adding a DNM label, there is a couple of issues with this PR as is that needs to be fixed.

PR #1188, opened 2019-05-20
KIALI-2851 Tracing tab when tracing is disabled

aljesusg:
Jira: [KIALI-2851](https://issues.jboss.org/browse/KIALI-2851)

PR #1189, opened 2019-05-20
KIALI-2849 Tracing options still appear when removed in the backend

aljesusg:
JIRA: [KIALI-2849](https://issues.jboss.org/browse/KIALI-2849)

- Now in authentication process if jaeger is disabled we set the values to false, the problem was in the persist configuration
- Fixed externalIcon , it wasn't working with PF4
- This is fixed in https://github.com/kiali/kiali-ui/pull/1188 where we check if the errTraces is not -1, anyway I added an extra control with the URL.


aljesusg:
Thanks @hhovsepy

PR #1191, opened 2019-05-20
KIALI-2728 Introduce Desired/Current/Available Pod Status details

lucasponce:
List Pods individually
This UI needs backend PR https://github.com/kiali/kiali/pull/1081.

Pod Status details desired/current/available info.

This will help to give user more information on Pod problems.

image

image

Also, another problem we had in our previous design, is that the Group Pods can't be really merged, as in situation like these, you want to see what are the failed Pods. So, now Pod table shows all individual Pods (as it happens with OpenShift Console).

image

Note that for an investigation fields like Phase or CreatedAt can be completely different, so we can't merge it.


skondkar:
Verified that the pods status shows - details desired/current/available info - and Pod table shows all individual Pods.

PR #1192, opened 2019-05-20
KIALI-2888 Fix regression of kiali-2827

jshaughn:
kiali-2827 made it such that the missing sidecar icon did not flash prior to completely rendering a workload/app/service detail.  It did a few other minor things as well.  The rendering changes were in PfInfoCard, which was lost when kiali-2840 stopped using that class and moved logic to PfTitle.

- don't assume a missing sidecar until it is proven that there is at least one pod with a missing sidecar
- consistently position "Show on graph" link to the right
- remove now-unused PfInfoCard and related tests


src/components/Pf/PfTitle.tsx
```diff
@@ -90,18 +90,16 @@ class PfTitle extends React.Component<PfTitleProps, PfTitleState> {
     return (
       <h2 className={PfTitleStyle}>
         {this.state.icon} {this.state.name}
-        {this.props.istio ? (
-          !this.props.hideGraphLink && (
-            <>
```
aljesusg: If we are never going to use  this.props.hideGraphLink we should remove from https://github.com/kiali/kiali-ui/blob/2b5aef79176372e58336e8d95343e7ffa18677e5/src/components/Pf/PfTitle.tsx#L22

aljesusg: If we are going to show the link only if there is a name it's ok for me. But I think that we don't need that check  we always expect a name in the url

jshaughn: +1 I don't even know where that came from. I'll remove.

aljesusg: It was in the code I didn't know if the author of the graph link wanted to control this

PR #1193, opened 2019-05-20
KIALI-2884 Add DurationDropdown to list pages

josejulio:
** Describe the change **

Allows to configure the rate interval on the list pages.

** Screenshot **

kiali-2884

### DIFF
--- a/src/pages/AppList/AppListComponent.tsx
+++ b/src/pages/AppList/AppListComponent.tsx
@@ -178,7 +178,7 @@ class AppListComponent extends ListComponent.Component<AppListComponentProps, Ap
             />
           </Sort>
           <ToolbarRightContent style={{ ...AlignRightStyle }}>
-            <DurationDropdownContainer />
+            <DurationDropdownContainer id="app-list-dropdown" />
             <Button onClick={this.updateListItems}>
               <Icon name="refresh" />
             </Button>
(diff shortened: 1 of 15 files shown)


src/pages/AppList/AppListComponent.tsx
```diff
@@ -16,14 +16,14 @@ import { AlignRightStyle, ThinStyle } from '../../components/Filters/FilterStyle
...
 
-interface AppListComponentState extends ListComponent.State<AppListItem> {
-  rateInterval: number;
-}
+interface AppListComponentState extends ListComponent.State<AppListItem> {}
 
 interface AppListComponentProps extends ListComponent.Props<AppListItem> {
   rateInterval: number;
   activeNamespaces: Namespace[];
+  onRateIntervalChanged: (rateInterval: number) => void;
```
jshaughn: Can we not use the "RateInterval" terminology and consistently use "Duration" for durations.  Someone correct me if I'm wrong, that this is not a rate interval as used in prometheus queries but rather the duration/time range of the queried data.  Also, the unit should likely be DurationInSeconds, not number.

src/pages/AppList/AppListPage.tsx
```diff
@@ -3,21 +3,42 @@ import { Breadcrumb } from 'patternfly-react';
...
-        currentSortField={ListPagesHelper.currentSortField(AppListFilters.sortFields)}
-        isSortAscending={ListPagesHelper.isCurrentSortAscending()}
-        rateInterval={ListPagesHelper.currentDuration()}
-      />
-    </>
-  );
+type AppListPageProps = {};
+
+type AppListPageState = {
+  rateInterval: number;
```
jshaughn: As discussed in the other Jira, the duration used here should likely be synced with the Overview and Graph pages because it is being used to determine health of the list elements.  As such this should likely be using the redux value.

josejulio:
@jshaughn Updated this one, but this one now requires my other pr #1174

PR #1171, opened 2019-05-13
KIALI-2840 Redesign Service Details

aljesusg:
JIRA: [KIALI-2840](https://issues.jboss.org/browse/KIALI-2840)

Design: https://github.com/kiali/kiali-design/issues/116

## Demo

redesign

cc @serenamarie125 @abonas

PR #1196, opened 2019-05-21
KIALI-2840 Rename Info tab to Overview

aljesusg:
JIRA: [KIALi-2893](https://issues.jboss.org/browse/KIALI-2893)

Continue of https://github.com/kiali/kiali-ui/pull/1171
- Rename info tab to Overview like design https://github.com/kiali/kiali-design/issues/116
- Clean code, remove empty ```<div></div>```

cc @abonas @skondkar

### DIFF
--- a/src/pages/ServiceDetails/ServiceDetailsPage.tsx
+++ b/src/pages/ServiceDetails/ServiceDetailsPage.tsx
@@ -262,15 +262,9 @@ class ServiceDetails extends React.Component<ServiceDetailsProps, ServiceDetails
         >
           <div>
             <Nav bsClass="nav nav-tabs nav-tabs-pf">
-              <NavItem eventKey="info">
-                <div>Overview</div>
-              </NavItem>
-              <NavItem eventKey="traffic">
-                <div>Traffic</div>
-              </NavItem>
-              <NavItem eventKey="metrics">
-                <div>Inbound Metrics</div>
-              </NavItem>
+              <NavItem eventKey="info">Overview</NavItem>
+              <NavItem eventKey="traffic">Traffic</NavItem>
+              <NavItem eventKey="metrics">Inbound Metrics</NavItem>
               {errorTraces !== undefined &&
                 errorTraces > -1 &&
                 (this.props.jaegerIntegration ? (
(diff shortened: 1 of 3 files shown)


src/pages/ServiceDetails/ServiceDetailsPage.tsx
```diff
@@ -262,15 +262,9 @@ class ServiceDetails extends React.Component<ServiceDetailsProps, ServiceDetails
...
-              </NavItem>
-              <NavItem eventKey="traffic">
-                <div>Traffic</div>
-              </NavItem>
-              <NavItem eventKey="metrics">
-                <div>Inbound Metrics</div>
-              </NavItem>
+              <NavItem eventKey="info">Overview</NavItem>
+              <NavItem eventKey="traffic">Traffic</NavItem>
+              <NavItem eventKey="metrics">Inbound Metrics</NavItem>
```
abonas: in another file in this PR Inbound metrics key is in_metrics and here it is just metrics. isn't it a bit confusing/not consistent?

src/pages/WorkloadDetails/WorkloadDetailsPage.tsx
```diff
@@ -183,27 +183,17 @@ class WorkloadDetails extends React.Component<RouteComponentProps<WorkloadId>, W
...
-              </NavItem>
-              <NavItem eventKey="in_metrics">
-                <div>Inbound Metrics</div>
-              </NavItem>
-              <NavItem eventKey="out_metrics">
-                <div>Outbound Metrics</div>
-              </NavItem>
+              <NavItem eventKey="info">Overview</NavItem>
+              <NavItem eventKey="traffic">Traffic</NavItem>
+              <NavItem eventKey="logs">Logs</NavItem>
```
abonas: is the order here OK? in other places it's traffic, then metrics, here it is traffic, logs, metrics.
(I know this existed prior to this PR, and doesn't need to be fixed as part of this PR)

src/pages/ServiceDetails/ServiceDetailsPage.tsx
aljesusg: Totally agree I am think about move all of this in a component to reuse it and avoid this, but we should do it in another PR

abonas: no objections from my side

PR #1202, opened 2019-05-22
KIALI-2901 Fix missing nulls in MergeJsonPatch

lucasponce:
This fix adds key: null into the mergeJsonPatch used by Kiali to update an Istio config resource.
Those null in the patch values are necessary when an attribute is removed from the yaml view and transformed into an non key. The Merge Patch needs a key with a null value to perform the delete.


burmanm:
Works fine for me (all the combinations I tried)

PR #1203, opened 2019-05-22
KIALI-2903 protect against missing session on logout

jshaughn:
There may be a better fix than this but this adds a simple guard.  It seems that during logout the session in redux gets set undefined, which can cause a re-render of the UserDropdown (due to the state change) before that element is unmounted, which then crashes because it assumes a session.


jshaughn:
Merging to fix the blocker, @aljesusg can follow up if he thinks there is a better solution.

PR #1206, opened 2019-05-24
KIALI-2165 Add npm-force-resolutions so that npm can behave like yarn

mtho11:
** Describe the change **
Add npm-force-resolutions so that npm can enforce the resolutions block just like yarn does. This allows npm to act like yarn 

_Cause_:
_The only reason this is needed is for internal Red Hat production build environments that do not allow yarn to be installed and instead use npm._

If we didn't have this then some CVE resolutions may not be enforced in the sub packages.

** Issue reference **
https://issues.jboss.org/browse/KIALI-2165
https://www.npmjs.com/package/npm-force-resolutions

** Backwards compatible? **
Yes

No

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,8 +24,6 @@ npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
 
-package-lock.json
-
 # compiled patternfly assets
 src/**/*.css
 src/**/*.css.map
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,1 @@
 # Avoid the generation of package-lock.json
-package-lock=false
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,19009 @@
+{
+  "name": "@kiali/kiali-ui",
+  "version": "0.21.0",
+  "lockfileVersion": 1,
+  "requires": true,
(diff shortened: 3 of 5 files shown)


josejulio:
If we are including a package-lock.json, we should configure snyk to scan both yarn and package-lock.json, don't you think?

mtho11:
@josejulio I suppose we could do that. Would it be any different? I mean it's the same package.json libraries; that is why we are adding the npm-resolutions is to make sure they use the same resolution policies between npm/yarn.

josejulio:
> @josejulio I suppose we could do that. Would it be any different?

I'm not sure, in the package description it says (emphasis mine):

> **similar** to yarn's selective dependency 

So I was wondering if is always the same or not.

If you are confident enough that they are the same i suppose is ok to only use the yarn one, but then, i think we have to configure snyk to use yarn's, because it might try to use the package-lock.json if it finds it. ( i recall i read it in their docs)

mtho11:
FYI: the docs https://snyk.io/docs/snyk-for-nodejs/#source-control-integration

josejulio:
@mtho11 going to close this as we don't really need this, thanks!

PR #1207, opened 2019-05-24
update the base url for relative url

inyee786:
** Describe the change **
 Update the base url to take the relative url not the absolute path
  while running the kiali under reverse proxy for the url like
 `abc.com/xyz/def/api/v1/namespaces/istio-system/services/kiali:20001/proxy/`
   it takes the wrong base url URL `abc.com/api/v1/namespaces/istio-system/services/kiali:20001/proxy/`

 To solve this I have taken the relative URL and it starts taking the relative url 

** Issue reference **
* If this is a fix for a GH-issue, please link it here
   https://github.com/kiali/kiali/issues/1059

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -4,7 +4,7 @@
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no" />
     <meta name="theme-color" content="#000000" />
-    <base href="/" />
+    <base href="./" />
     <script type="text/javascript" src="./env.js"></script>
     <!--
       manifest.json provides metadata used when your web app is added to the


rhqci:

jmazzitelli:
related server-side PR: https://github.com/kiali/kiali/pull/1090

jmazzitelli:
See my comment before merging: https://github.com/kiali/kiali/pull/1090#issuecomment-495638720

lucasponce:
@jmazzitelli worth if you can also add a review on this, to not merge by accident this one.

I've added also the "PR Required" label to easy check that this PR needs backend work.

mtho11:
Where does this PR sit?

jmazzitelli:
This PR (along with the UI side PR) did not work for the basic "/kiali" use case when I last tested this. I looked at the code and it didn't look right to me. I think we should close this and if someone else wants to try to get this to work, they can re-visit and come up with a different implementation.

PR #1209, opened 2019-05-27
KIALI-2911 Add options to menu in graph

aljesusg:
JIRA: [KILA-2911](https://issues.jboss.org/browse/KIALI-2911)

## Menu

menu

## Jaeger Disabled
jaeger_menu_disabled

## With Jaeger integration
jaeger_menu_integration

## Without Jaeger integration
jaeger_menu_no_integration

### DIFF
--- a/src/components/CytoscapeGraph/ContextMenu/NodeContextMenu.tsx
+++ b/src/components/CytoscapeGraph/ContextMenu/NodeContextMenu.tsx
@@ -1,29 +1,63 @@
 import * as React from 'react';
 import { NodeContextMenuProps } from '../CytoscapeContextMenu';
+import { JaegerSearchOptions, JaegerURLSearch } from '../../JaegerIntegration/RouteHelper';
 import history from '../../../app/History';
+import { Paths } from '../../../config';
 
-export class NodeContextMenu extends React.PureComponent<NodeContextMenuProps> {
-  // @todo: We need take care of this at global app level
-  private static makeDetailsPageUrl(props: NodeContextMenuProps) {
-    const namespace = props.namespace;
-    const nodeType = props.nodeType;
-    const workload = props.workload;
-    let app = props.app;
-    let urlNodeType = app;
-    if (nodeType === 'app') {
-      urlNodeType = 'applications';
-    } else if (nodeType === 'service') {
-      urlNodeType = 'services';
-    } else if (workload) {
-      urlNodeType = 'workloads';
-      app = workload;
+type NodeContextMenuState = {
+  nodeType: string;
+  app: string | undefined;
+};
+
+export class NodeContextMenu extends React.PureComponent<NodeContextMenuProps, NodeContextMenuState> {
+  constructor(props: NodeContextMenuProps) {
+    super(props);
+    let app = this.props.app;
+    let nodeType = '';
+    switch (this.props.nodeType) {
+      case 'app':
+        nodeType = Paths.APPLICATIONS;
+        break;
+      case 'service':
+        nodeType = Paths.SERVICES;
+        break;
+      case 'workload':
+        app = this.props.workload;
+        nodeType = Paths.WORKLOADS;
+        break;
+      default:
     }
-    return `/namespaces/${namespace}/${urlNodeType}/${app}`;
+    this.state = { nodeType, app };
+  }
+  // @todo: We need take care of this at global app level
+  makeDetailsPageUrl() {
+    return `/namespaces/${this.props.namespace}/${this.state.nodeType}/${this.state.app}`;
+  }
+
+  getJaegerURL() {
+    let tracesUrl = `/jaeger?namespaces=${this.props.namespace}&service=${this.state.app}.${this.props.namespace}`;
+    if (!this.props.jaegerIntegration) {
+      const url = new JaegerURLSearch(this.props.jaegerURL, false);
+      const options: JaegerSearchOptions = {
+        serviceSelected: `${this.state.app}.${this.props.namespace}`,
+        limit: 20,
+        start: '',
+        end: '',
+        minDuration: '',
+        maxDuration: '',
+        lookback: '3600',
+        tags: ''
+      };
+
+      tracesUrl = url.createRoute(options);
+    }
+    return tracesUrl;
   }
 
   render() {
     const version = this.props.version ? `${this.props.version}` : '';
-    const detailsPageUrl = NodeContextMenu.makeDetailsPageUrl(this.props);
+    const detailsPageUrl = this.makeDetailsPageUrl();
+    const nodetype = this.state.nodeType;
     return (
       <div className="kiali-graph-context-menu-container">
         <div className="kiali-graph-context-menu-title">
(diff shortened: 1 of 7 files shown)


pilhuhn:
Very cool!
This certainly needs some good UX review

abonas:
IMO the actions text in the context menu need to be aligned to the left.

abonas:
I added this to UX agenda for today

abonas:
UX review from Veethika:
1. left align the text
2. remove extra padding at the bottom of the menu
3. add a separator between the black title (the entity name) and the actions

aljesusg:
mmm I saw this https://github.com/kiali/kiali-design/issues/104 related with the context menu

aljesusg:
# Menu style
Screenshot from 2019-05-27 16-52-08

cc @abonas

hhovsepy:
1. Still see  this problem in some cases. When playing with left/right click. Moving from Service to App points in the same graph.
2. Verified.
3. Right click menu is disabled here, but when double click and go to Graph for Service, it is still there:
Screenshot from 2019-05-28 09-38-57

jshaughn:
Minor comments:
- The UX-adjusted menu looks much better,
- In which way was #2 fixed ("Maybe open here general "reviews" app or the "Workload: reviews-v1") ?  I think workload detail is probably best for a versioned app node.

hhovsepy:
>     * In which way was #2 fixed ("Maybe open here general "reviews" app or the "Workload: reviews-v1") ?  I think workload detail is probably best for a versioned app node.

It opens general "reviews" app.

src/components/CytoscapeGraph/CytoscapeReactWrapper.tsx
```diff
@@ -52,32 +56,41 @@ export class CytoscapeReactWrapper extends React.Component<CytoscapeReactWrapper
...
     if (e.target) {
       const anchor = e.target as HTMLAnchorElement;
       const href = anchor.getAttribute('href');
-      if (href) {
+      const newTab = anchor.getAttribute('target') === '_blank';
+      if (href && !newTab) {
         history.push(href);
+      } else {
+        if (href) {
+          window.open(href, '_blank');
```
israel-hdez: If  you are going to open a new tab, I think it is probably better to not use Javascript/Typescript to open the tab. Probably, it is better to not call `e.preventDefault()` and just let the browser to handle the redirection.

If  you open a new window through JS/TS, sometimes you may get that annoying warning about that the page is trying to open a new window, and the browser may prevent the window from opening until the user accepts opening the window.

src/components/CytoscapeGraph/CytoscapeReactWrapper.tsx
```diff
@@ -146,27 +159,82 @@ export class CytoscapeReactWrapper extends React.Component<CytoscapeReactWrapper
...
+  }
+
   private makeTippyForNode = (node: any) => {
     return tippy(node.popperRef(), {
       content: (() => {
         // Remember we on Tippy.js which is html over the cytoscape canvas -- not React JSX
         // @todo: Refactor this to ReactDOM.render() to use JSX
         const tippyDiv = document.createElement('div');
-        tippyDiv.setAttribute('class', 'kiali-graph-context-menu-container');
+        tippyDiv.setAttribute('class', 'popover-content');
```
israel-hdez: Where is this style defined?

If you replace the style, perhaps you want to remove the `kiali-graph-context-menu-container` in CytoscapeGraph.scss.

src/components/CytoscapeGraph/CytoscapeReactWrapper.tsx
aljesusg: We can't do it like is the code now with the a element, It's ok if we are pushing the story but not a new tab, maybe after KIALI-2919

src/components/CytoscapeGraph/CytoscapeReactWrapper.tsx
aljesusg: this is an eror thanks @israel-hdez I was triying to overwrite the theme of tippyDiv because I didn't know why CutoscapeGraph.scss didn't work, after that i added the css file to the index styles. I'll remove it

josejulio: FWIW, I'm working on adding support to use react Components instead of raw html code.

israel-hdez:
>> * In which way was #2 fixed ("Maybe open here general "reviews" app or the "Workload: reviews-v1") ?  I think workload detail is probably best for a versioned app node.

> It opens general "reviews" app.

Jay was probably talking about abonas comment:

> 2. remove extra padding at the bottom of the menu

src/components/CytoscapeGraph/CytoscapeReactWrapper.tsx
israel-hdez: I was meaning to do the exception only when target=_blank, just let the browser handle the click. For the other cases, it is OK to handle with JS/TS.

aljesusg: but @israel-hdez there is a control there when is "_blank"  with the variable newTab

```typescript
private static redirectContextLink(e: MouseEvent) {
    e.preventDefault();
    if (e.target) {
      const anchor = e.target as HTMLAnchorElement;
      const href = anchor.getAttribute('href');
      const newTab = anchor.getAttribute('target') === '_blank';
      if (href && !newTab) {
        history.push(href);
      } else {
        if (href) {
          window.open(href, '_blank');
        }
      }
    }
  }
```

aljesusg:
@hhovsepy are these errors in master?

hhovsepy:
@aljesusg I still see the errors:
1. menu not opened on FF
3. On istio-system service graph services, it now opens the menu and point to non-existing service.

aljesusg:
> @aljesusg I still see the errors:
> 
> 1. menu not opened on FF
> 2. On istio-system service graph services, it now opens the menu and point to non-existing service.

But are these errors in master or in this PR?

hhovsepy:
> But are these errors in master or in this PR?

They are both menu in graph related, so happens only on this PR.

josejulio:
@aljesusg from what i could see, the "hideOnClick: true" option from tippy is also reacting to the right-click "mouse up" event and hiding the menu as soon as it appears.
I think that for we would have to set hideOnClick to false.

mtho11:
I'll review once it's rebased with master which now has #1215

src/components/CytoscapeGraph/CytoscapeContextMenu.tsx
```diff
@@ -18,6 +20,8 @@ type ContextMenuContainer = HTMLDivElement & {
 type ContextMenuProps = {
   element: any;
   contextMenu: TippyInstance;
+  jaegerIntegration: boolean;
+  jaegerURL: string;
```
josejulio: I think we can move these to the `NodeContextMenu` and take these values from redux (they are in redux, right?) If not, use a HOC in the GraphPage, to keep this component a tad more generic.
I can do that if you are OK with it.

src/components/CytoscapeGraph/ContextMenu/NodeContextMenu.tsx
```diff
@@ -1,49 +1,109 @@
...
+        nodeType = Paths.SERVICES;
+        app = this.props.service;
+        break;
+      case 'workload':
+        app = this.props.workload;
+        nodeType = Paths.WORKLOADS;
+        break;
+      default:
+    }
+    this.state = { nodeType, app };
```
josejulio: I think that we need to use the componentDidUpdate to update this state, it could bite us later if this instance is re-used by redux when building the menu instead of creating it from scratch.

PR #1210, opened 2019-05-27
KIALI-2751 Include Gateways in Wizards

lucasponce:
This PRs allows to add support for Gateways in Kiali Wizards.
Before this work, any gateway addition should be added manually using the YAML editor.

With this change, now user can select a wizard or create a default new one from the "Advance Options".

So, now Advanced Options section is organized in the following layout:

image

User can modify/define the VirtualService Hosts section (by default, Wizard uses the short name of the service, the goal of this PR is not going to add more features here, but now user can also personalize this, as when selecting a gateway, it can be different).

Also, there is a switch to add a gateway, and also user can choose between selecting an existing gateway, or creating a default one (selecting the port and Gateway Hosts, which may can be different from VirtualService hosts).

Please @hhovsepy and @burmanm, we have commented about this, if you can test it and tell me if there is no missing use case, it would be great.

Thanks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
     "Instead update 'web_root' in Kaili config map in OpenShift console."
   ],
   "homepage": "./",
+  "proxy": "http://localhost:8000",
   "bugs": {
     "url": "https://issues.jboss.org/projects/KIALI/issues/"
   },
--- a/src/components/IstioWizards/GatewaySelector.tsx
+++ b/src/components/IstioWizards/GatewaySelector.tsx
@@ -1,0 +1,257 @@
+import * as React from 'react';
+import {
+  Col,
+  ControlLabel,
+  DropdownButton,
+  Form,
+  FormControl,
+  FormGroup,
+  HelpBlock,
+  MenuItem,
+  Radio,
+  Switch
+} from 'patternfly-react';
+import { style } from 'typestyle';
+
+type Props = {
+  serviceName: string;
+  hasGateway: boolean;
+  vsHosts: string[];
+  gateway: string;
+  gateways: string[];
+  onGatewayChange: (valid: boolean, gateway: GatewaySelectorState) => void;
+};
+
+export type GatewaySelectorState = {
+  addGateway: boolean;
+  vsHosts: string;
+  gwHosts: string;
+  gwHostsValid: boolean;
+  newGateway: boolean;
+  selectedGateway: string;
+  port: number;
+};
+
+enum GatewayForm {
+  SWITCH,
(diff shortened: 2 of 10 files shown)


burmanm:
A slight concern is with the ``mesh`` keyword which must be added for certain scenarios. And I guess also cross namespace gateways if we want to support those, as the description says:

```
If a list of gateway names is provided, the rules will not be applied to sidecars inside the mesh unless the reserved gateway name mesh is included in the list. Gateways defined in a different namespace can be selected by prefixing the gateway name with namespace/.
```

hhovsepy:
It allows to create routing with empty gateway:
Screenshot from 2019-05-28 09-12-01
Screenshot from 2019-05-28 09-12-27

lucasponce:
@burmanm yeah, your comment is right, I have a reasonable doubt with the cross namespace gateways, as wizards are focus within the same namespace, but I guess we are talking here from a different object and it's more common to define GWs on a different NS.

Also, an "Advanced" Wizard is planned for next step and perhaps we can cover the cross-namespace scenarios there. Or at least, we can let to select GW from others NSs but only creating on the current one.

@hhovsepy you are right ! Good catch. Yeah, if empty gw list, then it shouldn't allow to select it.

Another question @hhovsepy, in your example there is NO gws on the current namespace, right ?
(Just to confirm that there is a bug in other area).

hhovsepy:
> Another question @hhovsepy, in your example there is NO gws on the current namespace, right ?
> (Just to confirm that there is a bug in other area).
Right.

hhovsepy:
Make sure there are 2 gateways in the namespace.
Create Matching Routing and select a gateway.
Edit Matching Routing and change the gateway to another one.
VS disappears from Service Details. But still exist in Istio Config.
Happens only with Matching Routing.

lucasponce:
@hhovsepy I've sent a commit to fix the scenarios you have commented.

I haven't yet introduced @burmanm suggestions, that's next task on my todo list, but you could test this one ? To validate if the issues are gone ?

Thanks a lot.

lucasponce:
@hhovsepy @burmanm 
I have added Micke's suggestion, now wizard supports:
- Cross-namespace gateway selection
- Optional "mesh" gateway check.

Please @burmanm test/review it and let me know how it looks like to consolidate it.

Thanks in advance.

lucasponce:
Thanks @burmanm & @hhovsepy

hhovsepy:
Verified:
Screenshot from 2019-05-29 15-35-24
Screenshot from 2019-05-29 15-58-49

PR #1212, opened 2019-05-27
KIALI-2930 Show details of error in initialization screen

israel-hdez:
https://issues.jboss.org/browse/KIALI-2930

At first load of Kiali, a request is done to retrieve some settings. If this request fails, an error message is shown. Sometimes, this error message is not enough to troubleshoot the issue and the developer console is needed to see what the back-end is returning.

To improve this experience, adding a "Show details" pane to display the raw response sent by the back-end if first-load initialization fails.

showDetails

### DIFF
--- a/src/app/InitializingScreen.tsx
+++ b/src/app/InitializingScreen.tsx
@@ -10,11 +10,49 @@ import {
   LoginPageHeader,
   Spinner
 } from 'patternfly-react';
+import { style } from 'typestyle';
 import { isKioskMode } from '../utils/SearchParamUtils';
+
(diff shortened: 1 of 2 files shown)


israel-hdez:
@jshaughn It's not using it, because it is not available at login screen :)

Not sure, but I remember I saw that the message center can show the sandwich menu:
https://www.patternfly.org/pattern-library/communication/notification-drawer/

Probably, we can use that menu to place a "View details" option. I will create a Jira to investigate that. It would be nice to be able to see the detailed error.

PR #1215, opened 2019-05-28
KIALI-2919 Allows context menu to use react components.

josejulio:
The react components take the node/edge data as their props,
  the tippyInstance (contextMenu) and the node/edge element from cy

** Screenshot **

Warning: UI didn't change, the gif is only a basic example of what can be done given the following components:

```typescript
   class CEdge extends React.Component<ContextMenuEdgeProps> {
      render() {
        return (
          <div>
            Hello Edge from: {this.props.source} to: {this.props.target}{' '}
          </div>
        );
      }
    }

   const CNode = () => {
      return <div> Hello Node</div>;
    };

    class CGroup extends React.PureComponent<ContextMenuNodeProps> {
      render() {
        return <div>Hello Group with id {this.props.id}</div>;
      }
    }
```

context-menu-component

### DIFF
--- a/src/components/CytoscapeGraph/ContextMenu/NodeContextMenu.tsx
+++ b/src/components/CytoscapeGraph/ContextMenu/NodeContextMenu.tsx
@@ -1,10 +1,10 @@
 import * as React from 'react';
-import { ContextMenuNodeProps } from '../CytoscapeContextMenu';
+import { NodeContextMenuProps } from '../CytoscapeContextMenu';
 import history from '../../../app/History';
 
-export class ContextMenuNodeComponent extends React.PureComponent<ContextMenuNodeProps> {
+export class NodeContextMenu extends React.PureComponent<NodeContextMenuProps> {
   // @todo: We need take care of this at global app level
-  private static makeDetailsPageUrl(props: ContextMenuNodeProps) {
+  private static makeDetailsPageUrl(props: NodeContextMenuProps) {
     const namespace = props.namespace;
     const nodeType = props.nodeType;
     const workload = props.workload;
@@ -23,7 +23,7 @@ export class ContextMenuNodeComponent extends React.PureComponent<ContextMenuNod
 
   render() {
     const version = this.props.version ? `${this.props.version}` : '';
-    const detailsPageUrl = ContextMenuNodeComponent.makeDetailsPageUrl(this.props);
+    const detailsPageUrl = NodeContextMenu.makeDetailsPageUrl(this.props);
     return (
       <div className="kiali-graph-context-menu-container">
         <div className="kiali-graph-context-menu-title">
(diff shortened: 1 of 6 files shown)


josejulio:
Thanks for your detailed review, I'll try to answer the best i can.

> @josejulio OK, overall I think I like this but I'm not sure I totally understand it, Here is what I think is going on, please correct me...
> 
> * ContextMenuNodeComponent is responsible for generating the the actual context menu for a node or app box (group).  And at the moment the impl is just temporary as it only renders one option.  ContextMenuEdgeComponent would be similar, but there is no impl because we don't currently have any context menu on impls.

That's correct, these implementation would act as a regular React component, connecting to redux, having  its own state, etc. 
The props are taken from the node.data (or edge.data, depending the case).


>   * For me, it would be more clear with naming:  NodeContextMenu/EdgeContextMenu.  And  NodeContextMenuProps/EdgeContextMenuProps.

Sounds good, make it more easy to spot if we are talking about a node or an edge.

>   * I think this PR should wait on the work currently in review for the updated context menu options and then be updated with the actual menu impl.

I'm OK for that, unless this make it easier for @aljesusg to implement his work.

> * I'm a little confused that there is both a class and a type defined with the name ContextMenuNodeComponent, can you explain?

Ah, my bad.

One is the actual component that renders the view, while the other is the Class that it accepts.
It can accept any class that is a React.Component, function component or PureComponent as long as their props are NodeContextMenuProps/EdgeContextMenuProps

> * CytoscapeContextMenu is not actually a context menu but rather the "harness" or "wrapper" for the actual context menu?

Correct.

>   
>   * If that's correct perhaps a different name like CytoscapeContextMenuWrapper or something to distinguish it from being an actual context menu.

Sure, makes sense.

> * I think you've changed it such that there is only one visible context menu at any time. And it does not disappear with a timeout.  I like those changes although how do you get rid of the context menu without a click? An escape?

I ensure that there is only one div container, and attach to it an instance of the currently visible context menu, before showing anything i hide it: https://github.com/josejulio/swsui/blob/ba50cc96a9ac57e58b62d290c4615a8880f6a9b8/src/components/CytoscapeGraph/CytoscapeContextMenu.tsx#L58

> * I like that logic has been removed from CytoscapeReactWrapper

+1

jshaughn:
> I'm OK for that, unless this make it easier for @aljesusg to implement his work.

I think [his PR](https://github.com/kiali/kiali-ui/pull/1209) is close.  It seems stopped because of a possible issue rendering the context menu on Firefox.  Have you tried your impl on FF?

josejulio:
> I think [his PR](https://github.com/kiali/kiali-ui/pull/1209) is close. It seems stopped because of a possible issue rendering the context menu on Firefox. Have you tried your impl on FF?

I just did and seeing looking that the menu opens and closes as soon as you release the button.

josejulio:
> I think [his PR](https://github.com/kiali/kiali-ui/pull/1209) is close. It seems stopped because of a possible issue rendering the context menu on Firefox. Have you tried your impl on FF?

@aljesusg told me that is OK to merge this first. I just pushed some changes that fix the problem with FF and allows interaction with the context menu.

with Firefox:

firefox-tippy

PR #1216, opened 2019-05-28
KIALI-2918 Enable group nodes to have a context menu

josejulio:
** Describe the change **

- Moved some redux properties to redux-connect syntax
- In versioned graph, the versioned node (v1, v2) will be considered as workloads
- Droped the version from the title because it won't be valid to show a workload name with a version
- NodeType=unknown and isInaccessible won't show any menu

** Screenshot **
Peek 2019-05-31 11-42

### DIFF
--- a/src/components/CytoscapeGraph/ContextMenu/NodeContextMenu.tsx
+++ b/src/components/CytoscapeGraph/ContextMenu/NodeContextMenu.tsx
@@ -105,8 +105,8 @@ export class NodeContextMenu extends React.PureComponent<Props> {
   }
 
   render() {
-    // Disable context menu if we are dealing with a unknown node
-    if (this.props.nodeType === 'unknown') {
+    // Disable context menu if we are dealing with a unknown or an inaccessible node
+    if (this.props.nodeType === 'unknown' || this.props.isInaccessible) {
       this.props.contextMenu.disable();
       return null;
     }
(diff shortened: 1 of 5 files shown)


jshaughn:
@josejulio I think now that the other work is in you can update this PR appropriately and probably use it to make the adjustments you recommended in [The 1209 PR](https://github.com/kiali/kiali-ui/pull/1209).

src/components/CytoscapeGraph/ContextMenu/NodeContextMenu.tsx
```diff
@@ -72,27 +83,35 @@ export class NodeContextMenu extends React.PureComponent<NodeContextMenuProps, N
   }
 
   render() {
-    const version = this.props.version !== '' ? `:${this.props.version}` : '';
-    const detailsPageUrl = this.makeDetailsPageUrl();
-    const { nodeType, app } = this.state;
+    // Disable context menu if we are dealing with a unknown node
```
jshaughn: We should also disable content menu for isInaccessible nodes.  These nodes are not actually accessible to the user, they are rendered only to show a connection and are typically in a restricted namespace or may be service entries.

josejulio:
> @josejulio Looks good, just one comment to take deal with before approval. Also, in the video there are errors popping up, is that anything to be concerned with? 

Not really, these errors were related to a call my server didn't had at that poin.

> One last note, when you make the videos maybe move a little more slowly because the mouse movement often does not show up and it's hard to follow (at least for me).

My bad, I disabled the mouse cursor from appearing in the video. I'll record again with mouse on.

src/components/CytoscapeGraph/ContextMenu/NodeContextMenu.tsx
josejulio: Sounds good

josejulio: Done

josejulio:
@jshaughn updated the gif, hope is better now.

PR #1217, opened 2019-05-28
KIALI-2940 Restore Kiali colors

israel-hdez:
https://issues.jboss.org/browse/KIALI-2940

Restoring colors that were present in Kiali before the PF4 migration.

Login page:

image

Initializing screen:

image

Main console:

image

### DIFF
--- a/src/components/Nav/Menu.tsx
+++ b/src/components/Nav/Menu.tsx
@@ -2,6 +2,7 @@ import _ from 'lodash';
 import * as React from 'react';
 import { matchPath } from 'react-router';
 import { Link } from 'react-router-dom';
+import { style } from 'typestyle';
 import { Nav, NavList, NavItem, PageSidebar } from '@patternfly/react-core';
 import { ExternalLinkAltIcon } from '@patternfly/react-icons';
 
@@ -15,6 +16,12 @@ const ExternalLink = ({ href, name }) => (
     </a>
   </NavItem>
 );
+
+const sidebarStyle = style({
+  borderTopStyle: 'solid',
+  borderTopWidth: '1px',
+  borderTopColor: '#19506d'
+});
 
 type MenuProps = {
   isNavOpen: boolean;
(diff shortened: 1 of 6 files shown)


src/components/Nav/Menu.tsx
```diff
@@ -74,7 +81,7 @@ class Menu extends React.Component<MenuProps, MenuState> {
       </Nav>
     );
 
-    return <PageSidebar isNavOpen={isNavOpen} nav={PageNav} />;
+    return <PageSidebar className={sidebarStyle} isNavOpen={isNavOpen} nav={PageNav} />;
   }
```
aljesusg: Can we keep all the styles with the same pattern in files?

israel-hdez:
Not sure if this is really OK.
Feel free to reject this PR/Jira, if needed.

aljesusg:
@israel-hdez we must keep the styles like openshift console. I think that we shouldn't change this. Maybe @abonas or @serenamarie125 can comment here.

israel-hdez:
> I think that we shouldn't change the styles of the menu, this is the pattern for PF4 and aligned with openshift

I though RCUE styles were for that :(
Probably, at least the sample background images should be removed.

aljesusg:
> > I think that we shouldn't change the styles of the menu, this is the pattern for PF4 and aligned with openshift
> 
> I though RCUE styles were for that :(
> Probably, at least the sample background images should be removed.

there is not RCUE for PF4 yet, they are working on it .The changes required for RCUE now are the logos. maybe after GA.

serenamarie125:
As long as the branding changes do not affect the downstream colors, this is fine from a UX perspective.  Having a different color scheme for upstream is allowed :)

serenamarie125:
It may be easier to keep the difference between up and downstream localized to the top masthead, rather than proliferating it throughout the navigation.

israel-hdez:
Closing in favor of https://issues.jboss.org/browse/KIALI-2993

PR #1220, opened 2019-05-29
KIALI-2795 Automaticaly refresh the mesh-wide TLS lock in the masthead

xeviknal:
** Describe the change **
Automatically refresh the mesh-wide TLS status.

- The lock appears and disappears in the masthead. 
- In the graph, it switches the logic between using open locks and full locks. 
- In the Overview page, locks next to namespaces, changes accordingly.

Needs: https://github.com/kiali/kiali/pull/1102

** Issue reference **
https://issues.jboss.org/browse/KIALI-2795

** Backwards compatible? **
yes

** Screenshot **
Kiali Console (1)

Kiali Console (2)

### DIFF
--- a/src/actions/KialiAppAction.ts
+++ b/src/actions/KialiAppAction.ts
@@ -9,6 +9,7 @@ import { MessageCenterAction } from './MessageCenterActions';
 import { NamespaceAction } from './NamespaceAction';
 import { UserSettingsAction } from './UserSettingsActions';
 import { JaegerAction } from './JaegerActions';
+import { MeshTlsAction } from './MeshTlsActions';
 
 export type KialiAppAction =
   | GlobalAction
@@ -21,4 +22,5 @@ export type KialiAppAction =
   | MessageCenterAction
   | NamespaceAction
   | UserSettingsAction
-  | JaegerAction;
+  | JaegerAction
+  | MeshTlsAction;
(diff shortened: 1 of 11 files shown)


pilhuhn:
I am not sure I get that: on the 1st screenshot is the mesh-wide lock only visible if one selects "show security" in the drop-downs?

xeviknal:
@pilhuhn nope, the mesh-wide lock appears when after applying the MeshPolicy and DestinationRule. You are right, I guess it is not very clear. Let me elaborate :)

What happened on the first screen is:
1. I create the pair: MeshPolicy and DR enabling mesh-wide mtls.
2. I enabled the security layer (thought the dropdown).
3. There weren't mtls traffic yet, so no locks appeared.
4. after 10s, new mtls status fetch happend.
5. Masthead lock appears
6. Open locks in the edges (bc not 100% of the traffic uses mTLS)

pilhuhn:
On 29 May 2019, at 16:22, Xavier Canal i Masjuan wrote:

> 5. Masthead lock appears

Can we get this from the Kube api instead of waiting
for traffic? Something like `oc get MeshPolicy`  ?

xeviknal:
@pilhuhn masthead lock comes from k8s api. It was before 
What I am doing in this PR is updating that icon every N seconds instead of at login point only. Now, the users won't need to refresh the whole screen to get the lock updated.

PR #1154, opened 2019-05-02
KIALI-2529 Update from react-scripts-ts to React-scripts

cfcosta:
## Motivation

There are two motivations for this change:

* `react-scripts-ts` has been deprecated now what the original `react-scripts`
  now [supports typescript natively][1].
* `webpack-dev-server` is [vulnerable for development][2]. It doesn't seem to
  affect our use case, but it throws false positives on the vulnerability
  ratings we do have (`snyk`, `yarn audit`).

Because of that, it is something that we have to do, that will become harder as
we need for a perfect time to do the upgrade, but that isn't urgent enough that
we can't move it forward for after 1.0.

## Strategy

We did a huge PR in the beginning that included all the changes that we had to
do, but that revealed itself to be quite unwieldly. Long PRs require a lot of
context, can hide lots of small errors and is generally hard to review, so
people just don't do it.

To avoid that, we decided to go with two different "sets" of pull requests, one
for "code changes" and the other for library changes. Library changes are
updates on library versions, and are more risky because we don't control the
code that's being changed, while the code changes are usually things that did
compile before but does not anymore, mostly because the compilation is now a
little different than it was before, because `react-scripts` uses Babel for
transpiling, and as such has some differences from the old implementation
(namely namespaces).

The idea is that all the pull requests are independent from each other, and can
be merge at a more suitable time for each of them. Code changes are supposed to
keep working on the current version while also preparing the codebase for the
library changes.

## Dependencies

- [x] #1225 Remove deprecated `namespace`

  **Important**: This PR is not as huge as it seems, add `?w=1` to the URL to
  see the changes without the whitespace changes.

- [x] #1123 Remove deprecated `const enum`
- [x] #1124 Fix cases of type erasure in the codebase that don't compile anymore
- [x] #1236 Fix absolute imports
- [x] #1237 Fix moment import
- [x] #1238 Fix export problems
- [x] #1239 Fix values that can be undefined or null
- [x] #1240 Add type hints to composite types
- [x] #1241 Fix compilation error (this one is a treat, go read it)
- [x] #1222 Update to latest version of `typescript`
- [x] #1263 Fix compilation error still present after #1240 
- [x] Fix failing tests

[1]: https://vincenttunru.com/migrate-create-react-app-typescript-to-create-react-app/
[2]: https://www.npmjs.com/advisories/725

PR #1137, opened 2019-04-24
KIALI-2764 Update typescript to 2.4.5 to get --incremental compile feature

mtho11:
KIALI-2764 Update typescript to 2.4.5 from 3.3.4000 to get --incremental dev compilation performance increases and other things.

http://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html

> For a project the size of Visual Studio Code, TypeScript’s new --incremental flag was able to reduce subsequent build times down to approximately a fifth of the original.

** Issue reference **

https://issues.jboss.org/browse/KIALI-2764

** Backwards compatible? **
yes
[ ] Is your pull-request introducing changes in behaviour?
No

PR #1222, opened 2019-05-30
Update to Typescript 3.5

cfcosta:
Updates to latest version of typescript:
- https://devblogs.microsoft.com/typescript/announcing-typescript-3-5/

This update makes the compiler a little more strict, so some of the
things that we had before will not work. This update has the same
motivation as #1154, and having a more up to date typescript helps us
with it (some errors happen because of the lacking `unknown` type, which
is defined as `null | undefined | T extends {}`).

Some other changes:

* `@types/react-redux` has been updated because it generates a
  compilation error. This is build-only, so no impact. Also, we're already
  running the latest redux, so we're just keeping them in sync.
* `hoist-non-react-statics` had improper typescript typings and had to
  be forcefully updated. I did some investigations on the changes for
  the versions, nothing seems non-breaking.
* Health-related API namespaces were erasing types, so typescript just
  defaulted to `any` and ignored the typing. This has been fixed.
* Typescript 3.5 broke webpack ts-loader implementation when using
  `incremental`: microsoft/TypeScript#31447
  This is fixed by also setting the `tsbuildinfo` file on tsconfig.

Some tests are failing, and I found the same problems on #1154, so I'm
backporting those changes to the tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,7 +112,7 @@
     "@types/node": "11.11.4",
     "@types/react": "16.7.13",
     "@types/react-dom": "16.0.11",
-    "@types/react-redux": "6.0.11",
+    "@types/react-redux": "7.0.9",
     "@types/react-router-dom": "4.3.1",
     "axios-mock-adapter": "1.16.0",
     "babel-core": "6.26.3",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -128,11 +128,12 @@
...
     "source-map-explorer": "1.8.0",
     "tslint-no-circular-imports": "0.6.2",
-    "typescript": "3.4.5"
+    "typescript": "3.5"
   },
   "resolutions": {
     "@types/react": "16.7.13",
-    "js-yaml": "3.13.1"
+    "js-yaml": "3.13.1",
+    "hoist-non-react-statics": "3.3.0"
```
mtho11: @cfcosta what is this line for?

cfcosta:
```
hoist-non-react-statics had improper typescript typings and had to
be forcefully updated. I did some investigations on the changes for
the versions, nothing seems non-breaking.
```

From the description. But expanding, it is used by react-redux to do **stuff**, and the type definitions for it are broken, it fails compilation when `@types/react-redux` imports types from it.

cfcosta:
@gbaufake what do you think about this? I'm reasonably sure it is going to break anything, but what do you think?

The last time we did this change: #1137

cfcosta:
@mtho11 @jshaughn updated to latest master, checking build and tests and we can merge :)

PR #1228, opened 2019-05-31
KIALI-2950 Improve login - avoid unnecessary screen switching

israel-hdez:
https://issues.jboss.org/browse/KIALI-2950

After logging in, the "initializing" screen is no longer shown. Instead, the "Log in" button text is disabled and the text is changed to "Logging in...".

Openshift authentication is not affected.

Also, if user is already logged in, Kiali's first load is also unaffected.

**A succeeded login**

loginSucceed

**A failed login**

loginFail

### DIFF
--- a/src/app/AuthenticationController.tsx
+++ b/src/app/AuthenticationController.tsx
@@ -26,11 +26,11 @@ interface AuthenticationControllerReduxProps {
 
 type AuthenticationControllerProps = AuthenticationControllerReduxProps & {
   protectedAreaComponent: React.ReactNode;
-  publicAreaComponent: React.ReactNode;
+  publicAreaComponent: (isPostLoginPerforming: boolean, errorMsg?: string) => React.ReactNode;
 };
 
 interface AuthenticationControllerState {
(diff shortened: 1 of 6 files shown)


src/app/AuthenticationController.tsx
```diff
@@ -26,11 +26,11 @@ interface AuthenticationControllerReduxProps {
 
 type AuthenticationControllerProps = AuthenticationControllerReduxProps & {
   protectedAreaComponent: React.ReactNode;
-  publicAreaComponent: React.ReactNode;
+  publicAreaComponent: (isPostLoginPerforming: boolean, errorMsg?: string) => React.ReactNode;
 };
 
 interface AuthenticationControllerState {
-  stage: 'login' | 'post-login' | 'logged-in';
+  stage: 'login' | 'post-login' | 'logged-in' | 'logged-in-at-load';
```
mtho11: If you used an `enum` for stage you wouldn't need to assign strings to it below and then it would be typesafe.

israel-hdez:
@mtho11 Fixed your comment.

PR #1231, opened 2019-06-03
KIALI-2970 Removes fixed height of 95vh for page

josejulio:
** Describe the change **

Patternfly sets the height of the "Page" to 95vh, so we can our calc wrong in the GraphPage.
Additionally, sets the header height to the remaining 5vh, if not, when the screen is big (height about 1800px) a black bar appears on bottom.

### DIFF
--- a/src/components/Nav/Navigation.tsx
+++ b/src/components/Nav/Navigation.tsx
@@ -6,6 +6,7 @@ import { RouteComponentProps } from 'react-router';
 import Masthead from './Masthead/Masthead';
 import Menu from './Menu';
 import { Page, PageHeader, PageSection, Brand } from '@patternfly/react-core';
+import { style } from 'typestyle';
 
 import MessageCenterContainer from '../../components/MessageCenter/MessageCenter';
 import { kialiLogo, serverConfig } from '../../config';
@@ -28,6 +29,11 @@ type NavigationState = {
   isNavOpenDesktop: boolean;
   isNavOpenMobile: boolean;
 };
+
+const flexBoxColumnStyle = style({
+  display: 'flex',
+  flexDirection: 'column'
+});
 
 class Navigation extends React.Component<PropsType, NavigationState> {
   static contextTypes = {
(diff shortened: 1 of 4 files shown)


israel-hdez:
Hmm.... But on "normal" screens, it runs out of the viewport:

image

josejulio:
Ah, I must have something wrong in the calcs. I though i tested it on normal screens too :)

josejulio:
Ah yes, the topbar is bigger than 5vh in normal screen. The problem is that the page (defined by PF) is 95vh, I'll check this again later.

src/components/Nav/Navigation.tsx
```diff
@@ -107,9 +113,9 @@ class Navigation extends React.Component<PropsType, NavigationState> {
     );
 
     return (
-      <Page header={Header} sidebar={Sidebar} onPageResize={this.onPageResize}>
+      <Page style={{ backgroundColor: 'purple' }} header={Header} sidebar={Sidebar} onPageResize={this.onPageResize}>
```
israel-hdez: Just for curiosity, why purple?

josejulio: Debugging purpose :) 
Was to easily see if my Page filled the whole page.

josejulio: Will remove this

josejulio: I just removed this purple color

israel-hdez: Ok, looks better now :)

josejulio:
@mtho11 if you were still having the issue with the graph toolbar (legend, fit, etc) being hidden on chrome, could you check if this PR helps?

src/components/Nav/Navigation.tsx
mtho11: Purple, haha, I do pink.

mtho11:
The legend fit issues are still there on OS/X laptop:

overlap

But it is not any worse, so I'm not blocking or endorsing. Either way, I'm good.

PR #1232, opened 2019-06-03
KIALI-2972 Avoid infinite querying by protecting against 0 refresh interval

jshaughn:
@jmazzitelli Please try this if you get the chance, would be good to get a fix in for this.

### DIFF
--- a/src/components/MTls/MeshMTLSStatus.tsx
+++ b/src/components/MTls/MeshMTLSStatus.tsx
@@ -22,7 +22,7 @@ type Props = {
 };
 
 type State = {
-  intervalId: NodeJS.Timeout;
+  intervalId?: NodeJS.Timeout;
   refreshInterval: PollIntervalInMs;
 };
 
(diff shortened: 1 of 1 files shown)


josejulio:
I only have one question and not directly related to this, what does that tls refresh do? Should it be called with the manual refresh button? (because it does not)

jmazzitelli:
> I only have one question and not directly related to this, what does that tls refresh do? Should it be called with the manual refresh button? (because it does not)

I don't know but this is the PR that introduced this feature: https://github.com/kiali/kiali-ui/pull/1220

jshaughn:
@josejulio Good point.  I think it's just checking to see if the global TLS config has changed, it probably should get checked on a manual refresh.  Let's get this blocker fix in, though.  And we can follow up.

jshaughn:
@josejulio I created https://issues.jboss.org/browse/KIALI-2974 to capture the manual refresh issue.

PR #1236, opened 2019-06-05
KIALI-2529 Remove absolute imports

cfcosta:
Moving to `react-scripts` from `react-scripts-ts` moves our
transpilation layer to use babel instead of tsc, which forbids absolute
imports.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "baseUrl": "./",
+    "baseUrl": "src",
     "outDir": "build/dist",
     "module": "esnext",
     "target": "es5",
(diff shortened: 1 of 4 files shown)


josejulio:
Ah.. I liked the absolute imports.. :(

cfcosta:
> Ah.. I liked the absolute imports.. :(
> It felt it was easier to locate the resource we were importing.

You can use "go to definition" on the import, all editors should implement this :)
Also, the next step for us might be adding "paths" to transform something like:

```typescript
import { Foo } from '../../../types/foo';
```

Into:

```typescript
import { Foo } from 'types/foo';
```

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "baseUrl": "./",
+    "baseUrl": "src",
```
israel-hdez: is this change also needed?
Looks redundant if relative paths are being used now... Perhaps, is it to forbid absolute imports?

cfcosta: That’s a nice side effect, but it does not compile after the upgrade.

israel-hdez: Ok, I'm approving.

PR #1225, opened 2019-05-30
KIALI-2529 Remove namespaces.

cfcosta:
Extracted part from #1154 that deals with namespaces, which are deprecated because they cause problems like (https://github.com/improbable-eng/ts-protoc-gen/issues/39)[this one]. It does not compile anymore after the change to `react-scripts` from `react-scripts-ts`.

## Possible impact

Close to none, as all the changes are syntatic and typescript covers those kind of errors quite well. Build is ok and tests are (hopefully) green.

PR #1240, opened 2019-06-05
KIALI-2529 Add type hints to composite types

cfcosta:
Typescript has no way to know if the item has the health attribute or
not. Before, it just defaulted to any, but when we start using babel, it
becomes a compilation error.

This is one of those cases where the current implementation is incorrect
and only worked because it downcasted it to `any`.

### DIFF
--- a/src/pages/AppList/FiltersAndSorts.ts
+++ b/src/pages/AppList/FiltersAndSorts.ts
@@ -1,7 +1,7 @@
 import { ActiveFilter, FILTER_ACTION_APPEND, FilterType } from '../../types/Filters';
 import { AppListItem } from '../../types/AppList';
 import { SortField } from '../../types/SortFilters';
-import { getRequestErrorsStatus, WithHealth, hasHealth } from '../../types/Health';
+import { getRequestErrorsStatus, WithAppHealth } from '../../types/Health';
 import {
   istioSidecarFilter,
   healthFilter,
@@ -52,22 +52,18 @@ export namespace AppListFilters {
       title: 'Health',
       isNumeric: false,
       param: 'he',
-      compare: (a: AppListItem, b: AppListItem) => {
-        if (hasHealth(a) && hasHealth(b)) {
-          const statusForA = a.health.getGlobalStatus();
-          const statusForB = b.health.getGlobalStatus();
+      compare: (a: WithAppHealth<AppListItem>, b: WithAppHealth<AppListItem>) => {
+        const statusForA = a.health.getGlobalStatus();
+        const statusForB = b.health.getGlobalStatus();
 
-          if (statusForA.priority === statusForB.priority) {
-            // If both apps have same health status, use error rate to determine order.
-            const ratioA = getRequestErrorsStatus(a.health.requests.errorRatio).value;
-            const ratioB = getRequestErrorsStatus(b.health.requests.errorRatio).value;
-            return ratioA === ratioB ? a.name.localeCompare(b.name) : ratioB - ratioA;
-          }
-
-          return statusForB.priority - statusForA.priority;
+        if (statusForA.priority === statusForB.priority) {
+          // If both apps have same health status, use error rate to determine order.
+          const ratioA = getRequestErrorsStatus(a.health.requests.errorRatio).value;
+          const ratioB = getRequestErrorsStatus(b.health.requests.errorRatio).value;
+          return ratioA === ratioB ? a.name.localeCompare(b.name) : ratioB - ratioA;
         }
 
-        return 0;
+        return statusForB.priority - statusForA.priority;
       }
     }
   ];
@@ -139,8 +135,8 @@ export namespace AppListFilters {
     if (sortField.title === 'Health') {
       // In the case of health sorting, we may not have all health promises ready yet
       // So we need to get them all before actually sorting
-      const allHealthPromises: Promise<WithHealth<AppListItem>>[] = unsorted.map(item => {
-        return item.healthPromise.then((health): WithHealth<AppListItem> => ({ ...item, ...{ health } }));
+      const allHealthPromises: Promise<WithAppHealth<AppListItem>>[] = unsorted.map(item => {
+        return item.healthPromise.then((health): WithAppHealth<AppListItem> => ({ ...item, ...{ health } }));
       });
       return Promise.all(allHealthPromises).then(arr => {
         return arr.sort(isAscending ? sortField.compare : (a, b) => sortField.compare(b, a));
--- a/src/pages/AppList/FiltersAndSorts.ts
+++ b/src/pages/AppList/FiltersAndSorts.ts
@@ -1,7 +1,7 @@
 import { ActiveFilter, FILTER_ACTION_APPEND, FilterType } from '../../types/Filters';
 import { AppListItem } from '../../types/AppList';
 import { SortField } from '../../types/SortFilters';
-import { AppHealth, getRequestErrorsStatus } from '../../types/Health';
+import { getRequestErrorsStatus, WithHealth, hasHealth } from '../../types/Health';
 import {
   istioSidecarFilter,
   healthFilter,
@@ -9,8 +9,6 @@ import {
   getFilterSelectedValues,
   filterByHealth
 } from '../../components/Filters/CommonFilters';
-
-type AppListItemHealth = AppListItem & { health: AppHealth };
 
 export namespace AppListFilters {
   export const sortFields: SortField<AppListItem>[] = [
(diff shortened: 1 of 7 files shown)


josejulio:
A test is failing here

src/pages/ServiceList/FiltersAndSorts.ts
```diff
@@ -54,18 +55,22 @@ export namespace ServiceListFilters {
...
-        if (statusForA.priority === statusForB.priority) {
-          // If both services have same health status, use error rate to determine order.
-          const ratioA = getRequestErrorsStatus(a.health.requests.errorRatio).value;
-          const ratioB = getRequestErrorsStatus(b.health.requests.errorRatio).value;
-          return ratioA === ratioB ? a.name.localeCompare(b.name) : ratioB - ratioA;
+          return statusForB.priority - statusForA.priority;
         }
 
-        return statusForB.priority - statusForA.priority;
+        return 0;
```
josejulio: Should we throw an error here? we are comparing two different things and returning 0 means that they are equal.

cfcosta: I agree, but there is the same behavior on the other files, so I just made them consistent. Let's see what others think @mtho11 @israel-hdez @jotak

josejulio: Ah, OK.

src/types/Health.ts
```diff
@@ -151,9 +151,7 @@ export const getRequestErrorsViolations = (reqIn: ThresholdStatus, reqOut: Thres
 };
 
 export abstract class Health {
-  items: HealthItem[];
-
-  constructor(items: HealthItem[]) {
+  constructor(public items: HealthItem[], public requests: RequestHealth, public ctx: HealthContext) {
```
jotak: I will look at this PR more closely tomorrow, but why do we need to keep `requests` and `ctx` as class members? They should only be necessary while building the objects, and be discarded after, no?

src/pages/ServiceList/FiltersAndSorts.ts
jotak: I'll have a closer look tomorrow but if I remember correctly this comparison method should only be called with items that have associated health. I guess the problem is that it may not be enforced by types?

src/types/Health.ts
israel-hdez: Yes, I also understand that requests and ctx don't need to be class fields and that data can be discarded.

Looking at the child classes, these fields are public. So, this is only pulling up the existent definitions. But, definitely, those class fields can probably can be removed.

src/pages/WorkloadList/FiltersAndSorts.ts
```diff
@@ -270,12 +269,8 @@ export namespace WorkloadListFilters {
...
       // In the case of health sorting, we may not have all health promises ready yet
       // So we need to get them all before actually sorting
-      const allHealthPromises: Promise<WorkloadItemHealth>[] = unsorted.map(item => {
-        return item.healthPromise.then(health => {
-          const withHealth: any = item;
-          withHealth.health = health;
-          return withHealth;
-        });
+      const allHealthPromises: Promise<WithHealth<WorkloadListItem>>[] = unsorted.map(item => {
+        return item.healthPromise.then((health): WithHealth<WorkloadListItem> => ({ ...item, ...{ health } }));
```
israel-hdez: isn't this the same as: `{ ...item, health }` ?? (just to have less nesting)

src/pages/ServiceList/__tests__/ServiceListComponent.test.ts
```diff
@@ -1,15 +1,15 @@
...
+  const health = new ServiceHealth(reqErrs, { rateInterval: 60, hasSidecar: true });
+
   return {
     name: name,
-    health: new ServiceHealth(reqErrs, { rateInterval: 60, hasSidecar: true })
-  } as ServiceListItem & {
-    health: ServiceHealth;
-  };
+    health: health
+  } as any;
```
israel-hdez: I think I would prefer to set empty values for the other properties not being used to avoid casting as any.

src/pages/ServiceList/FiltersAndSorts.ts
israel-hdez: What @jotak is saying seems to be true, because the code that is below (which you also changed in this PR):

https://github.com/kiali/kiali-ui/blob/4cfcea6865ca355966fa1605a75c3a2fae445e86/src/pages/ServiceList/FiltersAndSorts.ts#L137-L146

is calling the `compare` function once all healths have been collected and the "augmented" object is ready. So, this `compare` should receive `WithHealth<ServiceItemList>` arguments.

Code is similar for Apps and Workloads case. Probably, to make things consistent, the other `{App,Workload}List/FilterAndSort.ts` are the ones that need the change to the type of its arguments.

israel-hdez:
> A test is failing here

The service list component is the only one with tests :( but, apparently, those tests are revealing that the `hasHealth` type guard function is not working as expected and always returning false.

src/pages/ServiceList/FiltersAndSorts.ts
cfcosta: @israel-hdez @jotak if that's true, why does the current `compare` impl check if the item has health or not? I imagine it was changed somewhere else and became always true, so I think I'll refactor those functions and assume that it always has health data.

src/types/Health.ts
cfcosta: True, I changed the constructors to keep the interface the same across subclasses and have a smaller diff without too many architectural changes, because doing it the other way (remove the args from the subclasses) breaks the public interface.

src/pages/WorkloadList/FiltersAndSorts.ts
cfcosta: Oh, that's a good one, will try. I tried `{ ...item, ...health }` but it didn't work, maybe this does.

src/pages/ServiceList/__tests__/ServiceListComponent.test.ts
cfcosta: Yeah, that's probably what I'm going to do, but I don't like it very much, adds information to the test that is not under subject of testing and make things messy in general, but I think that's the best option.

jotak:
Suggested changes : https://github.com/cfcosta/swsui/pull/2

jotak:
(oops i didn't want to close)

src/pages/ServiceList/FiltersAndSorts.ts
jotak: @cfcosta I can't be 100% sure but it was maybe some zealous defensive check, that we can get rid of.

src/pages/ServiceList/__tests__/ServiceListComponent.test.ts
jotak: In this situation I prefer simpler / more readable test than type accuracy... so as Caina did. It's subject to debate...

cfcosta:
@jotak I merged your PR, and the tests are now ok, can you take a look to see if the changes are sound or if we can make it smaller?

src/pages/AppList/FiltersAndSorts.ts
```diff
@@ -140,12 +135,8 @@ export namespace AppListFilters {
...
       // In the case of health sorting, we may not have all health promises ready yet
       // So we need to get them all before actually sorting
-      const allHealthPromises: Promise<AppListItemHealth>[] = unsorted.map(item => {
-        return item.healthPromise.then(health => {
-          const withHealth: any = item;
-          withHealth.health = health;
-          return withHealth;
-        });
+      const allHealthPromises: Promise<WithAppHealth<AppListItem>>[] = unsorted.map(item => {
+        return item.healthPromise.then((health): WithAppHealth<AppListItem> => ({ ...item, ...{ health } }));
```
jotak: Minor: `...{ health }` can be simplified to just `health` I think? @israel-hdez pointed it out in earlier review iirc.
Same thing occurs 3 times (app/workloads/service sorting)

src/pages/ServiceList/__tests__/ServiceListComponent.test.ts
israel-hdez: I think we miss the classes from other languages which set the defaults in constructors and then you only set the data you need ;)

cfcosta:
Great, will merge this after TP is branched out.

cfcosta:
@mtho11 ga has been branched, can we merge this?

cfcosta:
@jotak @israel-hdez  same question ^

israel-hdez:
@cfcosta Go ahead. Just check if Joel's comment can be fixed.

cfcosta:
@israel-hdez ok, am fixing the merge conflicts on #1225 but will do the change as soon as I finish it.

cfcosta:
@mtho11 @israel-hdez @jotak fixed the nitpick, will merge as soon as build is green.

PR #1241, opened 2019-06-05
KIALI-2529 Fix compilation error

cfcosta:
This should not compile before, and it does not after the update to `react-scripts`, and to be honest I have no idea if that has even worked at all. Either way, this PR fixes it.

### DIFF
--- a/src/pages/Overview/OverviewCardContentExpanded.tsx
+++ b/src/pages/Overview/OverviewCardContentExpanded.tsx
@@ -136,7 +136,7 @@ class OverviewCardContentExpanded extends React.Component<Props> {
         <>
           {'Traffic, ' + getName(this.props.duration).toLowerCase()}
           <SparklineChart
-            id={'card-sparkline-' + name}
+            id={'card-sparkline-' + this.props.name}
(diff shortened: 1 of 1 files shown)


josejulio:
Check line [55](https://github.com/kiali/kiali-ui/blob/757488a7e814c538891b600dd7720d23e95dd26f/src/pages/Overview/OverviewCardContentExpanded.tsx#L55)

cfcosta:
@josejulio that's on `renderLeft()`, the change is on `renderRight()`.

josejulio:
> @josejulio that's on `renderLeft()`, the change is on `renderRight()`.

Ah, that makes sense on the "I don't know it worked before :)"

josejulio:
This was taking name from "window.name"

cfcosta:
> This was taking name from "window.name"

oh... that's even worse.

josejulio:
> oh... that's even worse.

Its a good thing that `react-scripts` prevents it.

israel-hdez:
Curiously, there is on `id` attribute in the final DOM. So, that attribute is, probably, not having any effect.

PR #1248, opened 2019-06-06
KIALI-2990 Show empty page messaging when overview has no namespaces

jshaughn:
kiali-2990

### DIFF
--- a/src/pages/Overview/OverviewPage.tsx
+++ b/src/pages/Overview/OverviewPage.tsx
@@ -1,5 +1,16 @@
 import * as React from 'react';
-import { Breadcrumb, Card, CardBody, CardGrid, CardTitle, Col, Row } from 'patternfly-react';
+import {
+  Breadcrumb,
+  Card,
+  CardBody,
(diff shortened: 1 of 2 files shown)


mtho11:
Can we get a before/after?
This is a nit, but the PR doesn't follow the template just one animated gif.

josejulio:
Before: before

jshaughn:
gracias for the screen shot @josejulio !

abonas:
@jstickle can you please review this new sentence? I'm having doubts about "user account has no accessible namespaces", perhaps this should be changed to something like "user has no permissions to access namespaces?" 
It's not that the namespaces are accessible (or not), nor do they belong to user accounts, but the user accounts have no access/permissions to the namespaces.

JStickler:
@abonas I think you revision ( "user has no permissions to access namespaces") is an improvement.

jshaughn:
@abonas @JStickler Please see and approve https://github.com/kiali/kiali-ui/pull/1253

PR #1249, opened 2019-06-06
KIALI-1717 Overview default sort from namespace name to health (aka status)

jshaughn:
Now, by default, the most unhealthy namespace cards should appear first.  Note that the URL will trump the default, so when testing ensure the URL 'sort' or 'direction' params are not overriding the default.

@jotak Any reason this sort field is called 'status' and not 'health' ?


jotak:
@jshaughn I think I just followed the initial design mockup. There's probably no strong reason about that. We can probably rename.

israel-hdez:
The fix for the health sorting is good! Now Apps and Services sort as expected.

PR #1250, opened 2019-06-06
KIALI-2993 Remove PatternFly sample background images

israel-hdez:
https://issues.jboss.org/browse/KIALI-2993

Sample background images are at:
* Login page
* About dialog box
* Masthead
* Initializing screen

Also, fixing a Firefox issue about white background in the login page.

Screenshots:

image

image

image

image

### DIFF
--- a/src/styles/_overrides.scss
+++ b/src/styles/_overrides.scss
@@ -63,8 +63,8 @@ $grid-float-breakpoint: $screen-sm-min !default;
 
 .pf-c-page__sidebar {
   --pf-c-page__sidebar--BackgroundColor: #{$color-pf-black-900};
-  --pf-c-page__sidebar--PaddingBottom: 0;
-  --pf-c-page__sidebar--PaddingTop: 0;
+  --pf-c-page__sidebar-body--PaddingBottom: 0;
+  --pf-c-page__sidebar-body--PaddingTop: 0;
   --pf-c-page__sidebar--md--Width: 210px; // TODO: Remove. Test enlarged message center when removing.
   min-height: 100vh;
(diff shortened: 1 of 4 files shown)


src/styles/_overrides.scss
```diff
@@ -213,6 +213,26 @@ $grid-float-breakpoint: $screen-sm-min !default;
...
+
+.pf-c-login,
+.login-pf {
+  background-image: none;
+  background-color: #030303;
+}
+
+.pf-c-about-modal-box__hero {
+  background-image: none !important;
+  background-color: transparent;
```
aljesusg: I prefer remove the background in component or pass a className to override the style instead use !important

israel-hdez: Placing styles here is what I understood from your other comment: https://github.com/kiali/kiali-ui/pull/1217/files#r288317015. I probably misunderstood :)

The About component does not have a background image set. It is PF4 styles which are setting that image by default.

I'm trying with typestyle. I still need to do:

```js
const aboutBoxStyle = style({
  $nest: {
    '& .pf-c-about-modal-box__hero': {
      backgroundImage: 'unset !important',
      backgroundColor: 'transparent'
    }
  }
});
```
PF styles have the `!important` flag, so even when using a className the `!important` flag is needed to prevent the PF4 styles taking precedence. So, I'm not sure if it is any better.

aljesusg: Oh I saw that we can change the background in a newest release of PF4 we are in 3.2 :( ok so go ahead , after GA we can make changes

mtho11:
Good to have this one fixed. Especially since it looked different on each browser.

PR #1252, opened 2019-06-10
KIALI-3000 Improve find/hide for edge traffic expressions

jshaughn:
Before to match edge certain edge rates you needed to treat the zero-case with a NaN match.  For example:

  "%httperror < 5 or %httptraffic = NaN"

Now it works as expected:

  "%httperror < 5"


abonas:
cc @ceastman-ibm

israel-hdez:
LGTM!

PR #1253, opened 2019-06-10
KIALI-2990 UX wording change for original fix

jshaughn:


abonas:
LGTM, but what about the tests? shouldn't they also include the new phrase?

jshaughn:
@abonas Yep, of course there is just nothing known as a 1-line fix :)  I'll update the test...

jshaughn:
@abonas Test has been updated.

PR #1256, opened 2019-06-11
KIALI-3005 Hide nodes with only hidden edges

jshaughn:
Here is an example of the change. I have bookinfo and istio-system and I hide the GRPC traffic edges:

Before:
image
After:
image


jshaughn:
@josejulio , In the future it may be requested that we can toggle between "hide" and "remove", with the latter option causing the graph to condense.  But until requested by users I think we can wait.  I have updated the Help, both the preface text and also two new entries in Usage Notes. What do you think?

image

PR #1259, opened 2019-06-13
Graph toolbar truncated on chrome

mtho11:
** Describe the change **
Fixes some weird truncation issues with the graph toolbar that happen on Chrome on smaller screens.

** Issue reference **

https://issues.jboss.org/browse/KIALI-2743

** Backwards compatible? **
Yes
No

** Screenshot **

Before:

bad-toolbar

After:

toolbar-good


mtho11:
It looks like there is a similar but more complex issue with the actual graph Legend. The scrollbars on the smaller screens and repeating SVG make this one more complex. I will create another jira for it.

PR #1265, opened 2019-06-18
KIALI-3035 Code-splitting for main navigation components

israel-hdez:
Before code-splitting, build reports these sizes:

```
File sizes after gzip:

  1.35 MB (+555.98 KB)  build/static/js/main.86522805.js
  165.32 KB (+460 B)    build/static/css/main.1734869f.css
```

After code-splitting, build reports these sizes:

```
File sizes after gzip:

  828.63 KB (-7 B)  build/static/js/main.158c81c6.js
  239.87 KB         build/static/js/3.08cdbb49.chunk.js
  177.84 KB         build/static/js/4.5d409cea.chunk.js
  164.87 KB         build/static/css/main.7eaa2ccd.css
  135.46 KB         build/static/js/0.fe2dc93f.chunk.js
  115.75 KB         build/static/js/1.5419a62c.chunk.js
  109.91 KB         build/static/js/2.40757215.chunk.js
  8.83 KB           build/static/js/5.b70775d6.chunk.js
  7.91 KB           build/static/js/6.2642d802.chunk.js
  7.48 KB           build/static/js/7.2ff671c9.chunk.js
  7.07 KB           build/static/js/10.b29f83f1.chunk.js
  6.96 KB           build/static/js/8.d1a01ffb.chunk.js
  5.62 KB           build/static/js/9.e62e27ca.chunk.js
```

### DIFF
--- a/src/components/Nav/RenderPage.tsx
+++ b/src/components/Nav/RenderPage.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Suspense } from 'react';
 import { Redirect, Route } from 'react-router-dom';
 import SwitchErrorBoundary from '../SwitchErrorBoundary/SwitchErrorBoundary';
 import { pathRoutes, defaultRoute, secondaryMastheadRoutes } from '../../routes';
@@ -25,12 +25,16 @@ class RenderPage extends React.Component<{ needScroll: boolean }> {
   render() {
     const component = (
       <div className={`container-fluid ${containerStyle}`}>
-        <SwitchErrorBoundary
-          fallBackComponent={() => <h2>Sorry, there was a problem. Try a refresh or navigate to a different page.</h2>}
-        >
-          {this.renderPathRoutes()}
-          <Redirect from="/" to={defaultRoute} />
-        </SwitchErrorBoundary>
+        <Suspense fallback={<div>Loading...</div>}>
+          <SwitchErrorBoundary
+            fallBackComponent={() => (
+              <h2>Sorry, there was a problem. Try a refresh or navigate to a different page.</h2>
+            )}
+          >
+            {this.renderPathRoutes()}
+            <Redirect from="/" to={defaultRoute} />
+          </SwitchErrorBoundary>
+        </Suspense>
       </div>
     );
     return (
--- a/src/routes.tsx
+++ b/src/routes.tsx
@@ -1,16 +1,6 @@
-import WorkloadListPage from './pages/WorkloadList/WorkloadListPage';
-import ServiceListPage from './pages/ServiceList/ServiceListPage';
-import IstioConfigPage from './pages/IstioConfigList/IstioConfigListPage';
-import ServiceJaegerPage from './pages/ServiceJaeger/ServiceJaegerPage';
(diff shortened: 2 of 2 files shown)


src/routes.tsx
```diff
@@ -63,71 +53,74 @@ const navItems: MenuItem[] = [
...
 const defaultRoute = '/overview';
 
+const graphPage = lazy(() => import('./pages/Graph/GraphPage'));
+const istioConfigDetailsPage = lazy(() => import('./pages/IstioConfigDetails/IstioConfigDetailsPage'));
+
 const pathRoutes: Path[] = [
   {
     path: '/overview',
-    component: OverviewPageContainer
+    component: lazy(() => import('./pages/Overview/OverviewPage'))
```
josejulio: Why did you followed a different pattern here? (directly calling lazy here)

mtho11:
@israel-hdez can you also include before/after loading times as this is purely a performance PR?

mtho11:
@gbaufake can we also get QE test here?

src/routes.tsx
israel-hdez: No hard reason. I just called directly where I saw no need to introduce a variable.
I can use a consistent pattern if that looks better.

gbaufake:
I will test it

israel-hdez:
@mtho11 I think load times vary from env to env, so comparisons may not reflect the end user experience. But, using my local env, and using "dev mode" `yarn start`, the **initial** load time I get is:

Before:

image

After:

image

This is only **initial** load time. Since code-splitting produces several JS files, those ones are loaded on-demand when needed... So, I guess "load time" is only "deferred".

israel-hdez:
That test failure seems hard to fix :(

mattmahoneyrh:
Unable to login on OCP4 via OAUTH:

"The client is not authorized to request a token using this method."

mattmahoneyrh:
> Unable to login on OCP4.x via OAUTH:
> 
> "The client is not authorized to request a token using this method."

Update - No OAUTH issue here - the issue was a miss-match w/ the kiali oauthclient and deploying kiali w/ TP11.

israel-hdez:
@mtho11 @josejulio Luckily, test failure looks fixed with the migration to react-scripts. So, I guess this can move forward. Do you mind doing another pass?
I just rebased, no code changes from your last review.

src/components/Nav/RenderPage.tsx
```diff
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Suspense } from 'react';
```
cfcosta: Nice! Have been thinking on introducing `Suspense` on the codebase for a while.

src/routes.tsx
cfcosta: I like forcing `lazy` when possible.

jotak:
@gbaufake and me are seeing some issues today and I wonder if it is related to this PR.
Running kiali with `yarn start` works fine, but with `yarn build` + image build is not working, leaving a blank screen after login, JS console showing an error that points to: https://reactjs.org/docs/error-decoder.html/?invariant=130&args[]=undefined&args[]=

Building from v1.2 tag, however, works fine. As this is (almost) the only commit between 1.2 and master it is suspicious :-)

PR #1266, opened 2019-06-20
Mark master as 1.1.0

lucasponce:
We are going to have two active branches.
Master will host the 1.1.0 and following versions, so better to update the versions.

There will be a similar one for backend.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@kiali/kiali-ui",
-  "version": "0.22.0",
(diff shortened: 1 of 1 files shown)


lucasponce:
Added DNM label.
Github is having issues this morning and I couldn't prepare the PR for bump kiali/kiali yet.

lucasponce:
https://github.com/kiali/kiali/pull/1165

PR #1268, opened 2019-06-20
KIALI-3048 Fix vertical scroll appearing in workload logs tab

israel-hdez:


israel-hdez:
Issue was introduced in #1171 where the "title" was pulled out of the Info/overview tabs, reducing the space available for the logs pane. That PR went into v0.21.0 so we should backport.

israel-hdez:
...this is a little brittle. Perhaps we should find a way to make it adapt regardless of the controls that are present. But that may involve more changes to the markup and css.

josejulio:
> ...this is a little brittle. Perhaps we should find a way to make it adapt regardless of the controls that are present. But that may involve more changes to the markup and css.

Maybe with some flexbox

src/styles/_layout.scss
```diff
@@ -6,9 +6,9 @@ body,
 .pf-c-page__main-section {
   /*
    * --pf-c-page__header--MinHeight is the height of the masthead
-   * 82px is the vertical space being used by the breadcrumb and tab headers
+   * 130px is the vertical space being used by the breadcrumb, tab headers, and the name of the {workload,svc,app}
    */
-  --kiali-details-pages-tab-content-height: calc(100vh - 82px - var(--pf-c-page__header--MinHeight));
+  --kiali-details-pages-tab-content-height: calc(100vh - 130px - var(--pf-c-page__header--MinHeight));
```
mtho11: Yeah, I wish we could get away from all of the pixel calculations.

israel-hdez:
> Brittle yes, working yes.

Merging because it passes all checks :)

src/styles/_layout.scss
israel-hdez: I think that the right way to make this happen is to contribute to patternfly-react. Anything we do in kiali-ui will involve hacking the PF css.

PR #1272, opened 2019-06-21
KIALI-3053 Adds a scrollbar when needed when listing too many namespaces

josejulio:
** Describe the change **

It provides scrollbars to the namespace selector when needed.

This also add a class that computes the offset from the element to the screen, it allows us to avoid manual computations on that value, adjusting better if we ever make a change there.

If someone could find a better name for this component, please let me know.

** Issue reference **

Fixes https://github.com/kiali/kiali/issues/1166

** Screenshot **

namespaces

### DIFF
--- a/src/components/BoundingClientAwareComponent/BoundingClientAwareComponent.tsx
+++ b/src/components/BoundingClientAwareComponent/BoundingClientAwareComponent.tsx
@@ -1,0 +1,90 @@
+import * as React from 'react';
+
+type Rect = ClientRect | DOMRect;
+
+export enum PropertyType {
+  VIEWPORT_HEIGHT_MINUS_TOP
+}
+
+export type Property = {
+  type: PropertyType;
+  margin?: number;
+};
+
+type ComputeOffsetProps = {
+  className: string;
+  maxHeight?: Property;
+  handleBoundingClientRect?(rect: Rect): void;
+};
+
+type ComputeOffsetState = {
+  maxHeight?: string;
+};
+
+export const vhMinusTop = (rect: Rect, offset: number) => `calc(100vh - ${rect.top + offset}px)`;
+
+// Computes the BoundingClientRect of the container, this helps to calculate the remaining height without
+// going further off the screen and without having to fix the value in the code.
+// Note: This does re-compute when there is a change in this component, but external changes are not yet
+// managed, that might require to observe the offsets, for our current use case this seems OK, as the top
+// headers doesn't change in height.
+export class BoundingClientAwareComponent extends React.Component<ComputeOffsetProps, ComputeOffsetState> {
+  private readonly containerRef: React.RefObject<HTMLDivElement>;
+
+  constructor(props: ComputeOffsetProps) {
+    super(props);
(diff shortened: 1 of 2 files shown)


src/components/BoundingClientAwareComponent/BoundingClientAwareComponent.tsx
```diff
@@ -0,0 +1,90 @@
...
+};
+
+export const vhMinusTop = (rect: Rect, offset: number) => `calc(100vh - ${rect.top + offset}px)`;
+
+// Computes the BoundingClientRect of the container, this helps to calculate the remaining height without
+// going further off the screen and without having to fix the value in the code.
+// Note: This does re-compute when there is a change in this component, but external changes are not yet
+// managed, that might require to observe the offsets, for our current use case this seems OK, as the top
+// headers doesn't change in height.
+export class BoundingClientAwareComponent extends React.Component<ComputeOffsetProps, ComputeOffsetState> {
```
mtho11: Isn't this `BoundingClientAwareComponent` a little overcomplicated for showing a scrollable list of components?

josejulio: The plan is to use this component in other places where we are fixing stuff like:
`calc(100vh - hard_coded_number_of_px)`
That tends to break easily.

Plus on this case, the math is not direct, it depends on the size of the tip of the arrow, the size of the "clear all", the input, etc.

jshaughn: My 2cents is that we need something like this to consolidate these sorts or calculated bounds.  This is probably as good a time as any to introduce it.  [disclaimer: not an expert]

israel-hdez: Quite complicated, but in the long run, it may be easier to use this, rather than do the math manually. The complexity needs to be somewhere :)

abonas:
cc @serenamarie125 is there anything similar in the selector in openshift console?

abonas:
@josejulio it is looking good IMO!

lucasponce:
Not specifically marked, but do we want this one for the v1.0 branch ?

abonas:
> Not specifically marked, but do we want this one for the v1.0 branch ?

I would say from functionality perspective -  yes, it's fixing a bug and will be useful to customers will a lot of ns,  but I can't assess the code change risk here, seems it's not a one liner.

lucasponce:
Ok.
@josejulio I've added it to the v1.0. Code had no conflicts, but better if you can also take a quick look.

Thanks

josejulio:
This one has a bug in non webkit browsers, the fix easy and here: https://github.com/kiali/kiali-ui/pull/1281
I added the backport_v1.0 label.

serenamarie125:
here's what's done in OpenShift today (4.x)

mWZJknoXdX

josejulio:
What we did is similar, except that we "fix" the search filter on top. It doesn't gets hidden when scrolling.

PR #1270, opened 2019-06-21
KIALI-2907 service details: move endpoint name into tooltip + enzyme-to-json

jotak:
Adding dependency enzyme-to-json to reduce snapshot size

JIRA: https://issues.jboss.org/browse/KIALI-2907

sshot

PR #1273, opened 2019-06-24
Change the way we test snapshots

cfcosta:
Uses `enzyme-to-json` on all places instead of only the ones @jotak changed.

[1]: https://github.com/facebook/jest/issues/7802
[2]: https://github.com/adriantoine/enzyme-to-json


jotak:
@cfcosta I've already added enzyme-to-json with this PR : https://github.com/kiali/kiali-ui/pull/1270 :-(
So I guess you can revert that part of my commit (remove the dependency)

israel-hdez:
What I don't like too much is that attrributes are not always expanded, like:

```xml
<TablePfProvider columns={{...}} striped={true} bordered={true} hover={true} dataTable={true} className=\\"\\" condensed={false} inlineEdit={false} components={{...}}>
```

note the `columns={{...}}` and `components={{...}}` attributes. I also saw the same happening for `style` attributes.
If enzyme-to-json solves this better, I would prefer to include the dependency.

cfcosta:
@israel-hdez this happens because it does shallow instead of complete rendering, so it does not saves snapshots when you pass objects/functions to the components.

This is necessary to have some layering on the rendering. We can use full rendering to get those to be saved as well, but that's out of the scope.

cfcosta:
@jotak @israel-hdez changed the PR to start using `enzyme-to-json` on all places, can you review it again?

cfcosta:
@jotak I'd prefer to keep `debug()`, but we already have enzyme-to-json on the repo, and it works. The less we change the better.

Also, @israel-hdez found some issues where important data was being hidden. To fix those problems, we would need to use `.render().debug()`, but it requires redux to be "running" to do that. So I decided to go the simplest way and then we can try something more complex later.

PR #1274, opened 2019-06-24
Use tsc unused variable checking instead of tslint

cfcosta:
Related to #1154.

This was supposed to fix the warning that appears when you try to build:

```
no-unused-variable is deprecated. Since TypeScript 2.9. Please use the built-in compiler checks instead.
```

Even disabling the rule did not make the warning disappear, but enabling the checking on tsconfig showed some warnings that this PR fixes.
Those are also errors after the upgrade to `react-scripts`, but we can get the benefits now.

Those are the offending files:

```
ERROR: /home/cfcosta/Projects/go/src/github.com/kiali/kiali-ui/src/components/IstioWizards/WeightedRouting.tsx:161:11 - 'numLocks' is declared but its value is never read.
ERROR: /home/cfcosta/Projects/go/src/github.com/kiali/kiali-ui/src/components/JaegerIntegration/__tests__/LookBack.test.tsx:6:16 - 'onChangeCustom' is declared but its value is never read.
ERROR: /home/cfcosta/Projects/go/src/github.com/kiali/kiali-ui/src/components/JaegerIntegration/__tests__/RightToolbar.test.tsx:6:16 - 'onGraphClick' is declared but its value is never read.
ERROR: /home/cfcosta/Projects/go/src/github.com/kiali/kiali-ui/src/components/JaegerIntegration/__tests__/RightToolbar.test.tsx:6:30 - 'onSummaryClick' is declared but its value is never read.
ERROR: /home/cfcosta/Projects/go/src/github.com/kiali/kiali-ui/src/components/JaegerIntegration/__tests__/RightToolbar.test.tsx:6:46 - 'onMinimapClick' is declared but its value is never read.
```

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -16,6 +16,7 @@
     "noImplicitThis": true,
     "noImplicitAny": false,
     "strictNullChecks": true,
+    "noUnusedLocals": true,
     "suppressImplicitAnyIndexErrors": true,
     "allowSyntheticDefaultImports": true,
     "tsBuildInfoFile": ".tsbuildinfo"
--- a/src/components/IstioWizards/WeightedRouting.tsx
+++ b/src/components/IstioWizards/WeightedRouting.tsx
@@ -158,14 +158,12 @@ class WeightedRouting extends React.Component<Props, State> {
   onLock = (workloadName: string, locked: boolean) => {
     this.setState(prevState => {
       let maxWeights = 100;
-      let numLocks = 0;
       for (let i = 0; i < prevState.workloads.length; i++) {
         if (prevState.workloads[i].name === workloadName) {
           prevState.workloads[i].locked = locked;
         }
         // Calculate maxWeights from locked nodes
         if (prevState.workloads[i].locked) {
-          numLocks++;
           maxWeights -= prevState.workloads[i].weight;
         }
       }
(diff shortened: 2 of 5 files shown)


tsconfig.json
```diff
@@ -16,6 +16,7 @@
     "noImplicitThis": true,
     "noImplicitAny": false,
     "strictNullChecks": true,
+    "noUnusedLocals": true,
```
josejulio: I just want to point out that (i think) this won't allow to have unused locals on dev builds (e.g. `yarn start`), is possible to only have them in `yarn build` but not on `yarn start`? 
I'm going to test this later and confirm.

cfcosta: I don't think it is, but to be honest I don't like the idea of having compilation targets having different outputs/statuses.

josejulio: From what i remember, you could include one file into the other.
The idea is to make it easier to test changes without having to have a flawless code.
e.g. commenting a path to quickly debug something without the "unused variable" error that does not let you  proceed.

josejulio: Oh and you cal set them to "WARNS" instead of "ERRORS", so they don't fail but you still get the info that your code is "dirty".

cfcosta: You can, however warns are supposed to break CI :(
My opinion on this matter is that the more strict the compiler is, the earlier you catch problems that will take a little more time later.

cfcosta:
I talked with @josejulio about this PR, we agreed to merge it now and if people start complaining/having problems, we can move it to raise warnings instead of errors.

PR #1282, opened 2019-06-25
KIALI-3058 Add version label in app-based custom metrics page

jotak:
The "additionalLabels" param was missing in case version is not provided (app view)

Now back to normal:

view

JIRA: https://issues.jboss.org/browse/KIALI-3058


jotak:
no, this is based on post-ga work so that's fine

PR #1287, opened 2019-06-27
KIALI-3071 Split details and 3scale promises

lucasponce:
- Move 3scale error message to INFO

As similar strategy with Jaeger/Grafana, if user has not rights to query istio-system ns to check if 3scale adapter is deployed, it shouldn't fail the service details page, but show an informative message instead.

### DIFF
--- a/src/pages/ServiceDetails/ServiceDetailsPage.tsx
+++ b/src/pages/ServiceDetails/ServiceDetailsPage.tsx
@@ -292,6 +292,7 @@ class ServiceDetails extends React.Component<ServiceDetailsProps, ServiceDetails
         const objectValidations = validations.destinationrule[destinationRule.metadata.name];
         if (
           formatValidation !== null &&
+          objectValidations.checks &&
           !objectValidations.checks.some(check => check.message === formatValidation.message)
         ) {
           objectValidations.checks.push(formatValidation);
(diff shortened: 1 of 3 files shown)


src/pages/ServiceDetails/ServiceDetailsPage.tsx
```diff
@@ -249,7 +250,11 @@ class ServiceDetails extends React.Component<ServiceDetailsProps, ServiceDetails
         }
       })
       .catch(error => {
-        MessageCenter.add(API.getErrorMsg('Could not fetch Service Details', error));
+        MessageCenter.add(
+          API.getInfoMsg('Could not fetch 3scale info. Turning off 3scale integration.', error),
```
abonas: is it accurate to say that the integration is switched off in this scenario?

lucasponce: From the perspective of the user, yes.
He/she won't have access to 3scale integration through Kiali as he/she doesn't have right access.

abonas: It might give the impression that the integration will be disabled, which is not the case here. I think it's better to say something that describes the situation more accurately - user don't have the right permissions.

lucasponce: We are following the same message used where Jaeger and Grafana integration for consistency.
As it's similar scenario.

abonas: so perhaps the messages there need to be changed a bit as well :) 
@jotak @aljesusg

PR #1294, opened 2019-07-02
[v1.0] KIALI-2985 normalize timeseries on X axis

jotak:
+ Add tests

This is targeting v1.0 branch directly.
Note that a slightly different fix will be needed for master, as it's involving k-charted there (no cherry-pick). I'll open a separate PR.

Series are now correctly left-aligned when they start later (here, v3 started after v1):

shift

### DIFF
--- a/src/utils/Graphing.ts
+++ b/src/utils/Graphing.ts
@@ -1,5 +1,57 @@
 import { TimeSeries, Histogram } from '../types/Metrics';
 
+type Series = {
+  name: string;
+  values: number[];
+};
+
+type Normalized = {
+  x: number[];
+  ys: Series[];
+};
+
+const compare = (a: number, b: number) => {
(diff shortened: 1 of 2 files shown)


src/utils/Graphing.ts
```diff
@@ -1,39 +1,103 @@
...
+  values: number[];
+};
+
+type Normalized = {
+  x: number[];
+  ys: Series[];
+};
+
+const compare = (a: number, b: number) => {
+  if (Math.abs(a - b) < 1000) {
```
israel-hdez: One second tolerance, right?

jotak: Yes. Actually, all timestamps returned by prometheus, when several timeseries are returned, should in theory be aligned. But I prefer to have this security tolerance and not rely on this assumption too much.

PR #1295, opened 2019-07-03
KIALI-GH-1211 Improve layout popups

jshaughn:
- Also, add tooltips for zoom options

This is for one of the bullets in https://github.com/kiali/kiali/issues/1211.

kiali-gh-1211

### DIFF
--- a/src/components/CytoscapeGraph/CytoscapeToolbar.tsx
+++ b/src/components/CytoscapeGraph/CytoscapeToolbar.tsx
@@ -64,16 +64,16 @@ export class CytoscapeToolbar extends React.PureComponent<CytoscapeToolbarProps>
     return (
       <div className={cytoscapeToolbarStyle}>
         <ButtonGroup>
-          <Button onClick={this.zoomIn}>
+          <Button onClick={this.zoomIn} title="zoom in">
             <Icon type="fa" name="plus" />
           </Button>
 
-          <Button onClick={this.zoomOut}>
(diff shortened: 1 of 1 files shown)


jshaughn:
cc @abonas

src/components/CytoscapeGraph/CytoscapeToolbar.tsx
```diff
@@ -64,16 +64,16 @@ export class CytoscapeToolbar extends React.PureComponent<CytoscapeToolbarProps>
     return (
       <div className={cytoscapeToolbarStyle}>
         <ButtonGroup>
-          <Button onClick={this.zoomIn}>
+          <Button onClick={this.zoomIn} title="zoom in">
```
josejulio: I think we should capitalize the first word, e.g. `Zoom in` and the same for the rest of the titles.

mtho11: I think we want to use the full-on PF Tooltips. They allow us nicer formatting and font choices (and even icons if we want)

jshaughn: ok, I mean OK

jshaughn: I just knew someone would ask for that...

jshaughn:
@mtho11 @josejulio The video and PR have been updated. Are you OK with the tooltip text?

PR #1298, opened 2019-07-04
KIALI-3097 Add scrollbar (when required) to MessageCenter

josejulio:
** Describe the change **

Prevents the message center from growing over the view port.

** Screenshot **

messagecenter

### DIFF
--- a/src/components/MessageCenter/NotificationDrawer.tsx
+++ b/src/components/MessageCenter/NotificationDrawer.tsx
@@ -13,6 +13,17 @@ import {
 
 import { MessageType, NotificationMessage, NotificationGroup } from '../../types/MessageCenter';
 import moment from 'moment';
+import {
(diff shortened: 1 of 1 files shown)


josejulio:
> It's OK... although I'd prefer to have the action area always in the screen.
> But with the several "groups" I'm not sure if it is easily doable.

I could try something

josejulio:
Added the DNM tag, this is ready to be merged as is, but i want to try to improve it a little more. 
If anyone feels it should be in a separate PR, please merge this and I'll open other PR.

josejulio:
I tried some ideas with css and the libs that we currently have and none provided a satisfactory solution.
Tried to use position:sticky, adding the scrolls to each group (we lose the main scroll or have 2 scrolls) and other things.

I'll merge as is and maybe we can check again later.

PR #1301, opened 2019-07-07
KIALI-3107 Typescript requirement warning by  typescript-estree, typescript should be <3.5

aljesusg:
JIRA: [KIALI-3107](https://issues.jboss.org/browse/KIALI-3107)

Typescript should be less than 3.5 
https://github.com/kiali/kiali-ui/blob/master/package.json#L134

```bash
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=3.2.1 <3.5.0

YOUR TYPESCRIPT VERSION: 3.5.1

Please only submit bug reports when using the officially supported version.

=============
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -136,6 +136,8 @@
   },
   "resolutions": {
     "@types/react": "16.7.13",
+    "@typescript-eslint/parser": "1.11.0",
+    "@typescript-eslint/eslint-plugin": "1.11.0",
     "handlebars": "4.0.14",
     "js-yaml": "3.13.1",
     "hoist-non-react-statics": "3.3.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1650,6 +1650,11 @@
     "@types/cheerio" "*"
     "@types/react" "*"
 
+"@types/eslint-visitor-keys@^1.0.0":
+  version "1.0.0"
+  resolved "https://registry.yarnpkg.com/@types/eslint-visitor-keys/-/eslint-visitor-keys-1.0.0.tgz#1ee30d79544ca84d68d4b3cdb0af4f205663dd2d"
+  integrity sha512-OCutwjDZ4aFS6PB1UZ988C4YgwlBHJd6wCeQqaLdmadZ/7e+w79+hbMUFC1QXDNCmdyoRfAFdm0RypzwR+Qpag==
+
 "@types/geojson@*":
   version "7946.0.4"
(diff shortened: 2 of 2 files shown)


cfcosta:
This is expected from #1154, so I'm -1 on merging this (it works with 3.5.0 without warnings, but 3.5.1 does not have breaking change).

lucasponce:
I am not sure if I understand the context.
Project is warning that a library version is required.
I see that this PR just tries to adjust that version in the package.json.

I think when user install the depedencies, those should be correct with the required by project, to reduce the chances to some problems.

Please, @cfcosta @aljesusg clarify this one.

From user perspective, in my opinion, the expected goal should be that after a `yarn install` user should have the right versions required without warnings.

(And I think that was the motivation of this fix).

aljesusg:
+1 @lucasponce, actually there is an warning error in Travis build and yarn start @cfcosta

jotak:
If we want to continue with more recent Typescript (3.5.x) the alternative would be to upgrade typescript-eslint to 1.10.1 : https://github.com/typescript-eslint/typescript-eslint/releases/tag/v1.10.1

Maybe it's better than a language downgrade?

cfcosta:
@lucasponce yeah, that's correct, but we knew about the warning when we merged, and downgrading typescript is a bad option because of that. @jotak mentioned that upgrading typescript-eslint might fix the problem, and I'm +1 for that.

aljesusg:
My point was remove this warning, I am good upgrading so

cfcosta:
@aljesusg what do you think we should do then? Can I close this PR and you create another one with the upgrade to `typescript-eslint`?

aljesusg:
No, ley me make the changes in this PR. The target is the same remove the warning

package.json
```diff
@@ -64,6 +64,7 @@
   "dependencies": {
     "@patternfly/patternfly": "2.4.1",
     "@patternfly/react-core": "3.2.6",
+    "@typescript-eslint/eslint-plugin": "^1.11.0",
```
josejulio: This fixed the issue?
If so, could you please lock to "1.11.0" ?

mtho11:
@aljesusg where do we sit with this?

aljesusg:
Hi

@mtho11 @josejulio The problem here is
@typescript-eslint/parser is in 1.6 by "react-scripts": "3.0.1" (latest version). I need to check this

aljesusg:
This fix the problem @jotak @mtho11

mtho11:
@aljesusg great!

PR #1303, opened 2019-07-08
KIALI-3093 Fix reset pagination when NS changes

lucasponce:
** Describe the change **

Reset pagination when user changes the namespace to fetch.
There are some scenarios where user can move to a page that doesn't exist.

** Issue reference **

https://issues.jboss.org/browse/KIALI-3093


lucasponce:
Ok, thanks, I'm going to move this one to v1.0 branch as it happens there as well

PR #1304, opened 2019-07-08
 KIALI-3109 Fix service graph error

josejulio:
** Describe the change **

Fix error that appeared when using the service graph (e.g. by going to a service and clicking "Show on graph")

This was caused because the result of incomers() do not implement the [Symbol.iterator](https://www.typescriptlang.org/docs/handbook/iterators-and-generators.html#iterables) needed for the `for-of` syntax.

Changed to use the [forEach](http://js.cytoscape.org/#eles.forEach) of cytoscapejs.

[KIALI-3109 ](https://issues.jboss.org/browse/KIALI-3109)

Before:
image-2019-07-08-14-29-57-652

After:
Screenshot_2019-07-08_16-51-20


josejulio:
> Can we get before/after screenshots? It's hard to tell what this error looked like and how severe it was.

Done

> Also, a link to the jira issue.

And done

PR #1307, opened 2019-07-09
Trivial: Separate browserlist from all in one to PROD and DEV sections

mtho11:
** Describe the change **

Separate browserlist from all in one to production and development settings for use with 'yarn build' and 'yarn start' respectively. This means we will be testing the more current browsers in development while supporting more versions in production (and more kinds)

** Backwards compatible? **
Yes
No

https://github.com/browserslist/browserslist

### DIFF
--- a/package.json
+++ b/package.json
@@ -158,10 +158,17 @@
       "pre-commit": "yarn run pretty-quick --staged --pattern \"{src/**/*.{js,jsx,ts,tsx,json,yml,css,scss},travis.yml,*.json}\" && npm run lint:precommit"
     }
   },
-  "browserslist": [
-    ">10%",
-    "last 2 versions",
-    "not ie <= 11"
-  ],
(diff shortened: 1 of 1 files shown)


josejulio:
What does this means? Faster dev builds?
Won't this have the negative impact of some bugs happening in production and not in development due to some process not happening in dev?

mtho11:
It may be slightly faster because fewer browser settings will be loaded. Mainly, it will force developers to concentrate on the latest isolated version (features) so anything wrong with the latest version will be immediately obvious instead of masked by settings enabled in the last two versions.

mtho11:
Remember, debugging a production browser css/html issue means those features/browser prefixes were locked in at the time the production build was done. The last two versions at the time of the production build. Not at the time the dev build is being done (which depending on browser releases, may or may not be the same). This is all rolling time windows now.

PR #1312, opened 2019-07-10
KIALI-3102 Show TCP edges without traffic as "triangle-cross"

josejulio:
** Describe the change **

Shows the TCP edges without traffic using the `triangle-cross` shape (i.e. it looks like it has traffic but with gray color)

Depends on: https://github.com/kiali/kiali/pull/1234

** Issue reference **

[KIALI-3102](https://issues.jboss.org/browse/KIALI-3102)

** Screenshot **

Before: 
image

After:
tcp-inactive

### DIFF
--- a/src/store/Selectors/GraphData.ts
+++ b/src/store/Selectors/GraphData.ts
@@ -11,7 +11,8 @@ import {
   DecoratedGraphNodeWrapper,
   GraphEdgeWrapper,
   GraphElements,
-  GraphNodeWrapper
+  GraphNodeWrapper,
+  protocolTrafficHasData
 } from '../../types/Graph';
 
 // When updating the cytoscape graph, the element data expects to have all the changes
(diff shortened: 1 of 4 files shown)


src/store/Selectors/GraphData.ts
```diff
@@ -20,47 +21,47 @@ import {
...
+      httpIn4xx: NaN,
+      httpIn5xx: NaN,
+      httpOut: NaN,
       isDead: undefined,
       isGroup: undefined,
       isInaccessible: undefined,
-      isMisconfigured: undefined,
-      isOutside: undefined,
+      isMiscofigured: undefined,
+      isOutsde: undefined,
```
israel-hdez: Hmmm are these typos in the received JSON?

josejulio: Good catch, I didn't notice it.

PR #1316, opened 2019-07-11
Add typeRoot for resolving types from IDE

lucasponce:
I don't know if there is a better way, but it seems the current tsconfig.json doesn't help IDE to resolve dependencies.

Adding this one helps but not sure if it may have some non desired side effect.

Please, review.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -26,7 +26,8 @@
     "strict": true,
     "resolveJsonModule": true,
     "isolatedModules": true,
-    "noEmit": true
+    "noEmit": true,
+    "typeRoots": ["node_modules/@types"]
   },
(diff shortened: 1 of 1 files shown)


cfcosta:
What happens when you try getting types for dependencies that do not provide `@types`? I think `moment` is a good one to test that.

mtho11:
@cfcosta I think that is a corner case that you mentioned and shouldn't stop this from going in.

mtho11:
Just adding this for reference: https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#types-typeroots-and-types

cfcosta:
Tested here on vim with languageclient, seems to be working ok.

PR #1317, opened 2019-07-11
KIALI-3114 Refresh PfTitle when props change

lucasponce:
Fix bug on PfTitle as state is not properly refreshed when props change.

### DIFF
--- a/src/components/Pf/PfTitle.tsx
+++ b/src/components/Pf/PfTitle.tsx
@@ -30,10 +30,22 @@ interface PfTitleState {
   icon: JSX.Element;
 }
 
+const namespaceRegex = /namespaces\/([a-z0-9-]+)\/([a-z0-9-]+)\/([a-z0-9-]+)(\/([a-z0-9-]+))?(\/([a-z0-9-]+))?/;
+
 class PfTitle extends React.Component<PfTitleProps, PfTitleState> {
   constructor(props: PfTitleProps) {
     super(props);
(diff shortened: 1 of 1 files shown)


src/components/Pf/PfTitle.tsx
```diff
@@ -68,15 +79,30 @@ class PfTitle extends React.Component<PfTitleProps, PfTitleState> {
...
       type: type,
       name: name,
       graphType: graphType,
       icon: icon,
       cytoscapeGraph: cytoscapeGraph.build()
-    };
+    });
+  }
+
+  componentDidMount(): void {
```
cfcosta: I don't think we historically have added `void` type annotations for functions, but that's just a nitpick.

src/components/Pf/PfTitle.tsx
```diff
@@ -68,15 +79,30 @@ class PfTitle extends React.Component<PfTitleProps, PfTitleState> {
...
+    this.doRefresh();
+  }
+
+  componentDidUpdate(prevProps: Readonly<PfTitleProps>): void {
```
cfcosta: Same here.

src/components/Pf/PfTitle.tsx
lucasponce: We have them in other locations.

cfcosta: Np then :)

lucasponce:
@hhovsepy let me know if the issue is gone or if there is anything else not covered by the fix.

PR #1320, opened 2019-07-15
KIALI-3123 Disable Actions on Service for User without permissions

lucasponce:
Add a tooltip message when user has not permissions to perform actions.

### DIFF
--- a/src/components/IstioWizards/IstioWizardDropdown.tsx
+++ b/src/components/IstioWizards/IstioWizardDropdown.tsx
@@ -322,9 +322,10 @@ class IstioWizardDropdown extends React.Component<Props, State> {
 
   render() {
     const updateLabel = this.getVSWizardLabel();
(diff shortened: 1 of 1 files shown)


lucasponce:
> It shows disabled Actions even for cluster admin.

Ouch, thanks for testing, it seems my last change didn't go well.
I'll fix it.

hhovsepy:
Screenshot from 2019-07-16 19-03-49

lucasponce:
Can you test with a service with sidecars ?

In the attached screenshot it looks correct: actions are disabled as the service has not workloads with sidecars as it is described in the tooltip.

Test should be: a service like reviews, where you can't create/update/delete, then you should have actions disabled with a different message.

hhovsepy:
On a service with sidecars:
Screenshot from 2019-07-17 19-54-23

When user does not have update/delete role on istio-config:
Screenshot from 2019-07-17 19-54-14

When user has update/delete role:
Screenshot from 2019-07-17 19-58-30

PR #1325, opened 2019-07-17
KIALI-3135 Add more typesafety to node/edge data

josejulio:
** Describe the change **

Reduce direct calls to `ele.data(field)` and instead use the typesafe version of the call.
This adds more type safety, but we would need to close the loop (in a different PR) to ensure that the data we are receiving is in sync with our definitions. 
Probably something at runtime that can warns/errs about any difference found.

** Issue reference **

https://issues.jboss.org/browse/KIALI-3135

There are no visual changes.

### DIFF
--- a/src/components/CytoscapeGraph/CytoscapeGraphUtils.ts
+++ b/src/components/CytoscapeGraph/CytoscapeGraphUtils.ts
@@ -1,6 +1,6 @@
 import { Layout } from '../../types/GraphFilter';
 import * as LayoutDictionary from './graphs/LayoutDictionary';
-import { CytoscapeGlobalScratchNamespace } from '../../types/Graph';
+import { CytoscapeGlobalScratchNamespace, DecoratedGraphEdgeData, DecoratedGraphNodeData } from '../../types/Graph';
 import { DagreGraph } from './graphs/DagreGraph';
 
 export const CyEdge = {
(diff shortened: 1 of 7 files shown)


src/components/CytoscapeGraph/CytoscapeGraphUtils.ts
```diff
@@ -88,3 +88,11 @@ export const runLayout = (cy: any, layout: Layout) => {
 
   cy.scratch(CytoscapeGlobalScratchNamespace).showNodeLabels = showNodeLabels;
 };
+
+export const edgeData = (ele: any): DecoratedGraphEdgeData => {
```
mtho11: Since this PR is about type safety: maybe give `ele` a type. 
Trivial though, if you don't want to that is also fine.

cfcosta: `unknown` might be better suited here, or `<T extends { data: () -> unknown }>`.

cfcosta: Thinking a little more about it:
* The `DecoratedGraph*Data` types seem equal, even though some of the names are different. Maybe they should be generalized in a further PR?
* Since they are interfaces, we can check the "capabilities", like `has http` or `has grpc`, and then we can just receive the capabilities as extends on the generics.

josejulio: Yes, more work can be made in that direction. I think that currently we need a way to ensure if what we are receiving (from the server) is really what we are expecting, else this added type-safety will be just an illusion.

PR #1331, opened 2019-07-19
Frontend part for adding Api Documentation to the services

couniojc:
** Describe the change **

This is the frontend part of the API Documentation feature. It optionally adds one column in the ServiceList and an 'API Doc' tab in the ServiceDetails pages

** Issue reference **

https://github.com/kiali/kiali/issues/1237

** Backwards compatible? **

[X] Is your pull-request introducing changes in behaviour?

It adds new column and new tab if annotations are set on the services

** Screenshot **

Screen Shot 2019-07-19 at 3 41 29 PM
Screen Shot 2019-07-19 at 3 40 59 PM

** Documentation **

https://github.com/kiali/kiali/pull/1266/commits/94a0441f7976b7039f1d45dd0f941bba3a173876


rhqci:

package.json
```diff
@@ -54,7 +54,7 @@
     "start:dev": "npm run remove-rcue; sh -ac 'source ./.env.upstream; npm run start:kiali'",
     "start:kiali": "npm run copy-fonts && npm run copy-img && npm run build-css && REACT_APP_VERSION=$npm_package_version REACT_APP_NAME=$npm_package_name REACT_APP_GIT_HASH=$(git rev-parse HEAD) react-scripts start",
     "start:prod": "npm run restore-rcue; sh -ac 'source ./.env.downstream; npm run start:kiali'",
-    "test": "npm run build-css && TEST_RUNNER=1 react-scripts test --env=jsdom __tests__",
+    "test": "npm run build-css && TEST_RUNNER=1 craco test --env=jsdom __tests__",
```
cfcosta: Why this change?

src/components/ApiDocumentation/ApiTypeIndicator.tsx
```diff
@@ -0,0 +1,46 @@
...
+    return style({
+      marginTop: -2,
+      marginRight: 6,
+      width: 30
+    });
+  }
+
+  render() {
+    if (this.props.apiType) {
+        if (this.props.apiType === 'grpc' || this.props.apiType === 'graphql' || this.props.apiType === 'rest') {
```
cfcosta: This should be part of `renderIcon`.

src/components/ApiDocumentation/__tests__/ApiTypeIndicator.test.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+import { ApiTypeIndicator } from '../ApiTypeIndicator';
+
+describe('ApiTypeIndicator', () => {
+  it('renders when not known', () => {
+    let wrapper = shallow(<ApiTypeIndicator apiType="unknown" />);
+    expect(wrapper.html()).not.toContain('img');
+  });
+
+  it('renders when known', () => {
```
cfcosta: The test names do not explicit what is being under test.

src/pages/ServiceDetails/ServiceDetailsPage.tsx
```diff
@@ -352,6 +357,9 @@ class ServiceDetails extends React.Component<ServiceDetailsProps, ServiceDetails
                     </>
                   </NavItem>
                 ))}
+              {this.state.serviceDetailsInfo.apiDocumentation.baseUrl && (
```
cfcosta: What happens here if `apiDocumentation` is null or the base url is `''`? And also, why does it get it from the state instead of using the proper props from Redux?

src/pages/ServiceList/ItemDescription.tsx
```diff
@@ -6,20 +6,23 @@ import { DisplayMode, HealthIndicator } from '../../components/Health/HealthIndi
...
+  columnWidth: number;
 }
 
 export default class ItemDescription extends React.PureComponent<Props, State> {
   private promises = new PromisesRegistry();
 
   constructor(props: Props) {
     super(props);
-    this.state = { health: undefined };
+    this.state = { health: undefined, columnWidth: props.hasApiColumn?3:4 };
```
cfcosta: I don't really like the magic numbers here, specially without proper padding.

src/pages/ServiceList/ItemDescription.tsx
```diff
@@ -51,22 +54,25 @@ export default class ItemDescription extends React.PureComponent<Props, State> {
...
+        <Col xs={12} sm={12} md={this.state.columnWidth} lg={this.state.columnWidth}>
           <strong>Config: </strong>{' '}
           <ConfigIndicator
             id={this.props.item.name + '-config-validation'}
             validations={[this.props.item.validation]}
             size="medium"
           />
         </Col>
-        <Col xs={12} sm={12} md={4} lg={4} />
+        <Col xs={12} sm={12} md={this.state.columnWidth} lg={this.state.columnWidth}>
```
cfcosta: Does it help about the width if the code just hide the whole `Col` if it's  not present?

src/pages/ServiceList/ItemDescription.tsx
```diff
@@ -51,22 +54,25 @@ export default class ItemDescription extends React.PureComponent<Props, State> {
...
+          {this.props.item.apiType && <strong>Api: </strong>}
+          {this.props.item.apiType && <ApiTypeIndicator apiType={this.props.item.apiType} />}
```
cfcosta: And here both could be only one conditional (hopefully on the top element level).

src/pages/ServiceList/ServiceListComponent.tsx
```diff
@@ -188,6 +189,14 @@ class ServiceListComponent extends ListComponent.Component<
     const pageStart = (this.state.pagination.page - 1) * this.state.pagination.perPage;
     let pageEnd = pageStart + this.state.pagination.perPage;
     pageEnd = pageEnd < this.state.listItems.length ? pageEnd : this.state.listItems.length;
+    let hasApiColumn = false;
+    for (let i = pageStart; i < pageEnd; i++) {
```
cfcosta: `hasApiColumn = this.state.listItems.slice(pageStart, pageEnd).some(item => item.apiType)`, maybe?

package.json
couniojc: We are using (at least for this version) a webcomponent. When running the tests, the webcomponent is imported and called and it leads to a "customElements is not defined".
Using craco helps to override the import of the web element so tests can run.
Any alternate suggestion is welcomed

src/components/ApiDocumentation/ApiTypeIndicator.tsx
couniojc: I will move it

src/components/ApiDocumentation/__tests__/ApiTypeIndicator.test.tsx
couniojc: I will change it

package.json
cfcosta: I think we do something like that for `SVGPathElement`, or `SVGElement` (not at the computer right now to look).

A reference that might help (it did when I found this problem): https://github.com/jsdom/jsdom/issues/1423

src/pages/ServiceList/ItemDescription.tsx
couniojc: Actually there's already an empty column that doesn't show, but if something is in there,  then the total width is 16 and it breaks the columns.
To avoid to change the design for people not using the API functionality, I changed the columns to 3 3 3 3 instead of 4 4 4 (empty one being ignored). I'm opened to suggestions

src/pages/ServiceList/ItemDescription.tsx
couniojc: I tried with one conditional, but i had an error saying I needed one parent. Can you suggest something in the code ? I know it's not great as it is

src/pages/ServiceList/ServiceListComponent.tsx
couniojc: Wooo, I will try that :-) Thanks

src/pages/ServiceList/ItemDescription.tsx
cfcosta: You can wrap the elements like this: `<>Element here</>`.

src/pages/ServiceDetails/ServiceDetailsPage.tsx
couniojc: If apiDocumentation is not defined the tab is not displayed (it is probably displayed if baseUrl is '' but that shouldn't happen)

cfcosta: If we're checking for presence on those maybe a good idea would be just to tag it as nullable, and then check on the whole object.

src/pages/ServiceList/ItemDescription.tsx
couniojc: I will change to nbColumns and make the computing of size in the display part

package.json
couniojc: Thanks, I was able to do the same so I will remove the craco override

src/pages/ServiceDetails/ServiceDetailsPage.tsx
couniojc: Ok, I will add a check on the full object

couniojc:
@cfcosta Can you please have a look to the latest changes ? Thanks !

src/components/ApiDocumentation/ApiTypeIndicator.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  render() {
+    if (this.props.apiType) {
+      return this.renderIcon(this.props.apiType);
+    }
+    return <span />;
+  }
+
+  renderIcon(apiType: string) {
+    if (this.props.apiType === 'grpc' || this.props.apiType === 'graphql' || this.props.apiType === 'rest') {
```
cfcosta: This is somewhat better, and given that we have the `nameToSource` to see if there's an icon, maybe we should check if the key is there instead of enumerating all the possibilities here too?

src/pages/ServiceDetails/ServiceDetailsPage.tsx
```diff
@@ -352,6 +357,9 @@ class ServiceDetails extends React.Component<ServiceDetailsProps, ServiceDetails
                     </>
                   </NavItem>
                 ))}
+              { this.state.serviceDetailsInfo.apiDocumentation && this.state.serviceDetailsInfo.apiDocumentation.baseUrl && (
```
cfcosta: baseUrl should be not `''` and not `undefined`, empty strings are not false in javascript, as far as I know, and if they are, that's black magic :P

src/pages/ServiceDetails/ServiceDetailsPage.tsx
```diff
@@ -395,6 +403,14 @@ class ServiceDetails extends React.Component<ServiceDetailsProps, ServiceDetails
                   />
                 </TabPane>
               )}
+              {this.state.serviceDetailsInfo.apiDocumentation && this.state.serviceDetailsInfo.apiDocumentation.baseUrl && (
```
cfcosta: Same here!

src/pages/ServiceList/ItemDescription.tsx
```diff
@@ -49,24 +52,33 @@ export default class ItemDescription extends React.PureComponent<Props, State> {
   }
 
   render() {
+    var columnSize = 12 / this.state.nbColumns;
```
cfcosta: This is a lot more readable than before! :D

src/pages/ServiceList/ServiceListComponent.tsx
```diff
@@ -188,6 +189,7 @@ class ServiceListComponent extends ListComponent.Component<
     const pageStart = (this.state.pagination.page - 1) * this.state.pagination.perPage;
     let pageEnd = pageStart + this.state.pagination.perPage;
     pageEnd = pageEnd < this.state.listItems.length ? pageEnd : this.state.listItems.length;
+    let hasApiColumn = this.state.listItems.slice(pageStart, pageEnd).some(item => item.apiType)
```
cfcosta: Nice that it worked :D

src/components/ApiDocumentation/ApiTypeIndicator.tsx
couniojc: Good point, I fixed it

src/components/ApiDocumentation/ApiTypeIndicator.tsx
```diff
@@ -0,0 +1,48 @@
...
+  iconStyle() {
+    return style({
+      marginTop: -2,
+      marginRight: 6,
+      width: 30
+    });
+  }
+
+  render() {
+    if (this.props.apiType) {
```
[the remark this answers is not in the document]
couniojc: Changed to ternary

src/components/ApiDocumentation/ApiTypeIndicator.tsx
```diff
@@ -0,0 +1,48 @@
...
+const apiGraphqlIcon = require('../../assets/img/api-graphql-logo.svg');
+
+const nameToSource = new Map<string, string>([
+  ['grpc', apiGrpcIcon],
+  ['rest', apiRestIcon],
+  ['graphql', apiGraphqlIcon],
+]);
+
+export class ApiTypeIndicator extends React.Component<Props> {
+  iconStyle() {
```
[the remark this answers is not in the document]
couniojc: Fixed, thanks !

src/components/ApiDocumentation/ApiTypeIndicator.tsx
```diff
@@ -0,0 +1,48 @@
+import * as React from 'react';
+import { style } from 'typestyle';
+
+interface Props {
+  apiType: string;
+}
+
+const apiGrpcIcon = require('../../assets/img/api-grpc-logo.svg');
+const apiRestIcon = require('../../assets/img/api-rest-logo.svg');
+const apiGraphqlIcon = require('../../assets/img/api-graphql-logo.svg');
```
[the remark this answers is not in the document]
couniojc: Fixed, thanks !

src/pages/ServiceDetails/ServiceDetailsPage.tsx
couniojc: Let's fight black magic then, I added !== ''

src/pages/ServiceDetails/ServiceDetailsPage.tsx
couniojc: Fixed as well

couniojc:
@cfcosta Thanks for the review ! Super useful ! I committed latest fixes

src/pages/ServiceList/ItemDescription.tsx
```diff
@@ -49,24 +52,33 @@ export default class ItemDescription extends React.PureComponent<Props, State> {
...
           <strong>Config: </strong>{' '}
           <ConfigIndicator
             id={this.props.item.name + '-config-validation'}
             validations={[this.props.item.validation]}
             size="medium"
           />
         </Col>
-        <Col xs={12} sm={12} md={4} lg={4} />
+        <Col xs={12} sm={12} md={columnSize} lg={columnSize}>
+          // If no api, the column is empty and not displayed
```
jotak: these comments are displayed on screen :)
they should either be moved inside the `{...}` block or commented as html `<!-- -->`

couniojc: Wooo, how come I don't see it ? Let me change it

couniojc: Fixed

jotak:
I'm rebasing this PR and, i haven't seen it coming, the variable number of columns is not playing well with @aljesusg ' refactoring and virtual list. Don't worry we'll get that in.

jotak:
superseded by https://github.com/kiali/kiali-ui/pull/1356

PR #1292, opened 2019-07-01
KIALI-3067 PF4 migration: charts / k-charted

jotak:
JIRA: https://issues.jboss.org/browse/KIALI-3067

Ongoing work in k-charted repo: https://github.com/kiali/k-charted/issues?q=is%3Aissue+label%3APF4+

- [x] https://github.com/kiali/k-charted/pull/22
- [x] https://github.com/kiali/k-charted/issues/17
- [x] https://github.com/kiali/k-charted/issues/18
- [x] https://github.com/kiali/k-charted/issues/20
- [x] https://github.com/kiali/k-charted/issues/24
- [x] https://github.com/kiali/k-charted/issues/25
- [x] https://github.com/kiali/k-charted/issues/35
- [x] https://github.com/kiali/k-charted/issues/37
- [x] Crashing in some conditions: https://github.com/FormidableLabs/victory/issues/1355
- [x] Do some perf comparison: https://github.com/kiali/k-charted/issues/45

PR #1334, opened 2019-07-23
KIALI-3150 Fix blank screen

israel-hdez:
https://issues.jboss.org/browse/KIALI-3150

There is some issue with details pages. Lazy loading all them produces a blank screen bug in the containerized build (dev-mode `yarn start` works fine). It's enough if one of the detail pages is not lazy loaded to have the app working fine.

For now, leaving AppDetailsPage without lazy load to fix the issue. I'll try to find what's the root cause, but doing this to not delay the fix.

### DIFF
--- a/src/routes.tsx
+++ b/src/routes.tsx
@@ -1,4 +1,5 @@
 import { lazy } from 'react';
+import AppDetailsPage from './pages/AppDetails/AppDetailsPage';
 import { MenuItem, Path } from './types/Routes';
 import { icons, Paths } from './config';
 import DefaultSecondaryMasthead from './components/DefaultSecondaryMasthead/DefaultSecondaryMasthead';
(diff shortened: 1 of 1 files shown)


src/routes.tsx
```diff
@@ -104,7 +105,7 @@ const pathRoutes: Path[] = [
   },
   {
     path: '/namespaces/:namespace/' + Paths.APPLICATIONS + '/:app',
-    component: lazy(() => import('./pages/AppDetails/AppDetailsPage'))
+    component: AppDetailsPage
```
mtho11: I don't understand why the details page wouldn't lazy load. That is the page that one would think you would want to lazy load. Must be some shared component somewhere.

israel-hdez: Yes, most likely. That will be somewhat hard to find.

jotak:
FYI it happens again in this PR and I don't know why : https://github.com/kiali/kiali-ui/pull/1292#issuecomment-515387882

I've tried to not lazy-load all details pages (app+workload+service) + overview ... still with no luck

PR #1336, opened 2019-07-24
KIALI-3070 keep metrics settings on refresh + several improvements

jotak:
Various improvements:
- Update model (AllLabelsValues => LabelSettings) to keep more information in some transitional states
- Thanks to the above, performance improvements when labels are selected: no need to fetch metrics several times with updated labels (in IstioMetrics.tsx)
- Refactored component lifecycle of MetricsSettingsDropdown, more typical of the classical react way (props & state; no concistency check performed at render as before)
- Updated Traffic page links (now BY_LABELS param requires prom labels instead of display name)

JIRA: https://issues.jboss.org/browse/KIALI-3070


jotak:
@israel-hdez could you have a look please? I fear it requires more than just a quick look, as I went from small improvement to another and ended up in some bigger refactoring.
So if you could test the parts that you've implemented (e.g. traffic page links to metrics) to make sure I didn't miss anything.

@hhovsepy I think some general tests / no-regression would be nice on metrics pages (settings, group by labels, filtering, reloading from URL etc.) in addition to the original bug fix check.

Thanks!

israel-hdez:
I smoke tested and eveything worked.
Started the code review, but I haven't finished. I'll finish tomorrow.

israel-hdez:
I guess this is minor, or invalid. It depends on how much settings we want to preserve:

The series of the groupings are not preserved:
minorIssue

If we don't want to preserve the selection of the time series, that's OK. But behavior is inconsistent when coming from the traffic tab. Because that tab sets a pre-selection, at refresh the traffic tab pre-selection is preserved:
minorIssue2

I know this is not trivial to make consistent (in case we want to).

jotak:
@israel-hdez thanks for the reviews.
About the "data flow" : as you said, the data is first taken from URL and then augmented with whatever comes from the server. I'm not sure how it can be improved but if you have any idea, please go on :)

About the filters not being all preserved on refresh: I noticed that as well, but this is not a regression, I didn't want to change that behaviour. I'll enter a JIRA for enhancement. All we have to do is write the URL correctly, should be easy.

PR #1337, opened 2019-07-25
KIALI-3142 PoC - Check typesafety on some server responses

josejulio:
** Describe the change **

This is a PoC that aims to show one way to add typesafety on the received data from the server, actually checking the data without blindly believing it.

** Issue reference **

https://issues.jboss.org/browse/KIALI-3142

** Screenshot **

No changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "name": "@kiali/kiali-ui",
   "version": "1.3.0",
+  "proxy": "https://kiali-istio-system.127.0.0.1.nip.io/",
   "description": "React UI for [Kiali](https://github.com/kiali/kiali).",
   "keywords": [
     "istio service mesh",
@@ -75,6 +76,9 @@
     "cytoscape-dagre": "2.2.2",
     "cytoscape-popper": "1.0.2",
     "deep-freeze": "0.0.1",
+    "fp-ts": "^2.0.3",
+    "io-ts": "^2.0.0",
+    "io-ts-excess": "^1.0.1",
     "js-yaml": "3.13.1",
     "json-beautify": "1.0.1",
     "lodash": "4.17.13",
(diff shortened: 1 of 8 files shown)


jotak:
I understand the need but on the other hand, i think it hurts readability. Also we would need to consider the performance impact. We need to find the right trade-off.

Personally, although I've seen several times in the past some model mismatch between our go code & typescript, last time I complained about that was like maybe 1 year ago but it's been going pretty well since then.

josejulio:
> I understand the intention but on the other hand, i think it hurts readability.

That's a fair point, and i think I can fix that.

> Also we would need to consider the performance impact. We need to find the right trade-off.

I don't think there is a big performance impact, but we would need to measure that.

> Personally, although I've seen several times in the past some model mismatch between our go code & typescript, last time I complained about that was like maybe 1 year ago but it's been going pretty well since then.

It is likely that some issues go unseen, for example, in the graph code, we didn't use any typescript and were validating all by hand (i did some changes lately and found some mismatch between our typescript code), having a strong type could make the code cleaner IMO.

josejulio:
> > I understand the intention but on the other hand, i think it hurts readability.
> 
> That's a fair point, and i think I can fix that.
> 
> >

I though i could provide a workaround to this, but couldn't. 
Even if add manually the typescript interface, there is not a 100% way to ensure that they are in sync.

I'm looking alternatives.

jshaughn:
I'm more in line with @jotak on this topic.  I don't want to see complicated code and more results processing for response validation, which is something that should probably be handled via tests already.  I think the work done to ensure proper field access in the UI is probably good enough as the use of literals is an actual source of many bugs.

josejulio:
> I'm more in line with @jotak on this topic. I don't want to see complicated code and more results processing for response validation, which is something that should probably be handled via tests already. I think the work done to ensure proper field access in the UI is probably good enough as the use of literals is an actual source of many bugs.

I think the main issue here is that there isn't any test that ensure that the server data returned by the server is "in-sync" with our expected objects.

That said, current solution complicates a lot when we have complex types, I haven't found a satisfactory way to avoid it. The only thing i found is [typescript-is](https://www.npmjs.com/package/typescript-is), but currently typescript doesn't support the "transformers" that it uses out of the box which could lead to ejecting to configure (or using something like [craco](https://www.npmjs.com/package/@craco/craco) i think).

Closing.

PR #1339, opened 2019-07-26
KIALI-3150 Fix code splitting issue (undefined components)

israel-hdez:
When introducing code-splitting in Kiali, we are hitting an external bug or an unsupported scenario. The issue is either in PatternFly3 or in WebPack/Babel (or in something else).

For the NotificationDrawer component, PatternFly 3 uses the syntax that can be seen in: https://github.com/patternfly/patternfly-react/blob/patternfly-react%402.31.1/packages/patternfly-3/patternfly-react/src/components/Notification/NotificationDrawer/index.js#L21-L33

The PatternFly code is importing components and assigning them to a property of NotificationDrawer. This breaks the code-splitting mechanism and all those components assigned as properties of NotificationDrawer become undefined.

The issue is fixed by direct importing of those components in Kiali.

### DIFF
--- a/src/components/MessageCenter/NotificationDrawer.tsx
+++ b/src/components/MessageCenter/NotificationDrawer.tsx
@@ -10,6 +10,16 @@ import {
   EmptyStateTitle,
   EmptyStateIcon
 } from 'patternfly-react';
+import PfNotificationDrawerTitle from 'patternfly-react/dist/esm/components/Notification/NotificationDrawer/NotificationDrawerTitle';
+import PfNotificationDrawerAccordion from 'patternfly-react/dist/esm/components/Notification/NotificationDrawer/NotificationDrawerAccordion/NotificationDrawerAccordion';
+import PfNotificationDrawerPanel from 'patternfly-react/dist/esm/components/Notification/NotificationDrawer/NotificationDrawerAccordion/NotificationDrawerPanel';
+import PfNotificationDrawerPanelAction from 'patternfly-react/dist/esm/components/Notification/NotificationDrawer/NotificationDrawerAccordion/NotificationDrawerPanelAction';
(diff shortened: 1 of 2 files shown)


israel-hdez:
I'm unsure about this PR.
Perhaps, may be better to revert and forget the code-splitting.

This is a very weird issue. Even more, there are other PF components that are using a similar code to the offending one and those work OK. For example, Notification.Content:

https://github.com/patternfly/patternfly-react/blob/patternfly-react%402.31.1/packages/patternfly-3/patternfly-react/src/components/Notification/Notification.js#L46-L52

and this component is being used in the same file that I modified in this PR to fix the issue:

https://github.com/kiali/kiali-ui/blob/6a0afc345ebf2a31c7643a865245efcb34bcab45/src/components/MessageCenter/NotificationDrawer.tsx#L83-L84

The _difference_ I see, is that in the PatternFly code the NotificationDrawer is using that syntax in its index.js file, while the Notification component is using that syntax in the Notification.js file and its index.js file is re-exporting normally all sub-components (see https://github.com/patternfly/patternfly-react/blob/patternfly-react%402.31.1/packages/patternfly-3/patternfly-react/src/components/Notification/index.js)

So, I'm OK reverting the code-splitting because this looks too problematic.

israel-hdez:
Even more... The tests are not passing with those imports :/

israel-hdez:
> I wonder if we shouldn't just revert the code splitting PR?

Here is the PR reverting: https://github.com/kiali/kiali-ui/pull/1340

With the test failures, I'm more inclined to reverting. But leaving this PR open in case you find a better approach. You choose the direction :)

jotak:
Thanks @israel-hdez , i've merged the other one

PR #1343, opened 2019-07-30
KIALI-3172 Use require as a workaround in NotificationDrawer

jotak:
JIRA https://issues.jboss.org/browse/KIALI-3172

This should not break tests

As explained inline:
For some reason not fully explained, when loaded with "import" it happens that the NotificationDrawer
does not come with the expected ".Title", ".Accordion" (etc.) fields.
Which ends up in React error "Element type is invalid" as those components are undefined.
Using the "require" way is a workaround.
Note that it is unclear what triggers the error
(may happen with or without lazy loading, generally not seen using `yarn start` but seen with build)

See also https://github.com/kiali/kiali-ui/pull/1339 and https://github.com/kiali/kiali-ui/pull/1292

Also add prop-types as this is a peer dependency of several modules; hence fixes warnings

### DIFF
--- a/package.json
+++ b/package.json
@@ -128,6 +128,7 @@
     "node-sass": "4.12.0",
     "prettier": "1.16.4",
     "pretty-quick": "1.10.0",
+    "prop-types": "15.7.2",
     "redux-mock-store": "1.5.3",
     "snyk": "^1.161.1",
     "source-map-explorer": "1.8.0",
--- a/src/components/MessageCenter/NotificationDrawer.tsx
+++ b/src/components/MessageCenter/NotificationDrawer.tsx
@@ -1,16 +1,6 @@
 import * as React from 'react';
 import { Collapse } from 'react-bootstrap';
 
-import {
-  NotificationDrawer as PfNotificationDrawer,
-  Notification,
-  Icon,
-  Button,
-  EmptyState,
-  EmptyStateTitle,
-  EmptyStateIcon
-} from 'patternfly-react';
-
 import { MessageType, NotificationMessage, NotificationGroup } from '../../types/MessageCenter';
 import moment from 'moment';
 import {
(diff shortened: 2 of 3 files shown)


src/components/MessageCenter/NotificationDrawer.tsx
```diff
@@ -19,6 +9,14 @@ import {
 } from '../BoundingClientAwareComponent/BoundingClientAwareComponent';
 import { style } from 'typestyle';
 
+// KIALI-3172 For some reason not fully explained, when loaded with "import" it happens that the NotificationDrawer
+// does not come with the expected ".Title", ".Accordion" (etc.) fields.
```
jshaughn: Could you import those explicitly?

package.json
```diff
@@ -128,6 +128,7 @@
     "node-sass": "4.12.0",
     "prettier": "1.16.4",
     "pretty-quick": "1.10.0",
+    "prop-types": "^15.6.1",
```
josejulio: Is this required? If so, could you please lock the version?

jotak: Not exactly required, but it's removing some warning about peer dependencies during builds that we had for a long time.
Yes i'm going to lock the version.

src/components/MessageCenter/NotificationDrawer.tsx
jotak: You mean, the composed parts (.Title, .Accordion etc) ?
This is what @israel-hdez did here https://github.com/kiali/kiali-ui/pull/1339
But (as far as I understand) the problem is that it's pointing explicitly to the "esm" output in patternfly, which bypasses any module configuration in tsconfig and makes the tests fail.
I don't think it's a good idea to bypass the tsconfig modules settings.

package.json
jotak: done

src/components/MessageCenter/NotificationDrawer.tsx
israel-hdez: To add a little more detail: importing is not possible, because the _patternfly-react_ library is not individually exporting those components. It makes them available only through the NotificationDrawer.* properties.

jshaughn: OK, makes sense.  You cam only import something that is exported.

PR #1348, opened 2019-07-31
KIALI-3098 Fix icon spacing issue

jshaughn:
Before:
image
After:
image

### DIFF
--- a/src/components/Details/DetailedTrafficList.tsx
+++ b/src/components/Details/DetailedTrafficList.tsx
@@ -169,7 +169,7 @@ class DetailedTrafficList extends React.Component<DetailedTrafficProps> {
   };
 
   private renderWorkloadColumn = (node: TrafficNode, isProxyed: boolean) => {
-    const style = isProxyed ? { paddingLeft: '2em' } : {};
+    const style = isProxyed ? { paddingLeft: '2em', paddingRight: '4px' } : { paddingRight: '4px' };
     let icon = <Icon type="pf" name="unknown" style={style} />;
     let name = <>{node.name}</>;
 


src/components/Details/DetailedTrafficList.tsx
```diff
@@ -169,7 +169,7 @@ class DetailedTrafficList extends React.Component<DetailedTrafficProps> {
   };
 
   private renderWorkloadColumn = (node: TrafficNode, isProxyed: boolean) => {
-    const style = isProxyed ? { paddingLeft: '2em' } : {};
+    const style = isProxyed ? { paddingLeft: '2em', paddingRight: '4px' } : { paddingRight: '4px' };
```
abonas: any reason here for using px vs em? it might not be as responsive.

jshaughn: em is larger than desired.

abonas: OK

israel-hdez: FYI: em units allow floating point values... For next time :)

jshaughn: Did I miss a memo about how we are only supposed to use em?  We use px all over the place.

israel-hdez: You didn't. We haven't agreed on anything.

My personal preference is to use em if possible, because it is relative to the font-size being used in the containing HTML element. So, if at some point PatternFly project decides to change font-sizes again (or if we choose to change it), padding will also change and should proportionally look "ok". 

px is a hard unit. You always get 4px and if font-sizes change, spacing could look bad again.

But, since we haven't agreed on anything (and also because I don't think font sizes will be changed), I'm OK with the PR as is.

jshaughn: @israel-hdez OK, I'll use em going forward

PR #1351, opened 2019-08-01
KIALI-3160 Ignore sidecar and label violations for istio objects

jshaughn:
The [istio] control plane objects should not be expected to conform to the
same expectations of the data plane (mesh) objects.

For objects in the istio namespace (i.e. control plane):
- Remove missing sidecar info from App/Workload/Service lists.
- Remove missing sidecar info from App/Workload/Service details.
- Remove missing sidecar and label violation warnings from Workload detail.

Before:
kiali-3160-before

After:
kiali-3160-after

### DIFF
--- a/src/pages/AppDetails/AppInfo/AppDescription.tsx
+++ b/src/pages/AppDetails/AppInfo/AppDescription.tsx
@@ -58,7 +58,9 @@ class AppDescription extends React.Component<AppDescriptionProps, AppDescription
         <div className="ServiceList-Title">
           <div className="component-label">
             Workload{' '}
-            {!workload.istioSidecar && <MissingSidecar style={{ marginLeft: '10px' }} tooltip={true} text={''} />}
+            {!workload.istioSidecar && (
+              <MissingSidecar namespace={namespace} style={{ marginLeft: '10px' }} tooltip={true} text={''} />
+            )}
           </div>
           <Link to={this.workloadLink(namespace, workload.workloadName)}>{workload.workloadName}</Link>
         </div>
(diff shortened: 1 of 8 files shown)


mwringe:
Does this assume that everything in the control plane namespace is part of the control plane? What happens if we have data plane components running in the control plane namespace?

jshaughn:
> Does this assume that everything in the control plane namespace is part of the control plane?

It does, although this is not the first place in the code to make that assumption.  In my opinion it is a reasonable thing to assume no data-plane objects in the istio namespace.

pilhuhn:
I am with @jshaughn here. Users should not have the ability in a real system to deploy into `istio-system`. Even as the GWs are in reality only Envoys and thus part of the data-plane, they are from the Istio install and should thus be treated as such.

src/pages/AppDetails/AppInfo/AppDescription.tsx
```diff
@@ -58,7 +58,9 @@ class AppDescription extends React.Component<AppDescriptionProps, AppDescription
         <div className="ServiceList-Title">
           <div className="component-label">
             Workload{' '}
-            {!workload.istioSidecar && <MissingSidecar style={{ marginLeft: '10px' }} tooltip={true} text={''} />}
+            {!workload.istioSidecar && (
+              <MissingSidecar namespace={namespace} style={{ marginLeft: '10px' }} tooltip={true} text={''} />
```
mtho11: Why is the tooltip text blank? Otherwise, we don't even need a tooltip right?

jshaughn: I have no idea, it could probably be removed but it's not something I changed for this PR so I don't want to touch it.

jshaughn:
@mwringe Are you good with this?  If so please approve unless there is some other issue.

mtho11:
I see no jira link in the PR.

jshaughn:
@mtho11 I don't think I've ever linked back to the Jira, just deferring to the Jira ID in the summary.  I'll try to remember to put one one in the descriptions going forward.

PR #1352, opened 2019-08-06
KIALI-1400 Save state between sessions

josejulio:
- Adds: duration, refreshInterval and graph.layout

I added the messageCenter there, but someone else thinks that it could hold obsolete information (or maybe sensitive information?) so that data is restored on login.

https://github.com/kiali/kiali-ui/blob/d6a7fb41efaf523ec1e87f08b53c4c36a56fa354/src/reducers/MessageCenter.ts#L173-L176

Anyone thinks we should save the messagecenter or is OK to just clean it?

** Issue reference **

https://issues.jboss.org/browse/KIALI-1400

** Backwards compatible? **

No

** Screenshot **

No visual changes

### DIFF
--- a/src/store/ConfigStore.ts
+++ b/src/store/ConfigStore.ts
@@ -39,13 +39,19 @@ const namespacePersistFilter = whitelistInputWithInitialState(
   INITIAL_NAMESPACE_STATE
 );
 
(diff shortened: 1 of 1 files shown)


mtho11:
@josejulio I think @mattmahoneyrh is out this week. Assigning @gbaufake instead.

PR #1354, opened 2019-08-08
KIALI-3198 Migrate charts in overview cards to pf4

jotak:
JIRA: https://issues.jboss.org/browse/KIALI-3198

There's:
- sparkline chart
- bars chart

Splitting in dedicated components by the way.

!Note!: there's a regression with PF3 charts: we don't have tooltips here on the sparkline. So it now only represents a trend but we cannot really read values for it. This is because I couldn't find a good way to display tooltips on victory charts especially when the chart is small, because it requires huge padding around the chart to not be cut on borders.

I think we should investigate alternatives, such as a fixed-position tooltip, or debugging/forking victory's voronoi to "clamp" tooltip on borders, or maybe no tooltip but show y-axis values...

### DIFF
--- a/src/pages/Overview/OverviewCardBars.tsx
+++ b/src/pages/Overview/OverviewCardBars.tsx
@@ -1,0 +1,57 @@
+import * as React from 'react';
+import { Chart, ChartBar, ChartStack, ChartAxis } from '@patternfly/react-charts';
+
+import { NamespaceStatus } from './NamespaceInfo';
+import { FAILURE, DEGRADED, HEALTHY } from '../../types/Health';
+
+type Props = {
+  status: NamespaceStatus;
+};
+
+type State = {
+  width: number;
+};
+
+class OverviewCardBars extends React.Component<Props, State> {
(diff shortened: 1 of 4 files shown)


jotak:
Note also it should fix memory leak issue https://github.com/kiali/kiali/issues/1279

hhovsepy:
Verified, the regression testing passed.

josejulio:
> Note also it should fix memory leak issue [kiali/kiali#1279](https://github.com/kiali/kiali/issues/1279)

Should we also migrate other charts where the visibility workaround was used?

jotak:
@josejulio for the metrics page, the workaround is still in place and continues to prevent memory leak.
AFAIK same for the graph charts.
Only the overview page had no workaround in place for that memory leak, so that's where the patch is more urgent.

There's planned task to continue the PF4 migration in other areas.

aljesusg:
This need a rebase @jotak , in fact react-charts is in 4.4.7 in master after the merge of charts metrics

PR #1357, opened 2019-08-09
KIALI-3193 Add support for ServiceMeshPolicy,ServiceMeshRbacConfig

lucasponce:
Counterpart of https://github.com/kiali/kiali/pull/1326


abonas:
How does this influence the calculation of the visualization of masthead indicator for mtls?
cc @xeviknal

lucasponce:
> How does this influence the calculation of the visualization of masthead indicator for mtls?

Backend is now updated, on Maistra environment, it will use Maistra resources, on Istio upstream, it will continue using the Istio resources.

abonas:
> Backend is now updated, on Maistra environment, it will use Maistra resources, on Istio upstream, it will continue using the Istio resources.

so that means that each Kiali instance works only with one tenant because the masthead shows results from one entity. Is that correct?

lucasponce:
> so that means that each Kiali instance works only with one tenant because the masthead shows results from one entity. Is that correct?

Yes. There should be a Kiali instance per control plane, and a control plane per tenant. Now cluster wide resources are moved to namespace based, and Kiali logic will query cluster wide objects looking into the namespaces that a tenant have access.

PR #1360, opened 2019-08-12
KIALI-3213 More type-safety on Virtual Lists

jotak:
- Formalized type Renderer
- Strong link between ResourceType and Renderer
- Type guard on TResource w/ health ("hasHealth"), health not "any" anymore
- Generic on ResourceType, Renderer, VirtualList to avoid use of "any"

JIRA: https://issues.jboss.org/browse/KIALI-3213

### DIFF
--- a/src/components/VirtualList/Config.ts
+++ b/src/components/VirtualList/Config.ts
@@ -1,93 +1,118 @@
 import deepFreeze from 'deep-freeze';
+import { sortable } from '@patternfly/react-table';
+
 import { AppListItem } from '../../types/AppList';
 import { WorkloadListItem } from '../../types/Workload';
 import { ServiceListItem } from '../../types/ServiceList';
-import { sortable } from '@patternfly/react-table';
 import { IstioConfigItem } from '../../types/IstioConfigList';
 import { serverConfig } from '../../config';
+import * as Renderers from './Renderers';
+import { Health } from '../../types/Health';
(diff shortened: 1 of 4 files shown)


src/components/VirtualList/Config.ts
```diff
@@ -111,7 +136,7 @@ export const IstioTypes = {
 
 export type Resource = {
   name: string;
-  columns: ResourceType[];
+  columns: ResourceType<any>[];
```
jotak: Note: I tried, but aborted propagation of generic as `Resource<R>` because it's generating too many changes and in some cases leading to very noisy definitions.

josejulio: Would the code benefit from using `unknown` instead of `any`? 
You could use typeguards (or a explicit cast) when you want to access the data.
This way we ensure that you have to be explicit on what you expect it to be when you access the data.

jotak: thanks i'll try that tomorrow

jotak: @josejulio I couldn't do it. Using just `ResourceType<unknown>` is not possible as ResourceType needs type that extends `TResource` ; which is ok with `any` but not with `unknown`.

josejulio: Ok, thanks for trying!

PR #1356, opened 2019-08-09
 Frontend part for adding Api Documentation to the services [rebased]

jotak:
This is #1331 PR squashed & rebased from @couniojc , with an extra commit.

Fixing issues during rebase after @aljesusg 's work on Virtual Lists.
The API spec icon is now not displayed in a separate column but put inside the Details column.

Other modified files are only about code style changes by prettier & fixing sort of Details column.

PR #1362, opened 2019-08-12
KIALI-3217 Fix regression in MissingSidecar

jshaughn:
restores code presumably lost in rebase.


jotak:
@jshaughn my bad .. i moved that logic at another place and didn't notice that MissingSidecar was used other than in list views (see https://github.com/kiali/kiali-ui/pull/1356/commits/fa577f42e8949eb827c9bd004e32343d9d2301cd#diff-8833190642aa6841dbc86d95c281d8b4R43 )
Sorry about that

jotak:
BTW the problem with that MissingSidecar component is that it can take any kind/number of props ... so I removed "namespace" props and expected to get compile errors whenever it would be used with "namespace", but it didn't happen :-(

PR #1366, opened 2019-08-13
Use custom registry, if specified

israel-hdez:
Adds the possibility to use a custom package registry.

### DIFF
--- a/Makefile.jenkins
+++ b/Makefile.jenkins
@@ -88,6 +88,11 @@ endif
 
 ui-build:
 	rm -fr build/*
+	# If NPM_CONFIG_REGISTRY is set, change yarn.lock to use that mirror.
+	if [ -n "$$NPM_CONFIG_REGISTRY" ]; then \
+	  sed -i -e "s#https://registry.yarnpkg.com/#$$NPM_CONFIG_REGISTRY/#g" yarn.lock; \
+	  sed -i -e "s#https://registry.npmjs.org/#$$NPM_CONFIG_REGISTRY/#g" yarn.lock; \
+	fi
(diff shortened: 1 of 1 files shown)


Makefile.jenkins
```diff
@@ -88,6 +88,11 @@ endif
 
 ui-build:
 	rm -fr build/*
+	# If NPM_CONFIG_REGISTRY is set, change yarn.lock to use that mirror.
+	if [ -n "$$NPM_CONFIG_REGISTRY" ]; then \
+	  sed -i -e "s#https://registry.yarnpkg.com/#$$NPM_CONFIG_REGISTRY/#g" yarn.lock; \
+	  sed -i -e "s#https://registry.npmjs.org/#$$NPM_CONFIG_REGISTRY/#g" yarn.lock; \
```
josejulio: Do we have `npmjs` links in `yarn.lock` ?

israel-hdez: Nope. That is just in case :)

israel-hdez:
@josejulio @mtho11 Yes, the Jenkins job needs an additional parameter.
See the complimetary PR: https://github.com/kiali/kiali/pull/1336

After these PRs are merged, I'll change the config and it's done.

israel-hdez:
@mtho11 That's it, indeed.
Just setting the NPM_CONFIG_REGISTRY envvar will make both npm and yarn commands to use the specified registry. The Jenkins box will be setting that envvar.

The exception is when the yarn.lock file specifies a registry, then that overrides the envvar. So, this PR changes the lock file to properly use the custom registry.

PR #1367, opened 2019-08-14
Remove duplicated script "fetch-k-charted-pf4.sh"

jotak:
Keep only fetch-k-charted.sh and make it work for pf4


jotak:
@hhovsepy I've just noticed I had this script duplicated as it was introduced in two different PRs, one with pf3 and one with pf4. So this is removing the duplicate and keeping only the one for pf4.

PR #1369, opened 2019-08-15
KIALI-3230 Init screen to not use PF3

israel-hdez:
Init screen was using the LoginPage markup of PF3, which made sense when PF3 was being used since the page was "upgrading" to the Init to Login which looked quite similar.

With PF4 migrations, Login and Init screen designs are different, and it no longer looks as a smooth transition.

Since browser's default background color is usually white, this is moving the Init screen to a more simplistic design which uses almost no PatternFly, leading to less "flickering". The only PatternFly component being used now is the Alert, which is a component of PF4.

While loading:

image

When there is an error:

image


israel-hdez:
Merging. Thanks @mtho11 !

PR #1371, opened 2019-08-16
KIALI-2824 Update React and Redux

cfcosta:
Update the version for React, Redux and the related dependencies to it.

### Status

Build is ok, tests are failing.


src/components/DebugInformation/DebugInformation.tsx
```diff
@@ -154,7 +154,7 @@ const DebugInformationContainer = connect(
   mapStateToProps,
   null,
   null,
-  { withRef: true }
+  { forwardRef: true }
```
mtho11: Do you have any links that document this change?

src/components/MissingSidecar/MissingSidecar.tsx
```diff
@@ -3,11 +3,22 @@ import PropTypes from 'prop-types';
 import { Icon, OverlayTrigger, Tooltip } from 'patternfly-react';
 import { icons, serverConfig } from '../../config';
 
-const MissingSidecar = props => {
-  const { style, text, textTooltip, type, name, namespace, color, tooltip, ...otherProps } = props;
+type MissingSidecarProps = {
```
mtho11: Nice!

mtho11:
@mattmahoneyrh are you testing this one?

src/components/DebugInformation/DebugInformation.tsx
josejulio: https://github.com/reduxjs/react-redux/releases/tag/v6.0.0

> Breaking Changes
The withRef option to connect has been replaced with forwardRef. If {forwardRef : true} has been passed to connect, adding a ref to the connected wrapper component will actually return the instance of the wrapped component.

mtho11:
Ok good @israel-hdez it's confirmed it is not just me.

cfcosta:
> I'm hitting the "getWrappedInstance is not a function" error in the graph and also in the "View debug info" dialog.
> 

Same, I found this and am trying to fix.

cfcosta:
@mtho11 @israel-hdez errors are fixed.

PR #1372, opened 2019-08-20
KIALI-3240 Allow graph compression on Graph Hide

jshaughn:
- https://github.com/kiali/kiali/issues/1342 (related)

kiali-3240


jmazzitelli:
obviously, this needs some work due to the breaking of the animation (as seen in your animated gif)

josejulio:
> breaking of the animation

It is a new feature.

jshaughn:
@jmazzitelli Yes, this is a draft PR.  But, actually I think I've fixed the animation issue.  I've updated the video.  The video shows a more minor rendering issue that remains (you may not even notice...)

jshaughn:
> One minor comment is that the compress/expand button does run a layout when the selector is empty 

@josejulio good catch, I fixed it.

PR #1374, opened 2019-08-21
[KIALI-3245] Use the same component for refresh button in graph page

aljesusg:
Jira: [KIALI-3245](https://issues.jboss.org/browse/KIALI-3245)

Use the same component for refresh button in graphPage.

This will avoid having to change several parts


hhovsepy:
Verified

PR #1391, opened 2019-08-30
KIALI-3240 UX change to replace icon toggle with display option

jshaughn:
Note that in addition to changing from an icon toggle to a display option, I've changed the default.  Now the graph will compress by default.  Feedback is that despite re-positioning the smaller graph, or graph without gaps, is typically preferable.

Before:
image

After:
kiali-3240-3


src/store/Store.ts
```diff
@@ -39,6 +39,7 @@ export interface GraphFilterState {
   hideValue: string;
   showFindHelp: boolean;
   // Toggle props
+  compressOnHide: boolean;
```
mtho11: Wait there is one test failure that I can reproduce.

mattmahoneyrh:
Back-testing - looks good.

Opened bug: https://github.com/kiali/kiali/issues/1745

PR #1398, opened 2019-09-05
Dissociate grafana info fetching from login sequence

jotak:
So that login isn't slowed down when grafana doesn't answer quickly.
Removing all redux-based grafana state, replaced with a simple static promise.

Closes https://github.com/kiali/kiali/issues/1562


jotak:
Opening as draft, I still need to test several error paths

jotak:
@israel-hdez @mtho11 @josejulio are you ok with this removal of redux here?
I'm doing this because the redux state looks like an overkill to me here, logic was dispatched in 10 different files whereas the GrafanaState is being used only in one place.

israel-hdez:
@jotak Coming here just to say that I'm :+1: about reducing Redux usage only when neccessary :)
Clever move to using a static field in the IstioMetrics component.

mtho11:
@jotak yep. agreed.

PR #1405, opened 2019-09-06
Replace use of OS label styling in side panel with PF4 badge styling

jshaughn:
- The node info does not reflect the actual set of OS badges on the k8s object
  and so OS badge styling was confusing.  Simple Badge styling is still
  visually striking while not being misleading.

Fixes https://github.com/kiali/kiali/issues/1392

@serenamarie125 Because we are not actually displaying OS labels I've used a different styling, this is using PF4 Badge, what do you think:

Productpage Service Node:
image

Productpage Versioned App Node:
image

The connecting Edge:
image

### DIFF
--- a/src/pages/Graph/SummaryPanelCommon.tsx
+++ b/src/pages/Graph/SummaryPanelCommon.tsx
@@ -9,9 +9,9 @@ import * as M from '../../types/Metrics';
 import graphUtils from '../../utils/Graphing';
 import { Metric } from '../../types/Metrics';
 import { Response } from '../../services/Api';
-import Label from '../../components/Label/Label';
 import { serverConfig } from '../../config/ServerConfig';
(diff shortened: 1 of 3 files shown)


abonas:
@jshaughn version is a regular label afaik, no?
Namespace indeed isn't. 
So showing them with the same style is IMO confusing if what I wrote above is correct.

jshaughn:
> version is a regular label afaik, no?

Version is a label but I think the point here is that we don't have the full label set to display.  What we do have is information supplied by telemetry (whether it comes from a label or not is not really known).  I don't think we need the label set in the node summary, it makes sense in the detail.  In the summary I think we want to show the node info that we have:  the node type, the name, the namespace, and potentially a version for an versioned-app node.   This different look is to hopefully let the user know that we are not displayed labels.

jshaughn:
@abonas Are you OK with merge?  @jmazzitelli comments that the change solves his original issue.

PR #1415, opened 2019-09-12
PF3->PF4 Migrate Suspend Wizard

lucasponce:
Refactor of SuspendTraffic Wizard to use only PF4 components.
https://github.com/kiali/kiali/issues/1595

This is the easy one, as all components used have a 1to1 mapping with PF4:

image


lucasponce:
@aljesusg comments addressed

PR #1416, opened 2019-09-12
Auto-enable options as needed by the desired graph find/hide expression

jshaughn:
- generate an informational message when making the change, so the user
  is aware that the state has changed.

Auto-enabled options:
response time edge labels to support 'rt'/'responsetime' expression
security display option to support 'mtls' expression
unused nodes display option to 'unused' expression

Closes https://github.com/kiali/kiali/issues/1211

kiali#1295


mattmahoneyrh:
Back-testing - looks good.

PR #1417, opened 2019-09-12
Fix app counter in the side panel and add versions counter

israel-hdez:
Each version was being counted as an app, which was inconsistent with the overview page and the Apps list.

This fixes the app counter and, as suggested, a version counter is also added.

Fixes kiali/kiali#1490

image

### DIFF
--- a/src/pages/Graph/SummaryPanelGraph.tsx
+++ b/src/pages/Graph/SummaryPanelGraph.tsx
@@ -88,7 +88,7 @@ export default class SummaryPanelGraph extends React.Component<SummaryPanelPropT
 
     const numSvc = cy.$(`node[nodeType = "${NodeType.SERVICE}"]`).size();
     const numWorkloads = cy.$(`node[nodeType = "${NodeType.WORKLOAD}"]`).size();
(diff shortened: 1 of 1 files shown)


src/pages/Graph/SummaryPanelGraph.tsx
```diff
@@ -88,7 +88,7 @@ export default class SummaryPanelGraph extends React.Component<SummaryPanelPropT
 
     const numSvc = cy.$(`node[nodeType = "${NodeType.SERVICE}"]`).size();
     const numWorkloads = cy.$(`node[nodeType = "${NodeType.WORKLOAD}"]`).size();
-    const numApps = cy.$(`node[nodeType = "${NodeType.APP}"][!isGroup]`).size();
+    const { numApps, numVersions } = this.countApps(cy);
```
jshaughn: And suddenly I realize that javascript functions can return multiple args :)

israel-hdez: Technically, it can only return one arg -- in this case, an object. You "unwrap" that object with this syntax.

But, honestly, I also prefer to think that it's returning multiple args :)

israel-hdez:
I'll merge. It's a simple change and I think one review is enough.

PR #1421, opened 2019-09-17
Fix zoom buttons on the graph

cfcosta:
In Redux 6.0 `getWrappedInstance()` got removed, so it is always false,
and always sets the ref as null, so broken buttons.

Reference:

https://medium.com/@mehran.khan/using-refs-with-react-redux-6-how-to-use-refs-on-connected-components-4b80d4ea7300

Fixes https://github.com/kiali/kiali/issues/1661


cfcosta:
Also fixed a fun one:

josejulio:
> Also fixed a fun one:
> 
> 

`This zoom button is not working, I'm going to report it, lets attach the debug info..."

mtho11:
Since this is a big enough change and applicable elsewhere, it should probably deserve a Kiali-dev list callout.

PR #1429, opened 2019-09-18
PF3->PF4 Migrate Matching Routing Wizard

lucasponce:
Note, due problems with react-table and react-scripts v3 some ts validations are disabled.

Related:
https://github.com/kiali/kiali/issues/1595

cc @mtho11 this PR contains your suggested hack to disable the problems of react-table module.
UPDATE: we have discover the root cause: a bad import trigger the compilation of the react-table, so fixing imports fixed the problem and no need to modify tsconfig.json.


tsconfig.json
```diff
@@ -15,9 +15,9 @@
     "noImplicitReturns": true,
     "noImplicitThis": true,
     "noImplicitAny": false,
-    "strictNullChecks": true,
+    "strictNullChecks": false,
```
jotak: :-(
So there's no better solution? It's a serious regression in code quality enforcement.

Is there a way to deport this null-check / unused params checks rules to the linter so that we still check our own code and we aren't impacted by failures in 3rd party code?

lucasponce: I'm open for suggestions. But this is the only way to workaround the blocker.

lucasponce:
@hhovsepy okey, yes, if you try to type in the matching it will enable the add rule again, but yes, the "remove" action should reset the validation state.
I'll fix that.

lucasponce:
@hhovsepy updated your comments, let me know if you see anything else

@mtho11 @aljesusg @jotak reverted the hack with the checkers, finally it seems that the main problem was that the ide introduced a wrong dependency we didn't see in a first investigation.

tsconfig.json
cfcosta: Can't we just use `any` on the input params?

lucasponce: This change is already reverted.
So, this PR doesn't modify the tsconfig.json

src/pages/Graph/GraphPage.tsx
```diff
@@ -158,6 +158,7 @@ export class GraphPage extends React.Component<GraphPageProps, GraphPageState> {
     const workload = props.match.params.workload;
     const workloadOk = workload && workload !== UNKNOWN && workload !== 'undefined';
     if (!appOk && !namespaceOk && !serviceOk && !workloadOk) {
+      // @ts-ignore
```
jotak: I'm curious why this ts-ignore is necessary, whereas there's no change in tsconfig.json ?
If unsure, maybe after that PR we should have a global "ts-ignore-cleaning" task, wdyt?

lucasponce: +1 to have a followup tasks to review the "ts-ignore", perhaps we can clean the code in more places.
I guess this one was suggested by @mtho11 in the context to resolve the blocker we have.

I guess we can consolidate this and clean the ts-ignore in single / separted task.

PR #1432, opened 2019-09-19
Use Grafana Links in custom dashboards

jotak:
New GrafanaLinks component used in both IstioMetrics and CustomMetrics pages

WIP: needs https://github.com/kiali/k-charted/pull/50 + k-charted bump
Backend PR: https://github.com/kiali/kiali/pull/1691
     
Part of kiali/kiali#1676

### DIFF
--- a/src/components/Metrics/CustomMetrics.tsx
+++ b/src/components/Metrics/CustomMetrics.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import { connect } from 'react-redux';
 import { RouteComponentProps, withRouter } from 'react-router';
 import { Toolbar, ToolbarGroup, ToolbarItem } from '@patternfly/react-core';
-import { Dashboard, DashboardModel, DashboardQuery, Aggregator } from '@kiali/k-charted-pf4';
+import { Dashboard, DashboardModel, DashboardQuery, Aggregator, ExternalLink } from '@kiali/k-charted-pf4';
 
 import { serverConfig } from '../../config/ServerConfig';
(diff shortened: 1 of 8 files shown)


src/components/Metrics/CustomMetrics.tsx
```diff
@@ -67,13 +70,12 @@ export class CustomMetrics extends React.Component<CustomMetricsProps, MetricsSt
         const labelsSettings = MetricsHelper.extractLabelsSettings(response.data, this.state.labelsSettings);
         this.setState({
           dashboard: response.data,
-          labelsSettings: labelsSettings
+          labelsSettings: labelsSettings,
+          grafanaLinks: GrafanaLinks.buildGrafanaLinks(response.data.externalLinks, this.props.namespace, this.props.app, MetricsObjectTypes.APP, this.props.version)
```
xeviknal: Perhaps I miss something, but I don't understand why you build the grafanaLink outside of the grafana component. Shouldn't be the `response.data.externalLinks` the data stored in the state and the GrafanaLinks component apply internally the modifications to those externaLinks (aka calling `buildGrafanaLinks`)?

jotak: you're probably right :)
I guess I wanted to avoid having many props sent to that component, but probably it makes more sense as you suggest. I'm checking that.

jotak:
@xeviknal PR updated, much better like that, thanks

xeviknal:
@jotak something went wrong with the tests :/

PR #1435, opened 2019-09-23
Styling: don't break line on pods aggregation combo

jotak:
(css display flex)

Closes https://github.com/kiali/kiali/issues/1701

Capture d’écran de 2019-09-23 11-01-41


hhovsepy:
Verified

PR #1439, opened 2019-09-24
PF3->PF4 Move Weighted Routing content

lucasponce:
Maintained bootstrap-slider as a single dependency but removing PF3 components.
Related kiali/kiali#1595


lucasponce:
@hhovsepy I've sent a commit to unify the styles in the weighted routing and suspended wizard's tables. 
Using same ratio, so, then style is unified in all cases.
Let me know if now the style issue is fixed.

PR #1441, opened 2019-09-24
Show notification about warning/error validations for IstioConfig

josejulio:
Fixes #1500

** Screenshot **
Peek 2019-09-24 15-50


hhovsepy:
Screenshot from 2019-09-25 19-23-10

lucasponce:
I'm late to this feature.
I guess this notification when we enter in the details page is ok, but having notification in bunch where we are in the list is not ok.

@josejulio perhaps we can chat about it to refine it, I see you already merged it.

lucasponce:
Or if we want to maintain the warning in the list, I'd put it one single notification, not one per list item.

PR #1442, opened 2019-09-25
Re-enable tooltips in overview sparklines, upgrade k-charted

jotak:
- This is possible by upgrading victory/react-charts and using new prop constrainToVisibleArea
- Also fix an issue with axis on bars charts not being hidden as expected (not sure if this issue was introduced with the lib version upgrade)
- The k-charted upgrade allows the same improvement with tooltips in metrics pages (see https://github.com/kiali/k-charted/pull/54)

Closes https://github.com/kiali/kiali/issues/1702

Capture d’écran de 2019-09-25 14-24-28

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,9 +63,9 @@
     "postinstall": "sed -i -E 's#\\[.*eslint-config-react-app.*\\]#['\\'$PWD/.eslintrc\\'']#g' node_modules/react-scripts/config/webpack*.js"
   },
   "dependencies": {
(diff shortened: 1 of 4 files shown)


jshaughn:
I'm not sure, I hit this when hovering over the sparkline on Overview page, perhaps my build is not correct:

image

jotak:
@jshaughn yes the callback signature changed in latest victory charts, so the libs need to be updated locally.

PR #1447, opened 2019-09-26
[PF4] Migrate Labels to PF4

xeviknal:
Migration of Labels to PF4.
Closes https://github.com/kiali/kiali/issues/1634

### Screenshots:

Workload details:

ServiceDetails:

VirtualService details:

### DIFF
--- a/src/components/Label/Label.scss
+++ b/src/components/Label/Label.scss
@@ -24,6 +24,7 @@
     background-color: #bee1f4;
     color: #00659c;
     padding-top: 4px;
+    border-radius: 2px 0 0 2px;
     padding-bottom: 4px;
     margin-left: 0;
   }
@@ -32,6 +33,7 @@
     background-color: #7dc3e8;
     color: #fff;
     padding-top: 4px;
+    border-radius: 0 2px 2px 0;
     padding-bottom: 4px;
     margin-left: 0;
   }
--- a/src/components/Label/Label.tsx
+++ b/src/components/Label/Label.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { Label as PfLabel } from 'patternfly-react';
+import { Label as PfLabel } from '@patternfly/react-core';
 import './Label.css';
 import { canRender } from '../../utils/SafeRender';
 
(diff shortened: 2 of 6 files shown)


hhovsepy:
I see the old design in graph:
Screenshot from 2019-09-26 13-53-33

xeviknal:
Actually this is not wrong. As mention with others from graph, they said that they aren't actually labels from k8s. They are just information we want to show in the sidebar. 
Then, we agreed that labels in the those labels in the screenshot should be like shown.

jmazzitelli:
> As mention with others from graph, they said that they aren't actually labels from k8s.

Here's my comments I emailed earlier - I'll post here in this PR to keep a record of it here:

===

One thing we need to be clear about here - we should not be showing these using the "kubernetes labels" L&F if what we are showing are NOT actual k8s labels (and in many cases, they are NOT actually coming from k8s label fields - they may have the same key/values, but they aren't really coming from k8s labels.)

For example, many times when you see "app" and "version" in Kiali, they are coming from Prometheus. Although in some/most cases they will be 1-to-1 with the actual k8s labels, sometimes they are NOT (think of the times when people are using the newer format of the k8s labels like app.kubernetes.io/name and app.kubernetes.io/version). Also, we don't show ALL the labels on k8s objects - we normally are only showing app and version. So if you didn't know any better, you would think those are the ONLY k8s labels on their objects when most times that is not true (they will have other labels that we just do not show).

So if we know the data is NOT coming from queries that return actual k8s labels, then we should not visualize them using the PF L&F that OpenShift uses to visualize labels - otherwise, it will confuse people thinking their k8s objects have different labels when they don't.

jotak:
Not sure why, but unlike @hhovsepy I'm seeing the new design in the graph. To me it could be fine but it contradicts what @jmazzitelli says. So, not sure if this is ok or not.

xeviknal:
@jotak you are right. There was a bug in master at the time I create this branch. It was fixed https://github.com/kiali/kiali-ui/pull/1445
If you see the files changed in this PR, graph files haven't been changed.

xeviknal:
@jotak @hhovsepy I've rebased with master. Graph doesn't have this labels anymore. Now this branch is coherent with what @jmazzitelli and @jshaughn said.

PR #1457, opened 2019-10-01
Migrate graph side-panel charts to PF4

jotak:
Closes https://github.com/kiali/kiali/issues/1747


jotak:
Work in progress, but people can start having a look
cc @jshaughn @israel-hdez @josejulio @mtho11

jotak:
It's more or less ready ... yet not fully satisfied with the tooltips, I'll see if it can be improved

jshaughn:
@jotak It is a decent amount of work, thanks for this.  Unfortunately, I think the look and feel is not as nice as what we have currently.  Here are a few things that jump out at me:

1. Just a bug, I think, but colors don't match legend in the node summary:
image

2. The existing response time charts had an interactive legend and the user could disable unwanted percentiles. That is no longer the case (see below where I disabled "avg").

3. The graphs themselves are a bit taller vertically, which forces a more scrolling in the side panel.

4. The popups are not as clean.  They lack legend hints and the font is harder to read. They show only the line being hovered over, and not values for all lines at that time.  Also, if showing multiple values (if lines intersect over the hover spot), the time is repeated.  Unit is also repeated.

Before:
image

After
image

jotak:
> @jotak It is a decent amount of work, thanks for this. Unfortunately, I think the look and feel is not as nice as what we have currently.

Yes, this is all valid points, some can be easily fixed and others not.

> Here are a few things that jump out at me:
> 
>     1. Just a bug, I think, but colors don't match legend in the node summary:
>        image

I didn't see it, is it only for TCP? I had no running example with TCP.

> 
>     2. The existing response time charts had an interactive legend and the user could disable unwanted percentiles. That is no longer the case (see below where I disabled "avg").
> 

Indeed, the legend is not interactive anymore. We must decide how "acceptable" this regression is, given the legend is at the moment entirely handled by the patternfly component (in other words: it comes with the PF4 switch). We should open an issue in PF4 tracker I guess.

>     3. The graphs themselves are a bit taller vertically, which forces a more scrolling in the side panel.

Should be easy to fix.

> 
>     4. The popups are not as clean.  They lack legend hints and the font is harder to read. They show only the line being hovered over, and not values for all lines at that time.  Also, if showing multiple values (if lines intersect over the hover spot), the time is repeated.  Unit is also repeated.
> 
> 
> Before:
> image
> 
> After
> image

Yes that's my main concern currently. With the component provided by victory, tooltips content is not fully customizable, but I guess it's possible to use our own tooltip component built from scratch. Not sure how much work it is.

jotak:
WIP: interactive legend. Lots of headaches today because of that, I hope I can come up with something tomorrow!!

jotak:
I'm getting a design that is closer to the old charts by combining Area and Scatter. It helps to understand which point is being hovered.

Capture d’écran de 2019-10-03 18-13-23

Also, I've setup an interactive legend. When legend items are hovered, the corresponding line is highlighted. At the moment it's the only event I've implemented (mouseover) but if we want that, we could also trigger some behaviours on click. Unlike what we had before, these behaviours don't come out of the box, we need to implement them. But the more we implement, the further we walk away from the PF look and feel.

jotak:
@jshaughn I know there are some differences with PF3 but I think we should clarify, among them, which ones are blocking issues, which are regressions that could be addressed in the near future and which are acceptable differences.

jotak:
For the record, issue opened on PF4 side for built-in interactive legend: https://github.com/patternfly/patternfly-react/issues/3073

jshaughn:
> I know there are some differences with PF3 but I think we should clarify, among them, which ones are blocking issues, which are regressions that could be addressed in the near future and which are acceptable differences.

@jotak I'm not sure that's really my call, it depends on how much we're willing to lose in the name of progress.  I mean, we're doing all of this PF4 work and I can't think of anything that really looks or acts better than it did before, sometimes worse, so it's sort of annoying.  (this is not a criticism of the work, which has been fabulous).  I defer to the dev doing the PF4 work to declare that it's as good as it will get at this time.  If you say that about these charts I'm fine with it, I'll try it again and we can get it in.

jotak:
@jshaughn I understand the concerns (and share some of them, although I'm also seeing some good points :-) specifically for the charts, c3 was probably more mature and featured however it was poorly integrated with react, we had some memory leaks for which we wrote ugly workarounds, we didn't have full control as sometimes i wished ... ),
... so maybe we took that train a bit too early; on the other hand pf4 is growing and evolving, and the problems that we see can certainly be solved.

For the legend I think we can now have the same behaviour as we had with pf3.
For the tooltip, at the moment I'm having more troubles to make it look as nice as the previous one.

jshaughn:
@jotak, some good points, I just worry about the user experience, because in the end that's why we build this stuff :)  It's already fine for a first PF4 version, so just let me know when you are satisfied and I will do a final review.

jotak:
@jshaughn @aljesusg 
From my side I think it is good to review and/or test. I've reintroduced an interactive legend as we had before, with click to show and hide series. The tooltip maybe doesn't look as great, I hope there will be some improvements in victory and/or patternfly (or a custom solution of ours) - but I would say it's reviewable now and the tooltip design regression is quite acceptable.
Let me know what you think

src/components/Charts/SparklineChart.tsx
```diff
@@ -0,0 +1,213 @@
...
+      return <div ref={this.containerRef}>{this.renderChart()}</div>;
+    }
+    return this.renderChart();
+  }
+
+  buildEvents() {
+    return this.props.series.map((_, idx) => {
+      return {
+        childName: [this.props.name + '-legend'],
+        target: ['data', 'labels'],
```
dlabrecq: I believe event targets are strings, while targets for "external event mutations" are string or array types. 

In my POC, I had to build events separately for each target. I was encountering browser console warnings when specifying both targets. I found that hovering over the label works, but not the legend symbol.

jotak: That's weird, it works for me... I mean I get the hover effect when cursor is over the icon or the label. Could it differ depending on victory version? My yarn.lock shows `33.0.6`.

dlabrecq: Yes, same version. Hovering over the labels did work for me, but not both the label and legend symbol -- also the warnings in the browser console.

The shared-events type appears to be a string:
https://github.com/FormidableLabs/victory-core/blob/master/src/victory-shared-events/victory-shared-events.js#L39

If you look at Victory's code, they refer to `event.target === "parent"` instead of working with an array.

Anyway, just thought I would mention it...

PR #1459, opened 2019-10-01
Display Service Entries in traffic tab

israel-hdez:
Adds the ability to show ServiceEntries items in the traffic tab. Links to metric pages will pre-select the destination hostnames listed in the received graph data.

Fixes kiali/kiali#1613.

Only traffic tab of workload and apps are showing the service entries. The Services don't, because services only reach workloads.

This is how it looks:

image


package.json
```diff
@@ -175,5 +175,6 @@
     "last 2 versions",
     "not ie <= 11"
   ],
-  "snyk": true
+  "snyk": true,
```
israel-hdez: I'll revert this once I get dev approvals.

PR #1460, opened 2019-10-02
Fix issue when hide expression hides app box elements

jshaughn:
- can't use elems.jsons() to store element info because it loses styles, and
  we use style 'hidden' for graph hide.

Fixes https://github.com/kiali/kiali/issues/1745

Note that this issue only occurred when "compress hidden' option was disabled.

### DIFF
--- a/src/components/CytoscapeGraph/Layout/GroupCompoundLayout.ts
+++ b/src/components/CytoscapeGraph/Layout/GroupCompoundLayout.ts
@@ -1,11 +1,12 @@
 /*
   GroupCompoundLayout
 
-  This is a synthetic layout that helps to layout close to each other the contents of compound nodes,
-  in this way we ensure that the compound node itself is as small as possible, avoiding overlaps with other nodes.
+  This is a synthetic layout that helps to layout close to each other the contents of compound
+  nodes, in this way we ensure that the compound node itself is as small as possible, avoiding
+  overlaps with other nodes.
 
-  It requires a real layout to do the actual work, but there are some patches applied to the graph before and after the
-  real layout is run.
+  It requires a real layout to do the actual work, but there are some patches applied to the
+  graph before and after the real layout is run.
 
   Is composed of:
    - A compound layout (dagre in this case) does the layout of the children of a compound node.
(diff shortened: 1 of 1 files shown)


src/components/CytoscapeGraph/Layout/GroupCompoundLayout.ts
```diff
@@ -15,15 +16,17 @@
...
+    b. Get the resulting bounding box of the compound, set the width and height of the node
+       using `cy.style`, so that the real layout honors the size when doing the layout.
+    c. For every edge that goes to a child (or comes from a child), create a synthetic edge
+       that goes to (or comes from) the compound node and remove the original
        edge. We can cull away repeated edges as they are not needed.
-    d. Remove the children. This is important, else cytoscape won't honor the size specified in previous step.
-       "A compound parent node does not have independent dimensions (position and size), as those values are
-       automatically inferred by the positions and dimensions of the descendant nodes."
-       http://js.cytoscape.org/#notation/compound-nodes
+    d. Detach the children. This is important, else cytoscape won't honor the size specifie
```
josejulio:
```suggestion
    d. Detach the children. This is important, else cytoscape won't honor the size specified
```

src/components/CytoscapeGraph/Layout/GroupCompoundLayout.ts
```diff
@@ -172,29 +176,24 @@ export default class GroupCompoundLayout {
...
+      child.connectedEdges().each(edge => {
+        // (1.c) Create synthetic edges.
+        const syntheticEdge = this.syntheticEdgeGenerator.getEdge(edge.source(), edge.target());
+        if (syntheticEdge) {
+          syntheticEdges = syntheticEdges.add(this.cy.add(syntheticEdge));
+        }
+      });
+    });
+    // (1.d) Detach all child nodes from parents.
+    children.move('null');
```
josejulio: isn't better to call `remove` ?
```suggestion
    children.remove();
```

src/components/CytoscapeGraph/Layout/GroupCompoundLayout.ts
```diff
@@ -214,8 +213,8 @@ export default class GroupCompoundLayout {
 
       // Add and position the children nodes according to the layout
       parents.each(parent => {
-        // (4.a) Add back the children and the edges
-        this.cy.add(parent.scratch(CHILDREN_KEY));
+        // (4.a) Add back the child nodes (with edges still attached)
+        parent.scratch(CHILDREN_KEY).move(parent);
```
josejulio: If using `remove` above, we can use `restore` directly.
```suggestion
        parent.scratch(CHILDREN_KEY).restore();
```

src/components/CytoscapeGraph/Layout/GroupCompoundLayout.ts
```diff
@@ -172,29 +176,24 @@ export default class GroupCompoundLayout {
...
-            syntheticEdges = syntheticEdges.add(this.cy.add(syntheticEdge));
-          }
-          edges.push(edge);
-          return edges;
-        }, [])
-      );
-    }, []);
-    // (1.d) Remove children and edges that touch a child node.
-    this.cy.remove(this.cy.collection().add(elementsToRemove));
+    children.each(child => {
```
josejulio: I like the way you simplified this code!

src/components/CytoscapeGraph/Layout/GroupCompoundLayout.ts
jshaughn: Why?  I would think this has minimal overhead.

src/components/CytoscapeGraph/Layout/GroupCompoundLayout.ts
jshaughn: I still don't see how remove/restore is better than move (i.e. unset/set parent) ?  Is it just your preference?

src/components/CytoscapeGraph/Layout/GroupCompoundLayout.ts
josejulio: Is the `null` node document somewhere? If not, what is really happening in the background? is it creating a new node? 
I'm afraid that in the future this function would fail if the target node does not exist.

src/components/CytoscapeGraph/Layout/GroupCompoundLayout.ts
jshaughn: fixed

src/components/CytoscapeGraph/Layout/GroupCompoundLayout.ts
jshaughn: no, this is the standard way to move to no parent:  http://js.cytoscape.org/#eles.move :

"Move the elements with respect to graph topology (i.e. new source, target, or parent)."
"parent The ID of the new parent node (use null for no parent)"

josejulio: Ah, I missed that part.

PR #1467, opened 2019-10-07
Compile on test, fix tests

jotak:
Removing the temporary script I created `yarn tsc`, and run compilation on test so that errors will be visible on travis CI.

Fix not-compiling test, also remove some console.log


jotak:
thanks

PR #1470, opened 2019-10-07
Fix Kiosk mode not working

israel-hdez:
Because of the PF4 migration, styles changed a lot and, most likely, PF3 Kiosk mode styles lost effect. The new PF4 styles for Kiosk mode were omitted. This is adding the required styles for Kiosk mode to properly work.

Fixes kiali/kiali#1758.

### DIFF
--- a/src/styles/_overrides.scss
+++ b/src/styles/_overrides.scss
@@ -354,3 +354,16 @@ Inline Alerts: background transparent
 .stack_service_details {
   height: 20%;
 }
+
+/*
+Kiosk mode
+*/
+html.kiosk {
+  #page-sidebar {
+    display: none;
+  }
+
+  header[role='banner'] {
+    display: none;
+  }
+}


jotak:
Code lgtm but I haven't tested. How do we test kiosk mode?

aljesusg:
> Code lgtm but I haven't tested. How do we test kiosk mode?
> 
> EDIT OK, we just need to append `kiosk=true`. I've opened a ticket to document it ( [kiali/kiali#1770](https://github.com/kiali/kiali/issues/1770) )

To test this we should pass this options and be sure that #page-sidebar is hidden, maybe we should add some test here. What do you think @israel-hdez ?

israel-hdez:
@aljesusg Tests are very desirable. The blocker is that this is CSS. The test will need to resolve the final applied CSS on `#page-sidebar`, which our unit tests can't do AFAIK :(

Every now and then, I think a little how to improve testing on these areas. I still don't have a solution... I hope someday to have something.

josejulio:
> @aljesusg Tests are very desirable. The blocker is that this is CSS. The test will need to resolve the final applied CSS on `#page-sidebar`, which our unit tests can't do AFAIK :(

Perhaps we could use a headless chrome? I recall @vnugent did this for rendering some smoke tests, but we could use it for testing if some element has the desired style.

https://developers.google.com/web/updates/2017/04/headless-chrome

israel-hdez:
> Perhaps we could use a headless chrome? I recall @vnugent did this for rendering some smoke tests, but we could use it for testing if some element has the desired style.
> 
> https://developers.google.com/web/updates/2017/04/headless-chrome

I've thought about it. But I guess the full stack is needed.  I'll slowly work on it :)

jotak:
Well I guess we can create a new ticket for extra-work, and this PR can go in, right? (but there's conflicts now ...)

israel-hdez:
@jotak I'm waiting approval from gbaufake.

mattmahoneyrh:
Working with @pbajjuri20 to validate this bug.
All worked as expected except for one area that did not work as expected:

 1) Click on Istio Config
 2) Click on any object
 3) Add '&kiosk=true' to URL results in error:
image

israel-hdez:
@mattmahoneyrh @pbajjuri20 I think I can guess what was the issue.

The Istio Config pages seem to be the only ones that generate a URL without parameters at first access. Like this one when requesting the detail page of the details VirtualService:

`https://192.168.99.109/kiali/console/namespaces/bookinfo/istio/virtualservices/details`

If the URL does not have parameters, then `kiosk=true` will be the only parameter and you, instead, add `?kiosk=true` (note the **?** instead of **&**). The resulting URL looks like like:

`https://192.168.99.109/kiali/console/namespaces/bookinfo/istio/virtualservices/details?kiosk=true`

The image doesn't have the address bar, so this is only a guess.

PR #1472, opened 2019-10-08
[PF4] Jaeger Toolbar expand advanced options component

aljesusg:
Fix: https://github.com/kiali/kiali/issues/1665


aljesusg:
I fixed some styles to be aligned with the iframe

PR #1458, opened 2019-10-01
PF4 the GraphSettings display menu.

mtho11:
** Describe the change **
MIgrate the GraphSettings Display Popover to PF4 using Dropdown.

- [x] Migrate the GraphSettings to PF4

** Issue reference **

Patternfly Issue: https://github.com/patternfly/patternfly-react/issues/3043

fixes: https://github.com/kiali/kiali/issues/1746

** Screenshot **

**Previous(PF3):**
old-popover

**New(PF4):**

new-graph-settings

PR #1474, opened 2019-10-10
Move metrics settings to PF4

jotak:
Based on @mtho11 's work here: https://github.com/kiali/kiali-ui/pull/1458
There's the same odd artifact of having a gray background when mouse comes over.


jotak:
Capture d’écran de 2019-10-10 17-54-29

PR #1475, opened 2019-10-11
Move health indicator popover to PF4

jotak:
Relates to https://github.com/kiali/kiali/issues/1720 (but without mockup implementation)

Please note, there's a small regression, it seems in PF4 popovers cannot be triggered on mouseenter anymore. Also, it creates some divergence with the istio config indicator which uses Tooltip instead of Popover. I also tried Tooltip but it's supposed to be used for small texts - not the case here. Also, text in tooltips are centered, which is not desirable for health. Probably this needs to be harmonised.

Capture d’écran de 2019-10-11 10-08-26


jotak:
For the record: the regression is tracked here: https://github.com/patternfly/patternfly-react/issues/3116

PR #1478, opened 2019-10-11
Fix handling missing validation objecti

jotak:
Closes https://github.com/kiali/kiali/issues/1786

Still not sure why I'm the only one to see this issue, but anyway, here's a patch
Also remove unused function

### DIFF
--- a/src/pages/ServiceDetails/ServiceDetailsPage.tsx
+++ b/src/pages/ServiceDetails/ServiceDetailsPage.tsx
@@ -7,7 +7,7 @@ import ServiceId from '../../types/ServiceId';
 import * as API from '../../services/Api';
 import * as MessageCenter from '../../utils/MessageCenter';
 import { ServiceDetailsInfo } from '../../types/ServiceInfo';
-import { ObjectValidation, Validations } from '../../types/IstioObjects';
+import { Validations } from '../../types/IstioObjects';
 import IstioMetricsContainer from '../../components/Metrics/IstioMetrics';
 import ServiceTraces from './ServiceTraces';
 import ServiceInfo from './ServiceInfo';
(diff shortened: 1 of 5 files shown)


src/pages/ServiceDetails/ServiceDetailsPage.tsx
```diff
@@ -145,24 +145,6 @@ class ServiceDetails extends React.Component<ServiceDetailsProps, ServiceDetails
     return {};
   }
 
-  searchValidation(parsedSearch: ParsedSearch) {
```
xeviknal: I don't know why, it smells bad to me. Perhaps in the refactor was left unused but it is somewhat necessary.
Good test scenarios coverage would be useful. I would like to dig a little bit more on that and perhaps not merged right before today's release. Thoughts?

jotak: Well, I think if i'm having the bug, other could have to. I don't know why it happens to me. Not using maistra? Not having cluster rights? I see on the backend some differentiations related to that.

jotak: @xeviknal about the `searchValidation` being removed: that's unused anyway, I'm not removing any call to it. So that's going to be in the release.

jotak:
Fixed snapshots. Again, I think this should go in **before** the release, not after.

lucasponce:
> Well, I think if i'm having the bug, other could have to, and that could be a blocker. I don't know why it happens to me. Not using maistra? Not having cluster rights? I see on the backend some differentiations related to that.

Can we try to find a way to reproduce this ? I guess it's worth to investigate.

jotak:
@lucasponce a good start would be to know what json you get under `validations` while fetching a service, as said on irc. Because mine is empty and, looking at the code here on the UI, I don't see how it could not crash with an empty validations object. That's why I think in any case this patch is sane & safe, even without knowing yet the root cause ; because in typescript's model that `validations` object is explicitly declared as optional.

src/pages/ServiceDetails/ServiceDetailsPage.tsx
jshaughn: A quick search confirms this is unused.  And uncommented, so +1 to removal.

jotak:
thanks @lucasponce @jshaughn 
About the details: again, I don't know. I'm telling you what my config is (nothing fancy). Looking at the backend code, it seems that there's several paths that can result in an empty object. For instance, when an error occurs during validation. I've added some logs on the backend to see if there are some uncaught errors (and I think indeed errors during validation don't seem to be logged or forwarded to the frontend) but unfortunately I didn't see anything special in my logs.
But again : looking at backend code, it seems to be a valid scenario to have sometimes an empty object.

PR #1484, opened 2019-10-14
List views alignment issue

jshaughn:
This is a tricky issue (at least for me).  The problem occurs because of the width taken by the scroll bar when the virtual list requires scrolling.  Initially I thought it was due to problems with the cell rendering itself, but eventually I discovered the root cause.  A fix is difficult, or possibly just difficult for me given a lack of experience with visual orientation issues.  But because the Virtual table body is rendered independently of the table header, when the body width is reduced the cells compress slightly, but the header positions do not change.

On Chrome (webkit) the problem is solvable with a proprietary setting (overflowY: 'overlay'), but this does not exist on FF or Edge.  This PR contains the Chrome solution but this would still be a known issue for non-chrome browsers.  Perhaps someone else has a good idea, I'll set this to Draft and ask for review/suggestions from the UI heavy hitters...

note, there is also a small change/fix to the "Details' rendering, which may have contributed an additional issue, although I'm not positive it was a problem.

Finally, this updates the package for virtualized extension, because I was hoping it would resolve the issue. This ends up basically updating react.  The update is not required and I'll roll it back if and when this PR goes live.

Partial fix for https://github.com/kiali/kiali/issues/1710.

p.s. There is an unrelated fix to remove an unused and misleading 'style' property from MissingSidecar.
p.s. There is an unrelated fix to the version in package.json.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@kiali/kiali-ui",
-  "version": "1.8.0",
+  "version": "1.7.0",
   "description": "React UI for [Kiali](https://github.com/kiali/kiali).",
   "keywords": [
     "istio service mesh",
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "@patternfly/react-styles": "3.5.16",
     "@patternfly/react-table": "2.17.5",
     "@patternfly/react-tokens": "2.6.21",
-    "@patternfly/react-virtualized-extension": "1.1.97",
+    "@patternfly/react-virtualized-extension": "1.2.55",
     "@webcomponents/custom-elements": "1.2.4",
     "axios": "0.18.1",
     "bootstrap-slider-without-jquery": "10.0.0",
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "@patternfly/react-styles": "3.5.16",
     "@patternfly/react-table": "2.17.5",
     "@patternfly/react-tokens": "2.6.21",
-    "@patternfly/react-virtualized-extension": "1.2.55",
+    "@patternfly/react-virtualized-extension": "1.1.97",
     "@webcomponents/custom-elements": "1.2.4",
     "axios": "0.18.1",
     "bootstrap-slider-without-jquery": "10.0.0",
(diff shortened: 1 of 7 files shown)


mtho11:
Maybe https://modernizr.com/docs/#what-is-modernizr is a better way to detect this feature?

I looked for other css overflow polyfills buf didn't find any.

aljesusg:
We should report this to pf team, I'll open an issue about this

aljesusg:
Yes @mtho11, we can proceed with a workaround, my point is that we have a lot of lines in the css override file because components have different behaviour in Chrome than FF, PF4 team asked us why we import the css in this way (like openshift because we use PF3 & PF4) so maybe this things will be fixed . I think that the problem is in the PF3 css with the PF4 css so we prefer to finish the migration before change the css.

jshaughn:
> We should report this to pf team, I'll open an issue about this

@aljesusg , thanks.  Sorry, I was trying to reduce the PF4 work for you and still it ends up being work for you ;-(

Do you think we should proceed with the Chrome-only workaround until maybe we get a PF fix?  @mtho11 I'm a little nervous about introducing a new dep like modernizr (although it's neat).  With respect to the 'overlay' chrome feature, it' proprietary and it is not going to show up in FF.  The alignment problem is annoying but not critical.

aljesusg:
> > We should report this to pf team, I'll open an issue about this
> 
> @aljesusg , thanks. Sorry, I was trying to reduce the PF4 work for you and still it ends up being work for you ;-(
> 
> Do you think we should proceed with the Chrome-only workaround until maybe we get a PF fix? @mtho11 I'm a little nervous about introducing a new dep like modernizr (although it's neat). With respect to the 'overlay' chrome feature, it' proprietary and it is not going to show up in FF. The alignment problem is annoying but not critical.

No @jshaughn we should merge this if there is a problem about css in FF. My point is about to clean all these patchs after the migration with the PF4 team tips, so go ahead !!!.

mtho11:
@jshaughn can we get some before/after screenshots? It will help people just quickly scanning the issue to easily see the changes.

mtho11:
I still like the **Modernizer** approach to targeting browser _capabilities_ and not a specific browser hack. Browsers are constantly updating and their _capabilities_ constantly changing and some users change slower or faster (think ESR versions and even mobile/tablet versions). IMO browser capabilities are a better way to address cross-browser compatibility than browser types or versions.

jshaughn:
> can we get some before/after screenshots? 

Before (and currently on FF):
image

After (on WebKit browsers: chrome, safari. opera):
image

PR #1485, opened 2019-10-15
[PF4] Move MissingIstioSidecar component

aljesusg:
Fix https://github.com/kiali/kiali/issues/1799

### DIFF
--- a/src/components/MissingSidecar/MissingSidecar.tsx
+++ b/src/components/MissingSidecar/MissingSidecar.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import PropTypes from 'prop-types';
-import { Icon, OverlayTrigger, Tooltip } from 'patternfly-react';
+import { Tooltip, TooltipPosition } from '@patternfly/react-core';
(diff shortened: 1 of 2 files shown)


aljesusg:
is this in the GA release ? 

I don't remember if we changed this to show the missing SItioSidecar error if the service has some workload without istiosidecar or if this is a bug @lucasponce @xeviknal

xeviknal:
I vaguely remember that we decided to remove the MissingSideCar from service pages because sidecars apply only to workloads and application workloads. But again, I don't remember that quite well.

hhovsepy:
> I vaguely remember that we decided to remove the MissingSideCar from service pages because sidecars apply only to workloads and application workloads. But again, I don't remember that quite well.

Right, now I remember that as well.

aljesusg:
> I vaguely remember that we decided to remove the MissingSideCar from service pages because sidecars apply only to workloads and application workloads. But again, I don't remember that quite well.

But we are showing this icon in the servicelist in the column details...maybe we should remove it or indicate that this is because the service has workloads without istioSidecar

PR #1486, opened 2019-10-15
Add Jaeger configuration for namespaceSelector

aljesusg:
Fix: https://github.com/kiali/kiali/issues/1785

Require: https://github.com/kiali/kiali/pull/1801

### DIFF
--- a/src/app/AuthenticationController.tsx
+++ b/src/app/AuthenticationController.tsx
@@ -130,13 +130,14 @@ class AuthenticationController extends React.Component<AuthenticationControllerP
   };
 
   private setJaegerInfo = (jaegerInfo: JaegerInfo) => {
-    let jaegerState: JaegerState = { jaegerURL: '', enableIntegration: false };
+    let jaegerState: JaegerState = { jaegerURL: '', enableIntegration: false, namespaceSelector: true };
 
     if (jaegerInfo.url) {
       jaegerState = {
         jaegerURL: jaegerInfo.url,
         // If same protocol enable integration, otherwise new tab is open
-        enableIntegration: jaegerInfo.url.startsWith(window.location.protocol)
+        enableIntegration: jaegerInfo.url.startsWith(window.location.protocol),
+        namespaceSelector: jaegerInfo.namespace_selector !== undefined ? jaegerInfo.namespace_selector : true
       };
     }
 
--- a/src/app/History.ts
+++ b/src/app/History.ts
@@ -31,6 +31,7 @@ export enum URLParam {
   JAEGER_MAX_DURATION = 'maxDuration',
   JAEGER_MIN_DURATION = 'minDuration',
   JAEGER_SERVICE_SELECTOR = 'service',
+  JAEGER_NAMESPACE_SELECTOR = 'namespace_selector',
   JAEGER_TAGS = 'tags'
 }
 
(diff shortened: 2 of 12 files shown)


hhovsepy:
@aljesusg now those services which required service.namespace (for instance details.bookinfo) does not show traces in Kiali.

aljesusg:
@hhovsepy I did the changes that we talked, can you check them? Thanks

hhovsepy:
In Traces tab of service, it shows traces for bookinfo/details, but does not show for istio-system/istio-ingressgateway it does not. And for Distributed Tracing menu it behaves in the opposite way:
Screenshot from 2019-10-24 15-44-17
Screenshot from 2019-10-24 15-44-38
Screenshot from 2019-10-24 15-45-01
Screenshot from 2019-10-24 15-45-27

aljesusg:
@hhovsepy you should see a checkbox before the lookback drop-down with the namespace-selector option, it seems that the build is not for this PR

hhovsepy:
When Namespace Selector checkbox is selected/deselected the checked Service is reset and "Search Traces" button is disabled. Can the this functionality be removed? So Service selection is not reset.
This is not right from UX perspective.
Also in Service's Traces tab when checking or unchecking Namespace Selector, it disables the "Search Traces" button and keeps disabled forever.
Screenshot from 2019-10-24 18-09-58

xeviknal:
yep, I can agree with @hhovsepy . I think that toggling the checkbox shouldn't reset the dropdown. It is pretty odd to re-select the option. Specially for those who don't know about the feature.
Thinking out lout, but shouldn't this be on the advanced options? I don't think it is something you change many times. wdyt?

aljesusg:
cc @xeviknal @hhovsepy

hhovsepy:
@aljesusg so the namespace_selector config is always true by default?

hhovsepy:
Checking "Namespace Selector" checkbox causes the Traces data to reload, is this the intended behavior?.
But still it uses the checkbox's old (preclick) value in the refresh request. Only after click to "Search Traces" button sends correct request.
Also with smaller windows the view is broken.
Screenshot from 2019-11-13 13-14-59

hhovsepy:
Also in Distributed Tracing section, when first entering without having any namespace selected, then after selecting namespace, the service selection dropdown is still disabled:
Screenshot from 2019-11-13 13-21-59

aljesusg:
> @aljesusg so the namespace_selector config is always true by default?

Yes, the user must set in kiali configmap to false or kiali will set this to true

> Checking "Namespace Selector" checkbox causes the Traces data to reload, is this the intended behavior?.
> But still it uses the checkbox's old (preclick) value in the refresh request. Only after click to "Search Traces" button sends correct request.
> Also with smaller windows the view is broken.
> Screenshot from 2019-11-13 13-14-59

I did a trigger to relaod the iframe with the data onClick namespace Selector...maybe we can remove it and have the same behaviour than lookback


> Also in Distributed Tracing section, when first entering without having any namespace selected, then after selecting namespace, the service selection dropdown is still disabled:
> Screenshot from 2019-11-13 13-21-59

MMM let me check this I tested and it was working that.

aljesusg:
@hhovsepy I fixec the reload of the services, I removed the update testing :(. Can you check it ?

PR #1490, opened 2019-10-15
move Spinner to PF4

aljesusg:
Fix https://github.com/kiali/kiali/issues/1802

## Old Spinner
spinner_old

## New Spinner
new_spinner

### DIFF
--- a/src/components/PfSpinner.tsx
+++ b/src/components/PfSpinner.tsx
@@ -1,8 +1,8 @@
 import React from 'react';
 import { KialiAppState } from '../store/Store';
 import { connect } from 'react-redux';
-import { Spinner } from 'patternfly-react';
 import { style } from 'typestyle';
+import { SpinnerIcon } from '@patternfly/react-icons';
 
 type PfSpinnerProps = {
   isLoading: boolean;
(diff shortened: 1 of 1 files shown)


src/components/PfSpinner.tsx
```diff
@@ -22,7 +23,7 @@ export const PfSpinner: React.SFC<PfSpinnerProps> = props => {
   const { isLoading } = props;
   // It is more than likely it won't have any children; but it could.
   // @todo: Patternfly Spinner is not working here
-  return <Spinner className={spinnerStyle} loading={isLoading} inverse={true} />;
+  return isLoading ? <SpinnerIcon className={spinnerStyle} /> : <></>;
```
xeviknal: Nitpicking: dummy question, should be better to return 'undefined' instead of '<></>'?

aljesusg: mmm yep why not xD

aljesusg: hahaha not xD 
it should be a ReactElement
```ts
Type 'undefined' is not assignable to type 'ReactElement<any, string | ((props: any) => ReactElement<any, string | ... | (new (props: any) => Component<any, any, any>)> | null) | (new (props: any) => Component<any, any, any>)> | null'. 
```

PR #1491, opened 2019-10-15
 PF4 migration: stackbar charts (rates, in/out rates in side panel)

jotak:
Notable differences:
- Colors order on bars is reversed so that it matches legend order
- Legend mouseover is implemented, but not legend click events (is it useful here?)
- Removed some unused properties for simplicity... but can take them back if wanted

*Before*
Capture d’écran de 2019-10-15 14-50-55

*After*
Capture d’écran de 2019-10-15 14-49-56


src/components/SummaryPanel/InOutRateChart.tsx
```diff
@@ -1,88 +1,88 @@
 import * as React from 'react';
-import { StackedBarChart } from 'patternfly-react';
+import { Chart, ChartBar, ChartStack, ChartAxis, ChartTooltip } from '@patternfly/react-charts';
 import { PfColors } from '../../components/Pf/PfColors';
-import { LegendPosition, SUMMARY_PANEL_CHART_WIDTH } from '../../types/Graph';
+import { SUMMARY_PANEL_CHART_WIDTH } from '../../types/Graph';
+import * as Legend from 'components/Charts/LegendHelper';
 
 type InOutRateChartGrpcPropType = {
-  height?: number;
```
jotak: @israel-hdez @jshaughn For simplicity I've removed these properties as they were not used (so never tested), but if you don't like that and prefer that I take them back, I will.

jotak:
Hi @jshaughn @israel-hdez , please check my last commit, trying to address some of the issues.

@israel-hdez the different font size seemed to come from some rescaling due to a "Table" element that was present in `InOutRateTable.tsx` but not in `RateTable.tsx`, making the whole chart smaller, so I removed that.

@jshaughn I've investigated the tooltip issue and finally found some way to go. As a first step we now have some color indication back in toolip. I think before going too far here we should coordinate with UX / Patternfly (topic to be discussed today with UX meeting)

Also, I've increased the charts width so that they better match the width of the sparkline charts below ... Before that they were quite smaller, not sure if that was on purpose. I think it looks better like this, but this can be rollbacked if you prefer.

Capture d’écran de 2019-10-17 12-34-23

Legend is still left-justified. I haven't been able to center it, it doesn't seem that easy with svg (or i'm missing something?)

Last thing: I just realize i haven't re-established the legend "onclick" ... will do that later.

jshaughn:
@jotak All good changes, I still approve and am fine iterating improvements.

jotak:
Just added click on legend. But there's a small bug, I realize it also exists with the previous migrated chart & clickable legend: when you click on legend item to hide, then click again to show, then the mouseover event (with styling effect) doesn't work anymore on that item.

If you're fine to merge, I am too, I don't see this as a blocker. I'll open a ticket for this bug.

jshaughn:
@jotak Go ahead and merge.

jotak:
Thanks for review @jshaughn @israel-hdez

PR #1492, opened 2019-10-15
Bump versions of PF4 libs and migrate to Red Hat fonts

israel-hdez:
As suggested in https://www.patternfly.org/v4/documentation/react/overview/red-hat-font/, this is applying the required style to migrate from Overpass font to Red Hat {Text,Display} fonts.

Since the new fonts are present starting at PF v2.18.0, versions of all PF4 libraries were bumped to the 2019.08 release.

Closes kiali/kiali#1771.
Closes kiali/kiali#1807.


israel-hdez:
With the new version of PF, there is an issue with the dropdown. I'll try to fix.

mtho11:
I also noticed that the latest `@patternfly/react-table` breaks in our code.

mtho11:
@israel-hdez I created a separate issue for the PF4 upgrade: https://github.com/kiali/kiali/issues/1807

lucasponce:
Please, keep @aljesusg on these PRs.

Thanks.

jshaughn:
> I also noticed that the latest @patternfly/react-table breaks in our code.

I saw this and it looks like the issue is more stringent typing.  The issue results from this line in rules.tsx:

```
    const headerCells = [
      {
        title: 'Rule order',
        transforms: [cellWidth(10)],  <-- cellsWidth is a Decorator, but this needs to be a Transformer
        props: {}
      },
```

Unfortunately, it's not documented as to what we need to do to fix it, they have an open issue for docs: https://github.com/patternfly/patternfly-react/issues/2990

package.json
```diff
@@ -63,13 +63,13 @@
...
-    "@patternfly/patternfly": "2.4.1",
+    "@patternfly/patternfly": "2.36.0",
     "@patternfly/react-charts": "5.0.13",
-    "@patternfly/react-core": "3.89.2",
-    "@patternfly/react-styles": "3.5.16",
-    "@patternfly/react-table": "2.17.5",
-    "@patternfly/react-tokens": "2.6.21",
-    "@patternfly/react-virtualized-extension": "1.1.97",
+    "@patternfly/react-core": "3.104.0",
+    "@patternfly/react-styles": "3.5.22",
```
aljesusg: I would update react-core to 3.112.1. It's the latest productivized release https://www.npmjs.com/package/@patternfly/react-core

aljesusg: I would update react-core to 3.112.1. It's the latest productivized release https://www.npmjs.com/package/@patternfly/react-core

package.json
```diff
@@ -63,13 +63,13 @@
...
+    "@patternfly/react-table": "2.20.15",
+    "@patternfly/react-tokens": "2.6.26",
+    "@patternfly/react-virtualized-extension": "1.2.23",
     "@webcomponents/custom-elements": "1.2.4",
```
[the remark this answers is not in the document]
israel-hdez: I was preferring to use the set of released packages as stated in https://www.patternfly.org/v4/documentation/react/overview/notes/. I know each package is released independently, but the release notes page made me think that those versions are tested and "guaranteed" to work together.

That said, I'm not against upgrading individually if each package can work regardless of the version of other libs.

package.json
israel-hdez: Let me try.

israel-hdez: It's at latest now.

israel-hdez:
@jshaughn @mtho11 The table issue is fixed.

israel-hdez:
> These are two separate things can we have 2 different jiras for this?
UxD will have look over the Red Hat Fonts one much more carefully. And the upgrade of PF4 will require more QE attention that it doesn't break stuff.

OK, there are now two different issues, but I'm not sure if we want two different PRs.
The font change is trivial, but I think it requires a smoke test. The library upgrades would also need a smoke test. I think it's better to smoke test in one go to save some review time.

And about the fonts migration, I though that it wouldn't need UxD review given that it is PatternFly which is replacing Overpass font. So, I though the new fonts already have the :+1: of UxD :/ But let me know if this assumption is wrong.

mtho11:
fine by me @israel-hdez to have one PR

PR #1496, opened 2019-10-18
Create Pf4Icon function to consolidate all of the icons into a single place

mtho11:
Create a typesafe way to reuse Pf4Icons across Kiali-ui.  Also, makes changing the icons globally, very easy.

To use:
```
<div>
            <KialiIcon.Info /> Sparkline charts not supported for unknown node. Use edge for details.
            <hr />
</div>
```

** Issue reference **
closes https://github.com/kiali/kiali/issues/1813

### DIFF
--- a/src/pages/Graph/SummaryPanelNode.tsx
+++ b/src/pages/Graph/SummaryPanelNode.tsx
@@ -37,6 +37,7 @@ import { Reporter } from '../../types/MetricsOptions';
 import { icons } from '../../config/Icons';
 import { CyNode, decoratedNodeData } from '../../components/CytoscapeGraph/CytoscapeGraphUtils';
 import { PopoverPosition } from '@patternfly/react-core';
+import { KialiIcon } from 'config/KialiIcon';
 
 type SummaryPanelStateType = {
   loading: boolean;
@@ -399,7 +400,7 @@ export default class SummaryPanelNode extends React.Component<SummaryPanelPropTy
       return (
         <>
           <div>
-            <Icon type="pf" name="info" /> Sparkline charts not supported for unknown node. Use edge for details.
+            <KialiIcon.Info /> Sparkline charts not supported for unknown node. Use edge for details.
             <hr />
           </div>
         </>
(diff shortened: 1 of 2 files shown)


mtho11:
If this solution is acceptable we can do a global search/replace.

mtho11:
I'm sure I'm missing a couple of icons. This is just to illustrate the idea.

mtho11:
I tried to keep it as simple as possilble but eliminating any older PF3 `Icon` stuff.

josejulio:
Just throwing an idea, how about:
```typescript

export KialiIcons {
  Ok: () =>  <InfoAltIcon />,
  Info: () => <OkIcon color={PfColors.Green400} />
  // ... etc
}


// and you use it like:

<KialiIcons.Info/> Sparkline charts not supported for unknown node. Use edge for details.

```

This has less indirection and is easier to read IMO.

mtho11:
yeah, I like what @josejulio suggested. It seems simpler. Good idea. And I'll change the name to KialiIcon.

mtho11:
@jshaughn changes made. Please retest.

src/config/Pf4Icon.tsx
```diff
@@ -0,0 +1,33 @@
+import * as React from 'react';
```
jshaughn: Would it make sense to also name the file KialiIcon.tsx?

src/pages/Graph/SummaryPanelNode.tsx
```diff
@@ -399,7 +400,7 @@ export default class SummaryPanelNode extends React.Component<SummaryPanelPropTy
       return (
         <>
           <div>
-            <Icon type="pf" name="info" /> Sparkline charts not supported for unknown node. Use edge for details.
+            <KialiIcon.Info /> Sparkline charts not supported for unknown node. Use edge for details.
```
jshaughn: The solution is good, should this PR now apply the change everywhere, or is that a follow-up?

src/config/Pf4Icon.tsx
mtho11: Oh right. I forgot about changing that when I renamed.

src/pages/Graph/SummaryPanelNode.tsx
mtho11: I was thinking as a follow-up.

PR #1498, opened 2019-10-18
[v1.7] RpsChart: Fix array out of bound

jotak:
Backport kiali/kiali#1800 on 1.7


jotak:
@israel-hdez this is the backporting of https://github.com/kiali/kiali-ui/pull/1497

I'm not sure if we need to release 1.7.1 for that, but in case we prepare a release for something else or if we get people complaining about this bug, so we're ready to get it.

lucasponce:
I think this is good to merge.
Putting @israel-hdez as this may trigger a 1.7.1.

israel-hdez:
OK, it's merged. Please, ping me if you need the patch release.
As you say, nobody has reported it, so we could wait. But if you prefer the release, I can do it.

PR #1503, opened 2019-10-22
Trivial: Re-sync package.json version with back-end version

israel-hdez:
This is re-synching version number with the one found in https://github.com/kiali/kiali/blob/master/Makefile#L6.

Because version 1.8 is the next one to be released at end of sprint.


israel-hdez:
@jshaughn No problem. It became evident with the last snapshot that there was a problem :)

PR #1505, opened 2019-10-23
PF4 Message Center

jshaughn:
This PR replaces the PF3, NotificationDrawer-based, MessageCenter with a PF4 implementation.  It is meant to be an interim MessageCenter solution while we wait for a PF4 NotificationDrawer component, and helps us eliminate PF3 from the code base.

The general design is the same but look+feel obviously changes given the PF4 Card and Alert based impl.  

The initial PR has two feature losses that we need to acknowledge or rectify.  Given that this is an interim solution I decided to save time and complexity by omitting these two behaviors:

1. The toast notifications do not time out and instead require manual close. [FIXED !!!]
2. The notifications can not be marked 'seen' individually, only as a group.

For me, these two things are reasonable omissions, but if I had to add one, it would be item 1.  I'd have to investigate ways to add a timeout.  I don't think item 2 is all that important.

UPDATE: Issue 1 is solved!

This PR also refactors the MessageCenter impl to eliminate a lot of drill-down properties and let redux handle updates in a more typical fashion.  It also separates some components into their own files, letting each represent it's on redux container.  This refactor should hopefully still be relevant when we move to a real PF4 component.

As part of the PR there are fixes for 2 other issues, the "bell" alignment in the masthead, and inconsistent widths for the toast notifications.

Fixes https://github.com/kiali/kiali/issues/1639 // PF4
Fixes https://github.com/kiali/kiali/issues/1776 // bell
Fixes https://github.com/kiali/kiali/issues/1828 // widths

Before Video:
kiali#1639-before

After Video:
kiali#1639


mtho11:
Before/After images please.

src/pages/Graph/SummaryPanelNode.tsx
```diff
@@ -401,7 +401,7 @@ export default class SummaryPanelNode extends React.Component<SummaryPanelPropTy
       return (
         <>
           <div>
-            <KialiIcon.Info /> Sparkline charts not supported for unknown node. Use edge for details.
+            {KialiIcon.Info()} Sparkline charts not supported for unknown node. Use edge for details.
```
josejulio: Why? It wasn't working with the tag syntax?

josejulio: Ahh I see, is because the style? right?

jshaughn: I'd be happy if I didn't have to eval the KialiIcon but I didn't see a way to gave that work and have the style option.  Suggestions welcome...

src/config/KialiIcon.tsx
```diff
@@ -16,29 +16,40 @@ import {
...
+  AngleDoubleLeft: (style?: string) => <AngleDoubleLeftIcon className={style !== undefined ? style : iconStyle} />,
+  AngleDoubleRight: (style?: string) => <AngleDoubleRightIcon className={style !== undefined ? style : iconStyle} />,
+  AngleDoubleUp: (style?: string) => <AngleDoubleUpIcon className={style !== undefined ? style : iconStyle} />,
+  Applications: (style?: string) => <ApplicationsIcon className={style !== undefined ? style : iconStyle} />,
+  Bell: (style?: string) => <BellIcon className={style !== undefined ? style : iconStyle} />,
+  CircuitBreaker: (style?: string) => <BoltIcon className={style !== undefined ? style : iconStyle} />,
+  Error: (style?: string) => (
+    <ErrorCircleOIcon className={style !== undefined ? style : iconStyle} color={PfColors.Danger} />
+  ),
+  Info: (style?: string) => <InfoAltIcon className={style !== undefined ? style : iconStyle} color={PfColors.Info} />,
```
josejulio: I see two options here:

1. Create an `IconProps` type here that has the `style?: string` attribute and use it like this:
```
interface IconProps {
  style?: string;
}

Info: (props: IconProps) => <InfoAltIcon className={props.style !== undefined ? props.style : iconStyle} color={PfColors.Info} />,
```

2. Use object destructuring
```
  Info: ({style}: {style?: string}) => <InfoAltIcon className={style !== undefined ? style : iconStyle} color={PfColors.Info} />,

```
Both options allow to use it as a React component `<KialiIcon.Info style="foo"/>" or `<KialiIcon.Info/>`

I recommend to use option 1), as we have several Icons and I would use `className` instead of `style` to avoid mixing both semantics. Also, we can refactor the whole `style !== undefined ? style : iconStyle` out of the components, by giving the Icons defaultProps.

I'll write a PR to your branch with my suggested changes, but feel free to update as you see fit.

jshaughn: option 1 looks good to me, thanks for the pending PR...

src/pages/Graph/SummaryPanelNode.tsx
jshaughn: Thanks for the PR!

jmazzitelli:
once again, PF4 makes things look worse :/

PR #1512, opened 2019-10-29
Application Details, the missing Istio Sidecar is so close to workload name

aljesusg:
Fix https://github.com/kiali/kiali/issues/1852

## Before
app_details_view

## Now
now


abonas:
I would make the gap/space even bigger...

aljesusg:
now2

Done  @abonas

PR #1514, opened 2019-10-29
Use Select control in graph's Display dropdown

israel-hdez:
Replaces the Dropdown control with the Select control.
The aim is to stop using a free-form dropdown to a proper PF pattern.

image

Closes kiali/kiali#1849.


mtho11:
I actually like the number in the select box. It just gives you a rough idea of how many things you have going on. Kind of like the message count for the message center.

israel-hdez:
OK. Then the counter is fine!
Merging.

PR #1515, opened 2019-10-30
PF4 Help Tours

jshaughn:
This introduces a PF4 Tour impl to replace the existing PF3-based impl.  It is based on PF4 Popover.  I tried to maintain the basic approach used in the previous solution.  A Tour is still defined in its own component (GraphHelpTour.tsx).   But there are changes.  I now use redux to track the current tour and the current tour stop, and let updates be initiatd by changes to the redux state. The biggest change comes from the PF4 Popover requiring the 'children' property, and not being able to apply it (afaict) to a target html element like we did previously.  This means that each tour stop must be instrumented (i.e. you have to wrap the target element inside of the new TourStop element).

Fixes https://github.com/kiali/kiali/issues/1420

Before Video:
* note, from what I can see several tour stops are currently broken in the PF3 version in master...

kiali#1420-before

After Video:
kiali#1420-after


josejulio:
I found some problems with this, not sure if all are related to this change:

If one TourStop is not being rendered ( Bottom bar in this example) the tour will "continue" and won't be show. If trying to click on show it won't show anything (because is hiding), clicking again will show it.

Peek 2019-10-30 15-25

This is more strange, when opening the Tour and clicking on the Display menu, the top bar is hidden (Firefox and Chrome)
Peek 2019-10-30 15-28


If I try to click somewhere else, it prevents the popover from closing (probably intended, but looks bad as it flashes)

Peek 2019-10-30 15-26

src/components/Tour/TourStop.tsx
```diff
@@ -0,0 +1,183 @@
...
+        Done
+      </Button>
+    );
+  };
+
+  /*
+  private stopNumber = () => {
+    return <span className={stopNumberStyle}>{this.props.activeStop! + 1}</span>;
+  };
+  */
```
josejulio: Could we remove it If not needed anymore?

jshaughn:
@josejulio 

> If one TourStop is not being rendered ( Bottom bar in this example) the tour will "continue" and won't be show. If trying to click on show it won't show anything (because is hiding), clicking again will show it.

This is tricky.  It results from the approach of wrapping the child element with the TourStop.  If the child element is hidden, or maybe not even mounted, then there is nothing to render for the activeTourStop.   The issue is that I don't see a good way of detecting when the activeTourStop can't be rendered.  One way maybe would be to go back to configuring an element id for each stop.  And use a document,getElementById to validate the stop before setting it.  So when hitting Next or Back you'd skip over the ones that are't valid (this would be done in the redux state code).  It's unfortunate because then you have to assign an id in the config AND still wrap the target child element.  Any other ideas?

> This is more strange, when opening the Tour and clicking on the Display menu, the top bar is hidden...

This is weird.  I am suspicious maybe of the new Display component.  I think there is already an issue for someone reporting that the manifest disappears unexpectedly.  Not sure, will have to investigate.

jshaughn:
@josejulio @mattmahoneyrh The issue regarding unrendered tour stops is fixed, I think.   The second issue regarding the masthead is totally puzzling.  The same behavior is reported in https://github.com/kiali/kiali/issues/1834 but for a different action entirely.  I suspect maybe a PF4 issue.  I'm not sure whether that problem should block this PR, we can discuss.

mattmahoneyrh:
@jshaughn, validated that the first issue is fixed.

josejulio:
> I'm not sure whether that problem should block this PR

I don't think so, we should check it on a separate PR as it seems to be something not directly related to this PR

PR #1521, opened 2019-11-04
Fix Masthead disappears after scrolling on lists

aljesusg:
Fix https://github.com/kiali/kiali/issues/1834

cc @hhovsepy can you check if this fix the problem ?

### DIFF
--- a/src/components/Health/Health.scss
+++ b/src/components/Health/Health.scss
@@ -4,3 +4,16 @@ svg:not(:root).icon-healthy,
 svg:not(:root).icon-na {
   overflow: visible;
 }
+.tippy-popper[x-placement^='left'] .pf-tooltip-theme .tippy-arrow {
+  background-color: #fff;
+  border-color: black;
+  border-width: 1px;
+  border-style: solid;
+}
+.health_indicator .pf-c-tooltip__content {
+  background-color: #fff;
+  border-style: solid;
+  border-color: black;
+  border-width: 1px;
+  text-align: left;
+}
--- a/src/components/Health/HealthIndicator.tsx
+++ b/src/components/Health/HealthIndicator.tsx
@@ -1,9 +1,8 @@
 import * as React from 'react';
-import { Popover, PopoverPosition } from '@patternfly/react-core';
+import { Card, CardHeader, CardBody, PopoverPosition, Tooltip } from '@patternfly/react-core';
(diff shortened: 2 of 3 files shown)


aljesusg:
Ok it seems that there is a bug in the css in virtualList when we use PopOver.

Solutions : 

- Open an Issue and wait PF team
- Use Tooltip and maybe change the color of the background (See image)
Screenshot from 2019-11-04 13-32-42

cc @lucasponce

jshaughn:
@aljesusg This is also consistent with the scenario I recently added in the issue.  Popover is in use in both scenarios.  Although in mine there is no Virtual List, instead it has to do with the Select widget.  I think there is a PF4 issue that needs to be resolved.

jotak:
@aljesusg tooltips have text centered, that's one of the reasons why I didn't use it for health, as here we need to display several elements as a tree with indentation. I was not able to keep the indentation within tooltips.

jotak:
(btw, if it's meant to be reviewed & merged, can you change the name of this PR & commit messages please?)

aljesusg:
@jotak we can add styles to this, It's a Card so we can align the text to the left
@jshaughn Yes...this should be fixed by PF team I'll open the issue but we need to fix this while PF team work in a solution.

Thanks

hhovsepy:
Verified the fix of original issue.

aljesusg:
Done @jotak maybe we can merge this while PF team work in the fix

aljesusg:
> Not sure how urgent / critical the issue is, but the current solution looks too much of a regression to me.
> 
> * The "Card" is quite ugly inside a tooltip, with that big black border. Maybe we should remove card and have a TextContent with title as h2 or h3 instead?
> * Text should be left-aligned
> 
> I've just seen that there's a new flag in Tooltip, `isContentLeftAligned`, which seems to be exactly for that ( https://patternfly-react.surge.sh/patternfly-4/documentation/react/components/tooltip ) but it seems not present in our current version of PF

Ok let me try to have the same style, I can change the background color but no the arror I need to check how I can do it. Thanks @jotak I'll ping you later

aljesusg:
What do you think about this style @jotak ? It's text-aligned left and is close to the old design

Screenshot from 2019-11-05 13-29-03

aljesusg:
Ups sorry not, the exitdelay was added to inspect the css I need to remove it

PR #1522, opened 2019-11-04
Add referenced istio objects to validations

xeviknal:
** Describe the change **

Kiali has now istio object referenced in some validations. For example, multi match gateways validation. This PR adds visual list of references in the editor page, as a sidebar.

Closes https://github.com/kiali/kiali/issues/1419
Backend changes: https://github.com/kiali/kiali/pull/1912

P.S.: Adding the sidebar makes space for other referenced items (not exclusively from validations). It just popped in my mind this old jira: https://issues.jboss.org/browse/KIALI-1883

** Issue reference **
#1419

** Backwards compatible? **
yes

** Screenshot **

Screen-recording-_14_

Editor page with referenced objects:
Screenshot of Kiali Console (75)

Editor page without referenced objects:
Screenshot of Kiali Console (77)

Editor page without validations:
Screenshot of Kiali Console (76)

### DIFF
--- a/src/components/Link/IstioObjectLink.tsx
+++ b/src/components/Link/IstioObjectLink.tsx
@@ -19,7 +19,7 @@ const IstioObjectLink = (props: Props) => {
   if (!!subType) {
     to = to + '/' + istioType + '/' + subType + '/' + name;
   } else {
-    to = to + '/' + istioType.slug + '/' + name;
+    to = to + '/' + istioType.url + '/' + name;
   }
 
   return (
--- a/src/components/Link/IstioObjectLink.tsx
+++ b/src/components/Link/IstioObjectLink.tsx
@@ -1,0 +1,30 @@
+import * as React from 'react';
+import { Paths } from '../../config';
+import { Link } from 'react-router-dom';
+import { Badge, Tooltip, TooltipPosition } from '@patternfly/react-core';
+import { IstioTypes } from '../VirtualList/Config';
+
+interface Props {
+  name: string;
+  namespace: string;
+  type: string;
+}
+
+const IstioObjectLink = (props: Props) => {
+  const { name, namespace, type } = props;
+  //const objectType = type.charAt(0).toUpperCase() + type.slice(1);
+  const istioType = IstioTypes[type];
+  const url = '/namespaces/' + namespace + '/' + Paths.ISTIO + '/' + istioType.slug + '/' + name;
+  return (
+    <>
+      <Tooltip position={TooltipPosition.top} content={<>{istioType.name}</>}>
+        <Badge className={'virtualitem_badge_definition'}>{istioType.icon}</Badge>
+      </Tooltip>
+      <Link to={url}>
+        {namespace}/{name}
+      </Link>
+    </>
+  );
+};
+
+export default IstioObjectLink;
--- a/src/components/Link/IstioObjectLink.tsx
+++ b/src/components/Link/IstioObjectLink.tsx
@@ -8,18 +8,26 @@ interface Props {
   name: string;
   namespace: string;
   type: string;
+  subType?: string;
 }
 
 const IstioObjectLink = (props: Props) => {
-  const { name, namespace, type } = props;
+  const { name, namespace, type, subType } = props;
   const istioType = IstioTypes[type];
-  const url = '/namespaces/' + namespace + '/' + Paths.ISTIO + '/' + istioType.slug + '/' + name;
+  let to = '/namespaces/' + namespace + '/' + Paths.ISTIO;
+
+  if (!!subType) {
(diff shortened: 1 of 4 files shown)


hhovsepy:
Generally it looks good to me:
Screenshot from 2019-11-04 18-13-13
Screenshot from 2019-11-04 18-19-16
Screenshot from 2019-11-04 18-42-34

Only in one case the istio config page failed to load, when DR references to a host from other namespace:
"apiVersion: networking.istio.io/v1alpha3
kind: DestinationRule
metadata:
  name: reviews-dr1
spec:
  host: reviews.bookinfo2.svc.cluster.local
  trafficPolicy:
    loadBalancer:
      simple: RANDOM
  subsets:
  - name: v1
    labels:
      version: v1
  - name: v2
    labels:
      version: v2
  - name: v3
    labels:
      version: v3
"
Screenshot from 2019-11-04 18-20-10
No other error in logs.

xeviknal:
I have detected an error on the backend. Therefore this is taking me a little bit longer.
Here's the problem: there is one empty reference item in the references array. When we want to print it, it shows an error.

Screenshot from 2019-11-05 17-54-48

lucasponce:
@xeviknal then, this one needs to wait until some changes in backend are ready, isn't it ?

src/components/VirtualList/Config.ts
```diff
@@ -116,24 +116,24 @@ const istioType: ResourceType<IstioConfigItem> = {
...
-  policy: { name: 'Policy', icon: 'P' },
-  meshpolicy: { name: 'MeshPolicy', icon: 'MP' },
-  servicemeshpolicy: { name: 'ServiceMeshPolicy', icon: 'SMP' },
-  clusterrbacconfig: { name: 'ClusterRbacConfig', icon: 'CRC' },
-  rbacconfig: { name: 'RbacConfig', icon: 'RC' },
-  servicemeshrbacconfig: { name: 'ServiceMeshRbacConfig', icon: 'SRC' },
-  sidecar: { name: 'Sidecar', icon: 'S' },
-  servicerole: { name: 'ServiceRole', icon: 'SR' },
-  servicerolebinding: { name: 'ServiceRoleBinding', icon: 'SRB' }
+  gateway: { name: 'Gateway', slug: 'gateways', icon: 'G' },
```
lucasponce: slug ?

src/components/Link/IstioObjectLink.tsx
```diff
@@ -0,0 +1,30 @@
...
+  name: string;
+  namespace: string;
+  type: string;
+}
+
+const IstioObjectLink = (props: Props) => {
+  const { name, namespace, type } = props;
+  //const objectType = type.charAt(0).toUpperCase() + type.slice(1);
+  const istioType = IstioTypes[type];
+  const url = '/namespaces/' + namespace + '/' + Paths.ISTIO + '/' + istioType.slug + '/' + name;
```
lucasponce: Note that this is not going to work for templates/adapters.
As kiali supports/supported the old legacy way, we have a
```
http://localhost:3000/console/namespaces/istio-system/istio/templates/instances/attributes
```
for these special cases: adapter/templates.

My suggestion, don't change it here, but just add a comment for the future.

src/components/VirtualList/Config.ts
```diff
@@ -116,24 +116,24 @@ const istioType: ResourceType<IstioConfigItem> = {
...
+  virtualservice: { name: 'VirtualService', slug: 'virtualservices', icon: 'VS' },
+  destinationrule: { name: 'DestinationRule', slug: 'destinationrules', icon: 'DR' },
+  serviceentry: { name: 'ServiceEntry', slug: 'serviceentries', icon: 'SE' },
+  rule: { name: 'Rule', slug: 'rules', icon: 'R' },
+  adapter: { name: 'Adapter', slug: 'adapters', icon: 'A' },
```
lucasponce: Commented before: templates/adapters are special as we supported legacy formats (probably from Istio 1.4 we can refactor this).

src/components/VirtualList/Config.ts
xeviknal: It is the term used for humanizing IDs in the URL. Probably born because of all the SEO.
https://en.wikipedia.org/wiki/Clean_URL#Slug

For example, in `http://www.contentwithstyle.co.uk/content/clean-urls-for-a-better-search-engine-ranking` the `slug` is `clean-urls-for-a-better-search-engine-ranking`.

Anyway, I changed to follow the other places in the project where we use term `url`.

src/components/Link/IstioObjectLink.tsx
xeviknal: I changed the method itself to handle type and subtype. However the consumer component doesn't render type/subtype.

PR #1527, opened 2019-11-06
Show additional configured details (annotations)

jotak:
- Adding "additionalDetails" to workloads & services model
- Show as text or link
- Add a test

Backend PR: https://github.com/kiali/kiali/pull/1872/
Part of https://github.com/kiali/kiali/issues/1564

Capture d’écran de 2019-11-06 10-42-33


hhovsepy:
Screenshot from 2019-11-07 15-19-27
Screenshot from 2019-11-07 15-19-52

src/helpers/TextFormatting.tsx
```diff
@@ -0,0 +1,16 @@
+import * as React from 'react';
+
+export const formatTextOrLink = (text: string, urlTruncate?: number): JSX.Element => {
```
xeviknal: Does make sense to have a component link instead of a helper? I think it would be more idiomatic than a helper. wdyt

jotak: Good point

abonas:
@omerlh fyi

PR #1532, opened 2019-11-07
Use native js function to format date, fixing 12am/pm issue

jotak:
Sample output: _Oct 9, 2:42 PM_

Closes https://github.com/kiali/kiali/issues/1878


jotak:
@FilipB does it fix the test issue you had?

jotak:
Note that I'm keeping 'en-US' locale to stay on par with previous behaviour but we could use user locale as well

PR #1541, opened 2019-11-13
Adding validation status to Overview page

xeviknal:
** Describe the change **
Adding validation status to overview page.

Needs backend PR: https://github.com/kiali/kiali/pull/1932

** Issue reference **
https://github.com/kiali/kiali/issues/1556

** Screenshot **
Screenshot of Kiali Console (88)

Screenshot of Kiali Console (89)

### DIFF
--- a/src/components/Validations/ValidationSummary.tsx
+++ b/src/components/Validations/ValidationSummary.tsx
@@ -1,15 +1,13 @@
 import * as React from 'react';
-import { ObjectValidation, ValidationTypes } from '../../types/IstioObjects';
+import { ValidationTypes } from '../../types/IstioObjects';
 import { style } from 'typestyle';
 import { Text, TextVariants, Tooltip, TooltipPosition } from '@patternfly/react-core';
 import Validation, { severityToValidation } from './Validation';
-import { higherSeverity, highestSeverity } from '../../types/ServiceInfo';
 
 interface Props {
   id: string;
-  validations: ObjectValidation[];
-  definition?: boolean;
-  size?: string;
+  errors: number;
+  warnings: number;
 }
 
 const tooltipListStyle = style({
@@ -19,43 +17,19 @@ const tooltipListStyle = style({
 });
 
 export class ValidationSummary extends React.PureComponent<Props> {
-  numberOfChecks = (type: string) => {
-    let numCheck = 0;
-    this.props.validations.forEach(validation => {
-      if (validation.checks) {
-        numCheck += validation.checks.filter(i => i.severity === type).length;
-      }
-    });
-    return numCheck;
+  getTypeMessage = (count: number, type: ValidationTypes): string => {
+    return count > 1 ? `${count} ${type}s found` : `${count} ${type} found`;
   };
-
-  getTypeMessage = (type: string) => {
-    const numberType = this.numberOfChecks(type);
-    return numberType > 0
-      ? numberType > 1
-        ? `${numberType} ${type}s found`
-        : `${numberType} ${type} found`
-      : undefined;
-  };
-
-  checkCount() {
-    return this.props.validations.reduce((sum, current) => {
-      return sum + (current.checks ? current.checks.length : 0);
-    }, 0);
-  }
 
   severitySummary() {
     const issuesMessages: string[] = [];
 
-    if (this.props.validations.length > 0) {
-      const errMessage = this.getTypeMessage('error');
-      if (errMessage) {
-        issuesMessages.push(errMessage);
-      }
-      const warnMessage = this.getTypeMessage('warning');
-      if (warnMessage) {
-        issuesMessages.push(warnMessage);
-      }
+    if (this.props.errors > 0) {
+      issuesMessages.push(this.getTypeMessage(this.props.errors, ValidationTypes.Error));
+    }
+
+    if (this.props.warnings > 0) {
+      issuesMessages.push(this.getTypeMessage(this.props.warnings, ValidationTypes.Warning));
     }
 
     if (issuesMessages.length === 0) {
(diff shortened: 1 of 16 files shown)


abonas:
was the UI part discussed with UX?
From a brief look:
1. the icons might need to change so it won't be confused with health icons, although if the status is completely identical (err/warn...) then it makes sense to have same icons.
2. the location of the icon might seem odd.
3. I'd consider adding a clear text indication where the icon is that this is in context of validations. otherwise when looking at the screenshot, if I take the default namespace card as an example, it has a warning icon, but 0 apps, it looks strange. or taking istio-system namespace card, all apps are "green" but there is a warning icon on the left corner - also unclear why.

xeviknal:
> was the UI part discussed with UX?
It supposed to be part of the UX meeting for today, but meeting was cancelled. Touched in the next meeting. Here is the issue for UX/UI to discuss: https://github.com/kiali/kiali-design/issues/156

Location of the icon is odd. I am playing a bit now with that.
The warning icon for namespaces with no applications/services is odd however it is possible that the user has some istio objects living there without any application yet. It sounds strange, yep. Also, the environment there is pretty messed.
Text with clarification is a good idea to proceed.

jshaughn:
@abonas @xeviknal  I had this comment in the Jira, copying it here in case it was missed, Maybe this should go on the design issue, I see that UX has not yet added any design.

> If we are planning to have the same warning on Overview, Graph and in Config List, perhaps it should be centralized on the masthead? Could have a some sort of warning icon with a hover that if clicked brings you to the Config List?

lucasponce:
@jshaughn https://github.com/kiali/kiali-design/issues/156

I don't think this should go in the masthead, but focused on namespace.
Masthead is proper place for a globalcluster feature.

This one is similar as showing TLS per namespace.

I think that should be discussed in the Issue, instead of the PR, we are talking about a different feature.

lucasponce:
@xeviknal I like this feature and I guess that in parallel we can discuss with UX refinements about where to place the icon.
But this week we are not going to have UX meetings and I think it could be more interesting merging this and refine later, rather than put it on hold for 1-2 weeks.

If you are agree, we can proceed.

xeviknal:
yes, I agree. The issue is opened in kiali-design (https://github.com/kiali/kiali-design/issues/156) and ready to discuss even offline if needed.

lucasponce:
@xeviknal ok, let me review in detail both PRs locally.

abonas:
@jshaughn this is per namespace and as such, less suitable for the masthead IMO, but UX team will give their inputs to finalize the design.

lucasponce:
For consistency: "Config status" -> "Istio config status"

lucasponce:
Another regression:
image

I couldn't validate against master, but it seems that validations icon are now moved from their original position (well centered).

src/pages/Overview/OverviewPage.tsx
```diff
@@ -369,6 +416,26 @@ export class OverviewPage extends React.Component<OverviewProps, State> {
...
+          errors={ns.validations.errors}
+          warnings={ns.validations.warnings}
+          style={{ marginLeft: '5px' }}
+          wrapper={false}
+        />
+      );
+    }
+    return (
+      <>
+        <Text component={TextVariants.p}>Config status: {status}</Text>
```
lucasponce: Config -> Istio config

src/pages/Overview/OverviewPage.tsx
```diff
@@ -329,6 +375,7 @@ export class OverviewPage extends React.Component<OverviewProps, State> {
                   </CardHeader>
                   <CardBody>
                     {this.renderStatuses(ns)}
+                    {this.renderConfigStatus(ns)}
```
lucasponce: Less critical, but I'd also move IstioConfig here.

src/components/Validations/ValidationSummary.tsx
```diff
@@ -1,15 +1,16 @@
...
+import { CSSProperties } from 'react';
 
 interface Props {
   id: string;
-  validations: ObjectValidation[];
-  definition?: boolean;
-  size?: string;
+  errors: number;
+  warnings: number;
+  style?: CSSProperties;
```
lucasponce: Side comment.
Something is not right with this abstraction if we need to pass to a component css style that requires to understand the implementation details of the component.

No action comment, but leaving a comment here.

jshaughn: I'm not sure it needs to understand the impl, this just provides an override to the default style, which is a fairly common component feature, I think.  Maybe I misunderstand the comment.

xeviknal: That is exactly it. I think is a common approach for styling components.

lucasponce: No, it is not. As the "wrapper" prop affects as well in the final style.
But, as commented this is side comment about abstraction and best practices.
It doesn't require a change for this PR, but to me is a sympton that probably there are technical debt between layers.

jshaughn: I guess I don't really understand.  How is what is done here different than other components that support a style property?  Maybe you can answer this is dev-list if you think it will benefit people.

hhovsepy:
I see the validation icon a bit higher, but this happened before this PR, can this be rebased to include the fix for https://github.com/kiali/kiali-ui/pull/1546 ?
Screenshot from 2019-11-19 10-26-28

xeviknal:
@hhovsepy and  @lucasponce I see this happening before this PR also. Anyway, I just rebase with #1546 but no impact found. I compared with master and looks the same.
However, I pushed a tweak to follow same vertical-alignment as health icons. Perhaps it needs a bit more of tweaking.

hhovsepy:
Screenshot from 2019-11-19 15-14-11

lucasponce:
@xeviknal if you can review latest @hhovsepy comment we can consolidate this first.

xeviknal:
@lucasponce @hhovsepy forgot to push the commit from last night.

hhovsepy:
Verified:
Screenshot from 2019-11-20 10-58-29

PR #1542, opened 2019-11-13
Use PF4 css color variables more consistently

jshaughn:
An issue is that not every API accepts the variables and instead is wanting the hex value.  To handle this the PR provides a function to get the variable values.  The first call will determine the values (the document must exist), subsequent calls are faster.
- also, a little bit of renaming for clarity

Fixes https://github.com/kiali/kiali/issues/1839

This screenshot is just showing that the expected colors are rendered in the graph and side panel chart:
image

### DIFF
--- a/src/components/CytoscapeGraph/graphs/GraphStyles.ts
+++ b/src/components/CytoscapeGraph/graphs/GraphStyles.ts
@@ -1,5 +1,11 @@
 import { style } from 'typestyle';
-import { PfColors, withAlpha, PFAlertColorValsType, PfAlertColorVals } from '../../../components/Pf/PfColors';
+import {
+  PfColors,
+  withAlpha,
+  getPFAlertColorVals,
+  PFColorVal,
+  PFAlertColorVals
+} from '../../../components/Pf/PfColors';
 import { EdgeLabelMode } from '../../../types/GraphFilter';
 import { FAILURE, DEGRADED, REQUESTS_THRESHOLDS } from '../../../types/Health';
 import {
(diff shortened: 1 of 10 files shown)


src/components/CytoscapeGraph/graphs/GraphStyles.ts
```diff
@@ -18,15 +18,10 @@ import * as Cy from 'cytoscape';
...
-const ColorDanger = PfColors.Danger; // Same as Health.FAILURE.color
-const ColorHealthy = PfColors.Success; // Same as Health.HEALTHY.color
-const ColorWarning = PfColors.Warning; // Same as Health.DEGRADED.color
-
-const EdgeColor = ColorHealthy;
+let EdgeColor: string;
 const EdgeColorDead = PfColors.Black500;
-const EdgeColorDegraded = ColorWarning;
-const EdgeColorFailure = ColorDanger;
+let EdgeColorDegraded: string;
```
mtho11: It would be nice if we had a little stronger typing than just `string` for these colors. Even though it can just be a type alias for string. It aids in the documentation and prevents accidental invalid usage.

src/components/Pf/PfColors.tsx
```diff
@@ -89,26 +89,53 @@ export enum PfColors {
...
-export enum PFColorVars {
-  DangerColor = 'var(--pf-global--danger-color--100)',
-  // Patternfly has a new green that is darker but they have not yet
-  // updated the PF vars for it
-  // SuccessColor = 'var(--pf-global--success-color--100)',
-  SuccessColor = '#3e8635', // computed values like Green500 are not allowed in enum
-  WarningColor = 'var(--pf-global--warning-color--100)'
+// Health/Alert colors https://www.patternfly.org/v4/design-guidelines/styles/colors
+export type PFAlertColorValsType = {
+  Danger: string;
```
mtho11: Again, with the loosely-typed `string`

jshaughn:
@aljesusg just a cc in case you had any comment

PR #1543, opened 2019-11-14
Unmount non visibible children instead to mountonenter only

lucasponce:
Adjust better the behaviour of the tabs.
Instead to wait to first "enter" it will "unmount" non visible tabs.

This will fix the "back button" problem in some pages and will maintain the desired behavior to not load non-visible tabs that are heavy.

Fixes kiali/kiali#1863


lucasponce:
@xeviknal has pointed me a use case that we miss
I need to perform more tests, to see if I can fix both use cases.

PR #1551, opened 2019-11-18
Add support for Node.js 12.x (new LTS)

mtho11:
** Describe the change **
Now that Node.js 12.x is LTS we need to support that as well as the existing LTS which is 10.x.

- [ ] Make tests work with both Node.js 12.x and 10.x 

The issue here is that our tests fail when using the new version and fixing that makes the tests fail in the old version.

fixes: https://github.com/kiali/kiali/issues/1946


mtho11:
The real issue here is if we want to support both Node.js 10.x and 12.x for UI builds or just the latest 12.x version? It is further reaching because of transitive dependencies.

mtho11:
After fiddling with this stuff the question is: "Do we want to support both Node.js 10.x and 12.x, or should we just move on to 12.x"? 
Both are LTS versions with overlapping time periods.
Obviously, there is a cost to supporting both versus just the latest.

jotak:
As long as the 12.x support works just fine and doesn't cause any trouble to the developers, I don't see any reason to continue to support 10.x, no?

jshaughn:
@mtho11 Why did you close this?  If it makes sense and seems stable I think it should be fine for beginning of the next sprint.

mtho11:
There is a lot of work to be done to make this work with both 10 and 12.

jshaughn:
What about just 12?  Why support both?

israel-hdez:
I would vote to support node12 only if it is already available in downstream for building.
Note that *nodejs support* is only to support dev environments (and the build). The final package does not depend on nodejs.

mtho11:
Just supporting 12 would be much easier. That would be my vote.

PR #1553, opened 2019-11-19
remove null markers

jshaughn:
- no need for the null markers after the fixes for #1933.  And those metric fields should never be null.

Fixes : https://github.com/kiali/kiali/issues/1956


jshaughn:
@jotak Excellent catch, that's what I get for rushing at the end of the day.

PR #1556, opened 2019-11-20
Adjust CSS styles to fix wrong scroll on logs and yaml editor

lucasponce:
Fixes kiali/kiali#1955

Adjust CSS for Logs container and ACE Editor to adjust unnecessary scroll after PF4 migration.


hhovsepy:
The height is fixed, so with small window it still is not scrolling:
Screenshot from 2019-11-21 10-30-32

lucasponce:
@hhovsepy thanks, I'm going to refine the fix to cover this as well.

lucasponce:
I'm not sure if we can adjust this for small windows, as the same behaviour is similar to other components, like navigation:
image

So, I suggest if we consolidate this one, and we can revisit the global layout used in next iterations.

PR #1560, opened 2019-11-22
Traces (spans) in metrics

jotak:
- Fetch service spans
- Use new k-charted overlay feature
- Rolling-window for refreshes

Part of https://github.com/kiali/kiali/issues/1854

Pic:
Capture d’écran de 2019-11-22 15-09-14

Needs k-charted: https://github.com/kiali/k-charted/pull/59
Needs backend: https://github.com/kiali/kiali/pull/1994

### DIFF
--- a/src/components/Metrics/IstioMetrics.tsx
+++ b/src/components/Metrics/IstioMetrics.tsx
@@ -3,7 +3,7 @@ import { connect } from 'react-redux';
 import { RouteComponentProps, withRouter } from 'react-router';
 import { Card, CardBody, Grid, GridItem, Toolbar, ToolbarGroup, ToolbarItem } from '@patternfly/react-core';
 import { PfColors, PFAlertColor } from 'components/Pf/PfColors';
-import { Dashboard, DashboardModel, ExternalLink, toOverlay } from '@kiali/k-charted-pf4';
+import { Dashboard, DashboardModel, ExternalLink, toOverlay, OverlayInfo, Overlay } from '@kiali/k-charted-pf4';
 import { style } from 'typestyle';
 
 import RefreshContainer from '../../components/Refresh/Refresh';
@@ -25,7 +25,6 @@ import { GrafanaInfo } from '../../types/GrafanaInfo';
 import { MessageType } from '../../types/MessageCenter';
 import { GrafanaLinks } from './GrafanaLinks';
 import { Span, TracingQuery } from 'types/Tracing';
-import { OverlayInfo, Overlay } from '@kiali/k-charted-pf4/dist/pf4/src/types/Overlay';
 
 type MetricsState = {
   dashboard?: DashboardModel;
--- a/src/components/Metrics/IstioMetrics.tsx
+++ b/src/components/Metrics/IstioMetrics.tsx
@@ -151,7 +151,7 @@ class IstioMetrics extends React.Component<IstioMetricsProps, MetricsState> {
         // Update last fetch time only if we had some results
         // So that if Jaeger DB hadn't time to ingest data, it's still going to be fetched next time
         if (spans.length > 0) {
-          this.lastFetchMicros = nowMicros;
+          this.lastFetchMicros = Math.max(...spans.map(s => s.startTime));
         }
       })
       .catch(err => {
@@ -204,7 +204,7 @@ class IstioMetrics extends React.Component<IstioMetricsProps, MetricsState> {
           y: Number(span.duration / 1000000),
           error: hasError,
           color: hasError ? PFAlertColor.Danger : PfColors.Cyan300,
-          size: 5 + Math.min(span.traceSize, 25)
+          size: 4
         };
       });
       overlay = toOverlay(info, dps);
--- a/src/components/Metrics/IstioMetrics.tsx
+++ b/src/components/Metrics/IstioMetrics.tsx
@@ -141,16 +141,18 @@ class IstioMetrics extends React.Component<IstioMetricsProps, MetricsState> {
   fetchSpans(lastFetchMicros?: number) {
     const doAppend = lastFetchMicros !== undefined;
     const nowMicros = new Date().getTime() * 1000;
-    const frameStart = nowMicros - (this.options.duration || 600) * 1000000; // seconds to micros;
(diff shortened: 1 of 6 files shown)


lucasponce:
Note, that user should be able to enable/disable them.
Not sure if is better to have a separate control and not hide this feature inside the "Metric Settings"

jotak:
Currently they are disable by default and can be enabled by clicking on the legend item

lucasponce:
> Currently they are disable by default and can be enabled by clicking on the legend item

But that sounds a new way to interact with the Graph, isn't ?

I mean, perhaps is less obvious than a on/off switch as proposed in the mock.

(I mean, the final UI is less important, but my point is that user see that it's a first class new feature).

jotak:
@lucasponce I don't think it's new, as the legends items are already clickable to show/hide metrics, so this is the same here, the only difference being that this ones comes disabled at first.
The issue with the global show/hide is that it would affect all charts on the dashboard at once, which can cause some client-side slow down when there's a lot of data.

hhovsepy:
Does building this PR require any ongoing PR from k-charted?
Because it fails for me" Module '"../../../node_modules/@kiali/k-charted-pf4/dist/pf4/src"' has no exported member 'toOverlay'.  TS2305"

lucasponce:
@hhovsepy yes, this one https://github.com/kiali/k-charted/pull/59/files

hhovsepy:
Refresh (auto refresh or via manual click on refresh button) does not load the new Spans.
So after some time Spans are gone in page.
Only browser page refresh brings Spans back.
Screenshot from 2019-12-02 18-55-56
Screenshot from 2019-12-02 18-56-13

jotak:
@hhovsepy it's weird it was working for me... can you look at JS console, see if there's some calls to the ".../spans" endpoint? Do they return empty result?
Anything special in backend logs?

hhovsepy:
@jotak "spans/" endpoint returns empty list, no error in JS console, and backend logs "Found 0 spans in the 0 traces for service details
"

jotak:
@hhovsepy patched ; I'm not sure if it will fix it, just a guess

src/components/Metrics/IstioMetrics.tsx
```diff
@@ -141,16 +141,18 @@ class IstioMetrics extends React.Component<IstioMetricsProps, MetricsState> {
...
-        } else {
-          this.setState({ tracingSpans: res.data });
+        const spans = doAppend
+          ? this.state.tracingSpans.filter(s => s.startTime >= frameStart).concat(res.data)
+          : res.data;
+        this.setState({ tracingSpans: spans });
+        // Update last fetch time only if we had some results
+        // So that if Jaeger DB hadn't time to ingest data, it's still going to be fetched next time
+        if (spans.length > 0) {
+          this.lastFetchMicros = nowMicros;
```
jotak: todo / to fix: actually lastFetchMicros should not be the last time data was fetched, but the latest datapoint timestamp already fetched.

hhovsepy:
@jotak I still see the problem

lucasponce:
Some comment about feature:
image
When we have two different units, it's fine to have both y-axis in left/right with different units.

When both metric + spans are represented with same unit, it should use a single axis to also have same representation of the unit, otherwise perhaps a peak on the span is not same as a peak in the metric and that may lead some confusion.

lucasponce:
Another comment:
image

I don't know if we can use some z-level to bring the metrics to the front.
It seems that spans will populate a lot (which is fine) but I wonder if we can use some trick to put it more in the background.

In the previous example, I'd like to compare the metric with the span. As metric is represented with a line, I guess it has difficulties to be seen correctly with all the circles.

But perhaps just a z-level can help here.

lucasponce:
I know this doesn't depend on this PR as it's something may be need changes in the victory/pf4 framework, but it would be nice if the mouse pointer changes to let user know that "legend" can be enabled on/off.

(This comment is just to remind about it, but no action needed, and it's clear outside of this work)

lucasponce:
We have commented that probably size on spans is difficult to interpret as it represents the "size of the parent trace of that span".

Perhaps for metrics, we want to focus on spans only (without size) as we will have the "traces" tab next in the same location:

image

jotak:
> When we have two different units, it's fine to have both y-axis in left/right with different units.
> 
> When both metric + spans are represented with same unit, it should use a single axis to also have same representation of the unit, otherwise perhaps a peak on the span is not same as a peak in the metric and that may lead some confusion.

=> PR opened in k-charted: https://github.com/kiali/k-charted/pull/61

jotak:
@lucasponce , I've updated to have a fixed & smaller sized spans, see how it looks:

Capture d’écran de 2019-12-04 15-32-22

@hhovsepy I'm still not able to reproduce the broken refresh, but I've made another try in the last commit to fix it, hope it will.

lucasponce:
> @lucasponce , I've updated to have a fixed & smaller sized spans, see how it looks:

It looks great :-)

> @hhovsepy I'm still not able to reproduce the broken refresh, but I've made another try in the last commit to fix it, hope it will.

I couldn't see that error, either, but I was testing a local instance only.

hhovsepy:
@lucasponce @jotak can you try to reproduce that problem with maistra?

jotak:
@hhovsepy that's my plan :)

hhovsepy:
@jotak now it works fine with auto refresh.

lucasponce:
@hhovsepy @aljesusg thanks ! if it's ok, and it's ready from @jotak side I can start preparing k-charted and we can try to consolidate.

Just let me know.

jotak:
@lucasponce ok you can go on

lucasponce:
k-charted 0.4.1 released

jotak:
I've bumped to k-charted 0.4.2-rc0 ; I think this PR can be merged (as well as the backend one) ; I've added a maintenance task to bump to release: https://github.com/kiali/kiali/issues/2016

aljesusg:
How are the times displayed? UTC? LocalTime?

jotak:
@aljesusg it's local time iirc
It hasn't changed with this PR

jotak:
@lucasponce this too can be merged

PR #1561, opened 2019-11-22
Fix health tooltip so that first two tooltips show in Virtual Scroll.

mtho11:
** Describe the change **
The first two health icons do not have tooltips (all the rest show up fine).

** Issue reference **
fixes: https://github.com/kiali/kiali/issues/1935


abonas:
was something similar merged? why is this closed?

jshaughn:
@mtho11 did you see Alissa's question ^?

mtho11:
Ah, sorry I missed that one. I closed that at the start of break because my approach didn't work and I didn't want to leave it out there for a couple of weeks for folks to look at and get confused. I'll either reopen or create a new PR ounce a satisfactory solution is found.

PR #1562, opened 2019-11-22
Introduce 'is present' syntax sugar for Matching Routing

lucasponce:
Fixes kiali/kiali#1368

This PR introduces a "sugar syntax" in the Mathing Routing, declaring a "pseudo-operator" called "is present" that is translated to a generic regex were creating rules:

image

Note that we don't propagate the "is present" operator in all layers, because Wizard creates VirtualServices and DestinationRules that user can edit manually, so at the end user needs to know that "is present" === "/^.*$/" but I guess it's a compromise to help them to edit. Also, this only applies to headers.

Not yet finished, it's just a prototype tested locally.

### DIFF
--- a/src/components/IstioWizards/MatchingRouting.tsx
+++ b/src/components/IstioWizards/MatchingRouting.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import { WorkloadOverview } from '../../types/ServiceInfo';
 import Rules, { MOVE_TYPE, Rule } from './MatchingRouting/Rules';
 import RuleBuilder from './MatchingRouting/RuleBuilder';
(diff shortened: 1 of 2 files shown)


hhovsepy:
This is how the VS details looks like.
Screenshot from 2019-11-22 18-47-39
The Camel case is not related to this PR.
Screenshot from 2019-11-22 18-50-25
Screenshot from 2019-11-22 18-51-27

lucasponce:
@hhovsepy a minor change in the regex

hhovsepy:
Verified:
Screenshot from 2019-11-29 10-13-31
Screenshot from 2019-11-29 10-13-55

PR #1567, opened 2019-11-26
Add support for Node.js 12 LTS 

mtho11:
** Describe the change **
Now that Node.js 12.x is LTS we need to support that as well as the existing LTS which is 10.x

** Issue reference **
fixes: kiali/kiali#1946

** Backwards compatible? **
Looks like it is backward compatible with 10.x

No

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 dist: trusty
 node_js:
-  - '10'
+  - '12'
 cache:
   yarn: true
   directories:
--- a/package.json
+++ b/package.json
@@ -118,7 +118,7 @@
     "@types/enzyme": "3.10.3",
     "@types/jest": "23.3.10",
     "@types/lodash": "4.14.123",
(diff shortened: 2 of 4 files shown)


mtho11:
@jotak there are 3 failing tests in OverviewPage.test.tsx that I'm just skipping for now (with `xit`)

mtho11:
It looks like we can maintain node.js 10.x and 12.x if we can find a way to fix the 3 failing tests in a compatible way.

jotak:
@mtho11 I've just sent a commit for you here : https://github.com/mtho11/swsui/pull/29

pbajjuri20:
Tested and passed on #Kiali-ui 1.11.0 (893fb47217ee15edbb5f96c10d634b016903ad2b)

mtho11:
@israel-hdez that is great! We can drop support for Node.js 10.x as soon as you tell us the releases support 12.x!

PR #1570, opened 2019-11-28
Reset state when Wizard is closed

lucasponce:
Fixes kiali/kiali#1993


lucasponce:
Now the fix should be good.
Let me know @hhovsepy if this solves the issue without any corner case.

PR #1571, opened 2019-11-29
Fix missing icon and wrap phase status

lucasponce:
Fixes kiali/kiali#1989
Fixes kiali/kiali#1977


lucasponce:
@hhovsepy I've decided to add a related commit for style and fixe the kiali/kiali#1977 here as well.

Please, let me know if that is ok, I think it's minor and it can go in a single PR.

PR #1572, opened 2019-11-29
Implement token authentication

israel-hdez:
This is providing a new login screen with only one "Token" field. The user is expected to paste a service account token to login into Kiali.
    
The UI will pass the provided token to the back-end. In turn, the back-end will provide a cookie as usual (if the Token is valid). The back-end should use the provided token during the session of the user to make calls to the Kubernetes/OpenShift API, so that the cluster's RBAC takes effect.

image
    
Fixes kiali/kiali#1139

### DIFF
--- a/src/pages/Login/LoginPage.tsx
+++ b/src/pages/Login/LoginPage.tsx
@@ -1,7 +1,18 @@
 import React from 'react';
 import { connect } from 'react-redux';
 import { ThunkDispatch } from 'redux-thunk';
-import { Button, ListItem, LoginFooterItem, LoginForm, LoginPage as LoginNext } from '@patternfly/react-core';
+import {
+  ActionGroup,
+  Button,
+  Form,
+  FormGroup,
+  FormHelperText,
+  ListItem,
+  LoginFooterItem,
+  LoginForm,
+  LoginPage as LoginNext,
+  TextInput
+} from '@patternfly/react-core';
 import { ExclamationCircleIcon, ExclamationTriangleIcon } from '@patternfly/react-icons';
 import { KialiAppState, LoginSession, LoginStatus } from '../../store/Store';
 import { AuthStrategy } from '../../types/Auth';
@@ -25,6 +36,7 @@ type LoginState = {
   password: string;
   isValidUsername: boolean;
   isValidPassword: boolean;
+  isValidToken: boolean;
   filledInputs: boolean;
   showHelperText: boolean;
   errorInput?: string;
@@ -42,6 +54,7 @@ export class LoginPage extends React.Component<LoginProps, LoginState> {
       password: '',
       isValidUsername: true,
       isValidPassword: true,
+      isValidToken: true,
       filledInputs: false,
       showHelperText: false,
       errorInput: ''
@@ -72,6 +85,25 @@ export class LoginPage extends React.Component<LoginProps, LoginState> {
     if (authenticationConfig.strategy === AuthStrategy.openshift) {
       // If we are using OpenShift OAuth, take the user back to the OpenShift OAuth login
       window.location.href = authenticationConfig.authorizationEndpoint!;
+    } else if (authenticationConfig.strategy === AuthStrategy.token) {
+      if (this.state.password.trim().length !== 0 && this.props.authenticate) {
+        this.props.authenticate('', this.state.password);
+        this.setState({
+          showHelperText: false,
+          errorInput: '',
+          isValidToken: true,
+          filledInputs: true
+        });
+      } else {
+        let message = 'Please, provide a Service Account token.';
+
+        this.setState({
+          showHelperText: true,
+          errorInput: message,
+          isValidToken: false,
+          filledInputs: false
+        });
+      }
     } else {
       this.setState({
         isValidUsername: !!this.state.username,
@@ -155,13 +187,15 @@ export class LoginPage extends React.Component<LoginProps, LoginState> {
 
     const messages = this.getHelperMessage();
     const isLoggingIn = this.props.isPostLoginPerforming || this.props.status === LoginStatus.logging;
+    const isLoginBtnDisabled =
+      isLoggingIn || (this.props.postLoginErrorMsg !== undefined && this.props.postLoginErrorMsg.length !== 0);
 
     // Unfortunately, typescripg typings are wrong in the PatternFly
     // library. So, this casts LoginForm as "any" so that it is
     // possible to use the "isLoginButtonDisabled" property.
-    const Form = LoginForm as any;
+    const CredentialsForm = LoginForm as any;
     const loginForm = (
-      <Form
+      <CredentialsForm
         usernameLabel="Username"
         showHelperText={this.state.showHelperText || this.props.message !== '' || messages.length > 0}
         helperText={<>{messages}</>}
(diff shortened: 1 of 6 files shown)


src/pages/Login/LoginPage.tsx
```diff
@@ -155,13 +187,15 @@ export class LoginPage extends React.Component<LoginProps, LoginState> {
 
     const messages = this.getHelperMessage();
     const isLoggingIn = this.props.isPostLoginPerforming || this.props.status === LoginStatus.logging;
+    const isLoginBtnDisabled =
```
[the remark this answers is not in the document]
jmazzitelli: +1 ! It isn't so hard to type out the additional 3 characters for Button :) and it makes the code more readable.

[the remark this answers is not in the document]
mtho11: Yeah, this isn't `golang` no need to abbreviate. 😝

src/pages/Login/LoginPage.tsx
```diff
@@ -155,13 +187,15 @@ export class LoginPage extends React.Component<LoginProps, LoginState> {
...
+      isLoggingIn || (this.props.postLoginErrorMsg !== undefined && this.props.postLoginErrorMsg.length !== 0);
 
     // Unfortunately, typescripg typings are wrong in the PatternFly
     // library. So, this casts LoginForm as "any" so that it is
     // possible to use the "isLoginButtonDisabled" property.
-    const Form = LoginForm as any;
+    const CredentialsForm = LoginForm as any;
```
mtho11: This could be a little bit more typesafe.

src/services/Api.ts
```diff
@@ -76,13 +77,17 @@ export const extendSession = () => {
 };
 
 export const login = async (
-  request: LoginRequest = { username: ANONYMOUS_USER, password: 'anonymous' }
+  request: LoginRequest = { username: ANONYMOUS_USER, password: 'anonymous', token: '' }
 ): Promise<Response<LoginSession>> => {
+  const params = new URLSearchParams();
```
mtho11: not sure if `const` conveys the right thing here. As the next line mutates the `params`.

jmazzitelli:
So far, working for me. There is one thing that I think can be better, but I do not know how to fix it :) When I log in with a token, the "username" in the top right just says "token":

image

Is there no way for us to determine what service account or user this belongs to? If there is not, then is there some way we can ask the user what they want to be called? Because just looking at "token" I (as a user) have no way of knowing what user this is and thus I am not sure what permissions I should have.

Maybe have a "Username" field in the login screen (or maybe call it something else since it is just a label you give your token - it does not have to have any relationship to the token's service account or user - maybe "Token Label" or "What is your Name?" or something). This way, when you log in, that top right field does not just say "token" but has a more human-understandable value - even if it is just a custom name that the user typed in at login time.

src/pages/Login/LoginPage.tsx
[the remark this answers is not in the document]
israel-hdez: Fixed! No more abbreviations.

src/pages/Login/LoginPage.tsx
```diff
@@ -72,6 +85,25 @@ export class LoginPage extends React.Component<LoginProps, LoginState> {
...
+      if (this.state.password.trim().length !== 0 && this.props.authenticate) {
+        this.props.authenticate('', this.state.password);
+        this.setState({
+          showHelperText: false,
+          errorInput: '',
+          isValidToken: true,
+          filledInputs: true
+        });
+      } else {
+        let message = 'Please, provide a Service Account token.';
```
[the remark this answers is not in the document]
israel-hdez: Yes. I changed it.

src/pages/Login/LoginPage.tsx
israel-hdez: I see the typing issue on LoginForm is no longer present. So, I removed that cast, and also the preceding comment.

src/services/Api.ts
israel-hdez: Typescript transpiler is not complaining 🤷‍♂️
I think `const` is only avoiding reassignment.

So, you choose. I can change to `let` if it looks better.

israel-hdez:
> Is there no way for us to determine what service account or user this belongs to?

@jmazzitelli I just found that the ServiceAccount token has this data encoded in it. :)
I'll do the change in the back-end PR.

src/pages/Login/LoginPage.tsx
```diff
@@ -72,6 +85,25 @@ export class LoginPage extends React.Component<LoginProps, LoginState> {
     if (authenticationConfig.strategy === AuthStrategy.openshift) {
       // If we are using OpenShift OAuth, take the user back to the OpenShift OAuth login
       window.location.href = authenticationConfig.authorizationEndpoint!;
+    } else if (authenticationConfig.strategy === AuthStrategy.token) {
```
cfcosta: I think we are reaching a point here where either polymorphism or at least a `switch` case would be the better idea here...

src/pages/Login/LoginPage.tsx
```diff
@@ -72,6 +85,25 @@ export class LoginPage extends React.Component<LoginProps, LoginState> {
...
+      if (this.state.password.trim().length !== 0 && this.props.authenticate) {
+        this.props.authenticate('', this.state.password);
+        this.setState({
+          showHelperText: false,
+          errorInput: '',
+          isValidToken: true,
+          filledInputs: true
+        });
+      } else {
+        const message = 'Please, provide a Service Account token.';
```
cfcosta: Is this comma necessary? (commas in english are weird)

src/services/Api.ts
cfcosta: +1 to changing it to `let`. Also, iirc `const` only guards against reassignment, you need to freeze the object to protect against mutations (which only happens at runtime :D).

src/services/Api.ts
```diff
@@ -68,6 +68,7 @@ const newRequest = <P>(method: HTTP_VERBS, url: string, queryParams: any, data:
 interface LoginRequest {
   username: UserName;
   password: Password;
+  token: Password;
```
cfcosta: I think the token could be optional instead of being set to `''` every time, what do you think?

src/pages/Login/LoginPage.tsx
jshaughn: The comma should be removed, but not a blocker, only if there is another revision.

jmazzitelli:
I just rebuilt and ran it in my browser. I still see "token" at the top right.

jmazzitelli:
OK, this is interesting. If I get a service account token, I see its name in the top right.
But if I login with a user account (on OpenShift, do "oc login" to login as a user and then do "oc whoami -t" to get the user's token), I see "token" in the top right.

This probably isn't a big deal because if you are on OpenShift then you most likely aren't using token strategy but instead just using openshift strategy (and then you can log in as the user normally).

src/pages/Login/LoginPage.tsx
israel-hdez: I agree. But given that this doesn't change a lot, let's wait. I'll be :+1: if one more auth method comes in.

israel-hdez:
@jmazzitelli 

> This probably isn't a big deal because if you are on OpenShift then you most likely aren't using token strategy but instead just using openshift strategy (and then you can log in as the user normally).

I agree :) Moreover, as it's mentioned in the PR description, a ServiceAccount token is expected. Side effect is that it also works with OS user tokens. 

I'm merging as is.

@cfcosta I won't dismiss your other two comments. I'll take note to fix on my next PR and merge this one as is.

PR #1574, opened 2019-12-05
Update Wizards to use FQDN

lucasponce:
Fixes kiali/kiali#1879

This PR allows Wizards to use FQDN like "reviews.bookinfo.svc.cluster.local" instead of short names "reviews" for the VirtualServices and DestinationRules generated by Kiali Actions.

### DIFF
--- a/src/components/IstioWizards/IstioWizard.tsx
+++ b/src/components/IstioWizards/IstioWizard.tsx
@@ -15,6 +15,7 @@ import SuspendTraffic, { SuspendedRoute } from './SuspendTraffic';
 import { Rule } from './MatchingRouting/Rules';
 import {
   buildIstioConfig,
+  fqdnServiceName,
   getInitGateway,
   getInitHosts,
   getInitLoadBalancer,
@@ -37,7 +38,6 @@ import ThreeScaleIntegration from './ThreeScaleIntegration';
 import { ThreeScaleServiceRule } from '../../types/ThreeScale';
 import GatewaySelector, { GatewaySelectorState } from './GatewaySelector';
 import VirtualServiceHosts from './VirtualServiceHosts';
-import { serverConfig } from '../../config';
 
 const emptyWizardState = (fqdnServiceName: string): WizardState => {
   return {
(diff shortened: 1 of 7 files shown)


lucasponce:
Please, @xeviknal @hhovsepy, I have tested locally and I don't see any regression, but I'd like if you can test this one with more scenarios, to check that validations or other potential cornercase is not affected.

Let me know, thanks !

src/components/IstioWizards/IstioWizard.tsx
```diff
@@ -74,7 +75,7 @@ const emptyWizardState = (serviceName: string): WizardState => {
 class IstioWizard extends React.Component<WizardProps, WizardState> {
   constructor(props: WizardProps) {
     super(props);
-    this.state = emptyWizardState(props.serviceName);
+    this.state = emptyWizardState(props.serviceName + '.' + props.namespace + '.' + serverConfig.istioIdentityDomain);
```
xeviknal: What do you think about moving this operation into a new function?

src/components/IstioWizards/IstioWizardActions.ts
```diff
@@ -82,6 +82,10 @@ const SERVICE_UNAVAILABLE = 503;
 
 export const KIALI_WIZARD_LABEL = 'kiali_wizard';
 
+export const fqdnServiceName = (serviceName: string, namespace: string): string => {
+  return serviceName + '.' + namespace + serverConfig.istioIdentityDomain;
```
xeviknal: Am I having a configuration issue or there is one `.` missing between the namespace and the domain?

lucasponce: Ouch ! you are absolutely right, my bad !!

prachiyadav:
while creating a route of a service if we change/edit the hostname it is not updating the YAML file with the new hostname, instead of that it is showing default hostname. However, we tried to update the same route we can see the edited host name. Please find the attached screenshot for the same.
Screenshot from 2019-12-06 14-23-01
Screenshot from 2019-12-06 14-24-20
Screenshot from 2019-12-06 14-25-15

lucasponce:
Thanks @prachiyadav for your comments, let me review that use case locally to see if a fix is missing in this PR.

lucasponce:
@prachiyadav I think you are seeing the correct behaviour.
The "spec/hosts" field is not the same as "spec/http/route/destination/host" and may be different.
For example, typically the "spec/hosts" can be refered to a wildcard or a dns domain meanwhile the "spec/http/route/destination/host" defines a target service.

hhovsepy:
@lucasponce  we thought that Virtualservice Hosts" field is switched to set values to "spec/http/route/destination/host".

lucasponce:
Great.
Just to document this.
"Advanced Options" -> "VirtualService Hosts" maps with the "spec/hosts" for the VS:
https://istio.io/docs/reference/config/networking/virtual-service/#VirtualService

And in "Weight" the workloads map with the "Destination" host:
https://istio.io/docs/reference/config/networking/virtual-service/#Destination

PR #1575, opened 2019-12-05
Graph Replay

jshaughn:
I have an initial working POC for graph replay and so am creating this draft PR for discussion, to help deliver a first-pass at a replay feature.

Fixes https://github.com/kiali/kiali/issues/1421

Here's a video:
kiali#1421

Some initial feedback after scrum demo:
- need more time indication on the slider: start/end ticks, times as opposed to frame numbers
  - +1, I'll see what I can do
- confusion around "last" terminology for duration and "ago" terminology for replay interval
  - +1, may initially just try adding a help popup
- idea: launch custom time range graph from another view like chart or timeline
  - +1, would be cool for correlation.  The plumbing is basically there but we'd probably need url bookmarkability for the start/end time.
- idea: FF/REW controls?
  - +1, could be interesting as real-time refresh can be slow.
- comment: would a "brush" be better than a slider?
  - I'm not really sure, maybe, but I'm not sure how that would work.  Not planning to change for initial feature.
- need more flexible time setting
  - +1, but we don't have a time picker.  Not sure if a brush approach could be helpful.
- still needs UX pass
  - +1


abonas:
Great to see this.
Definitely needs UX team inputs.
Some thoughts, trying not to repeat the input that is already listed in the top description comments:
1. "stop replay" might better be a square icon and not an X. X is typically close or delete. 
2. Play/replay would be more clear if represented by a triangle "play" icon as in many music/video apps.
3. I think that having displayed at the same time the past replay time selection and the regular (what exists today) controls on the right side of the screen (the frequency and the "last X min..." part) is confusing.
4. Replay also needs a frequency/skip setting unless I missed it. In other words, when replaying, what are the "jumps" between frames. For example looking at yesterday and seeing a replay that shows the situation in skips of 10 minutes (data points of every 10 minutes)

jshaughn:
Here is a video of the latest POC.   It tries to make the replay controls more intuitive and adds some more options.  The main weakness is still around establishing the replay "window".  Without a decent date/time picker from PF4, I'm not sure of a good way to establish the start/end times.  So the current approach relies on endTime being set to "now" and the start time being an offset.  I'm sure this can be improved in the future, but I'm mainly wondering if this will be OK in a first release of the feature.

kiali#1421-2

jshaughn:
The latest video:   Updated control layouts for left-to-right organization.  Start with a default replay window.  Allow to shift the window by setting start time into the past via modal.

kiali#1421-3

jshaughn:
Closing this PR as the work has moved to a new branch.

PR #1577, opened 2019-12-10
Replace traces iFrame with a Kiali React component

aljesusg:
Require https://github.com/kiali/kiali/pull/2017

Backend changes:

- UrlEndopoint
- Integration (Check if kiali has integration internally)
- NamespaceSelector (Let user find traces by service wothout the namespace selector)
- IntegrationMessage ( Message about why kiali is not integrated with jaeger)

## UI

1. JaegerTracing Menu now link to the Jaeger UI, if there is not Jaeger URL the user won't see the menu item

2. In traces tab in service
   
    - If user has integration the user will see the images below
    - If user has not integration but he have the jaegerURL, the user will be linked to the traces for this service

3. The new traces tab
    -  User can click the point and get  the information related with this trace
    -  Still working in the information displayed
    -  The points will be red if the user has error traces ( We are thinking about to have a range of colors for other cases like the trace has more duration)
    - If the user has the jaeger Url set he will see a link to   the jaeger UI Span information

### Chart with Traces

- PF4 Charts from K-charted with legend to select only error traces or all traces
- Filters added:
    - Show Errors
    - Limits results (We provided the limits)
    - Span Interval (We check the duration of the results to provided the best options for the user, the range can be 2,,5 or 10)
    -Status Code, we added the error status codes that we think are useful for the user.
    - Link to see the same results in JaegerUI and perform the search

Screenshot from 2020-01-13 15-27-05

### OnClick a trace you will see the information about the trace

- We display the number of spans, errors, duration and a link to check them in JaegerUI
- We sorted the spans by operation showing which one have errors.

Screenshot from 2020-01-13 15-27-17

### You can click a ServiceOperation to check the spans

- We created a customWrapper to add a red border if the span has the error and an icon.
- Link to see the traces of the service Name
- For each span we can see the service name, duration, metrics of that service or the logs of the node that created the span
Screenshot from 2020-01-13 15-27-30

### You can see the tags of a span
- We can expand the row to check the tags and processes that the span has.
- We can copy the json value of each tag
- We show the values with different colors depending of the type of data (number string...)
- For the tag error we display the true value with red color.
Screenshot from 2020-01-13 15-27-48

## Trace without Errors
- This is a sample trace without error
Screenshot from 2020-01-13 15-29-04


src/components/JaegerIntegration/JaegerScatter.tsx
```diff
@@ -0,0 +1,68 @@
...
+      datapoints.push({
+        x: seed,
+        y: Math.random(),
+        name: `Trace ${_i}`,
+        unit: 'seconds',
+        size: 8
+      });
+      seed += 1000000;
+    }
+
```
aljesusg: Hi @jotak maybe you can help me here, I used the same sample that we have in the storybook https://github.com/kiali/k-charted/pull/66 (It's working in onClick method)

src/components/JaegerIntegration/JaegerScatter.tsx
```diff
@@ -0,0 +1,68 @@
...
+    };
+    return (
+      <ChartWithLegend
+        data={[data]}
+        unit="seconds"
+        fill={true}
+        seriesComponent={<ChartScatter />}
+        onClick={dp => alert(dp.name)}
+      />
+    );
```
aljesusg: but in the Scatter chart the onClick method show the information of other point :(

jotak: As discussed, this needs more investigation but if we don't find a fix, there's an alternative solution to compute ourselves the closest point to cursor among all the data points in chart.

aljesusg:
Last Update 

Screenshot from 2019-12-13 11-58-07

lucasponce:
I've tried to setup and I have a first problem:
image

Perhaps @aljesusg you have something in local not yet committed.

aljesusg:
now we show jaeger errors

## When there is no traces
notraces

### when there is some error
Screenshot from 2019-12-13 16-36-04

lucasponce:
I will start adding some comments as soon as review:
image

This input box should be a fixed dropdown. For example, if I add a high number of traces, I can easily block Kiali, so I guess it makes sense to add some fixed value to select (10, 20, 100). 

For more things, then user can go to Jaeger.

lucasponce:
This seems some issue:
image

The filter is between min span 5 ms and max span 10 ms but there are traces greater.

lucasponce:
It would be nice that spans could be also "linkable" 
image
and allow to browse to specific service.

That's a low effort that can add a lot of value here.

lucasponce:
I wonder if now that in this page we show in a first level the trace:
image

We could enrich this and when clicking in a specific span we could also to show the details of the span with a link as well if the span is a kiali service to browse between spans.

I think that info is collected in the same query and it could be interesting to show it.

Also, this is just a preliminar step, the idea is to prepare the foundation to correlate this information with other Kiali's info, like details of the service and also specific Istio config that may affect to a specific span, but that is something to iterate later.

lucasponce:
Also, I think that Kiali should try to provide a better context between span and service:
image

I mean, the main "trace" is called productpage, but we are under the ratings service, so in this case perhaps we can show/mark which is the span that belongs to this service, just to help user to put in context (different label color, format, I don't know, this also could be interesting to get some UX help).

lucasponce:
I'm also wonder if we can calculate and prepopulate the min / max values to avoid introducing text in an input:
image

Also, I would address user to jaeger for advanced tagging searching

image

I guess we can remove the tag search box and maintain some built-in queries (errors, max traces, etc).

So, in this built-in queries is where Kiali can add value, because in next iterations we can correlate with metrics and other signals and prepare a "built-in" with that unique data.

But if user wants to refine a jaeger search using custom tags, they should do that inside Jaeger, not Kiali.

lucasponce:
It seems there are some some issue when clicking in the trace:
image

Tooltip information doesn't match with trace details.

aljesusg:
> It seems there are some some issue when clicking in the trace:
> image
> 
> Tooltip information doesn't match with trace details.

@jotak I found the problem, in the onclick method of ChartWithLegend. Why are you looking in a ratio? I think that you can't use onclick without click a point because there is not target.

About why storybook runs ok is by the type of the X, the series return the values for X of type ==> Mon Dec 16 2019 18:30:25 GMT+0100 and the ratio applied in my case is  + 0.43 * 168418 so the method can't find the point.

```typescript
onClick: function (event, target) {
                        var series = target.data;
                        var pos = event.clientX - padding.left;
                        var size = _this.state.width - padding.left - padding.right;
                        var ratio = pos / size;
                        var numFunc = (typeof series[0].x === 'object' ? function (x) { return x.getTime(); } : function (x) { return x; });
                        var xLength = numFunc(series[series.length - 1].x) - numFunc(series[0].x);
                        var clickedX = numFunc(series[0].x) + ratio * xLength;
                        console.log(series[0].x);
                        console.log('ratio');
                        console.log(ratio);
                        console.log('length'):
                        console.log(xLength);

                        console.log('Clicked');
                        console.log(clickedX);

                        console.log()
                        // Find closest point
                        var closest = series.reduce(function (p, c) {
                            if (p === null) {
                                return c;
                            }
                            var dist = Math.abs(clickedX - numFunc(c.x));
                            var prevDist = Math.abs(clickedX - numFunc(p.x));
                            return dist < prevDist ? c : p;
                        });
                        _this.props.onClick(closest);
                        return [];
                    }
```

jotak:
@aljesusg I've created this: https://github.com/kiali/kiali/issues/2039

aljesusg:
> I think you already discussed this with @jotak:
> image
> 
> But time format in the x-axis is odd, not fully clear if it's seconds/minutes.

related with https://github.com/kiali/k-charted/pull/66#issuecomment-566448555

src/components/JaegerIntegration/JaegerResults/JaegerTraceTitle.tsx
```diff
@@ -0,0 +1,41 @@
...
+  onClickLink: string;
+}
+
+const DEFAULT_DURATION_PERCENT = 0;
+
+const durationBar = style({
+  backgroundColor: '#d7e7ea'
+});
+
+export class JaegeTraceTitle extends React.Component<JaegerScatterProps> {
```
jotak: typo: `JaegerTraceTitle`

src/app/AuthenticationController.tsx
```diff
@@ -130,17 +130,25 @@ class AuthenticationController extends React.Component<AuthenticationControllerP
   };
 
   private setJaegerInfo = (jaegerInfo: JaegerInfo) => {
-    let jaegerState: JaegerState = { jaegerURL: '', enableIntegration: false, namespaceSelector: true };
+    let jaegerState: JaegerState = {
+      jaegerURL: '',
+      integration: false,
+      namespace_selector: true,
```
jotak: that should be camel case I guess, no? (as we do everywhere)

src/components/JaegerIntegration/JaegerResults/transform.ts
```diff
@@ -0,0 +1,288 @@
+import _isEqual from 'lodash/isEqual';
+import _round from 'lodash/round';
+import moment from 'moment';
+import { KeyValuePair, Span, SpanData, JaegerTrace, TraceData } from '../../../types/JaegerInfo';
+
```
jotak: It seems this file comes mainly from Jaeger console, correct?
Could you put a comment & link to the original? So that if we have to dig into this code, we know where it refers to

src/components/JaegerIntegration/JaegerScatter.tsx
jotak: sorry I missed this comment.
I suspect that the time conversion is wrong, either here, or when you fetch data. Because even when I select something like "last 30 minutes", I only see plots for the last minute.

src/components/JaegerIntegration/RouteHelper.ts
```diff
@@ -37,112 +22,33 @@ const convTagsLogfmt = (tags: string) => {
...
-  const end = `${endDate} ${endDateTime}`;
-  return {
-    start: `${moment(start, 'YYYY-MM-DD HH:mm')
-      .utc()
-      .valueOf()}000`,
-    end: `${moment(end, 'YYYY-MM-DD HH:mm')
-      .utc()
-      .valueOf()}000`
+export const getQueryJaeger = () => {
+  let params: any = {
```
jotak: When fetching the spans I created the `TracingQuery` type (in types/Tracing.ts), thinking it could be used also for traces queries. I guess it could be reused here (you can extend it with whatever is needed, like limit / tags / etc)

src/components/JaegerIntegration/JaegerScatter.tsx
jotak: Ok, I finally get it, and I better understand why @lucasponce was seeing weird x-axis plots in this comment: https://github.com/kiali/kiali-ui/pull/1577#issuecomment-565505249

This happens when there are many traces produced, because we're limiting to a small number (20 by default) so only the 20 last traces are shown. So, say we have 10 traces per second and we ask for traces in the last 30 minutes, the X axis will show only the 2 last seconds, making it difficult to understand.

I would say we have 2 problems here:
- the chart range adapts automatically to the data it contains, regardless the queried time frame. So we ask for 30 minutes but it can show much less than that, due to limiting. There's already an issue opened in k-charted about that: https://github.com/kiali/k-charted/issues/15 ; we could do that.

- when there's many traces, the query will return only the last ones. I don't know how this is worked around in Jaeger UI, but it doesn't seem to happen there: it looks like the traces are randomly ordered and not ordered by date. In the metrics page with spans overlay, I tried to reduce this problem by showing only spans above some percentile, but it's still imperfect.

lucasponce:
@aljesusg I'm testing the latest kiali/kiali#2017 and this one.

Adding some comments:

image

image

In my example the title is confusing. It says "Error Traces (8)" but even trying to find them in the chart is hard.

Probably those 8 error traces are out the boundaries, that may create some confusion I think.

lucasponce:
Another comment, the detailed view requires headers:
image

Would be nice to have a table there, and not to bring on each list a "Duration:" for example

lucasponce:
Minor but for UI consistency:
image
And
image
Should have the icon of an external link, like:
image

At least to use the same way to represent an external link (then tomorrow same way can be changed but at least to apply same design in same external links).

lucasponce:
In the detail of a trace:
image

It points to:
image

And then it splits the spans between traces.
It could be nice if "error" traces are marked to bring attention to the user and link to jaeger UI for more details if necessary.

Otherwise the number of errors lacks some additional info.

lucasponce:
Style looks also non consistent here:
image

aljesusg:
> @aljesusg I'm testing the latest [kiali/kiali#2017](https://github.com/kiali/kiali/pull/2017) and this one.
> 
> Adding some comments:
> 
> image
> 
> image
> 
> In my example the title is confusing. It says "Error Traces (8)" but even trying to find them in the chart is hard.
> 
> Probably those 8 error traces are out the boundaries, that may create some confusion I think.
> 
> Ok, are out of the time selected:
> 
> image
> 
> But not sure about how we can improve the title, I guess. I don't have a strong suggestion, still thinking on it.
> 
> Perhaps the Error Traces (X) should honor the same interval queried.

About this actually we are checking the last 100 traces in the last hour for errors.

So the options are:

- Show the errors in the interval showed.
- If there are errors in the last hour show them in the chart (actually we are showing traces of the interval showed)

lucasponce:
> About this actually we are checking the last 100 traces in the last hour for errors.

I think all info showed needs to be synced with the interval chosen by the user, or if the complexity on the title increases, I think it's not enough tradeoff, then I see fine to disable the # Errors in the title if we need to pass more info.

I.e. other tabs don't have this feature and now that we are refactoring this component, the errors would be in the precanned query.

aljesusg:
> > About this actually we are checking the last 100 traces in the last hour for errors.
> 
> I think all info showed needs to be synced with the interval chosen by the user, or if the complexity on the title increases, I think it's not enough tradeoff, then I see fine to disable the # Errors in the title if we need to pass more info.
> 
> I.e. other tabs don't have this feature and now that we are refactoring this component, the errors would be in the precanned query.

Ok so I prefer to show if there are errors in the interval selected by the user instead remove this feature, are you ok with this ?

lucasponce:
Forgot to add that I'm still seeing some inconsistencies when changing the data input and I need to press "refresh" to perform the query.

So perhaps is worth to maintain the "Search" button as well, or everytime the inputs are changed, perform a refresh.

lucasponce:
> Ok so I prefer to show if there are errors in the interval selected by the user instead remove this feature, are you ok with this ?

+1

aljesusg:
LAST UPDATE 

### Chart with Traces

- PF4 Charts from K-charted with legend to select only error traces or all traces
- Filters added:
    - Show Errors
    - Limits results (We provided the limits)
    - Span Interval (We check the duration of the results to provided the best options for the user, the range can be 2,,5 or 10)
    -Status Code, we added the error status codes that we think are useful for the user.
    - Link to see the same results in JaegerUI and perform the search

Screenshot from 2020-01-13 15-27-05

### OnClick a trace you will see the information about the trace

- We display the number of spans, errors, duration and a link to check them in JaegerUI
- We sorted the spans by operation showing which one have errors.

Screenshot from 2020-01-13 15-27-17


### You can click a ServiceOperation to check the spans

- We created a customWrapper to add a red border if the span has the error and an icon.
- Link to see the traces of the service Name
- For each span we can see the service name, duration, metrics of that service or the logs of the node that created the span
Screenshot from 2020-01-13 15-27-30

### You can see the tags of a span
- We can expand the row to check the tags and processes that the span has.
- We can copy the json value of each tag
- We show the values with different colors depending of the type of data (number string...)
- For the tag error we display the true value with red color.
Screenshot from 2020-01-13 15-27-48

## Trace without Errors
- This is a sample trace without error
Screenshot from 2020-01-13 15-29-04

lucasponce:
There is an annoying behaviour when "hovering" on the controls:
image

There is some style change on hover that shows a scrolling bar and creates a bad behaviour.

lucasponce:
image

Metrics and Logs shouldn't have title in the column, but just a tittle like "View metrics" "View logs"

Similar as other tabs:
image

Also, how workload logs are calculated ? Not sure if this is correct:
image

aljesusg:
> image
> 
> Metrics and Logs shouldn't have title in the column, but just a tittle like "View metrics" "View logs"
> 
> Similar as other tabs:
> image
> 
> Also, how workload logs are calculated ? Not sure if this is correct:
> image

If you go to the tags of the span you can check that the node_id is the one where the trace is generated/propagated so ion this one you can check the logs, in other case there is no way to check for example in the reviews which workload you must check, v1, v2 or v3

src/components/JaegerIntegration/RouteHelper.ts
aljesusg: I use some transforms and conversions from jaeger Types this is why I created them in types/JaegerInfo

lucasponce:
I think perhaps this is related to k-charted, but it has to be fixed.
User should know that a element is in the traces chart is "touchable" to show the detail.
Like it was fixed in the metrics + span traces, where "hover" on the legend the mouse pointer is changed.
I'd expect something similar in the charts as well.

lucasponce:
Minor one:
image
Could you move headers as: "Operation, Service, Duration" and next the actions links ?
Also, minor, "View Metrics -> View metrics".

lucasponce:
Also, I have other scenario that is annoying from user perspective.
When I click in the detail, I guess tracing should "pause" de refresh, otherwise, after 10s or similar the details is refreshed and hidden.

Personally I'd like to continue to refresh and maintain the previous detail open (if possible) but if that is tricky, pausing the refresh when examining the detail of a trace also works.

lucasponce:
This one seems a bug:
image

lucasponce:
Also, I think that @jotak introduced in master the time window fix:
image

So, if I select last 30m, I'd expect to see the 30m interval in the x-axis (perhaps it needs to rebase).

aljesusg:
> Also, I think that @jotak introduced in master the time window fix:
> image
> 
> So, if I select last 30m, I'd expect to see the 30m interval in the x-axis (perhaps it needs to rebase).

The problem here is that you set the limit results to 20 so jaeger return the last 20, if you increase these number you will see the last 30 min.

Jaeger returns inside the last 30 minutes the last X (limit results) requested.

aljesusg:
> Also, I have other scenario that is annoying from user perspective.
> When I click in the detail, I guess tracing should "pause" de refresh, otherwise, after 10s or similar the details is refreshed and hidden.
> 
> Personally I'd like to continue to refresh and maintain the previous detail open (if possible) but if that is tricky, pausing the refresh when examining the detail of a trace also works.

It's easy to check, I can check in refreshTraces if tab is traces so...Should we stop refresh if user is not in traces tab ?

aljesusg:
> This "NaN" one seems a bug:
> image
> This happens when changing the limit results.

mmm Annoying, let me check

lucasponce:
I've tested the features and it looks like in good shape:
image

I'm going to review the PR from code perspective to see if there is anything preventing to consolidate this.

lucasponce:
It seems there is a regression in the last commits:
image
When I use the interval span:
image
It looks like that query returned is empty, when it shouldn't be as without filter we see there are traces with that rt.

src/components/DefaultSecondaryMasthead/DefaultSecondaryMasthead.tsx
```diff
@@ -17,7 +17,7 @@ export default class DefaultSecondaryMasthead extends React.Component {
     }
 
     if (path === 'jaeger') {
```
lucasponce: Perhaps with the changes "path == 'jaeger'" won't never happen ? I think.
Unless I miss some cornercase here.

aljesusg: you're right

src/components/JaegerIntegration/JaegerResults/SpanDetail.tsx
```diff
@@ -0,0 +1,97 @@
...
+  };
+
+  goService = (service: string = this.props.spans[0].process.serviceName) => {
+    const ns = service.split('.')[1] || serverConfig.istioNamespace;
+    const srv = service.split('.')[0];
+    return '/namespaces/' + ns + '/services/' + srv;
+  };
+
+  render() {
+    console.log(this.props.spans);
```
lucasponce: Probably this "console.log" entries can go away once the PR is debugged and in good shape.

aljesusg:
> It seems there is a regression in the last commits:
> image
> When I use the interval span:
> image
> It looks like that query returned is empty, when it shouldn't be as without filter we see there are traces with that rt.

MMM I can't reproduce this. Did you check if the output is the same in the Jaeger UI ?

src/components/JaegerIntegration/JaegerResults/JaegerItem.tsx
```diff
@@ -0,0 +1,134 @@
...
+        <JaegerTraceTitle
+          trace={trace}
+          duration={duration}
+          durationPercent={(trace.duration / maxTraceDuration) * 100}
+          onClickLink={this.props.jaegerURL !== '' ? `${this.props.jaegerURL}/trace/${this.props.trace.traceID}` : ''}
+        />
+        <CardBody>
+          <Grid style={{ marginTop: '20px' }}>
+            <GridItem span={2}>
+              <Label isCompact>
```
lucasponce: I find this compact labels hard to read in some standard resolution (100% zoom):
image
I wonder if normal label looks better, but not a blocker, just a comment.

src/components/JaegerIntegration/JaegerResults/SpanTable.tsx
```diff
@@ -0,0 +1,217 @@
...
+import { PfColors } from '../../Pf/PfColors';
+
+interface SpanDetailProps {
+  spans: Span[];
+  namespaceSelector: boolean;
+}
+
+interface SpanDetailState {
+  spanSelected?: Span;
+  isModalOpen: boolean;
```
lucasponce: Now that modals are not used, this property is being used ?

src/components/JaegerIntegration/JaegerResults/SpanTable.tsx
```diff
@@ -0,0 +1,217 @@
...
+    };
+  }
+
+  componentDidUpdate(prevProps: Readonly<SpanDetailProps>): void {
+    if (prevProps.spans !== this.props.spans) {
+      this.setState({ rows: this.getRows() });
+    }
+  }
+
+  goService = (service: string = this.props.spans[0].process.serviceName, extra: string = '') => {
```
lucasponce: Not sure if the caller checks that this.props.spans is always populated, otherwise would send some undefined in the console. It didn't fail in the tests and this case didn't happen, but perhaps add a comment to the caller on these scenarios (for easy maintenance in future iterations).

src/components/JaegerIntegration/JaegerResults/SpanTable.tsx
```diff
@@ -0,0 +1,217 @@
...
+
+  goLogsWorkloads = (workload: string, srv: string) => {
+    const ns = srv.split('.')[1] || serverConfig.istioNamespace;
+    return '/namespaces/' + ns + '/workloads/' + workload + '?tab=logs';
+  };
+
+  getNodeLog = (sp: Span) => {
+    const node = sp.tags.filter(tag => tag.key === 'node_id')[0].value;
+    const srv = sp.process.serviceName.split('.')[0];
+    const regex = new RegExp(`${srv}-v[0-9]*`);
```
lucasponce: I don't this will work in all cases. It's pretty tight to "v1/v2, but one with "version=prod" in Istio will fail.

So, if node_id format is:
sidecar~10.130.2.14~productpage-v1-5fc4d7dbc9-9j6bk.bookinfo~bookinfo.svc.cluster.local

better to match something like "productpage-prod" so RegExp perhaps needs to take anything between two dashes I guess.

src/components/JaegerIntegration/JaegerResults/SpanTable.tsx
```diff
@@ -0,0 +1,217 @@
...
+      return <> We can't find logs</>;
+    }
+  };
+
+  getRows = () => {
+    let rows: (IRow | string)[] = [];
+    this.props.spans.map(span => {
+      let number = rows.push({
+        isOpen: false,
+        props: 'Hi',
```
lucasponce: is this used ?

src/components/JaegerIntegration/JaegerResults/SpanTable.tsx
```diff
@@ -0,0 +1,217 @@
...
+    this.setState({
+      rows
+    });
+  };
+
+  render() {
+    const { columns, rows } = this.state;
+    return (
+      <Table
+        aria-label="Collapsible table"
```
lucasponce: Span Table instead ?

src/components/JaegerIntegration/JaegerResults/SpanTable.tsx
aljesusg: mmm yeah it could be a solution, but if user create some kind of tag like v1-dev or v1-prod we'll have the same problem :(

src/components/JaegerIntegration/JaegerResults/transform.ts
lucasponce: +1
A reference would help to maintain & understand some decisions.

src/components/JaegerIntegration/JaegerScatter.tsx
lucasponce: I think this one is addressed in the lastest commits:
image
image

Due nature of the query of Jaeger traces (where you can select "give me last 20 traces from last 6h" and those "20 traces can be in the first minutes").

There is now an "adjust time" checkbox to see this in the x-asis with the interval asked (i.e. 6h) or just to adjust the x-axis with the size of the query (i.e. last 20 traces).

It's more an experimental way to handle this, I think it would allow user to put in context the results.

src/pages/ServiceDetails/ServiceDetailsPage.tsx
```diff
@@ -182,14 +195,19 @@ class ServiceDetails extends React.Component<ServiceDetailsProps, ServiceDetails
...
       this.setState({ trafficData: null });
       this.fetchBackend();
     }
 
     if (currentTab === trafficTabName) {
       this.fetchTrafficData();
     }
+
+    if (currentTab === tracesTabName) {
+      this.fetchBackend();
```
lucasponce: Do you need to fetch the backend when accessing to traces tab ?

aljesusg: now not, I added this to update the number of errors traces when user go to traces. I'll remove it

lucasponce:
Also another minor one:
image

I think it'd be better to use same wording here as well "View in Jaeger" / "View in Tracing" but in both places.

prachiyadav:
Screenshot from 2020-01-16 00-59-26
Screenshot from 2020-01-16 00-54-38
Screenshot from 2020-01-16 01-02-34

hhovsepy:
No error Traces number or icon on tab:
Screenshot from 2020-01-16 12-03-04

hhovsepy:
The icon of link "View traces" is a style of external service, but in fact the link reloads another service traces, which is confusing:
Screenshot from 2020-01-16 12-05-22

hhovsepy:
Tooltip on copy icon is missing:
Screenshot from 2020-01-16 12-08-53

aljesusg:
> The icon of link "View traces" is a style of external service, but in fact the link reloads another service traces, which is confusing:
> Screenshot from 2020-01-16 12-05-22



> No error Traces number or icon on tab:
> Screenshot from 2020-01-16 12-03-04

Now there is not a number only an icon. Can you check this again ?

aljesusg:
> The icon of link "View traces" is a style of external service, but in fact the link reloads another service traces, which is confusing:
> Screenshot from 2020-01-16 12-05-22

In this case the link should link you to the traces tab of the service details of the namespace bookinfo, I must remove the icon because is not an external link. sorry

aljesusg:
> Tooltip on copy icon is missing:
> Screenshot from 2020-01-16 12-08-53

About this, there are not tooltips here...but it's a good idea , let me add them

hhovsepy:
"Interval Span" select values are interactive, based on the previous selected results?
For instance when it is "none", and there are some traces available, it shows values, but when I select some of these values, which does not have traces, the selection values are reset:
Screenshot from 2020-01-16 14-07-40
Screenshot from 2020-01-16 14-07-44

hhovsepy:
Traces error icon is not shown, and also the pop-over on traces are different:
Screenshot from 2020-01-16 14-13-40
Screenshot from 2020-01-16 14-13-59

aljesusg:
> Traces error icon is not shown, and also the pop-over on traces are different:
> Screenshot from 2020-01-16 14-13-40

About this, you have an error in the backend, when kaili try to get the information about productpage (in thsi endpoint we check the error traces) the backend reutrn {"error":"rules.config.istio.io \"threescale-bookinfo-productpage\" not found"}

> Screenshot from 2020-01-16 14-13-59

About this trace (https://jaeger-istio-system.apps.ocp4-kqe1.maistra.upshift.redhat.com/trace/385f2be93ecd213dfc0889a57bd69ea4),  there is not name, maybe we should indicate like jaeger with <trace-without-root-span>

aljesusg:
Fixed the problem with the span of istio-ingress @hhovsepy we must to deal with this different for be in the istio-system namespace

hhovsepy:
Post merge check, still see this error:
Screenshot from 2020-01-17 15-37-51
Screenshot from 2020-01-17 15-38-06
Screenshot from 2020-01-17 15-38-44

jshaughn:
Congrats, that was a big one!

PR #1580, opened 2019-12-11
Remove link redundancy in cards of Overview page

israel-hdez:
Implements hiding a redundant link in the links area of the cards of the Overview page. Hidden item is redundant because the middle of the card has a text based on the view type which is a link to the relevant list page.

Peek 2019-12-11 14-51

Fixes kiali/kiali#1474

### DIFF
--- a/src/pages/Overview/OverviewCardLinks.tsx
+++ b/src/pages/Overview/OverviewCardLinks.tsx
@@ -20,40 +20,63 @@ const iconStyle = style({
 class OverviewCardLinks extends React.Component<Props> {
   render() {
     const tooltipProps = { distance: 10, entryDelay: 100, exitDelay: 100 };
-    return (
-      <div style={{ marginTop: '10px' }}>
-        <Tooltip key="ot_graph" content={<>Go to graph</>} {...tooltipProps}>
-          <Link
-            to={`/graph/namespaces?namespaces=${this.props.name}&graphType=${this.toGraphType(
-              this.props.overviewType
-            )}`}
-            className={iconStyle}
-          >
-            <TopologyIcon />
-          </Link>
-        </Tooltip>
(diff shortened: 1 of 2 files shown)


mtho11:
@israel-hdez the _blue_ on your screenshots that is _purple_ always makes me think there is something wrong. But I realize it is your screen 😆

lucasponce:
I'm with Joel here.
No blockers to merge this one, but I don't see redundancy here a problem to address.

jshaughn:
Going back to the original Jira, the issue was reported by Alissa and apparently approved by UX.  We could re-ask UX given potential revisions in Patternfly l+f but it did previously have approval and obviously was on the backlog.  @israel-hdez , either schedule a UX review or ask for approval from @jotak .

israel-hdez:
I was going to mention what @jshaughn already said about that the original Jira went through UX and was never rejected. I think we should quickly ask in today's UXD.

jotak:
Ok fair enough. I lost track of the original JIRA, but it seems I said exactly the same thing but this comes from UX so that's all good! :-)

pbajjuri20:
Tested and passed on #Kiali-ui 1.11.0 (799b6f361b61eb810a84b66a60b45dc5f18156de)

serenamarie125:
UX suggests that the link on the bottom of the card should be dynamic.

israel-hdez:
@mtho11 

> @israel-hdez the blue on your screenshots that is purple always makes me think there is something wrong. But I realize it is your screen laughing

My screen shows the right colors, but in the resulting GIF colors aren't right. So, I think the problem is on the screen recording app. Most likely, the app I'm using is color blinded :D

israel-hdez:
I'm merging, because this behavior is confirmed with UXD.

PR #1582, opened 2019-12-16
Add Support to AuthorizationPolicies CRD

xeviknal:
** Describe the change **
Adding AuthorizationPolcies CRD to IstioConfig section
Needs https://github.com/kiali/kiali/pull/2028

** Issue reference **
https://github.com/kiali/kiali/issues/1344


lucasponce:
Not related but it affects to this PR:
image
As the types starts to increase here, I wonder if we may want to oder that list alphabetically for easy search.

PR #1583, opened 2019-12-16
Make Graph Settings display settings more consistent 

mtho11:
The **visiibility** section of the Graph Settings **Display** dropdown doesn't have a label like the **Badges** section.

** Describe the change **

Make GraphSettings display settings more consistent 
by adding a 'Visibility' group heading label to match the 'Badges' group heading label.
It looked a little off without the balance of the visibility section; then they look similar.

** Backwards compatible? **
Yes

No

** Screenshot **
Before:
Kiali_Console

After:
Kiali_Console

### DIFF
--- a/src/pages/Graph/GraphToolbar/GraphSettings.tsx
+++ b/src/pages/Graph/GraphToolbar/GraphSettings.tsx
@@ -168,7 +168,7 @@ class GraphSettings extends React.PureComponent<GraphSettingsProps, GraphSetting
         isGrouped={true}
         selections={selection}
       >
-        <SelectGroup label="" key="visibilityLayers">
+        <SelectGroup label="Show" key="visibilityLayers">
           {visibilityLayers.map((item: VisibilityLayersType) => (
             <SelectOption
               isChecked={item.value}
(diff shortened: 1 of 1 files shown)


abonas:
That would be a good thing to add, but I am wondering if "visibility" is the appropriate name here. the top level menu is already called "display".
Perhaps "Details" would be better?

jshaughn:
+1, not sure about "visibility".   "Details" is OK but I'm not sure each selection really relates to a detail. Maybe just "options"?   Then you read it as:

Display options:
Display badges:

abonas:
> +1, not sure about "visibility". "Details" is OK but I'm not sure each selection really relates to a detail. Maybe just "options"? Then you read it as:
> 
> Display options:
> Display badges:

The thing is, that service nodes and unused nodes are entities, not options.
If anything, when I go over all the 5 items in that section, they are of a very different characteristics, I'm not sure that even grouping them is a good idea. Traffic animation is not the same as displaying entities of type service nodes, for example.

mtho11:
> Traffic animation is not the same as displaying entities of type service nodes, for example.

I still see _traffic animation_ as a graph display option.

mtho11:
A point could be made that **Node Names** is a _badge option_ instead of a _display option_.

mtho11:
I created a design issue for it: https://github.com/kiali/kiali-design/issues/162

mtho11:
@jshaughn updated with UxD comments from today's meeting. Please review.

mtho11:
@mattmahoneyrh since this is just a label change, I don't think it needs QE review.

PR #1585, opened 2019-12-18
Graph Replay

jshaughn:
POC under review for Graph Replay

Fixes https://github.com/kiali/kiali/issues/1421


jshaughn:
The latest video:   Updated control layouts for left-to-right organization.  Start with a default replay window.  Allow to shift the window by setting start time into the past via modal.

kiali#1421-3

cfcosta:
Did some testing on it, verified that it does what it is supposed to do correctly.

- I did variations into the request amount and could verify they changing by time.
- I removed and added new services into the cluster and it did show it appearing/disappearing.
- I forced some errors on code for one of the nodes and the deployment was shown as expected.

I did not test errors induced by Istio (delays, circuit breakers). but there shouldn't be any reason they would not work.

Am verifying the code right now.

src/components/DurationDropdown/DurationDropdown.tsx
```diff
@@ -29,16 +32,26 @@ type DurationDropdownProps = BasicDurationDropdownProps & {
 };
 
 export class DurationDropdown extends React.Component<DurationDropdownProps> {
+  static defaultPrefix = 'Last';
+  private prefixedDurations: Durations;
+
+  constructor(props) {
+    super(props);
+    const prefix = !!this.props.prefix ? this.props.prefix : DurationDropdown.defaultPrefix;
```
cfcosta: shouldn't `this.props.prefix || DurationDropdown.defaultPrefix` work?

src/components/IstioWizards/Slider/BootstrapSlider.tsx
```diff
@@ -45,8 +48,13 @@ class BootstrapSlider extends React.Component<Props> {
       this.props.onSlide(value);
       this.slider.setValue(value);
     };
+    const onSlideStop = value => {
+      value = value >= this.props.max ? this.props.max : value;
```
cfcosta: Since you've used lodash somewhere else: `_.max([this.props.max, value])` sounds better to me, wdyt?

src/components/IstioWizards/Slider/BootstrapSlider.tsx
```diff
@@ -74,12 +82,14 @@ class BootstrapSlider extends React.Component<Props> {
...
+        value = value >= this.props.max ? this.props.max : value;
+        this.props.onSlideStop(value);
+        this.slider.setValue(value);
+      };
       this.slider.on('slide', onSlide);
-      this.slider.on('slideStop', onSlide);
+      this.slider.on('slideStop', onSlideStop);
       this.slider.setAttribute('formatter', this.props.formatter);
-      if (this.props && this.props.orientation && orientation) {
-        console.log('TODELETE orientation check');
```
cfcosta: :D

src/components/IstioWizards/Slider/Slider.tsx
```diff
@@ -60,6 +68,11 @@ class Slider extends React.Component<Props, State> {
     };
   }
 
+  componentDidMount() {
+    // This empty setState forces a re-render which resolves an issue with initial tick_label placement
```
cfcosta: Why? What exactly is the issue here?

src/components/Time/DateTimeInput.tsx
```diff
@@ -0,0 +1,151 @@
...
+interface DateTimeInputProps {
+  minTime: TimeInSeconds;
+  steps?: number;
+  time: TimeInSeconds;
+  tooltip: string;
+
+  onTimeChange: (time: TimeInSeconds) => void;
+}
+
+type DateTimeInputState = {
```
cfcosta: Just for consistency, this should be an interface too.

src/components/Time/DateTimeInput.tsx
```diff
@@ -0,0 +1,151 @@
...
+  customTime: TimeInSeconds;
+  showCustomTime: boolean;
+};
+
+const defaultSteps: number = 100;
+
+const sliderStyle = style({
+  marginBottom: '2em',
```
cfcosta: We have `csx` that we can use for this: `csx.em(2)`, `csx.percent(15)` and so on.

I'm not seeing it being used anywhere else and I have no idea why. @mtho11 do you know why?

src/components/Time/DateTimeInput.tsx
```diff
@@ -0,0 +1,151 @@
...
+const inputStyle = style({
+  width: '11em'
+});
+
+const buttonStyle = style({
+  height: '36px'
+});
+
+export default class DateTimeInput extends React.PureComponent<DateTimeInputProps, DateTimeInputState> {
+  private sliderMin: number;
```
cfcosta: Shouldn't this be in the state?

src/components/Time/LocalTime.tsx
```diff
@@ -1,27 +1,69 @@
...
+  }
+  return date.toLocaleString('en-US', { ...formatOptions, ...options });
+};
+
+export const toRangeString = (
+  start: TimeInMilliseconds,
+  end: TimeInMilliseconds,
+  startOptions?: any,
+  endOptions?: any
+): string => {
```
cfcosta: Dates suck :(

src/components/Time/Replay.tsx
```diff
@@ -0,0 +1,309 @@
...
+
+const stopStyle = style({
+  margin: '0 10px 0 0',
+  height: '37px',
+  width: '10px'
+});
+
+const sliderStyle = style({
+  width: '100%',
+  margin: '0 20px 0 15px'
```
cfcosta: Same as my other command, `csx` has some helpers to make this more typesafe.

src/components/Time/Replay.tsx
```diff
@@ -0,0 +1,309 @@
...
+      isReplaying: false,
+      refresherRef: undefined,
+      replayFrame: Replay.queryTimeToFrame(props),
+      replayFrameCount: Replay.getFrameCount(props.replayWindow.interval),
+      replaySpeed: defaultReplaySpeed
+    };
+  }
+
+  componentDidMount() {
+    if (!!!this.props.replayWindow.startTime) {
```
cfcosta: Wait what.

src/pages/Graph/GraphPage.tsx
```diff
@@ -402,6 +368,7 @@ export class GraphPage extends React.Component<GraphPageProps> {
     if (this.loadPromise) {
       this.loadPromise.cancel();
     }
+    const queryTime: TimeInSeconds | undefined = !!this.props.replayQueryTime ? this.props.replayQueryTime : undefined;
```
cfcosta: shouldn't this be `undefined` by default?

src/pages/Graph/GraphToolbar/GraphToolbar.tsx
```diff
@@ -208,13 +234,33 @@ export class GraphToolbar extends React.PureComponent<GraphToolbarProps> {
       this.props.setEdgeLabelMode(mode);
     }
   };
+
+  /*
```
cfcosta: There's commented code here ;(

jshaughn:
@cfcosta Thanks for the feedback but for now things are still in UX iteration and subject to change, let's hold off on more code review.  I'll address these when it's closer to merge.

jshaughn:
@serenamarie125 I've been looking at [the design feedback ](https://docs.google.com/presentation/d/1G69mOJWQL-t1aECAAv1z4AbuocIbeGQiffWl8YP4qo4/edit#slide=id.g47a44af4cd_0_19) and here are my comments and results:

- I've rearranged things such that:
  - controls are on the left and reorganized
  - text is restrained to tooltips 
- I replaced the Clock icon with the history icon (thoughts?)
- I replaced Play icon with PlayCircle, Pause icon with PauseCircle
- I revised tooltip text and used consistent sentence-case
- I cleaned up the replay progress bar by:
  - removing the the text labels (you can get that with hover)
  - adding ticks for every frame
  - repositioning the play/pause control under the slider
  - adding elapsed time under the slider
- I right-adjusted the Close Replay 'X' and pulled it closer to the slider

What I have is getting closer to Option B than A.  Some suggestions have varying levels of difficulty given the code structure.  We may want to consider a first version and then iterating:

- Moving controls below the graph
  - Because the replay stuff is currently an extension of the entire toolbar, it would take some re-architecting to split it out and under the graph.
- Adding speed control buttons under the slider
  - this may be doable but I didn't want to work on it until after a review of what I have now
- Moving controls up and right and extending the slider to the width of the page
  - may be doable but I feel it may more easily cause wrapping, and it's work as it likely splits up the single "replay" component I have now. Again, didn't want to do it until after a review.
- Set start time slider
  - I agree this slider is not great but I don't want to go down the road of building a datetime picker.  One possibility is to see wif we could wrap a bootstrap component but I really feel like PF should provide a widget.  I did not yet change the color of the setting dot but it's likely doable.
- The visual cue is still up for debate
  - I actually tried the background color change and imho it looks pretty bad given the boundaries of our various components:

image

It would probably look better if we had a bottom margin, but even then, I can't say I like this very much.

Instead I tried something different for your feedback.  In the video below see the dashed border and also the badged chip.  Take a look at the video and we can discuss in the next UX meeting...

kiali#1421-4

src/components/DurationDropdown/DurationDropdown.tsx
jshaughn: I think this one is just a personal preference.  I've gotten into the habit of using !! and !!! to explicitly show that a non-boolean variable is truthy or falsey, respectively.  Let me know if there is a technical reason to avoid this approach.

src/components/IstioWizards/Slider/BootstrapSlider.tsx
jshaughn: +1, may as well use lodash as we already import it. It's more clear. But, actually we want `_.min`.

src/components/IstioWizards/Slider/BootstrapSlider.tsx
jshaughn: No points for these until a DNM label is removed :)

src/components/IstioWizards/Slider/Slider.tsx
jshaughn: I'm not sure what else to say, without the re-render prior to initial display the tick label placement was incorrect.

src/components/Time/DateTimeInput.tsx
jshaughn: [Even though it took me longer than I want to admit] This component will be replaced by a DateTimePicker, so no more work merited here.

src/components/Time/Replay.tsx
jshaughn: I know, `!!!` is basically equivalent to `!` but I find it more readable to show it's explicitly a negation  of truthy/falsey for non-boolean types.

src/pages/Graph/GraphPage.tsx
jshaughn: How would you propose to rewrite this keeping it a const?

jshaughn:
@serenamarie125 Here are some potential changes given the latest UX feedback.  Tell me what you think.  I put the video here but feel free to comment in slack if that is easier.

- Increased the play/pause button size
- got the datetimepicker working and simplified times by trimming away the seconds.
- changed the graph border to solid, made it just the left side, and extended it to group the replay toolbar
- migrated the "duration" dropdown back to the top toolbar and changed the text from, for example "Last 5m" to "5m Traffic".  Duration is not a replay-specific setting and should probably stay in the top toolbar.  I changed the text to hopefully be applicable to both normal and replay scenarios.  Also, "Traffic" is more descriptive of what the setting actually does; it sets the amount of traffic data pulled per rendering. The tooltips are situational.
- Pulled the "floating X" up to the top toolbar and gave it a text description.  Now the open and close of replay is located on the same horizontal.
- as a side-effect of the changes the progress bar gets more space.

kiali#1421-5

src/components/Time/LocalTime.tsx
```diff
@@ -1,27 +1,69 @@
...
 }
 
+const defaultOptions = {
+  month: 'short',
+  day: 'numeric',
+  hour: '2-digit',
+  minute: '2-digit'
+} as any;
+
+export const toString = (time: TimeInMilliseconds, options?: any): string => {
```
jotak: Changes look good, but I think these exported funcs would better be in a utility file rather than staying in component

src/components/MTls/MeshMTLSStatus.tsx
```diff
@@ -13,11 +13,11 @@ import * as API from '../../services/Api';
 import { KialiDispatch } from '../../types/Redux';
 import { bindActionCreators } from 'redux';
 import { MeshTlsActions } from '../../actions/MeshTlsActions';
-import { RefreshIntervalInMs } from '../../types/Common';
+import { TimeInMilliseconds } from '../../types/Common';
 import Namespace from '../../types/Namespace';
 
 type ReduxProps = {
-  lastRefreshAt: RefreshIntervalInMs;
+  lastRefreshAt: TimeInMilliseconds;
```
jotak: good catch

src/components/Time/Replay.tsx
```diff
@@ -0,0 +1,384 @@
...
+
+// key represents replay interval in seconds
+const replayIntervals = {
+  60000: '1 minute',
+  300000: '5 minutes',
+  600000: '10 minutes',
+  1800000: '30 minutes'
+};
+
+// key represents refresh interval in seconds
```
jotak: "in seconds"? Isn't it miliseconds instead?
Can you elaborate on how speed is used?

src/components/Time/Replay.tsx
```diff
@@ -0,0 +1,384 @@
...
+  };
+
+  formatFrame = (frame: number): string => {
+    const elapsedTime: IntervalInMilliseconds = frame * frameInterval;
+    const elapsedSec: number = Math.floor((elapsedTime / 1000) % 60);
+    const elapsedMin: number = Math.floor((elapsedTime / 1000 - elapsedSec) / 60);
+    const zeroPadSec: string = elapsedSec < 10 ? '0' : '';
+    const zeroPadMin: string = elapsedMin < 10 ? '0' : '';
+    const elapsed: string = `${zeroPadMin}${elapsedMin}:${zeroPadSec}${elapsedSec}`;
+    return `${elapsed}`;
```
jotak: that's just `return elapsed;` :-)

src/components/Health/__tests__/__snapshots__/HealthDetails.test.tsx.snap
```diff
@@ -12,7 +12,7 @@ exports[`HealthDetails renders deployments failure 1`] = `
       size="sm"
       title={null}
     />
-     Error Rate over last 1m: 
+     Error Rate over 1m:
```
jotak: @jshaughn is it an intended change or not? Do you think it's fine to remove "last" here? I can understand when reading "Error rate over 1m" that it's implicitly the *last* 1 minute, but just want to make sure it's ok with you too.

src/pages/Graph/GraphPage.tsx
```diff
@@ -82,12 +88,12 @@ type ReduxProps = {
     edgeLabelMode: EdgeLabelMode,
     showSecurity: boolean,
     showUnusedNodes: boolean,
-    node?: NodeParamsType
+    node?: NodeParamsType,
+    queryTime?: TimeInMilliseconds
```
jotak: I was pretty sure query time was in milliseconds in our rest api, but turns out i was wrong. If you choose to use ms instead of seconds because of my previous comment, feel free to do as you prefer.

src/components/Time/LocalTime.tsx
jshaughn: Makes sense, I'll create Time/utils.tsx and move to there.

src/components/Time/Replay.tsx
jshaughn: Stale comment, I originally had this in seconds but converted to ms when consolidating on ms.  Because this is a replay we don't have to, and really don't want to, refresh in real-time.  It's too tedious to wait.  So "speed" is basically how fast the replay moves.  It  maps to a refresh rate but is a more simple interface than explicitly setting a number of seconds.

src/components/Health/__tests__/__snapshots__/HealthDetails.test.tsx.snap
jshaughn: Good eye, I have no idea wht that "last" was omitted.  I'll add it back.

jshaughn: Oh wait, this is a snapshot.  I must have just updated snapshots.  It took me a moment but now I see the unintended consequence of my change to the duration dropdown values.  If we are looking at live data then the "last" prefix makes sense but if we are looking at historical data it doesn't.  And health may very well [soon] be reporting historical health given the support we are adding for custom time ranges.  I think maybe RateIntervals.getName() should take a prefix  parameter, or just never add "last".   Let me know what you want and I can update it in the PR.

src/pages/Graph/GraphPage.tsx
jshaughn: I think I prefer that we still use seconds in the query parameters, meaning that we use seconds in our backend API.  But up to the point that we make the API call I think it's fine to consistently use milliseconds in the client code, because javascript works best with ms (Date.now(), Date.getTime(), etc).  The parameter above is still a js method param, so I have it as ms.  The conversion it done just when constructing the url params.

src/components/Health/__tests__/__snapshots__/HealthDetails.test.tsx.snap
jotak: It's fine for me to remove "last" here

jshaughn: Well in the time since my last reply I noticed that thechange I made created a problem because MetricsDuration was depending on the text value in the serverConfig.durations.  The dropdpowns for metrics became just time vaues like. "1m", "5m".

Sooooo, in this latest revision I've mde the values consistent again, but instead of the very situational "Last XX", its now "XX Traffic".  This imo is a better value in general, as it is more descriptive of what the dropdown actually represents (the duration of traffic data per query).  UX is OK with the change, what do you think?

mceledonia:
This is looking like great progress. I like the changes, the play button feels better, the close button is more noticeable/clear, larger progress bar helps to visually remind the user they're in replay mode. A couple follow up questions:

1. Is there any control over the visual aspects of the new date picker? Even changing the font to ours would be a huge improvement. It sticks out a bit right now but seeing it as a temporary solution until we get the PF version available seems fine to me, what do you think @serenamarie125?
2. I want to find you a new color to replace the blue used in the border, it works alright here but isn't in the range of what we typically have in PF. Stay tuned for that!

jshaughn:
@mceledonia Thanks for the feedback. If there are visual changes to the datepicker we can override the css to some degree.  Below is a change to our font and font-size.  Do you like it better?

Before:
image

After:
image

Additionally, one more change for your consideration.  Taking some informal feedback from the team, the request was for a few simple replay options and then a toggle to a custom start time only when requested.  So now it starts with replay "Last 5m", with a few other options.  Button click to set the custom range. Thoughts:

kiali#1421-6

src/components/DurationDropdown/DurationDropdown.tsx
```diff
@@ -36,9 +37,10 @@ export class DurationDropdown extends React.Component<DurationDropdownProps> {
         disabled={this.props.disabled}
         handleSelect={key => this.props.setDuration(Number(key))}
         value={String(this.props.duration)}
-        label={this.props.durations[this.props.duration]}
-        options={this.props.durations}
+        label={serverConfig.durations[this.props.duration]}
```
jotak: it seems `this.props.durations` isn't used anywhere now, or did I miss one? So it could be removed from props

src/components/Health/__tests__/__snapshots__/HealthDetails.test.tsx.snap
jotak: Fine!

beaumorley:
@serenamarie125 is at an onsite this week. She’s asked me to give some feedback.

I agree with @mceledonia  that this is looking really good. I like how you have been able to extend the border on the left to include the tool bar and are using the same color for the border and the Replay labels.

Really just one thing caught my eye to consider:

- When you first go into Replay Mode, the blue label/box on the top left says “(Replay) waiting to play…”. I am worried that the “…waiting to play” part will make users think they just have to wait for it to start?  Should we say something like “(Replay) tap Play to start”. 

I like the change to the calendar font. Using this font consistently ties it all together!

serenamarie125:
@jshaughn this is looking great!  I agree with everything that Beau & Michael said as well.  Couple of quick things I observed.

1. can you capitalize Time in the date/time picker?
1. were you able to visually change the slow/medium/fast toggle to be a bit more similar to those in the topology view?  I wonder if you could have vertical separators between them, and just have the font color be used to indicate selection?  @mceledonia thought on that ?
1. when setting the custom time, i'm not sure that you need to show it in the text field ... it looks like it's hidden by default - and when you set it the field shows up until you click on that icon again.  That feels a bit odd to me.  I wonder if you can just show the date/time picker immediately ?  If you really wanted to provide and editable field, you could put it in the modal below the picker maybe?
1. Re: Beaus' comment about changing the label, I think  “(Replay) *click* Play to start”.

mtho11:
@jshaughn 38 files! this is getting a bit large for a single PR?

jshaughn:
> 38 files! this is getting a bit large for a single PR?

@mtho11 It is large but I don't think it's too unwieldly.  It touches a lot of files but that includes some new file, some snapshot updates, some redux/state additions, and some minor refactors that touch several places.  TBH, I'd be scared to break it up at this point.

src/components/DurationDropdown/DurationDropdown.tsx
jshaughn: Seems unused, I'll remove it.

jshaughn:
@serenamarie125 @mceledonia @beaumorley...

> can you capitalize Time in the date/time picker?

done

> were you able to visually change the slow/medium/fast toggle to be a bit more similar to those in the topology view? I wonder if you could have vertical separators between them, and just have the font color be used to indicate selection?

I went with the vertical separator approach and use color/bold to indicate the selection.  Hopefully good.

> when setting the custom time, i'm not sure that you need to show it in the text field ... it looks like it's hidden by default - and when you set it the field shows up until you click on that icon again. That feels a bit odd to me.

It's a toggle between simple, predefined replay options for recent times (like replay the last 5 minutes)  and the ability to set a custom start time anytime in the valid time range give prom config.  For now let's go with it unless you highly object.

> Re: Beaus' comment about changing the label, I think “(Replay) click Play to start”.
done

I'm really just waiting now on any color/border changes UX may have.  So let's see if we can get any final feedback by Monday, or Tuesday at the UX meeting.  Looking to get this into upstream ASAP, early next week barring any bugs from QE.

There have been some other changes given a round of QE testing (@mattmahoneyrh).  There were a couple of nasty things in tooltip behavior and also we've added the ability to "Play Again" when at the end.  Here is the latest video showing relevant changes:

kiali#1421-7

One last thing, I am going to take the DNM label off for final approval and testing.

serenamarie125:
@jshaughn thanks for doing all of these iterations, this feature is awesome and I think all the iterations have paid off.  I've asked Michael to see if he can get back to you early Monday.  Completely agree with your suggestions above, and the need to get this reviewed/merged.  I'd also suggest that we do some usability testing on this, even if it's "in house"!

Great job!

mceledonia:
If it's not too late, I can recommend the color variable for the blue highlight color:
`$pf-global--active-color--200` 
or, if that feels too light,
`$pf-global--active-color--300`

jshaughn:
OK, here are the final updates barring anything unexpected in review.
UX Feedback (@mceledonia)
- thicken replay border
- set replay color to --pf-global--active-color--300
  - border
  - slider selection background (I added it here for consistency)
  - badged time range background
- set active speed to --pf-global--primary-color--200
- set active custom start icon to --pf-global--active-color--400
  - also, for consistency, set same for active layout icon in cytoscape toolbar

Full video for review:
kiali#1421-8

And also a screenshot showing that this is part of the graph help tour:
image

PR #1588, opened 2019-12-20
Custom time search in metrics

jotak:
- Add new FROM and TO url params
- Rename the previous TimeRange component (used in Graph and Overview) to TimeControls (as it also involves refresh control & replay)
- Replace old MetricsDuration component with new TimeRangeComponent, that handles both Duration and custom From/to with datepickers. This is now decorelated from Metrics as it was actually used in other places (traffic, logs...)
- New sum type TimeRange = DurationInSeconds | BoundsInMilliseconds with some guarding and conversion functions

Capture d’écran de 2020-01-24 14-19-56

Fixes https://github.com/kiali/kiali/issues/1859


jotak:
@jshaughn & saturn you'll tell me if you're fine the the various renaming I did

src/types/Common.ts
```diff
@@ -26,3 +26,25 @@ export type TimeInSeconds = number;
 
 export type RefreshIntervalInMs = TimeInMilliseconds;
 export type DurationInSeconds = TimeInSeconds;
+export type FromToInMilliseconds = {
+  from: TimeInMilliseconds;
+  to?: TimeInMilliseconds;
+};
+
+export type TimeRange = DurationInSeconds | FromToInMilliseconds;
```
jshaughn: I guess FromTo could be convenient for describing a time range in the past, but conceptually it's the same as queryTime + duration, which is what we use now in our APIs and which works naturally with Prometheus.  So why change to To and From parameters?  I could be missing something but it seems like TimeRange could just be duration and optional queryTime (default=now).   Helper functions could still convert to FromToInMilliseconds if necessary.  As is maybe it's over-complicated.

Either way, is there a benefit to FromToInMilliseconds as opposed to FromToInSeconds?   Seconds seems to be used more often in the API (maybe not charts, I'm not sure).

jotak: You're probably correct, I didn't realize I could have the same with what we already have.
As about ms versus seconds: I find on the contrary more stuff being in milliseconds, for instance queryTime is in ms, no? DurationInSeconds being more the anomaly .. anyway the question isn't relevant anymore if I remove this FromTo thing.
I'll do that after the break

jotak: @jshaughn I think the best is to use the existing http API with duration/queryTime as you said, but still have a specific "from-to" model like the one I suggest because it maps 1-to-1 with the actual component that we're building (a "from" field and a "to" field) so the app state is more accurately described.

To take an example that shows a behavioural difference: if "from" is defined but "to" is undefined, if this is accurately described in the app state, "from" will stay fixed upon refreshes whereas "to" will keep pointing to "now" (which is I think a valid scenario that users would expect). On the contrary, if it gets converted to a duration in the app state, the initial "from" that the user explicitly selected would be lost after the first refresh.

For this kind of things I think it's better to have the model accurately describing the UI component. Conversions to duration would then be performed only when fetching data (ie. in a stateless process)

jshaughn: @jotak I'd like to have a meeting with you to show you what I've done in My PR for graph replay.  We're in an area of overlap with, I think, slightly different goals to support custom time range for charts and custom time range for the graph.   We should talk it out to make sure we come up with something that satisfies all of the foreseeable Kiali needs.  Of course any interested person is invited as well.  I'll contact you...

jshaughn: By the way, today I decided to use TimeInMilliseconds and IntervalInMilliseconds for basically all of the client-side time manipulations in graph (in my PR).  Leaving TimeInSeconds only for the rest api for use with queryTime param to the backend API, which expects seconds because that works well with Prom (which actually uses time.Time).  I still like seconds for the queryTime param but our API is not yet supported, so if there is a movement to use millis in the rest api we could still make that change.

src/components/Time/TimeControls.tsx
```diff
@@ -5,14 +5,14 @@ import RefreshContainer from 'components/Refresh/Refresh';
...
-type TimeRangeProps = ReduxProps & {
+type TimeControlsProps = ReduxProps & {
   disabled: boolean;
   id: string;
 
   handleRefresh: () => void;
 };
 
-export class TimeRange extends React.PureComponent<TimeRangeProps> {
+export class TimeControls extends React.PureComponent<TimeControlsProps> {
```
jotak: @jshaughn what do you think about this renaming of `TimeRange` component to `TimeControls`? Or do you have other suggestions?
There's two motivations:
- first because it's not just a range, it also manages the refresh rate. And your PR will also add the replay option. So much more than just a range
- this is also because I wanted a name for my new from/to pickers component, and TimeRange looks actually not too bad for that, I think.
Or any other suggestion?

jshaughn: @jotak I agree that that is a better name given the evolving nature of the component.  We can rename to "TimeControls" and free up "TimeRange" for another purpose.   Should I do it in my PR or will you make it part of your work?

jotak: As you want. I'm already doing that here, so you can leave it untouched

jshaughn: I will defer the change, thanks.

jotak:
I think this is pretty much ready for test & review

lucasponce:
image

This Timezone is calculated from server (i.e. backend in US) or client (local browser in Europe zone) ?

lucasponce:
With this option, I think we miss the URL bookmark, so, from/to are in the state but not propagated to the urls.

image

I don't consider this a blocker, but something to address.

lucasponce:
Two UI issues:
image

On my screen, I think it could be nice to have a little of space between the "custom" controls and the first graph, it seems too close.
Also, if you see the right chart, it overlaps some legend in the y-axis with the left chart, too.

lucasponce:
The datepicker looks pretty good:
image

I mean, it can change in the future with some pf4 alignment, but it is a very decent first approach.

aljesusg:
can we move the dropdown and refresh out the metrics tabs like overview and traffic tab?

jotak:
@lucasponce 

> This Timezone is calculated from server (i.e. backend in US) or client (local browser in Europe zone) ?

This is client-based time & timezone.

> With this option, I think we miss the URL bookmark, so, from/to are in the state but not propagated to the urls.
> 

It *should* be in URL, if not then something isn't working as i would expect, i'll check that.

> 
> On my screen, I think it could be nice to have a little of space between the "custom" controls and the first graph, it seems too close.

OK

> Also, if you see the right chart, it overlaps some legend in the y-axis with the left chart, too.

Hmm that's not related to this PR then, that's something probably in master as well

jotak:
> can we move the dropdown and refresh out the metrics tabs like overview and traffic tab?

We can discuss it in another ticket then, ok? (The difference here with the other pages is that there's more items in the toolbar (runtimes metrics must be considered as well, with bigger toolbar), which leaves less space when it's moved above.) but i'd rather discuss that not on this PR.

prachiyadav:
Can not see the changes after updating PR. Also verified yaml file. 
Screenshot from 2020-01-28 23-27-00
Screenshot from 2020-01-28 23-26-35

jotak:
Hi @prachiyadav ,
You need to select "custom" in the duration dropdown, on the right. Can you see it?

prachiyadav:
Verified! 

Screenshot from 2020-01-29 13-55-29
Screenshot from 2020-01-29 13-54-18

jotak:
> @lucasponce

> > With this option, I think we miss the URL bookmark, so, from/to are in the state but not propagated to the urls.
> 
> It _should_ be in URL, if not then something isn't working as i would expect, i'll check that.
> 

Tested again and I do have it in URL: `http://localhost:3000/console/namespaces/default/applications/ball?tab=in_metrics&from=1580376000989&to=1580389500989` for example... so not sure what happened for you

> > On my screen, I think it could be nice to have a little of space between the "custom" controls and the first graph, it seems too close.
> 
> OK
> 

I've added some padding

> > Also, if you see the right chart, it overlaps some legend in the y-axis with the left chart, too.
> 
> Hmm that's not related to this PR then, that's something probably in master as well

This is fixed in master

jotak:
@skondkar :

> Looks good to me, however, below are some minor observations
> 
>     1. The date picker arrow button for Next(to select next month) positioned in 'Time' section. It would be nice to have it in month/date selection section.
> 

it's moved

>     2. User can select 'To' date less than 'From' date.

This is fixed although the solution might not be 100% satisfying:
- Calendar days are now greyed out when no valid because "from" would be > than "to" => that's good
- But invalid times can still be selected in component. There's an issue about that in react-datepicker but I'm not sure it's actively worked on (?) ( https://github.com/Hacker0x01/react-datepicker/issues/1135 with a 8-months old PR that isn't merged. ) so not sure what's the status of this issue; anyway, I did a workaround to not have "from" > "to" after they are changed in component.

PR #1590, opened 2019-12-23
IstioConfigList - Istio type selector now is typeahead variant

xeviknal:
** Describe the change **
The list of Istio Objects has increased substantially since we started the project.
https://github.com/kiali/kiali-ui/pull/1582#issuecomment-567601410 suggested to sort types alphabetically. This is step further to ease the task of filtering objects by type.
This PRs add the ability to pick istio types by typing its name. It uses the [PF4 typeahead selector](https://www.patternfly.org/v4/documentation/core/components/select#single-with-typeahead).

** Issue reference **
https://github.com/kiali/kiali-ui/pull/1582#issuecomment-567601410

** screenshot **

Screen recording (16)

### DIFF
--- a/src/components/Filters/StatefulFilters.tsx
+++ b/src/components/Filters/StatefulFilters.tsx
@@ -229,7 +229,7 @@ export class StatefulFilters extends React.Component<StatefulFiltersProps, State
           isExpanded={this.state.isExpanded}
           aria-label="filter_select_value"
           placeholderText={currentFilterType.placeholder}
-          width={'65%'}
+          width={'auto'}
         >
           {currentFilterType.filterValues.map((filter, index) => (
             <SelectOption key={'filter_' + index} value={filter.id} label={filter.title} />
@@ -242,7 +242,7 @@ export class StatefulFilters extends React.Component<StatefulFiltersProps, State
           value={'default'}
           onChange={this.filterValueSelected}
           aria-label="filter_select_value"
-          style={{ width: '65%' }}
+          style={{ width: 'auto' }}
         >
           <FormSelectOption key={'filter_default'} value={'default'} label={currentFilterType.placeholder} />
           {currentFilterType.filterValues.map((filter, index) => (
@@ -259,7 +259,7 @@ export class StatefulFilters extends React.Component<StatefulFiltersProps, State
           placeholder={currentFilterType.placeholder}
           onChange={this.updateCurrentValue}
           onKeyPress={e => this.onValueKeyPress(e)}
-          style={{ width: '65%' }}
+          style={{ width: 'auto' }}
         />
       );
     }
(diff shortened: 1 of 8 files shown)


src/components/Filters/StatefulFilters.tsx
```diff
@@ -150,7 +155,7 @@ export class StatefulFilters extends React.Component<StatefulFiltersProps, State
     }
   };
 
-  filterValueSelected = (value: string) => {
+  filterValueSelected = (_: any, value: any) => {
```
cfcosta: Can the `_` be something with a more meaningful name?

xeviknal: yep, it can be. However, it is not used and then, the linter jumps and says it should use it. So the only way you can avoid this one is using that underscore.

cfcosta: You can actually use `_foo` as the name and the compiler will know it is not supposed to be used.

mtho11:
Nice feature!

src/components/Filters/StatefulFilters.tsx
xeviknal: Good point!

hhovsepy:
> Verified

Regression is found: With rest of the type of selectors it does not apply the value.

lucasponce:
> Regression is found: With rest of the type of selectors it does not apply the value.

Please, if that's true and merging this can creeate some problem, don't hesitate to ask changes on the PR.

xeviknal:
@hhovsepy thanks for catching this up. It should work now :+1: 
Reviewers, this should be ready to review now. Thanks

lucasponce:
Let me check if this also happens in master:
image
I see that for some filters, the title is cut.

lucasponce:
Yeah, in master looks ok:
image

lucasponce:
Another issue, but perhaps it's the component:
image
Once you have some data selected, it seems the type-ahead doesn't show a result (at least in my test).

hhovsepy:
I saw the same behaviour, when trying to type several filters in a row without unfocusing/focusing. It seemed the component itself.

> Another issue, but perhaps it's the component:
> image
> Once you have some data selected, it seems the type-ahead doesn't show a result (at least in my test).
> 
> I'd need to exit the focus and click again in the component to get the typeahead.
> 
> This issue is more related to the component, I think, it's NOT a blocker for this PR, but just worth to comment.

xeviknal:
@lucasponce I could approach the first comment. However, the second one it is a bit tricky.
The alternative could be the following although I don't like that much because it is not consistent with the logic we've got in the other selectors.

Screen recording (19)

Also regarding to this comment, the bug happens when you hit the 'return' key. If you click the option, the issue is no present:

Screen recording (18)

Thanks!

xeviknal:
Opened the following issue to smooth the UX of this:
https://github.com/patternfly/patternfly-react/issues/3488

PR #1592, opened 2019-12-23
Remove prop-types

cfcosta:
** Describe the change **

Removes prop-types and it's dependency, as it is not used on Typescript.

** Issue reference **

Closes https://github.com/kiali/kiali/issues/1579

** Backwards compatible? **

### DIFF
--- a/src/components/IstioWizards/Slider/Boundaries.tsx
+++ b/src/components/IstioWizards/Slider/Boundaries.tsx
@@ -1,7 +1,6 @@
 // Clone of Slider component to workaround issue https://github.com/patternfly/patternfly-react/issues/1221
 
 import React from 'react';
-import PropTypes from 'prop-types';
 
 const Boundaries = props => {
   const { children, min, max, reversed, showBoundaries, slider } = props;
@@ -26,15 +25,6 @@ const Boundaries = props => {
       {children}
     </div>
   );
-};
-
-Boundaries.propTypes = {
-  children: PropTypes.array,
-  min: PropTypes.number,
-  max: PropTypes.number,
-  reversed: PropTypes.bool,
-  showBoundaries: PropTypes.bool,
-  slider: PropTypes.object.isRequired
 };
 
 Boundaries.defaultProps = {
--- a/src/components/IstioWizards/Slider/Boundaries.tsx
+++ b/src/components/IstioWizards/Slider/Boundaries.tsx
@@ -25,8 +25,8 @@ class Boundaries extends React.Component<BoundariesProps, {}> {
     const minElement = <b>{min}</b>;
     const maxElement = <b>{max}</b>;
 
-    let leftBoundary: any = null;
-    let rightBoundary: any = null;
+    let leftBoundary: JSX.Element | null = null;
+    let rightBoundary: JSX.Element | null = null;
 
     if (showBoundaries) {
       if (reversed) {
--- a/src/components/IstioWizards/Slider/Boundaries.tsx
+++ b/src/components/IstioWizards/Slider/Boundaries.tsx
@@ -1,38 +1,52 @@
 // Clone of Slider component to workaround issue https://github.com/patternfly/patternfly-react/issues/1221
-
 import React from 'react';
 
-const Boundaries = props => {
-  const { children, min, max, reversed, showBoundaries, slider } = props;
-  const minElement = <b>{min}</b>;
-  const maxElement = <b>{max}</b>;
-  let leftBoundary: any = null;
-  let rightBoundary: any = null;
-  if (showBoundaries) {
-    if (reversed) {
-      leftBoundary = maxElement;
-      rightBoundary = minElement;
-    } else {
-      leftBoundary = minElement;
-      rightBoundary = maxElement;
-    }
-  }
-  return (
-    <div className="slider-pf">
-      {leftBoundary}
-      {slider}
-      {rightBoundary}
-      {children}
-    </div>
-  );
+type BoundariesProps = {
+  children: JSX.Element[];
+  min: number;
+  max: number;
+  reversed: boolean;
+  showBoundaries: boolean;
+  slider?: JSX.Element;
 };
 
-Boundaries.defaultProps = {
-  children: [],
-  min: 0,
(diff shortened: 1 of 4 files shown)


src/components/IstioWizards/Slider/Boundaries.tsx
```diff
@@ -28,15 +27,6 @@ const Boundaries = props => {
...
 
-Boundaries.propTypes = {
-  children: PropTypes.array,
-  min: PropTypes.number,
-  max: PropTypes.number,
-  reversed: PropTypes.bool,
-  showBoundaries: PropTypes.bool,
-  slider: PropTypes.object.isRequired
-};
-
```
xeviknal: Should we add the types in the props definition? Otherwise the type check won't happen.

cfcosta: Done, added. I also changed the components to be class-based as the other ones on the repo.

src/components/IstioWizards/Slider/Boundaries.tsx
```diff
@@ -1,48 +1,52 @@
...
+    showBoundaries: false
+  };
+
+  render() {
+    const { children, min, max, reversed, showBoundaries, slider } = this.props;
+
+    const minElement = <b>{min}</b>;
+    const maxElement = <b>{max}</b>;
+
+    let leftBoundary: any = null;
```
mtho11: This could be a little more strict than `any` since we know what it is.

src/components/IstioWizards/Slider/Slider.tsx
```diff
@@ -80,7 +80,7 @@ class Slider extends React.Component<Props, State> {
     this.updateNewValue(newValue - 1);
   };
 
-  onInputChange = value => {
+  onInputChange = (value: any) => {
```
mtho11: We know this is a `string` type.

src/components/IstioWizards/Slider/Boundaries.tsx
cfcosta: Done!

src/components/IstioWizards/Slider/Slider.tsx
cfcosta: Done!

src/components/IstioWizards/Slider/Slider.tsx
```diff
@@ -165,15 +174,7 @@ class Slider extends React.Component<Props, State> {
...
-          {this.props.showLock && (
-            <Button
-              className={pinButtonStyle}
-              variant={this.props.locked ? ButtonVariant.primary : ButtonVariant.secondary}
-              onClick={() => this.props.onLock(!this.props.locked)}
-            >
-              <ThumbTackIcon />
-            </Button>
-          )}
+          {this.props.showLock ? LockIcon : <></>}
```
israel-hdez: Probably, use `null` instead of `<></>`?

aljesusg:
```ts
{this.props.showLock && LockIcon}
```

cfcosta: That does not work because showLock is not `null`, just `false`, unfortunately.

src/components/IstioWizards/Slider/Boundaries.tsx
```diff
@@ -1,48 +1,52 @@
...
 
-Boundaries.propTypes = {
-  children: PropTypes.array,
-  min: PropTypes.number,
-  max: PropTypes.number,
-  reversed: PropTypes.bool,
-  showBoundaries: PropTypes.bool,
-  slider: PropTypes.object.isRequired
+type BoundariesProps = {
+  children: JSX.Element[];
```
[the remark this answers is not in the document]
cfcosta: Let me see if removing it works, I remember having to add it because it does not inherit the component props correctly.

src/components/IstioWizards/Slider/Slider.tsx
aljesusg: If showLock is a Boolean you can do it, in fact if showLock is false or null the component won't render  LockIcon

cfcosta:
```
TypeScript error in /home/cfcosta/Projects/go/src/github.com/kiali/kiali-ui/src/components/IstioWizards/Slider/Slider.tsx(177,11):
Type 'false | Element' is not assignable to type 'Element'.
  Type 'false' is not assignable to type 'Element'.  TS2322
```

Nope, it does not work :(

aljesusg: Mm but the prop is Boolean not element.

```ts
type Props = {
  id: string;
  orientation: string;
  min: number;
  max: number;
  maxLimit: number;
  step: number;
  value: number;
  tooltip: boolean;
  onSlide: (value: number) => void;
  input: boolean;
  sliderClass: string;
  inputFormat: string;
  locked: boolean;
  showLock: boolean;
  onLock: (locked: boolean) => void;
};
```

cfcosta: Yes, but the `LockIcon` is.

src/components/IstioWizards/Slider/Boundaries.tsx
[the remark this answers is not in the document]
cfcosta: Removing it did work, done.

src/components/IstioWizards/Slider/Slider.tsx
mtho11: Do we still have an issue here? @aljesusg  I think we are waiting on your approval.

aljesusg: It's not a blocker @mtho11 , It's an improvement, feel free to merge this.

My only comment is that LockIcon is always true because is a Element defined, so {this.props.showLock && LockIcon} should work

cfcosta: I'm going to merge this then and we can change this later if necessary.

aljesusg:
Please @cfcosta nextime use squash and merge

PR #1595, opened 2020-01-08
Make sparkline tooltips overflowing (svg scope)

jotak:
Also fix an issue with the tooltip arrow not pointing exactly to the hovered datapoint

Fixes https://github.com/kiali/kiali/issues/2068

### DIFF
--- a/src/components/Charts/Charts.scss
+++ b/src/components/Charts/Charts.scss
@@ -1,0 +1,3 @@
+.pf-c-chart svg {
+  overflow: visible !important;
+}
--- a/src/components/Charts/CustomFlyout.tsx
+++ b/src/components/Charts/CustomFlyout.tsx
@@ -4,19 +4,13 @@ import { Flyout } from 'victory';
 const squareSize = 10;
 
 export const CustomFlyout = (props: any) => {
-  const { x, width, center, datum } = props;
+  const { width, center, datum } = props;
   const left = center.x - width / 2;
   const top = center.y - squareSize / 2;
   const extraWidth = squareSize + 5;
   return (
     <>
-      <Flyout
-        {...props}
-        width={width + extraWidth}
-        x={x - extraWidth}
-        center={{ x: center.x - extraWidth / 2, y: center.y }}
-        style={{ ...props.style, fillOpacity: 0.6 }}
-      />
+      <Flyout {...props} width={width + extraWidth} style={{ ...props.style, fillOpacity: 0.6 }} />
       <rect width={squareSize} height={squareSize} x={left} y={top} style={{ fill: datum.color }} />
     </>
   );
--- a/src/components/Charts/SparklineChart.tsx
+++ b/src/components/Charts/SparklineChart.tsx
@@ -15,6 +15,8 @@ import { PfColors } from 'components/Pf/PfColors';
 import * as Legend from './LegendHelper';
 import { CustomFlyout } from './CustomFlyout';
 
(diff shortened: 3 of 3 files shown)


jotak:
Capture d’écran de 2020-01-08 14-35-44

src/components/Charts/Charts.scss
```diff
@@ -0,0 +1,3 @@
+.pf-c-chart svg:not(:root) {
+  overflow: visible;
+}
```
israel-hdez: I'd prefer:
```css
.pf-c-chart svg {
￼  overflow: visible !important;
}
```

instead of using those weird pseudo-classes. Not a blocker, so you may merge without this change.

mtho11: I did not personally test out @israel-hdez preference here but it does make sense if there are no other side effects.

jotak: TBH I'm far from being a CSS expert - but I found this pattern `svg:not(:root)` on several places (I know, it's copy-paste (or stack-overflow) -based development) but also in our own codebase: https://github.com/kiali/kiali-ui/blob/master/src/components/Health/Health.scss (and I'm sure it's not me who wrote that).

So, given this, do you maintain this should be changed or not? I let you judge @mtho11 @israel-hdez

israel-hdez: > ( ~and I'm sure it's not me who wrote that~ - actually this IS me who wrote that ... lol).
>

LOL!
 
> So, given this, do you maintain this should be changed or not? I let you judge @mtho11 @israel-hdez

Yes, I still prefer `!important` instead of the pseudo-classes.

To give a little explanation, I think you know, CSS have a _cascading_ nature. So, if you have two identical selectors that match, the "last" one is applied, for example, in:

```css
svg { display: none }
[...more css...]
svg { display: block }
```

`display: block` will be applied. Also, if more than one class matches, the more specific one is applied. So, if you have:

```css
a:hover { color: red }
a { color: blue }
```

You will see color red when you hover on a link regardless of the order in the CSS file, because `a:hover` is more specific. This is the rule being applied with those weird pseudo-classes, so if the generated CSS was like:

```css
.pf-c-chart svg:not(:root) { overflow: visible; }
.pf-c-chart svg { overflow: hidden; }
```

basically both selectors will _always_ match because an `svg` node can never be the _root_ element (in an HTML doc, only the `<html>` tag matches against CSS `:root`). But since you added `:not(:root)` you are turning it to a _more specific selector_ than the simple `svg`. Thus, the `overflow: visible` property is applied regardless of the order.

So, given that you just want to give more priority to the style, I'd lean towards using `!important` because it's clearer.

israel-hdez: But as @mtho11 said, check that there are no side effects when using `!important` :)

jotak: So after some tests, your suggestion works, but I had some doubt because the "not root" trick is actually already defined somewhere to hide overflow: `svg:not(:root) { overflow: hidden; }` (but without any class specifier). Not sure where that's defined, patternfly or victory. Anyway, it seems that "!important" wins!

jotak:
@mattmahoneyrh sorry, I'm not sure to understand what you mean. Don't you have the tooltip updated when you move the cursor over different datapoints? You have to go out & in?

jotak:
I'm merging this but @mattmahoneyrh please let me know for any follow-up

PR #1603, opened 2020-01-14
Undo filter renaming on List Pages

xeviknal:
** Describe the change **

When there are two filters with same names, the filter keeps applied between pages. That is the case of Name filter for Services/Apps/Workloads.

The bug was introduced here: https://github.com/kiali/kiali-ui/pull/1590


hhovsepy:
Fixes https://github.com/kiali/kiali/issues/2099

PR #1604, opened 2020-01-15
Fix scroll in Overview page

lucasponce:
Fixes kiali/kiali#2094

This only fixes the Overview pages, list pages will be fixed in a following PR.
image
image


lucasponce:
@hhovsepy can you check if there is any regression in other details pages with scroll ?

PR #1611, opened 2020-01-19
Remove capitalize class from DetailObject component

kinsolee:
** Describe the change **

Remove the `text-capitalize` class from the DetailObject component. Because the match rule's keys are case-sensitive, users cannot distinguish if their original case when using `text-capitalize` class.

** Screenshot **

Before change:
Jietu20200119-111038

After change:

Jietu20200119-112510


rhqci:

PR #1612, opened 2020-01-20
Add a whitelist for services in istio-system

aljesusg:
Require https://github.com/kiali/kiali/pull/2225

Fix https://github.com/kiali/kiali/issues/2230

### DIFF
--- a/src/components/JaegerIntegration/JaegerResults/SpanTable.tsx
+++ b/src/components/JaegerIntegration/JaegerResults/SpanTable.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { Span } from '../../../types/JaegerInfo';
+import { JaegerInfo, Span } from '../../../types/JaegerInfo';
 import { Table, TableHeader, TableBody, IRow, expandable, RowWrapperProps } from '@patternfly/react-table';
 import { ExclamationCircleIcon } from '@patternfly/react-icons';
 import { Tooltip } from '@patternfly/react-core';
@@ -17,7 +17,7 @@ import { PfColors } from '../../Pf/PfColors';
 
 interface SpanDetailProps {
   spans: Span[];
-  namespaceSelector: boolean;
+  jaegerInfo?: JaegerInfo;
 }
 
 interface SpanDetailState {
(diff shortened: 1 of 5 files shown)


src/components/JaegerIntegration/JaegerResults/SpanTable.tsx
```diff
@@ -26,6 +26,8 @@ interface SpanDetailState {
   rows: any;
 }
 
+const whiteListIstioSystem = ['istio-ingressgateway', 'jaeger-query'];
```
lucasponce: Mmm, this is not right, it should come from the config server backend instead and configurable from the CR as well.

aljesusg: ok let @hhovsepy if this fix the problem with jaeger-query and we can think about hot to deal with this, because this only happens with the services of the istio-system namespace. If jaeger is in another namespace it should run without problem

hhovsepy:
I still see the problem
Screenshot from 2020-01-20 11-17-06
Click on span:
Screenshot from 2020-01-20 11-17-22

aljesusg:
> I still see the problem
> Screenshot from 2020-01-20 11-17-06
> Click on span:
> Screenshot from 2020-01-20 11-17-22

I need your backend, it has not sense for me

lucasponce:
@aljesusg perhaps is this related ?

image

jaeger-query doesn't honor the <service>.<namespace> format.

aljesusg:
> @aljesusg perhaps is this related ?
> 
> image
> 
> jaeger-query doesn't honor the . format.

Yes I know this is why I need to parse the services inside the istio-system, no it's not related whit this

hhovsepy:
Verified with the last commit.

aljesusg:
@hhovsepy Is this happening in master now? I can't reproduce it now I think that we can close it

aljesusg:
Ok so I am going to update this to merge it asap with the backend  change proposed by @lucasponce

aljesusg:
Done, this require https://github.com/kiali/kiali/pull/2225

Can you check with the backend and this one that if fixed @hhovsepy ? I tested in local and it works.

PR #1613, opened 2020-01-20
Traces tab should be visible just with InClusterURL configured

aljesusg:
Require : https://github.com/kiali/kiali/pull/2118

Fix: https://github.com/kiali/kiali/issues/2117


aljesusg:
> LGTM
> nit: `let jaegerState` could now be `const`.
> but +1 anyway, thanks

Changed to const xD

PR #1617, opened 2020-01-20
Add structure to create new Istio config resources at namespace level

lucasponce:
Today, Kiali creates IstioConfig at Service details level only using "context" wizards, where Services/Workloads are combined in a high level abstraction to allow user to define Istio configuration in a set of driven scenarios.

This new PR allows to create objects that are out of the context of a Service, for example, on this work we have introduced Gateways and Sidecars.

This new option is enabled at Istio Config list level:
image

It moves to a new Form page where user can select the type of resource to create:
image

On this first step, only Gateways and Sidecars can be generated. 
Another goal of this work is to structure these subforms in a way that is easy to add more.
Also, the forms don't cover all the potential scenarios (some objects as Sidecars may have huge combination of possibilities) but to simplify the creation of the most common use cases (forms have been designed based on common examples). 
In any case, once a resource is created, the user can only update it using YAML, as this form are used only for creation step.

For Gateways:
image

For Sidecars:
image

Validation of input fields are supported and permissions are checked per user/serviceaccount as usual.

Fixes kiali/kiali#1537
Fixes kiali/kiali#1380

### DIFF
--- a/src/pages/ServiceDetails/ServiceDetailsPage.tsx
+++ b/src/pages/ServiceDetails/ServiceDetailsPage.tsx
@@ -161,7 +161,7 @@ class ServiceDetails extends React.Component<ServiceDetailsProps, ServiceDetails
   }
 
   componentDidMount() {
-    this.doRefresh();
+    this.promises.cancelAll();
   }
 
   componentDidUpdate(prevProps: ServiceDetailsProps, _prevState: ServiceDetailsState) {
--- a/src/pages/ServiceDetails/ServiceDetailsPage.tsx
+++ b/src/pages/ServiceDetails/ServiceDetailsPage.tsx
@@ -161,7 +161,7 @@ class ServiceDetails extends React.Component<ServiceDetailsProps, ServiceDetails
   }
 
   componentDidMount() {
-    this.promises.cancelAll();
+    this.doRefresh();
   }
 
   componentDidUpdate(prevProps: ServiceDetailsProps, _prevState: ServiceDetailsState) {
--- a/src/utils/IstioConfigUtils.ts
+++ b/src/utils/IstioConfigUtils.ts
@@ -64,3 +64,39 @@ export const getIstioObject = (istioObjectDetails?: IstioConfigDetails) => {
   }
   return istioObject;
 };
+
+const nsRegexp = /^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[-a-z0-9]([-a-z0-9]*[a-z0-9])?)*$/;
+const hostRegexp = /(?=^.{4,253}$)(^((?!-)(([a-zA-Z0-9-]{0,62}[a-zA-Z0-9])|\*)\.)+[a-zA-Z]{2,63}$)/;
+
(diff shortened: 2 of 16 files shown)


src/pages/ServiceDetails/ServiceDetailsPage.tsx
```diff
@@ -161,7 +161,7 @@ class ServiceDetails extends React.Component<ServiceDetailsProps, ServiceDetails
   }
 
   componentDidMount() {
-    this.doRefresh();
+    this.promises.cancelAll();
```
jotak: iirc there's several scenarios in `doRefresh` (traces, traffic, details ... ) .. did you check all scenarios are still ok on first load / page refresh?
(I can test that later, now just looking at the code)

src/utils/IstioConfigUtils.ts
```diff
@@ -64,3 +64,39 @@ export const getIstioObject = (istioObjectDetails?: IstioConfigDetails) => {
   }
   return istioObject;
 };
+
+const nsRegexp = /^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[-a-z0-9]([-a-z0-9]*[a-z0-9])?)*$/;
+const hostRegexp = /(?=^.{4,253}$)(^((?!-)(([a-zA-Z0-9-]{0,62}[a-zA-Z0-9])|\*)\.)+[a-zA-Z]{2,63}$)/;
+
+// Used to check if Sidecar and Gateway host expressions are valid
+export const isServerHostValid = (serverHost: string): boolean => {
```
jotak: I wonder if we should not relax a little bit the hosts validation. If I understand correctly, when hosts are detected as invalid we will disable the "create" button, right? That makes us very dependant on these regexes and on any special handling done by Istio or not (e.g. wildcards). What if we display the warning message but without blocking creation?

src/pages/ServiceDetails/ServiceDetailsPage.tsx
lucasponce: Oh, I think this is wrong, this PR shouldn't touch the ServiceDetailsPage.
Let me review, fix it.

src/utils/IstioConfigUtils.ts
lucasponce: Main problem is that Istio is quite strict on creation here, so my first version addressed the relaxed version, but it catches back the Istio error.
So, now the logic is mapped with the Istio requeriment for this hosts, where also the wildcards supported by Istio are included in the validation logic.
One obvious case is that, for example, Istio doesn't permit for these objects to create a "short" name, so I think is better that Kiali should allow user to add the correct pattern.

Give it a try, perhaps you see something I missed, but I think it's a 1:1 in the expected format.
Note that this helper method doesn't cover all validation, some of full validation are in the Gateway and Sidecar specific forms.

src/pages/ServiceDetails/ServiceDetailsPage.tsx
lucasponce: This is fixed now.

lucasponce:
Thanks !

PR #1618, opened 2020-01-20
Sparklines: show several colors on tooltip (hacky)

jotak:
See inline comment for hack description.
Note that I'll suggest a patch in victory to get a better solution

Fixes https://github.com/kiali/kiali/issues/2097

Capture d’écran de 2020-01-20 19-36-11


jotak:
Issue opened on victory: https://github.com/FormidableLabs/victory/issues/1473

PR #1634, opened 2020-01-27
Service Details load Error when Jaeger is configured wrongly

aljesusg:
jaeger error didn't appear in MessageCenter.

Now we can see the message related with the integration.

Fix: https://github.com/kiali/kiali/issues/2155


aljesusg:
@xeviknak @hhovsepy this is the same for master branch

PR #1635, opened 2020-01-27
Service Details load Error when Jaeger is configured wrongly

aljesusg:
backport to v1.12

jaeger error didn't appear in MessageCenter.

Now we can see the message related with the integration.

Fix: kiali/kiali#2155


hhovsepy:
This is what I see, is it the error message expected:
Screenshot from 2020-01-28 09-09-18

aljesusg:
Yes @hhovsepy and after that you shouldn't see any message related with traces in service details because the integration is off.

PR #1636, opened 2020-01-27
Enable traces correlation with runtime metrics

jotak:
- Move spans-fetching related code from istio metrics component to a separate class
- Re-use in custom metrics

Fixes https://github.com/kiali/kiali/issues/2160

Capture d’écran de 2020-01-27 15-11-25


lucasponce:
Feature works fine, but I find a side effect that I guess it needs @aljesusg here.

With this PR I can see tracing integrated in Metrics, as expected:

image

But not in the service, which would be expected:

image

So, I'm going to approve this PR, but I think that we need to address this consistency: when tracing is enabled in the metrics charts, it should be available under the service as well.

PR #1638, opened 2020-01-28
Fix undefined value not caught on jaeger rendering

lucasponce:
When `this.props.jaegerUrl` is `undefined` then `this.props.jaegerUrl !== ''` returns a true which triggers to render the Distributed Tracing menu and the Traces tab.

I've tested disabling the the tracing service in config.yaml.

There is also an error of the backend when fetching service details (but I'm using a non-master branch, so I'll let @aljesusg address that).


aljesusg:
Why did you close it?

PR #1641, opened 2020-01-29
Decouple data of CytoscapeGraph component from redux

israel-hdez:
Fixes kiali/kiali#2150


src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -138,22 +140,34 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
...
+      isLoading: this.props.dataSource.isLoading,
+      isError: this.props.dataSource.isError
+    };
   }
 
   componentDidMount() {
     this.cyInitialization(this.getCy());
+
+    // Listen to data source events.
+    this.props.dataSource.on('loadStart', this.loadStartHandler);
```
mtho11: My only comment is that it is a major architecture change away from Redux dispatching to this event emitter stuff. So we probably need to discuss it as a full team here.

mtho11: The EventEmitter style is more of a Node.js style than browser client style architecturally.

israel-hdez: Yes, indeed this is the reason I'm asking for a quick preliminary review. It's subject to change per your feedback.
If @jshaughn can do his quick review, we could have a discussion today. Then, discuss tomorrow in the water cooler with the rest of the team.

mtho11:
I'm -1 on this change. This is not a standard architecture such as Redux (which is one of the reasons we chose it at the start). So my question is "this problem can't be solved within the current architecture or something else is better and we should migrate to it"? Which opens up a whole new conversation.

jshaughn:
As a side comment, I'm perfectly happy with Redux. I love the way it avoids property drill-down and triggers re-rendering on global state change.  I don't think this current issue casts a negative shadow on Redux nor do I want us to completely revisit our state persistence.  Redux does a good job of storing shared state and maintaining state across component mount/unmount.

The issue here is how to handle multiple instances of the same component.  I'm not sure if redux has any capability around segmenting state by some predictable key.  The thing is that I think we still need a copy of the main graph's settings in redux, right?  Otherwise how do we navigate away and back and maintain settings?

israel-hdez:
Promoting this PR, because it's ready for final review.

pbajjuri20:
When deploying this PR, it can be seen that HTTP traffic is showing "0% error" on the spark-line graph, even though the HTTP edge is 'red'.


error1

israel-hdez:
Hi @pbajjuri20.
I'm seeing the same issue in master. I'll create another issue for your find.
So, if you didn't see another issue, I think we can let this PR go in.

pbajjuri20:
Hi @israel-hdez, I don't see any other issue related to this PR.

Tested and passed on Kiali UI
1.14.0 (542af82c0344c83f38101660794e392b5e1a7382)

PR #1642, opened 2020-01-29
[Metrics] Clicking on a span should lead to Traces view

aljesusg:
Fix https://github.com/kiali/kiali/issues/2107

Require K-charted dump 

Screenshot from 2020-01-29 16-04-56

## Demo

ezgif-6-02293daedd97

### DIFF
--- a/src/components/Metrics/IstioMetrics.tsx
+++ b/src/components/Metrics/IstioMetrics.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import { connect } from 'react-redux';
 import { RouteComponentProps, withRouter } from 'react-router';
 import { Card, CardBody, Grid, GridItem, Toolbar, ToolbarGroup, ToolbarItem } from '@patternfly/react-core';
-import { Dashboard, DashboardModel, ExternalLink, Overlay } from '@kiali/k-charted-pf4';
+import { Dashboard, DashboardModel, ExternalLink, Overlay, VCDataPoint } from '@kiali/k-charted-pf4';
 import { style } from 'typestyle';
 
 import RefreshContainer from '../../components/Refresh/Refresh';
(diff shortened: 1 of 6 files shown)


jotak:
Tested it and looks good! I've found 1 issue though (comment inline)

src/components/Metrics/IstioMetrics.tsx
```diff
@@ -161,6 +161,13 @@ class IstioMetrics extends React.Component<IstioMetricsProps, MetricsState> {
     this.fetchMetrics();
   };
 
+  onClickDataPoint = (_, datum: VCDataPoint) => {
+    if ('traceId' in datum) {
+      HistoryManager.setParam(URLParam.TAB, 'traces');
+      HistoryManager.setParam(URLParam.JAEGER_TRACE_ID, datum.traceId);
```
jotak: Here this works fine when we look at metrics from service details, but we can also see span from app page, so it would fail because it's going to find trace tab in app details, which doesn't exist.
I think it should always redirect to service details/traces, so we need to rebuild the full url, something like:

`/namespaces/${this.props.namespace}/services/${this.props.object}?${URLParam.TAB}=traces&${URLParam.JAEGER_TRACE_ID}=${datum.traceId}`

jotak: Also the same behaviour should be done in CustomMetrics.tsx for runtimes

aljesusg: Wow I didn't think this cases...so I need to do more changes thanks @jotak

jotak:
there's some conflicts now.
I can merge when it's resolved

aljesusg:
Done, this is ready @jotak . Thanks

PR #1644, opened 2020-01-31
Improved error handling in Spans overlay

jotak:
- Check integration flag to know if Jaeger should be fetched
- Do not repeat consecutive errors

Fixes https://github.com/kiali/kiali/issues/2187 (part-of)

### DIFF
--- a/src/components/Metrics/CustomMetrics.tsx
+++ b/src/components/Metrics/CustomMetrics.tsx
@@ -37,16 +37,21 @@ type CustomMetricsProps = RouteComponentProps<{}> & {
   template: string;
 };
 
+type Props = CustomMetricsProps & {
+  // Redux props
+  jaegerEnabled: boolean;
+};
+
 const displayFlex = style({
   display: 'flex'
 });
 
-export class CustomMetrics extends React.Component<CustomMetricsProps, MetricsState> {
+export class CustomMetrics extends React.Component<Props, MetricsState> {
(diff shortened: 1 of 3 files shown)


jotak:
Among the identified scenarios we want to address, I'm putting in bold what this PR fixes, ticking what works:

- [ ] 1. If jaeger is disabled from config, integration is disabled and Kiali should not even try to do anything with Jaeger. No error displayed.
=> (not fully addressed here; needs another fix as @aljesusg explained: https://github.com/kiali/kiali/issues/2187#issuecomment-580651699 )

- [x] 2. Jaeger enabled from config and available/correctly configured: Kiali shows tracing features

- [x] 3. Jaeger enabled from config, unavailable or misconfigured (wrong url): Kiali tries to show tracing features, but it fails, so show errors. **A mechanism must be in place to avoid too many errors being displayed**

- [ ] 4. Jaeger enable from config, available (=> works fine) then shut it down: Kiali consistently should show same behaviour as 3.
=> does not work unless user refreshes the page

- [ ] 5. Jaeger enable from config, unavailable (=> same as 3.) then turn it on: Kiali should show working tracing features.
=> does not work unless user refreshes the page


About "A mechanism must be in place to avoid too many errors being displayed" : I've just added a bool to track consecutive errors. I guess we should do the same in Trace Details view (@aljesusg , fine with you?)

PR #1645, opened 2020-01-31
Reworked some jaeger error handling

jotak:
(Follow-up on https://github.com/kiali/kiali-ui/pull/1644 )

- Some errors were silently ignored when fetching traces, do not ignore them anymore
- Avoid repeating consecutive trace fetching errors
- Changed JaegerInfo model to match backend changes; no more "integrationMessage"
- Use directly JaegerInfo object in redux state as it was strictly identical to JaegerState
- Remove some unused jaeger actions

Backend PR: https://github.com/kiali/kiali/pull/2192
Fixes https://github.com/kiali/kiali/issues/2187 + various issues discussed in that ticket


jotak:
Putting here again the considered scenarios:

1. If jaeger is disabled from config, integration is disabled and Kiali should not even try to do anything with Jaeger. No error displayed.
  - [x] metrics screen => fixed
  - [x] traces screen => fixed

2. Jaeger enabled from config and available/correctly configured: Kiali shows tracing features
  - [x] metrics screen => still working
  - [x] traces screen => still working

3. Jaeger enabled from config, unavailable or misconfigured (wrong url): Kiali tries to show tracing features, but it fails, so show errors. A mechanism must be in place to avoid too many errors being displayed.
  - [x] metrics screen => still working
  - [x] traces screen => still working

4. Jaeger enabled from config, available (=> works fine) then shut it down: Kiali consistently should show same behaviour as 3.
  - [x] metrics screen => fixed
  - [x] traces screen => fixed

5. Jaeger enabled from config, unavailable (=> same as 3.) then turn it on: Kiali should show working tracing features.
  - [x] metrics screen => fixed
  - [x] traces screen => fixed

hhovsepy:
/home/jenkins/go/src/github.com/kiali/kiali-ui/src/pages/Graph/SummaryPanelNode.tsx
TypeScript error in /home/jenkins/go/src/github.com/kiali/kiali-ui/src/pages/Graph/SummaryPanelNode.tsx(658,60):
Property 'integration' does not exist on type 'JaegerInfo'.  TS2339

    656 | 
    657 | const mapStateToProps = (state: KialiAppState) => ({
  > 658 |   jaegerIntegration: state.jaegerState ? state.jaegerState.integration : false,
        |                                                            ^
    659 |   namespaceSelector: state.jaegerState ? state.jaegerState.namespaceSelector : true,
    660 |   jaegerURL: state.jaegerState ? state.jaegerState.jaegerURL : ''
    661 | });

jotak:
@hhovsepy probably some conflicts with latest master? I'm going to rebase & fix

jotak:
rebased

jotak:
Note @hhovsepy @aljesusg I didn't run regression tests about combinations with/without the external URL set.
Also, since jaeger integration is only driven by the "enabled" config flag, it should be a valid config to set "enabled: false" but still provide an external URL => kiali would show hard links to jaeger.

aljesusg:
I'll create some tests in backend about this too

src/components/CytoscapeGraph/ContextMenu/NodeContextMenu.tsx
```diff
@@ -171,9 +171,9 @@ export const getOptions = (
 };
 
 const mapStateToProps = (state: KialiAppState) => ({
-  jaegerIntegration: state.jaegerState ? state.jaegerState.integration : false,
+  jaegerIntegration: state.jaegerState ? state.jaegerState.enabled : false,
   namespaceSelector: state.jaegerState ? state.jaegerState.namespaceSelector : true,
```
aljesusg: I prefer to move this to jaegerEnabled if we are going to remove the integration check.

src/components/Nav/Navigation.tsx
```diff
@@ -122,8 +122,8 @@ class Navigation extends React.Component<PropsType, NavigationState> {
 
 const mapStateToProps = (state: KialiAppState) => ({
   navCollapsed: state.userSettings.interface.navCollapse,
-  jaegerUrl: state.jaegerState ? state.jaegerState.jaegerURL : '',
-  jaegerIntegration: state.jaegerState ? state.jaegerState.integration : false
+  jaegerUrl: state.jaegerState ? state.jaegerState.url : '',
+  jaegerIntegration: state.jaegerState ? state.jaegerState.enabled : false
 });
```
aljesusg: I prefer to move this to jaegerEnabled if we are going to remove the integration check.

src/pages/Graph/SummaryPanelNode.tsx
```diff
@@ -655,9 +655,9 @@ export class SummaryPanelNode extends React.Component<SummaryPanelNodeProps, Sum
 }
 
 const mapStateToProps = (state: KialiAppState) => ({
-  jaegerIntegration: state.jaegerState ? state.jaegerState.integration : false,
+  jaegerIntegration: state.jaegerState ? state.jaegerState.enabled : false,
   namespaceSelector: state.jaegerState ? state.jaegerState.namespaceSelector : true,
```
aljesusg: I prefer to move this to jaegerEnabled if we are going to remove the integration check.

src/pages/ServiceDetails/ServiceDetailsPage.tsx
```diff
@@ -566,8 +585,8 @@ class ServiceDetails extends React.Component<ServiceDetailsProps, ServiceDetails
 
 const mapStateToProps = (state: KialiAppState) => ({
   duration: durationSelector(state),
-  jaegerUrl: state.jaegerState ? state.jaegerState.jaegerURL : '',
-  jaegerIntegration: state.jaegerState ? state.jaegerState.integration : false
+  jaegerUrl: state.jaegerState ? state.jaegerState.url : '',
+  jaegerIntegration: state.jaegerState ? state.jaegerState.enabled : false
 });
```
aljesusg: Same

hhovsepy:
Jaeger disabled: links in Traces tab and in Menu, no error.
Jaeger disabled, wrong external URL: wrong links in Traces tab and in Menu, no error.
Jaeger enabled, wrong external URL: Traces working, link in menu is redirecting to given wrong url, no error.
Jaeger enabled, wrong in_cluster_url: Error "cannot fetch traces/spans" shown in each Traces/Metrics tabs change/reload.
Screenshot from 2020-02-03 13-19-56

Jaeger enabled, Traces are down: Error "cannot fetch traces/spans" shown in each tab change.
Jaeger enabled, Traces are down than UP: Traces are shown back without a problem.

jotak:
@hhovsepy not sure to read if there's some scenario not working or if it's fine?

lucasponce:
> Jaeger disabled: links in Traces tab and in Menu, no error.

Why if Jaeger is disabled do we have links to Jaeger ? I don't understand.

jotak:
> > Jaeger disabled: links in Traces tab and in Menu, no error.
> 
> Why if Jaeger is disabled do we have links to Jaeger ? I don't understand.

It's a consequence of merging the previous "integration" flag into the "enabled" config as we discussed in the issue. If we want to keep the possibility to define an external URL but without in-cluster access, then it means we have to set  `{ Enable: false, URL: 'something' }` in config. It's a bit weird indeed...

We could have other options to tell that we want Jaeger only through links to external URL. For instance:

-  implicitely when InClusterURL is empty? `{ Enable: true, InClusterURL: '', URL: 'something' }`
-  or a new flag "in cluster" explicitly set? `{ Enable: true, InCluster: false, URL: 'something' }`

TBH I don't find any of them very nice.

lucasponce:
Let's keep this simple.

If tracing is "disabled" Kiali doesn't have any visibility and all Jaeger features are ignored.
If tracing is "enabled" Kiali try to work with Jaeger, if there is any problem, then that's logged.

That's would expect, any combination of "disabled" but "I may define some external url" is a little bit weird and it creates confusion.

aljesusg:
> bled" Kiali doesn't have any visibility and all Jaeger feature

> Let's keep this simple.
> 
> If tracing is "disabled" Kiali doesn't have any visibility and all Jaeger features are ignored.

+1

> If tracing is "enabled" Kiali tries to work with Jaeger, if there is any problem, then that's logged.
> 

Yes and not, If tracing is "enabled" we'll have the external links if there is a URL set, and we'll have the itnegration if there is a in_cluster_url, so here is where we must create the logic.

1. Enabled + URL + In_cluster_url    => Show tracing menu link and traces tab integration
2. Enabled + URL and not in:cluster => Show only the menu link
3. Disabled + URl o in_cluster           => Show nothing

> That's what I would expect as a user, any combination of "disabled" but "I may define some external url" is a little bit weird and it creates confusion.

jotak:
@lucasponce we can simplify as you say - we did already simplify quite a bit - but this is a step further, as it means we would stop to do anything for off-cluster jaeger.

Today, it exists and it works. So as Alberto says we could do a distinction depending on "in cluster url" being defined or not. But the reason why I don't like it so much is that it's not very intuitive to understand that we disable some features by using URL instead of in_cluster_url. Whereas `Enable` is quite clear about enabling/disabling features.

So in the end, among the different options we have, I think I still prefer to keep `enabled: false` but still show external links when URL is defined; this is anyway an edge case that I don't expect to see often (I think most of the time Jaeger will be in cluster, or not present at all)

In this situation we say that the Enable flag controls Kiali talking to Jaeger ; when false, Kiali won't talk to Jaeger; but that doesn't prevent us from showing the URL in the UI if the user provided it, given Kiali won't verify or do anything with that URL, it's purely for link generation and nothing else.

This option allows to not loose what currently exists for off-cluster jaeger.

jotak:
@hhovsepy this should be fixed

@aljesusg i've implemented this new endpoint for traces in error right now .. to avoid temporary regression

aljesusg:
Ok here we go 

## Jaeger Disabled
### With Url and In_Cluster
jaeger_disabled_with_url_and_in_cluster
### With no Url
jaeger_disable_no_url
### With URL
jaeger_disabled_but_url

**In this way I think that if jaeger is disabled we shouldn't show the external Link** What do you think @jotak @lucasponce ?

## Jaeger Enabled
### Not Url not in_cluster 
jaeger_enabled_not_url_not_in_cluster
### With Url and not in_cluster
jaeger_enabled_with_url_and_not_in_cluster
### With not url but in_cluster
jaeger_enabled_with_not_url_and_in_cluster
### with all configuration
jaeger_enabled_with_url_and_in_cluster

On the other hand...I don't know if it could be a good idea that we change the traces tab to external link if there is a problem fetching errors or show the content traces tab a message instead the scatter plot saying " Hey There is not integration this is the error but you can go to tracing tool with this link"

jotak:
@aljesusg I'm also fine to do as you suggest. So I'm going to reintroduce the "integration" flag because if we do this we need to dissociate "enabled" and "enable-in-cluster" , the latter being basically the "integration" flag that we had. This is not of course a full rollback: being "in-cluster-disabled" is only if the user explicitly set "in_cluster_url" to an empty string, are we ok on this? We're not going to check connectivity and decide on behalf of the user. Are we on page?

aljesusg:
> @aljesusg I'm also fine to do as you suggest. So I'm going to reintroduce the "integration" flag because if we do this we need to dissociate "enabled" and "enable-in-cluster" , the latter being basically the "integration" flag that we had. This is not of course a full rollback: being "in-cluster-disabled" is only if the user explicitly set "in_cluster_url" to an empty string, are we ok on this? We're not going to check connectivity and decide on behalf of the user. Are we on page?

Yeah with that it would be fine. So, If jaeger is disabled => remove all jager stuff.

If jaeger is enabled but there is not in_cluster_url so show the links in menu and traces tab.
If jaeger is enabled but there is not url but in_cluster_url so the traces tab
If jaeger is enabled and all configuration are set but the integration fails show a message content instead the ScatterPlot showing the error and a link to the external tracing tool if is set. with this way id is a "temporal issue" next time user will see the traces integration.

jotak:
@hhovsepy @aljesusg I've rebased and pushed both back&front. Is it fine for merging? @aljesusg you had a "changes requested" review that I think was addressed.

hhovsepy:
@jotak verified on my side

src/components/CytoscapeGraph/ContextMenu/NodeContextMenu.tsx
```diff
@@ -164,25 +153,25 @@ export const getOptions = (
...
 
   return options;
 };
 
 const mapStateToProps = (state: KialiAppState) => ({
-  jaegerIntegration: state.jaegerState ? state.jaegerState.integration : false,
-  namespaceSelector: state.jaegerState ? state.jaegerState.namespaceSelector : true,
-  jaegerURL: state.jaegerState ? state.jaegerState.jaegerURL : ''
+  jaegerInfo: state.jaegerState || undefined
 });
```
aljesusg: Good one

jotak:
thanks @aljesusg @hhovsepy

PR #1646, opened 2020-01-31
Show Istio Config status only when objects present. Show # objects counted

xeviknal:
Needs backend PR: https://github.com/kiali/kiali/pull/2193

Kiali doesn't compute the validity of a namespace, Kiali analyzes Istio objects living in a namespace. Because of that, the behavior in the Overview page and Graph regarding Istio Objects is changed in this PR.

The new behavior is:
- Kiali shows 'N/A' with a tooltip saying: "No Istio Objects found in this namespace" **when** there aren't istio Objects for that namespace.
- Kiali shows an Icon (Validation higher severity) with a tooltip saying: "N objects analyzed - Error - 4 warnings, 2 errors found", **when** there is one or more Istio Objects.

In the picture below we have:
1. default ns - bookinfo example running on the mesh and we have istio objects (in warning).
2. desert ns - no applications running, no istio objects. It shows N/A.
3. empty ns - it has applications running but no istio objects. It shows N/A.
4. only-objects ns - it hasn't contain any running app but it has istio objects. As a result, Kiali shows the highest severity of objects analyzed.

Screenshot of Kiali Console

Screenshot of Kiali Console - 2020-01-31T175139 067

The list of objects is left unchanged:
Screenshot of Kiali Console - 2020-01-31T175421 226

Whilst the Graph has been updated with same tooltip wording:
Screenshot of Kiali Console - 2020-01-31T175359 787

Fixes https://github.com/kiali/kiali/issues/2113
Fixes https://github.com/kiali/kiali/issues/1930

### DIFF
--- a/src/components/Validations/ValidationSummary.tsx
+++ b/src/components/Validations/ValidationSummary.tsx
@@ -9,6 +9,7 @@ interface Props {
   id: string;
   errors: number;
   warnings: number;
+  objectCount?: number;
   style?: CSSProperties;
 }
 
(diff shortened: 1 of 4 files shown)


abonas:
Is Istio objects the correct terminology here? Not Istio *configuration* objects?

src/components/Validations/ValidationSummary.tsx
```diff
@@ -9,6 +9,7 @@ interface Props {
   id: string;
   errors: number;
   warnings: number;
+  objectCount?: number;
```
jotak: I wonder why it is optional, is it that when there's 0 object, it's undefined instead of 0 ? In that case I think it would be better to explicitly initialize to 0 server-side, and not make it optional here

xeviknal: I am using the optional because this component is used besides the overview and graph page. It is used in the Istio Config list. It is kind of a flag for showing one thing on the tooltip or another.
Perhaps, I could add one extra prop in the component to be more explicit and more readable.
The behavior is: when there is ObjectCount then the tooltip subtitle says the number of analyzed objects. Otherwise, it shows the aggregation severity (highest one).

xeviknal:
@abonas good point. Changed in code and in the description screencapture.

src/components/Validations/ValidationSummary.tsx
jotak: fair enough :)

xeviknal:
thank you all! waiting for QE to merge this one :)

PR #1648, opened 2020-02-04
Fix getDefaultWeights function

kinsolee:
** Describe the change **

Fix the `getDefaultWeights` function to make sure the sum of weights equals 100.

When the `workloads.length` comes to numbers which will be increased 1 after calling `Math.round(100 / workloads.length)` (e.g., 6, 8, 13, etc.), the sum will exceed 100.

e.g., when workloads.length is 6:
100/6=16.666666...
Math.round(100/6) = 17
100%6=4
sum: 17*6+4 = 106

** Screenshot **

image


rhqci:

lucasponce:
Thanks for the PR !

PR #1649, opened 2020-02-04
Summary panel actions: allow open in new window

jotak:
It happens when jaeger is not configured in cluster: external link is used in this case. It must be dealt with something like window.open

Fixes https://github.com/kiali/kiali/issues/2202

@jshaughn I've put the click handler in a common function for wherever it's used. I could have opted for a dedicated component as well... but did the quickest way, if it's ok with you.


pbajjuri20:
Tested and passed on Kiali UI 1.14.0 (94dfae6d8257ed55ccd485b998b70c6f046b0995)
Kiali Server v1.14.0-SNAPSHOT (bedc74754e8d9021b66da2c1995a5f143f6fd021)

PR #1652, opened 2020-02-07
For non-istio apps make correct the namespace/validation spacing

mtho11:
** Describe the change **
For regular k8s apps, meaning non-istio (without sidecars) apps, correct the namespace spacing and validations to look proper.

fixes #https://github.com/kiali/kiali/issues/2217

** Screenshot **

Before:
Kiali_Console_-_Firefox_Developer_Edition

After:
Kiali_Console

### DIFF
--- a/src/pages/Graph/SummaryPanelGraph.tsx
+++ b/src/pages/Graph/SummaryPanelGraph.tsx
@@ -275,7 +275,7 @@ export default class SummaryPanelGraph extends React.Component<SummaryPanelPropT
               NS
             </Badge>
           </Tooltip>
-          {ns}
+          {ns}{' '}
           {!!validation && (
             <ValidationSummary
               id={'ns-val-' + ns}


src/pages/Graph/SummaryPanelGraph.tsx
```diff
@@ -275,7 +275,7 @@ export default class SummaryPanelGraph extends React.Component<SummaryPanelPropT
               NS
             </Badge>
           </Tooltip>
-          {ns}
+          {ns}{' '}
```
jshaughn: Doesn't this add space in all cases?  This is minor and I guess ir doesn't really matter but this way there still may be inconsistent space between the namespace name and the validation status. No big deal, can leave it.

mtho11: Yes, it does but I think it is irrelevant given it keeps the codes base simple and the human eye can barely discriminate the extra 5 pixels:

Kiali_Console

mtho11: I'm totally willing to remedy this if anyone feels it warrants the extra 'if' clause.

mtho11: Here is an image without the extra 5px spacing:
Kiali_Console

Hardly discernible to the human eye.

But again, I'm open to whatever ppl think.

xeviknal: It make sense. And it keeps the code cleaner, I guess.

pbajjuri20:
Tested and passed on Kiali UI1.14.0 (d4faab6b59ca7c710b11ecd37276be6dcfac4ebe)
Kiali Server v1.14.0-SNAPSHOT (069101f97dbe96bfeb0bc9172702076c5ad09898)

PR #1654, opened 2020-02-07
Fix race condition when sorting by Health

cfcosta:
Closes https://github.com/kiali/kiali/issues/1935.

This, combined with https://github.com/kiali/kiali-ui/pull/1561 fixes all the problems with the health tooltips and sorting.

The main problem this tackles is that, when doing the sort, the health for the services/apps/workloads might not be available yet, so it fails. It is ittermitent but it does not happen anymore after the change.

### DIFF
--- a/src/types/Health.ts
+++ b/src/types/Health.ts
@@ -332,11 +332,8 @@ export type NamespaceServiceHealth = { [service: string]: ServiceHealth };
 export type NamespaceWorkloadHealth = { [workload: string]: WorkloadHealth };
 
 export type WithAppHealth<T> = T & { health: AppHealth };
-
-export const hasHealth = <T>(val: T): val is WithAppHealth<T> => {
-  return !!val['health']['requests'];
-};
-
 export type WithServiceHealth<T> = T & { health: ServiceHealth };
 export type WithWorkloadHealth<T> = T & { health: WorkloadHealth };
+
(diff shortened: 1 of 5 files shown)


cfcosta:
@jotak @mtho11 this PR is also a requirement to the typescript 3.7 update.

src/types/Health.ts
```diff
@@ -332,5 +332,11 @@ export type NamespaceServiceHealth = { [service: string]: ServiceHealth };
 export type NamespaceWorkloadHealth = { [workload: string]: WorkloadHealth };
 
 export type WithAppHealth<T> = T & { health: AppHealth };
+
+export const hasHealth = <T>(val: T): val is WithAppHealth<T> => {
```
mtho11: Ah, the magic!

src/types/Health.ts
```diff
@@ -332,5 +332,11 @@ export type NamespaceServiceHealth = { [service: string]: ServiceHealth };
...
+  return !!val['health']['requests'];
```
jotak: would it work with just `!!val['health']` ?
I know all kind of health today does have `requests` field, but just in terms of logic it seems to me that checking for presence of "health" field would be more correct?

Also, function signature says `val is WithAppHealth<T>`, shouldn't it be `val is WithHealth<T>` instead?

PR #1662, opened 2020-02-17
Support Weigthed Routing Wizard when workloads are added/removed

lucasponce:
Fixes kiali/kiali#2205
This PR address the use cases we have discussed.

I'm going to review 2205 in the context of the other wizards and see here what is missing.

But probably the Weigthed Routing Wizard can be tested. @hhovsepy


lucasponce:
Not yet finished.
I found some issues in the Matching Routing:
image

PR #1663, opened 2020-02-17
Decouple settings of CytoscapeGraph component from redux

israel-hdez:
This is pulling up redux properties and disconnecting CytoscapeGraph and EmptyGraphLayout components from Redux. The behavior of the components look unchanged and it makes sense, because the same properties are still being provided to the changed components. Properties are just coming from GraphPage rather than directly from Redux.

Fixes kiali/kiali#2151


pbajjuri20:
Tested and passed on Kiali UI 1.14.0 (4040fada0bcc79b695e1dd333e082b1c0b7c8c55)

PR #1666, opened 2020-02-18
Double-click focus node to navigate to detail

mtho11:
** Describe the change **
Today in the graph, a single click selects a node and a double-click drills down into the node focus graph (aka node graph, node detail graph). On the node focus graph a double-click on the focus node has no effect because we are already drilled down. A natural progression would be to have that double-click navigate to the relevant detail page.

** Issue reference **
fixes https://github.com/kiali/kiali/issues/2046

** Screenshot **

dbl-click

### DIFF
--- a/src/components/CytoscapeGraph/CytoscapeGraph.tsx
+++ b/src/components/CytoscapeGraph/CytoscapeGraph.tsx
@@ -673,6 +673,27 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
     this.graphHighlighter!.onClick(event);
   };
 
+  // This allows us to navigate to the service details page when zoomed in on nodes
+  private handleDoubleTapSameNode = (targetNode: NodeParamsType) => {
+    const makeAppDetailsPageUrl = (namespace: string, nodeType: string, name?: string): string => {
+      return `/namespaces/${namespace}/${nodeType}/${name}`;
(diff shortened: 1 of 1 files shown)


mtho11:
I think the hardest part was figuring out where it should take you depending on app/versioned app/service/workflow. The end result is that it takes you to the app detail screen which has links to all of the others:

Kiali_Console_-_Firefox_Developer_Edition

mtho11:
ok @lucasponce I'll make those changes...

mtho11:
@jshaughn can you review when you get a chance. I have made @lucasponce changes.

pbajjuri20:
Tested and passed on Kiali UI 1.14.0 (f9249b8f7fc051a6fe1c8f88bbbfc88c4bf14b6b)

src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -721,6 +742,7 @@ export class CytoscapeGraph extends React.Component<CytoscapeGraphProps, Cytosca
       version: version,
       service: service
     };
+    console.dir(targetNode);
```
jshaughn: Needs removal I think.

mtho11: Good catch!

PR #1671, opened 2020-02-19
Use "auto" tooltip positioning for graph side panel

jshaughn:
Use "auto" tooltip positioning for the graph side panel badges to reduce tooltip covering.

Fixes https://github.com/kiali/kiali/issues/2196

image

@abonas A couple of comments.  First, as seen in the screenshot the tooltip pointer does not have a black background when positioned to the left.  That seems to be a feature of Tooltip, perhaps a minor bug.  I am going to show UX and create a PF bug if necessary.   Second, I have very occasionally seen the second part of this issue, the "transparent" tooltip.   I can't reproduce it reliably, or sometime at all.  This looks like a PF issue for sure.  The Tooltip looks like it's stuck in a "fade" animation.  We could try to workaround this by explicitly setting animation="none" but I think it 's probably better to generate a PF issue.  Again, I'll pass it by UX.  I think this issue should just focus on the positioning.


pbajjuri20:
Tested and passed on Kiali UI 1.14.0 (fb1f6f80b52ca53e682419bc40cb0fef9b8a6a26)

PR #1675, opened 2020-02-27
Move react-scripts as a devDependency

israel-hdez:
This is a small change to *help* Snyk to work properly. Apparently, react-scripts has a large number of dependencies and is causing a large dependency tree. By moving to the devDependencies, the tree generated by Snyk is pruned because it does not scan this section.

It, anyway, makes sense to move react-scripts as a devDepencency, because it's not used for coding (i.e. not imported in any file under src/), but it is just part of the dev/build framework.


israel-hdez:
Ommiting QE from this PR, because it does not involve any behavior change.

jotak:
+1 for this PR, I'm actually surprised that react-scripts wasn't in dev deps.

I'm also quite surprised that snyk doesn't scan dev deps by default. I believe it's a potential attack vector too, when malicious code is injected (like in [this epic hack](https://snyk.io/blog/a-post-mortem-of-the-malicious-event-stream-backdoor/) back in 2018 ).

Anyway +1 for this PR

israel-hdez:
Merging. Thanks for the thumbs up!

PR #1678, opened 2020-03-02
IstioConfigDetails: notify user when resource changes outside Kiali

lucasponce:
Fixes kiali/kiali#1383

This PR introduces a poll mechanism to refresh the current editable resource with the aim to identify changes on the resourceVersion or errors (delete situation).

On this cases send a warning to the user, in the case when a resource is deleted, it disables "delete/save" actions only allowing to reload/cancel the page to move back to the list page.

Note that this PR uses a fixed timeout, as a first try, this also brings the scenario to enable "refresh" in list/details pages for similar use case: automatically updates.

But this last scenario will be discussed/implemented in a separate issue/PR.

### DIFF
--- a/src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
+++ b/src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
@@ -170,12 +170,12 @@ class IstioConfigDetailsPage extends React.Component<RouteComponentProps<IstioCo
             );
             if (changed) {
               AlertUtils.addWarning(
-                'Istio object type [' +
-                  type +
-                  '] name [' +
+                type +
+                  ':' +
                   props.object +
-                  '] has been modified externally to Kiali. New ResourceVersion:  ' +
-                  newResourceVersion
+                  ' has a newer version (' +
+                  newResourceVersion +
+                  '). Reload to see a new version.'
               );
             }
           }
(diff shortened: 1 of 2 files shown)


src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
```diff
@@ -136,10 +149,51 @@ class IstioConfigDetailsPage extends React.Component<RouteComponentProps<IstioCo
...
+            );
+            if (changed) {
+              AlertUtils.addWarning(
+                'Istio object type [' +
+                  type +
+                  '] name [' +
+                  props.object +
+                  '] has been modified externally to Kiali. New ResourceVersion:  ' +
+                  newResourceVersion
+              );
```
xeviknal: nitpick: The message is shown even if the user has another tab opened and changed the instance there. Meaning that the change has happened within Kiali console. 
Therefore, I'd put a wider message covering this scenario too. Something like 'sidecar:no-ip-tables has a newer version (9182178). Reload to see the new version.'

src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
```diff
@@ -136,10 +149,51 @@ class IstioConfigDetailsPage extends React.Component<RouteComponentProps<IstioCo
...
+            }
+          }
+        })
+        .catch(error => {
+          this.setState({
+            isRemoved: true
+          });
+          AlertUtils.addError(
+            'Could not fetch Istio object type [' +
+              props.objectType +
```
xeviknal: Should we use the name naming as in the alerts for the validations `type:name`?

Screenshot of Kiali Console (23)

Screenshot of Kiali Console (24)

src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
lucasponce: +1 true

src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
lucasponce: +1

PR #1680, opened 2020-03-02
Add graph setting to show headless services

jotak:
Fixes https://github.com/kiali/kiali/issues/976
(Part of https://github.com/kiali/kiali/issues/2197)

Basically, showing headless services is equivalent as not running the dead node appender.
@jshaughn @israel-hdez I'm opening this as draft as there wasn't much discussion yet (since this: https://github.com/kiali/kiali/issues/976). I need to have a good solution to show headless services, another option (as @jshaughn said previously) could be to have a "cold" setting via config map.

Or yet a third option could be to display this "Show headless services" option *only* if we find that there are some (after first graph-fetch). But ... i guess it could generate some tricky situations...


jshaughn:
>  showing headless services is equivalent as not running the dead node appender

This isn't exactly true.  My [comment](https://github.com/kiali/kiali/issues/976#issuecomment-582899298) in the issue said, " We could add a display option ... that could *alter* the appender and not filter these headless service nodes".   I wasn't actually recommending to not run the deadNode appender because it actually does more than just filter on this headless case.  By not running the appender we'd potentially allow through too many nodes.  I'd suggest we continue running the appender but thinking on it a bit more, here are a couple of potential ways we could maybe go:

- Show headless service nodes as long as they have traffic.
  - Pro: See headless services.
  - Con: Don't see valid headless services without active traffic for the time range.
  - Con:  See bad telemetry.  But for smaller time ranges it will come an go quickly.
- Flag headless service nodes.  Instead of filtering them server-side just set a flag (i.e. IsHeadless=true) that can then be acted on client-side, perhaps using the same toggle you are setting up in the PR already.
  - Pro: Optionally see headless nodes.
  - Con: May still see bad telemetry
- Combination: Flag headless services and filter out those w/o traffic.  The filtering could be done server-side or client-side.

So, to summarize, I don't think we want to skip running DeadNodeAppender.  But there are some changes we could make to its behavior.

jotak:
@jshaughn I'd like to better understand what you call "bad telemetry". Is that ... something actually wrong / that shouldn't exist? Or is it real requests but that for some reason don't make sense to show in Kiali? (if so, what's those reasons?)
Do you have some reproducible steps to generate that "bad telemetry" ?

jshaughn:
@jotak There is _bad_ telemetry and _unwanted_ telemetry.  _Bad_ telemetry is telemetry generated by mistake.  Historically, and it's possible this is no longer an issue, istio was generating partial telemetry during various pod lifecycle changes. For example, only the source or destination information was set.  So, in our code we had to detect and remove it.  Unfortunately, this bad telemetry looks just like headless telemetry (i.e. successful traffic to a terminating service node). _Unwanted_ telemetry is different, it's telemetry that we remove to improve the user's experience.  This is generally telemetry that does not map to an existing workload.

If you simply disable the DeadNode appender you'll start getting unwanted telemetry as well as, bad telemetry (if there is any).  I can't say how to reproduce it but you could try wiping your app pods and (delete pods --all -n <ns>) and letting them come back up.   If we are convinced that bad telemetry is no longer a concern we can just remove that code from DeadNode appender and voila, headless service nodes will appear.   My suggestion above was keep the check and flag those nodes as headless, allowing the client to optionally display them.

jotak:
I've tried to see that bad telemetry, unsuccessfully so far. What I *do* see during pod restart or such, is some telemetry to unknown app & workload ; but that telemetry shows errors (503; envoy code UH), so given our implementation of dead node appender (it only filters "dead" nodes that have no errors), these nodes are still visible, the appender being running or not. They are even still visible when the edge goes greyed.

Capture d’écran de 2020-03-09 12-06-35

Is it something that you would have expected to see filtered out by this appender?

jotak:
@jshaughn what about solving that problem of "dead nodes" differently: in telemetry today, there are the fields `destination_service` and `destination_service_name`

When telemetry is normal, they are like:
- destination_service = ui.default.svc.cluster.local
- destination_service_name = ui

On Envoy's UH, it is:
- destination_service = ui.default.svc.cluster.local
- destination_service_name = ui.default.svc.cluster.local

So, perhaps in graph data we could store both `destination_service` and `destination_service_name`, and when we identify missing data (Envoy UH case) we're able to detect that this service is actually the same than "ui", hence merge them. Which would be, I think, the correct way to show data; ie. not duplicating nodes that are actually identical.

If we do that, we can remove some of the dead node appender checks, which would also enable viewing the headless services without any change in the frontend.

wdyt?

jshaughn:
@jotak I think what we're seeing now is a change given V2 telemetry.  We don't want to see all of those terminating service nodes labeled with the service name.  I think in the past those would have consolidated into a single "unknown" destination.  I think we should revisit our dead node handling for V2 (i.e. 1.5+).  I will look more closely at your proposal above.  If I understand what what you are saying you're suggesting we use destination_service to aggregate when we have multiple nodes with the same destination_service but different destination_service_name?

jotak:
@jshaughn yes, I could imagine two different ways of doing that: Either merging / aggregating nodes when an existing `destination_service` is detected. Or perhaps even better, having `destination_service` as the only identifier for services instead of `destination_service_name`. But, looking at the code, changing this would be a big refactoring I guess - so maybe the first option is more pragmatic.

jshaughn:
@jotak I have also considered settling on _service and eliminating the use of _service_name.  And then deferring to the client to extract the name as needed (we don't really want to use the full _service string when labeling).  Let me think on it.

jmazzitelli:
It seems we need to now worry about supporting multiple versions of telemetry, correct?

It sounds like istio telemetry is diverging quite a bit from previous versions of Istio. Are we going to try to refactor our internal implementation to support multiple versions? Or is this going to become a rat's nest trying to handle different versions of telemetry?

jotak:
@jmazzitelli this `destination_service` isn't actually new, its format changed in istio 1.0 iirc, but even at that time we could have chosen to use it as the service ID. I think the choice of `destination_service_name` was kind of arbitrary (certainly more convenient as it provides the name to be displayed). In other words, perhaps it won't breaking anything with backward versions. That's what I hope :)

jshaughn:
@jmazzitelli what do you mean, "will become a rat's nest" :)   Honestly, some of Kiali's value is that "we do the work so you don't have to".  It's never been super clean, we have always tried to handle under the covers a variety of telemetry quirks.  Having said that, the move from mixer to V2 has been challenging.  We now support a variety of things in a backward compatible way.  From a new response time metric, to new telemetry fields, etc.  Mixer is now deprecated.  In Istio 1.7 mixer will be removed.  At that time we could optionally drop support for Istio version <= 1.4, and clean up our code.

jshaughn:
@jotak What does a good Kafka/headless service look like in telemetry for the _service and _service_name fields?  Does _service == _service_name or is _serviceName a short name?  I wonder if we could just filter away nodes where _service_name is set to the host.  As far as I know that shouldn't really happen.

jotak:
well well well ... what do we have here today? It seems that.. everything now just works fine. Without having anything to do. Headless services now seem to report telemetry down to the workload node:

```
istio_tcp_sent_bytes_total{connection_security_policy="unknown",destination_app="kafka",destination_canonical_revision="v1_strimzi-0.16.0_kafka-2.4.0",destination_canonical_service="strimzi",destination_principal="spiffe://cluster.local/ns/kafka/sa/messaging-kafka",destination_service="messaging-kafka-brokers.kafka.svc.cluster.local",destination_service_name="messaging-kafka-brokers",destination_service_namespace="kafka",destination_version="v1_strimzi-0.16.0_kafka-2.4.0",destination_workload="messaging-kafka",destination_workload_namespace="kafka",instance="172.17.0.29:15090",job="envoy-stats",namespace="default",pod_name="web-j11hotspot-5f8588b46f-dccxv",reporter="source",request_protocol="tcp",response_flags="-",source_app="web",source_canonical_revision="j11hotspot",source_canonical_service="web",source_principal="spiffe://cluster.local/ns/default/sa/default",source_version="j11hotspot",source_workload="web-j11hotspot",source_workload_namespace="default"}
```

Capture d’écran de 2020-03-09 17-28-48

All these discussions, perhaps for nothing !
I'll double-check tomorrow with a fresh reinstall of my cluster.

jshaughn:
OK, well that's interesting, now we just need to know why :)

jotak:
closing, superseded by https://github.com/kiali/kiali-ui/pull/1680

PR #1686, opened 2020-03-09
Add new filter in list by label

aljesusg:
Related with https://github.com/kiali/kiali/issues/2233
Fix https://github.com/kiali/kiali/issues/1511 & https://github.com/kiali/kiali/issues/1522

Require https://github.com/kiali/kiali/pull/2496

Added Tooltip in app and version label validation with the value
Added filter by label

ezgif-6-6d75aa400bb0

### DIFF
--- a/src/components/Filters/LabelFilter.tsx
+++ b/src/components/Filters/LabelFilter.tsx
@@ -1,42 +1,57 @@
 import * as React from 'react';
-import { TextInput } from '@patternfly/react-core';
+import { Button, ButtonVariant, Popover, PopoverPosition, TextInput } from '@patternfly/react-core';
+import { defaultIconStyle, KialiIcon } from '../../config/KialiIcon';
 
-export interface LabelFiltersProps {
+interface LabelFiltersProps {
   onChange: (value: any) => void;
   value: string;
-  onKeyPress: (key: any) => void;
+  filterAdd: (value: string) => void;
+  duplicatesFilter: (value: string) => boolean;
 }
 
 export class LabelFilters extends React.Component<LabelFiltersProps> {
-  updateCurrentValue = (key, value) => {
-    this.props.onChange(`${key}=${value}`);
+  onkeyPress = (e: any) => {
+    if (e.key === 'Enter') {
+      if (this.props.value && this.props.value.length > 0) {
+        this.props.value.split(' ').map(val => !this.props.duplicatesFilter(val) && this.props.filterAdd(val));
+      }
+    }
   };
 
   render() {
-    const values = this.props.value.split('=');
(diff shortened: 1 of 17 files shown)


lucasponce:
This should return all items with a label, instead of nothing:
image

And this case should return similar as if you add "d" in the name filter:
image

Example of workload name filter:
image

hhovsepy:
the tooltip on applied label filter is not shown always.
Copy the label key=value from OCP and paste in Kiali UI label filter, hit enter.
It will distribute key and value fields, apply, and display the tooltip (but keeping = on the end).
Screenshot from 2020-03-11 11-02-38
But then manually typing key and value in filter, the tooltip is not shown.
Screenshot from 2020-03-11 11-02-52

aljesusg:
> This should return all items with a label, instead of nothing:
> image

Ok...I can change this...but I disagree about that we should add another kind of filter type "Label presence"

> 
> And this case should return similar as if you add "d" in the name filter:
> image

This is going to return all items with label app=d 

> 
> Example of workload name filter:
> image

aljesusg:
> the tooltip on applied label filter is not shown always.
> Copy the label key=value from OCP and paste in Kiali UI label filter, hit enter.
> It will distribute key and value fields, apply, and display the tooltip (but keeping = on the end).
> Screenshot from 2020-03-11 11-02-38
> But then manually typing key and value in filter, the tooltip is not shown.
> Screenshot from 2020-03-11 11-02-52

The idea is set key and value in separate inputs...I can change this to one input but we should add some help icon to show format etc...

lucasponce:
> The idea is set key and value in separate inputs...I can change this to one input but we should add some help icon to show format etc...

I like to allow enter pattern "label" or "label=value" in a single input, place holder perhaps it's the best way to add the format

aljesusg:
Ok now We can check label presence or if label=value, we can also add more than one check
ezgif-6-68a7f7a8217b

prachiyadav:
Screenshot from 2020-03-13 00-15-11
Screenshot from 2020-03-13 00-12-53

aljesusg:
Ok @lucasponce @hhovsepy This is ready.

Now we check: 

- If label value start by the value you added
- If you set more values ( version=v1,v2,v3) we look if one of them is 
- In Apps we check or possible values (if reviews has version=v1,v2,v3 we check if value is in )
- If you set only the label without value (Example: ap) Is going to look if has a label startsWith

prachiyadav:
Screenshot from 2020-03-20 01-18-27
Screenshot from 2020-03-20 01-18-31

aljesusg:
Hooray

PR #1687, opened 2020-03-09
Fixing Icon from info validations

xeviknal:
Add the info icon back to cross-ns validations.

!! There is a weird snippet of css in the slider.scss file that interferes with other styles in the console. I guess it is right to remove that file. However, I had to add an extra css to fix the namespace selector and page title. @aljesusg do you mind check this out?

Updated: Slider css are embeded here. It looks like there is another iteration to find out the problem. this is a workaround. Only repeated styles are being removed from that blob.

Fixes https://github.com/kiali/kiali/issues/2492

Sliders are working:

Due too duplication of kiali css code, the app details page was slightly decomposed. It is working now:

Screenshot of Kiali Console (38)


lucasponce:
It fixes the problem, but CI needs to be green, worth to investigate what is happening and adjust it.

xeviknal:
@aljesusg do you mind taking a look into this one? I've moved the copy-pasted styles into a separate file and prevented the prettier to run it in this new file.

aljesusg:
> @aljesusg do you mind taking a look into this one? I've moved the copy-pasted styles into a separate file and prevented the prettier to run it in this new file.

ok. What do you think load the slider css file like you did with appdescription only when we use the component ?

xeviknal:
@aljesusg you are right. Approached. :+1:

hhovsepy:
post merge verification, regression testing for icons pass:
Screenshot from 2020-03-11 11-53-33

PR #1688, opened 2020-03-11
Allow root nodes to show metrics to other namespaces

jotak:
Fixes https://github.com/kiali/kiali/issues/2498

Metrics are now visible in summary panel:

Capture d’écran de 2020-03-11 11-32-58

Note: for testing you can use this setup: https://github.com/ia3andy/game-of-stacks/blob/master/ISTIO-QUICK-START.md

### DIFF
--- a/src/pages/Graph/SummaryPanelNode.tsx
+++ b/src/pages/Graph/SummaryPanelNode.tsx
@@ -172,7 +172,9 @@ export class SummaryPanelNode extends React.Component<SummaryPanelNodeProps, Sum
       // - istio namespace nodes (no source telemetry)
       const reporter: Reporter = nodeData.nodeType === NodeType.UNKNOWN || nodeData.isIstio ? 'destination' : 'source';
       // note: request_protocol is not a valid byLabel for tcp filters but it is ignored by prometheus
-      const byLabels = nodeData.isRoot ? ['destination_service_namespace', 'request_protocol'] : ['request_protocol'];
+      const byLabels = nodeData.isOutside
+        ? ['destination_service_namespace', 'request_protocol']
+        : ['request_protocol'];
       promiseOut = getNodeMetrics(
         nodeMetricType,
         target,
@@ -195,6 +197,9 @@ export class SummaryPanelNode extends React.Component<SummaryPanelNodeProps, Sum
       // comparator in getNodeDatapoints).
       const isServiceDestCornerCase = this.isServiceDestCornerCase(nodeMetricType);
       const byLabelsRps = isServiceDestCornerCase ? ['destination_workload', 'request_protocol'] : ['request_protocol'];
+      if (nodeData.isOutside) {
+        byLabelsRps.push('source_workload_namespace');
+      }
       const promiseRps = getNodeMetrics(
         nodeMetricType,
         target,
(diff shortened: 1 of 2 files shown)


jotak:
DNM label, question needs to be answered first: why did we exclude metrics from root nodes to other namespaces?

cc @jshaughn @israel-hdez

src/pages/Graph/SummaryPanelNode.tsx
```diff
@@ -243,26 +243,14 @@ export class SummaryPanelNode extends React.Component<SummaryPanelNodeProps, Sum
...
       comparator = (metric: Metric, protocol?: Protocol) => {
         return (protocol ? metric.request_protocol === protocol : true) && metric.destination_workload === UNKNOWN;
       };
-    } else if (data.isRoot) {
-      comparator = (metric: Metric, protocol?: Protocol) => {
-        return (
-          (protocol ? metric.request_protocol === protocol : true) &&
-          this.isActiveNamespace(metric.destination_service_namespace)
-        );
-      };
```
jshaughn: The graph has many use cases and it's not always easy to recall intent.  I should leave myself even more inline comments...   I think the intent here is this:  Root nodes are leaf nodes and leaf nodes can be outside of the selected/active namespaces.  The graph wants to show traffic flowing into, within, and out of the selected namespaces.  It does not want to include traffic completely outside of the selected namespaces.  I think the code above is making an assumption that the root node is outside of the selected namespaces and is filtering traffic from it to other nodes outside of the selected namespaces.  But your problem case is different, the root node is inside and traffic is flowing out of the selected namespaces.

I don't think we want to remove this block of code.  But it does need to change. I think line 259 should change to:

`} else if (data.isRoot && data.isOutside) {  // filter out traffic completely outside the active namespaces`

jotak: As discussed: I understand better, but then it doesn't cover all the various off-namespace cases :-)

Say we have A(x) ( = Service A in namespace x ), B(y) and K(k). Say our active namespace is x only.

If connections are:
- K(k) talks to A(x)
- K(k) talks to B(y)

Then it would be fine, the second connection is filtered out because K is outside and y isn't active.

But if connections are:
- A(x) talks to K(k)
- B(y) talks to K(k)

Then it's not filtered out because K is not a root.

I think the fact that it's root shouldn't influence decision here. What should influence is just `isOutside` and the *other* namespace (not especially destination namespace as it's currently coded, it could be the source namespace for inbound metrics).

jotak: put differently, we could just take the source & destination namespaces, and discard metrics when both of them are not active.

jotak:
@jshaughn I've run several tests, I think it looks good like this. Hoping to not forget any special case

jotak:
@israel-hdez yes, I was thinking the same about isRoot, although I only checked a small part of the code, I haven't seen any fundamental reason to have it. However it seems to be used to avoid unnecessary queries when we know that there's no inbound connection. So, it's still useful as an optimization flag.

jotak:
@jshaughn @israel-hdez actually I pushed a last commit because there was a missing case: when there's several active namespaces and node is outsider, valid metrics have to be summed.

jshaughn:
@jotak gets a master badge for side-panel expertise.  Nice find!   As for isRoot, that flag was at one point more significant as we used a different node style and such.  But even now it can be useful for optimization, I think we can keep it.

jotak:
Thanks

PR #1690, opened 2020-03-13
Don't rely on path value for "Istio Config" title.

jshaughn:
Fixes https://github.com/kiali/kiali/issues/2471

After:

image


jshaughn:
@jotak If it's too distasteful to put that string literal test in VirtualList let me know.  Not sure what else to do other than maybe add a util function in masthead that takes the path 'type ' and returns a display value.

pbajjuri20:
Before:

istio-config1


Tested and passed on Kiali UI 1.15.0 (8582f5e4673c99def411f62a3f1e6bdf577fbab9)

After:

istio config

pbajjuri20:
Thanks @jshaughn  It having the perfect information now.  

Tested and passed on Kiali UI 1.15.0 (f32ce4421853eb0932405533ddfa481b5a0bee5d)

Istio

PR #1693, opened 2020-03-16
Adding pficon FaceFont to be able to load icon into graph

xeviknal:
Adjusting the icons in the graph to use pficon font.

Fixes https://github.com/kiali/kiali/issues/2513

The icons are been able to load now. However, the size of the text is pretty small that chrome shows it as a transparent square. When users zoom in, the lock is displayed properly.
In firefox, instead of showing the transparent square, the icon doesn't appear at all.

@jshaughn @mtho11 is this behavior familiar to you?

Screen recording (24)

cc @aljesusg


src/styles/index.scss
```diff
@@ -11,8 +11,6 @@ $font-size-base: 14px;
 
 //PF4 Styles
 $pf-global--enable-reset: false;
-$pf-global--disable-fontawesome: true;
-$pf-global--font-path: '@patternfly/patternfly/assets/fonts/';
```
aljesusg: PF team told me about add this values...I don't know if we need them now that we migrated 100% to PF4-. Let me check it

jshaughn:
@xeviknal @aljesusg I'm playing with the branch.  The current behavior seems to be similar but worse to what we  see sometimes even in older versions (see [this OSSM issue](https://issues.redhat.com/browse/OSSM-166)).  And it's similar to [this Cytoscape issue](https://github.com/cytoscape/cytoscape.js/issues/2592).

It's hard to understand the behavior of cytoscape here.  It seems like it's something to do with maybe pre-loading/caching, but also maybe the size allocated by cy for the label.  If you change label options the icon tends to appear and then stay.  But the initial rendering is a problem.

mtho11:
At this point should we revert the PR just to get the release out? Might take the pressure off.

xeviknal:
I found what happened. It is about the order between the font is loaded and when it is rendered.
 
Screen recording (25)

I did something not really elegant. I'll give it a second chance to have something to be comfy with.

xeviknal:
It is working now. Two different issues fixed:

1. Iconic fonts have changed in the PF4 upgrade. Thus, It needs to change both the unicode char and the font family.
2. Cytoscape has a rendering issue when it uses a font that has not been loaded yet. See previous issue: KIALI-2817. Thus, `pficon` has been forced to load as soon as the page is loaded.

xeviknal:
I'd like to test virtual service, circuit breaking and sidecar badges as well. DNM

xeviknal:
Third issue in this PR:
FontAwesome was upgraded from 4 to 5. Some icons were renamed.

Screenshot of Kiali Console (40)

xeviknal:
Can we consider that this PR fixes https://github.com/kiali/kiali/issues/2513 ?
@jshaughn @mtho11 @hhovsepy  do you mind to confirm it, please?

src/components/CytoscapeGraph/graphs/GraphStyles.ts
```diff
@@ -33,7 +33,7 @@ const EdgeIconMTLS = icons.istio.mtls.ascii; // lock
 const EdgeIconDisabledMTLS = icons.istio.disabledMtls.ascii; // broken lock
 const EdgeTextOutlineColor = PfColors.White;
 const EdgeTextOutlineWidth = '1px';
-const EdgeTextFont = 'Verdana,Arial,Helvetica,sans-serif,FontAwesome,PatternFlyIcons-webfont';
+const EdgeTextFont = 'Verdana,Arial,Helvetica,sans-serif,FontAwesome,PatternFlyIcons-webfont,pficon';
```
jshaughn: Maybe we can remove ",FontAwesome,PatternFlyIcons-webfont", I'll test it when I run the PR.

src/styles/_fonts.scss
```diff
@@ -0,0 +1,9 @@
+// PF4 needed config for fonts
+$pf-global--disable-fontawesome: true;
+$pf-global--font-path: '@patternfly/patternfly/assets/fonts/';
+
+// This is necessary to have icons into Graph edge
+@font-face {
+  font-family: 'pficon';
+  src: url(../../node_modules/@patternfly/patternfly/assets/pficon/pficon.woff2);
```
jshaughn: Do we want to reference this from under node_modules or should we copy this under src in package.json and use a local  path?

src/components/CytoscapeGraph/graphs/GraphStyles.ts
mtho11: @jshaughn I tested that out and it looks like you are correct we can remove that

Kiali_Console

src/styles/_fonts.scss
mtho11: Personally, I think this is fine the way it is.

src/components/CytoscapeGraph/graphs/GraphStyles.ts
mtho11: If we change this we will want to get QE @hhovsepy to re-test just to be sure.

src/styles/_fonts.scss
xeviknal: I am being consistent with other places where we load PF assets in the same way.
https://github.com/kiali/kiali-ui/blob/82f39359463a45f1a42b34a80e02be6d701ee22a/src/styles/index.scss#L10

I am not very familiar with the way you are suggesting. And copying assets from dependency to our source makes me think that we will be more error-prone in next pf upgrades.

src/components/CytoscapeGraph/graphs/GraphStyles.ts
xeviknal: I thought I had removed those two. They are removed in the last commit.
Both of you have tested, me and alberto. I think it is good to go.

src/styles/_fonts.scss
jshaughn: This way is fine!

PR #1694, opened 2020-03-17
Make traffic animation proportional to volumetry

jotak:
- Needs to calibrate amplitude as first step
- Use linear relation between rate and dots generator
- Use consistent code between http & tcp rates

Fixes https://github.com/kiali/kiali/issues/2469

Capture d’écran de 2020-03-17 09-52-25

### DIFF
--- a/src/components/CytoscapeGraph/TrafficAnimation/AnimationTimerConfig.ts
+++ b/src/components/CytoscapeGraph/TrafficAnimation/AnimationTimerConfig.ts
@@ -1,22 +1,44 @@
 import { clamp } from 'utils/MathUtils';
 
+// Some information about the mathematical problem here:
+// We try to project a set of RATE values from an upper-unbounded interval ]0, +Inf[ to a set of TIMER DELAY values in
+//    a bounded interval [40ms, 5000ms] (arbitrary bounds).
+// TIMER DELAY is the delay for dots generation in the animation. The lower it is, the more dots there are.
+//
+// The constraints are:
+// - We'd like the relation to be inversely proportional: if RATE 1 is twice as big as RATE 2, DELAY 1 must be half of DELAY 2.
+// - The animations should be quite representative of volumetry both _relatively_ and _absolutely_: _relatively_ is the
+//    previous constraint (edges compared to each other); but _absolutely_ is another thing: regardless edges compared to
+//    each other, we should have a notion of low and high traffic.
+//
+// To better understand the issue, we can imagine two rates R1 = 0.001 rps and R2 = 0.1 rps. R2 should have 100 times more
+//    dots than R1. So it's likely to be "crowded", and looks like a high volumetry if we only consider the relative
+//    relationships. But thinking in absolute, 0.1 rps is small, it shouldn't be crowded.
+
 const initialThreshold = 50;
 
-class AnimationTimerConfig {
-  threshold: number;
+export class AnimationTimerConfig {
+  private threshold: number;
+  private scaleFactor: number;
 
-  // baseDelay: for a rate of 1 rps/bps, there will be one dot every `baseDelay` second.
-  constructor(public baseDelay: number) {
+  // baseDelay: for a rate of 1 rps/bps, there will be one dot every `baseDelay` seconds.
+  constructor(private baseDelay: number) {
     this.threshold = initialThreshold;
+    this.scaleFactor = 1;
   }
 
   resetCalibration() {
     this.threshold = initialThreshold;
+    this.scaleFactor = 1;
   }
 
   calibrate(rate: number) {
+    // We make this.threshold grow with max rate
+    // (there's some "magic numbers" here, not really explained, perhaps this formula can be changed, but keep threshold following max rate)
+    // The scale factor for this graph is updated with the changed threshold.
     if (rate > this.threshold / 1.5) {
       this.threshold = 3 * rate;
+      this.scaleFactor = this.computeScaleFactor();
     }
   }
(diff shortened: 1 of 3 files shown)


jotak:
cc @josejulio

josejulio:
It needs to be tested with http also, it was affected by this PR IIUC.


> I'm fine with the code changes although I'm not sure we quite have the settings we want. 

I agree, perhaps we need a some thresholds so when our current data has so little traffic that we want to show "little traffic" and not "lots of traffic".

jotak:
I think it can be improved by just setting a bigger initial amplitude, let me try

jotak:
> I agree, perhaps we need a some thresholds so when our current data has so little traffic that we want to show "little traffic" and not "lots of traffic".

yes, that seems to be what we should do

jotak:
I've tried something quite different, that should adapt to low or high traffic equally. Could you retest?

israel-hdez:
@jotak do you need a *good* review?
Because I don't understand what this code is doing. So, it may take me a little time to get familiar with it (blame me for not following closely @josejulio progress on this).

I'll take a closer look tomorrow to give better help here.

jotak:
I'll try to better comment the code.

The mathematical problem here, is that we try to project a set of RATE values from an upper-unbounded interval [0, +Inf[ to a set of TIMER DELAY values in a bounded interval [40ms, 5000ms] (arbitrary bounds).

TIMER DELAY is the delay for dots generation in the animation, so the lower, the more dots there are.

The constraints are:

- We'd like the relation to be inversely proportional: if RATE 1 is twice as big as RATE 2, DELAY 1 must be half of DELAY 2. This constraint is basically the reason of the change, currently it isn't proportional, which makes sometimes the animation not being clearly representative of volumetry.

- The animations should be quite representative of volumetry both _relatively_ and _absolutely_: _relatively_ is the previous constraint (edges compared to each other); but _absolutely_ is another thing: regardless edges compared to each other, we should have a notion of low and high traffic. I struggled a little bit to achieve this although I think it's not too bad now.

To better understand the issue, imaging two rate R1 = 0.001 rps and R2 = 0.1 rps. R2 should have 100 dots more than R1. So it's likely to be "crowded", and looks like a high volumetry if we only consider the relative relationships. But thinking in absolute, 0.1 rps is small, it shouldn't be crowded.

jotak:
..., now, to explain the code:

- Constraint 1: expression of relative volumetry:
This is mostly achieved by this formula:

```javascript
    const d = (1000 * this.baseDelay * scaleFactor) / rate;
```

Both `baseDelay` and `scaleFactor` are constant within the set of edges that we consider, so we have the inversely proportional relations that we want.

The constraint 1 is only broken at boundaries, because of clamping (`clamp(d, 40, 5000)`) => proportionality not guaranteed between two very low / two very high rates.

- Constraint 2:
(TBC)

jotak:
@israel-hdez @jshaughn unit tests and code comments in last commit

src/components/CytoscapeGraph/TrafficAnimation/AnimationTimerConfig.ts
```diff
@@ -0,0 +1,71 @@
...
+      return Math.pow(this.threshold / initialThreshold, 0.95);
+    }
+    return this.threshold / initialThreshold;
+  }
+}
+
+// HTTP config: 1 RPS => 1 dot every second
+export const timerConfig = new AnimationTimerConfig(1);
+// TCP config: 4 bps => 1 dot every second
+export const tcpTimerConfig = new AnimationTimerConfig(4);
```
jshaughn: I feel like this is still too small.  Requests are going to have at least a few bytes sent and even with the very low traffic in bookinfo (1rps) I get a lot of dots.  I tried 20 and 25 locally and they both seemed maybe better as initial values.  Of course bookinfo (with mongo) only has a single TCP edge, so perhaps it's not the best example.  But have you tried maybe 20 as an initial value in your example?

israel-hdez: I was thinking that dots should reflect kb/sec rather than bytes/sec.
Sending a simple "HELLO" message is already more than 4bytes. So I'm not sure how well byte/sec would "scale" on environments that receive lot of requests.

But I won't argue. This is just a comment.

israel-hdez:
OK, I understand this better.
I don't like having magic numbers for this :) Perhaps we should use stats for finding the "magic range", but let's leave that for the next iteration (if any).

I'm fine with the code. I think using kb/sec for TCP instead of bytes/s will solve Jay's comment. Perhaps you could try and if that solves the issue, we could then approve/merge.

jotak:
@jshaughn @israel-hdez no problem, I can change the numbers.
Remember that the initial value (4) is there to account behaviour on very low traffic. So, I can change 4 to 25, it's going to be fine, but I wouldn't set higher values such as kbps (would be something like 4096) because then it wouldn't correctly show traffic for values below that.
The upscaling mechanism is used to represent higher traffic.

About the magic numbers in upscaling (`if (rate > this.threshold / 1.5) { this.threshold = 3 * rate; }`) ; they're not essential and I can remove them if they make you feel uncomfortable. I just think that they kinda "smooth the curve" at thresholds but I admit I didn't think a lot about it (that was very empirical). I can simplify to `if (rate > this.threshold) { this.threshold = 2 * rate; }` . The essential thing is to make threshold bigger than max rate.

jotak:
Thanks @jshaughn @israel-hdez !

PR #1705, opened 2020-03-19
Better wording for Dropdown menus

cfcosta:
Closes kiali/kiali#2145.

Changes the duration dropdowns to support prefixes and suffixes, and also make so the duration dropdown is aware of replay.

Graph without replay:
image

Graph with replay:
image

Replay length when there's no start time:
image

Replay length with start time:
image


pbajjuri20:
Tested and passed on Kiali UI 1.16.0 (ed8dbfd3b4e770bfe1845b5e2e79d72fb48bc9e1)

cfcosta:
@jshaughn shouldn't be the case, no, it should be exactly like it was before :) I might have glossed over them or they might have different components for that, let me take a look.

jshaughn:
> shouldn't be the case, no, it should be exactly like it was before :)

@cfcosta actually we don't want it exactly like before because it was part of the motivation for the issue.  We want it to be improved.  We want it to probably go back to "Last xxx".

cfcosta:
So, did the change on the Workload pod logs, if there's any other place I might have missed please do tell me.

Before:
image

After:
image

pbajjuri20:
LGTM!

Before

workload before

After

Workload after

src/components/Time/TimeRangeComponent.tsx
```diff
@@ -77,7 +77,8 @@ export default class TimeRangeComponent extends React.Component<Props> {
   }
 
   renderDuration(d?: DurationInSeconds) {
-    const options = this.props.allowCustom ? { custom: 'Custom', ...serverConfig.durations } : serverConfig.durations;
+    const durations = humanDurations(serverConfig, undefined, 'Traffic');
```
jshaughn: This does not work for logs and is not as good for custom runtimes. Change to:
` const durations = humanDurations(serverConfig, 'Last', undefined);`

abonas:
I think that "Tail" might be not the greatest word here.
Not all users know what that means in this context.
To make it more clear, it can be changed imo to "show last X lines"

jshaughn:
> To make it more clear, it can be changed imo to "show last X lines"

I think maybe just "Show X lines" because we already use "Last" in the duration box.

mtho11:
I guess I'm surprised to see that this issue wasn't a UxD issue.

cfcosta:
@mtho11 agree, we can either get some UX review on this one or do this specific part of the change on another PR, what would you think is the best way to go?

mtho11:
@cfcosta why don't you put on tomorrow's UX meeting agenda, then you can still have it in by end of day.

lucasponce:
@mtho11 @cfcosta 
Please, mark the issue (not only the PR) with the "waiting ux" label and also added to the agenda.
We have defined a common flow to identify thess requests.

This is described in kiali-internal mailing list some time ago and with UX agreement.

src/components/Time/TimeRangeComponent.tsx
jshaughn: @cfcosta this change is independent of that UX wording issue brought up by Alissa.

cfcosta: @jshaughn I changed it to "X Traffic" because that was what was there before. Should I change it either way?

jshaughn: Yes, please change it, because before "X Traffic" we had "Last X" and that's what we want, and is part of the motivation of this issue.

cfcosta:
Got review from the UX team on the weekly meeting.

- "Tail X lines" will be changed to "Show Last X Lines"
- Will also investigate a way to add a checkmark or some other way to show which option you selected on those dropdowns.

jshaughn:
I still think "Show Last X Lines" is too verbose and I don't like the redundant "Last" in the two options.  I'm just whining :)

mtho11:
This can be merged right? @jshaughn @cfcosta

cfcosta:
@mtho11 yes, think this should be good to merge. Just did a last test.

PR #1709, opened 2020-03-23
Show labels ui

aljesusg:
Fix https://github.com/kiali/kiali/issues/2547

## Workload List
workloads

Now in workloads we remove the label validation and we add the warning in details
missing

## App List
apps
## Service List
services
## when you Filter by a Label we light up the label applied
lighted 
## Overview List
Screenshot from 2020-03-23 11-12-44

### Overview Card
overview_card

#### Overview Card Labels
overview_card_labels

### DIFF
--- a/src/pages/Overview/Filters.ts
+++ b/src/pages/Overview/Filters.ts
@@ -52,7 +52,7 @@ export const labelFilter: FilterTypeWithFilter<NamespaceInfo> = {
   id: 'label',
   title: 'Label',
   placeholder: 'Filter by Label',
-  filterType: FilterTypes.custom,
+  filterType: FilterTypes.label,
   customComponent: LabelFilters,
   action: FILTER_ACTION_APPEND,
   filterValues: [],
--- a/src/pages/Overview/OverviewPage.tsx
+++ b/src/pages/Overview/OverviewPage.tsx
@@ -10,12 +10,13 @@ import {
   EmptyStateVariant,
   Grid,
   GridItem,
-  Title
+  Title,
+  Tooltip,
+  TooltipPosition
 } from '@patternfly/react-core';
 import { style } from 'typestyle';
 import { AxiosError } from 'axios';
 import _ from 'lodash';
-
 import { FilterSelected } from '../../components/Filters/StatefulFilters';
 import * as FilterHelper from '../../components/FilterList/FilterHelper';
 import * as API from '../../services/Api';
@@ -50,6 +51,8 @@ import ValidationSummary from '../../components/Validations/ValidationSummary';
 import { DurationInSeconds, IntervalInMilliseconds } from 'types/Common';
 import { Link } from 'react-router-dom';
 import { Paths } from '../../config';
+import { PfColors } from '../../components/Pf/PfColors';
+import VirtualList from '../../components/VirtualList/VirtualList';
 
 const gridStyle = style({
   backgroundColor: '#f5f5f5',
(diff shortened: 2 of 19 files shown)


jotak:
I think I need more context to understand this.

I have some microservices deployed, with labels. In workloads list, I see these labels (so far so good). It's quite interesting to be able to filter by label [1].

But in Applications list, I don't see any labels, why? "Applications" isn't a k8s object, it's an abstract view, so how am I supposed to define labels on Applications?

Similarly in Overview page I only see "0 labels" whereas I do have some labels defined on workloads, so I'm not sure why.

[1] A small note in filtering labels: labels are displayed as "key: value", but for filtering it needs an equal: "key=value", so it seems inconsistent

prachiyadav:
Screenshot from 2020-03-24 23-40-49
Screenshot from 2020-03-24 23-40-42
Screenshot from 2020-03-24 23-39-54
Screenshot from 2020-03-24 23-39-23
Screenshot from 2020-03-24 23-39-01

src/pages/Overview/OverviewPage.tsx
```diff
@@ -409,6 +414,34 @@ export class OverviewPage extends React.Component<OverviewProps, State> {
     );
   }
 
+  renderLabels(ns: NamespaceInfo): JSX.Element {
+    const button = (
+      <div style={{ color: ns.labels && PfColors.Blue400 }}>
+        {ns.labels ? Object.entries(ns.labels).length : 0} labels
```
jshaughn: I think it's worth handling the "1 label" case.

aljesusg:
> I think I need more context to understand this.
> 
> I have some microservices deployed, with labels. In workloads list, I see these labels (so far so good). It's quite interesting to be able to filter by label [1].
> 
> But in Applications list, I don't see any labels, why? "Applications" isn't a k8s object, it's an abstract view, so how am I supposed to define labels on Applications?
> 
> Similarly in Overview page I only see "0 labels" whereas I do have some labels defined on workloads, so I'm not sure why.
> 
> [1] A small note in filtering labels: labels are displayed as "key: value", but for filtering it needs an equal: "key=value", so it seems inconsistent

For Applications I am showing all the workloads labels of that application. See for example bookinfo the application review as label version=v1,v2,v3

src/pages/Overview/OverviewPage.tsx
aljesusg: If there is 1 label so ns labels will be true and then Object.entries(ns.labels).length will be 1 .

I don't understand your proposal @jshaughn

jshaughn: Sorry, it's just that I don't want to see "1 labels".  I'd prefer to add a test for the singular case and display "1 label". So,
0 labels
1 label
N labels

aljesusg: Ohhh understand sorry I didn't see it

aljesusg:
@jotak @jshaughn 

Opened for repalce = by : https://github.com/kiali/kiali/issues/2550
Opened for wrap text https://github.com/kiali/kiali/issues/2558

The others comment s I'll address them here to be related with the visualization of labels.

Thanks

jotak:
> For Applications I am showing all the workloads labels of that application. See for example bookinfo the application review as label version=v1,v2,v3

Forget it, I wasn't using the latest backend code, that's why I didn't see them.
I would have some other questions or remarks but it's more about the whole feature & backend code than about this particular PR, so not going to create some noise here :)

aljesusg:
> @jotak @jshaughn
> 
> Opened for repalce = by : [kiali/kiali#2550](https://github.com/kiali/kiali/issues/2550)
> Opened for wrap text [kiali/kiali#2558](https://github.com/kiali/kiali/issues/2558)
> 
> The others comment s I'll address them here to be related with the visualization of labels.
> 
> Thanks

I'll do this issues in https://github.com/kiali/kiali/issues/2550

src/pages/Overview/Filters.ts
```diff
@@ -52,7 +52,7 @@ export const labelFilter: FilterTypeWithFilter<NamespaceInfo> = {
   id: 'label',
   title: 'Label',
   placeholder: 'Filter by Label',
-  filterType: FilterTypes.custom,
+  filterType: FilterTypes.label,
```
jotak: As discussed  on IRC I think it would be better to have a different filter type here, something like "nsLabel", to not mix up with labels on workloads

jotak: (since it seems the problem was already there before, I can open a separate ticket)

lucasponce:
In the overview page, it's visually inconsistency to use two different ways of tooltip:
image
image

We need to unify that, @kiali/ux-reviewers could suggest, but I would unify it.

I'm +1 to align the labels with the header:
image
This looks moved.

Also, in Service List I would move Details as last column, for consistency with Apps/Workloads:
image

I know column was in this position before, but now I guess it makes sense.

I'm undecided if I like to change the "=" per ":", because to me these are two different things.
One is an operator in an expression (used also in other text inputs like find/hide in graph where "=" is used) and ":" is used to represent a value in an object.

Also Kubectl use "=" in all places (on label operator and in --show-labels) only uses ":" when using json object.

lwrigh:
I would suggest using the black tooltip in this case. I also want to take a closer look at the placement of the labels within the card when we do another pass at a card layout.

aljesusg:
## Next iteration let a list view in overview to display more information ( we will be able to add more information in the future with this )

ezgif-4-8948fc97a164


## Proposal

Change the buttons
buttons

by icons, for example in ManageIQ we used icons to show views
Screenshot from 2020-03-30 16-52-13

aljesusg:
@jotak @lucasponce I did all the changes proposed by UX. 

The problems related with replace = by : and others are in the fix PR.

Please review it

aljesusg:
Screenshot from 2020-04-01 13-04-30
Screenshot from 2020-04-01 13-04-08

lucasponce:
In the Overview page I don't know what to do with this:
image
image

If we split the Istio Config icon in a column, the mtls should have same consistency in the list view.

lucasponce:
I know these changes are in master and are not in the goal of the PR:
image
But as we are updating this toolbar, I'd appreciate if we can align the "Namespaces" with the filter and the "Show health for" with the control.

lucasponce:
Also, I'm 50/50 with this scenario:
image

When showing namespaces in a card, the sorting control makes sense.

But when using the list, as a user it seems weird and I guess I'd expect to sort with the table headers:

image

I'm undecided 50/50 on this, as pros: consistency within the page, cons: introduce a weird scenario or a table (when in other pages, tables are sorted by headers).

aljesusg:
> Also, I'm 50/50 with this scenario:
> image
> 
> When showing namespaces in a card, the sorting control makes sense.
> 
> But when using the list, as a user it seems weird and I guess I'd expect to sort with the table headers:
> 
> image
> 
> I'm undecided 50/50 on this, as pros: consistency within the page, cons: introduce a weird scenario or a table (when in other pages, tables are sorted by headers).

We can disable the sort when is a list, maybe we can ask this to UX

lucasponce:
> We can disable the sort when is a list, maybe we can ask this to UX

Yes, that was a possibility, to disable/hide it when it's a list and just enable the columns for sorting.
But probably UX should give an opinion, I'm 50/50, so both options work for me.

lwrigh:
Thought about this a bit more and it might make sense to keep the sort option for the card views since the namespace name is still present on the card and it would still make sense to be able to  sort that name by alphabetical or reverse alphabetical order.

aljesusg:
Ok I did the changes and is updated . Thanks @lwrigh @lucasponce 

- Fixed tag show Health in dropdown
- Now we use virtualList component to
- Vertical align in all cells ( include workloads, apps etc...)
- Hide sort dropdown and enabled sortable headers in list view
- Added TLS column in List View

Screenshot from 2020-04-03 11-35-36

Screenshot from 2020-04-03 11-35-48

Screenshot from 2020-04-03 11-35-59

lwrigh:
Looks good to me! Having the sortable headers enabled only for list view looks good to me, it looks way cleaner than having the sort field in the toolbar.

lucasponce:
Almost perfect, but one single detail:
image

If we can align "Namespaces" title with the dropdown below on the left, it would be in sync with the rest of pages.

I.e. Graph:

image

Apps:

image

PR #1714, opened 2020-03-24
Fix duplicate results when filter by labels

aljesusg:
Fix https://github.com/kiali/kiali/issues/2550

Go to workload list and try to filter by label with "app=reviews version=v2,v3"

## Now
results

## Expected
expected

### DIFF
--- a/src/components/Filters/LabelFilter.tsx
+++ b/src/components/Filters/LabelFilter.tsx
@@ -32,21 +32,21 @@ export class LabelFilters extends React.Component<LabelFiltersProps> {
         />
         <Popover
           headerContent={<span>Label Filter Help</span>}
+          position={PopoverPosition.right}
           bodyContent={
             <>
               To set a label filter you must enter values like.
               <br />
               <ul style={{ listStyleType: 'circle', marginLeft: '20px' }}>
                 <li>Filter by label presence: label</li>
-                <li>Filter by label and value: label=value</li>
+                <li>Filter by label and value: label:value</li>
                 <li>
-                  Filter by more than one label and one or more values: label=value label2=value2,value2-2 (separate
+                  Filter by more than one label and one or more values: label:value label2:value2,value2-2 (separate
                   with ' ')
                 </li>
               </ul>
             </>
           }
-          position={PopoverPosition.bottom}
         >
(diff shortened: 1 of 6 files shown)


src/pages/AppList/FiltersAndSorts.ts
```diff
@@ -122,15 +122,17 @@ const filterByLabel = (items: AppListItem[], filter: string[]): AppListItem[] =>
...
       // Check Values
       values[1].split(',').map(
         val =>
           (result = result.concat(
             items.filter(item => {
               if (values[0] in item.labels) {
-                return item.labels[values[0]].split(',').some(appVal => appVal.startsWith(val));
+                return item.labels[values[0]]
+                  .split(',')
+                  .some(appVal => appVal.startsWith(val) && !result.includes(item));
```
lucasponce: Perhaps it could be a good idea to trim the val, it's easy to introduce a label:     value1,  value2.
At least it would protect some cases.

lucasponce: Note that trim should apply to more places where value is read, just comment on this line.

aljesusg:
Tasks:

- Added trim to values
- Moved method to one file "same method"
- Changed = by :
- Removed duplication

Ready for merge

src/components/Filters/LabelFilter.tsx
```diff
@@ -32,21 +32,21 @@ export class LabelFilters extends React.Component<LabelFiltersProps> {
...
             <>
               To set a label filter you must enter values like.
               <br />
               <ul style={{ listStyleType: 'circle', marginLeft: '20px' }}>
                 <li>Filter by label presence: label</li>
-                <li>Filter by label and value: label=value</li>
+                <li>Filter by label and value: label:value</li>
                 <li>
-                  Filter by more than one label and one or more values: label=value label2=value2,value2-2 (separate
+                  Filter by more than one label and one or more values: label:value label2:value2,value2-2 (separate
```
lucasponce, line 44: I suggest to add a couple of <br />
```
                <li>
                  Filter by more than one label and one or more values:
                  <br />
                  label:value label2:value2,value2-2
                  <br />
                  (separate with ' ')
                </li>
```
image

In that way the legend is not broken and I think it can be read better.

It's minor

PR #1715, opened 2020-03-25
Fix graph error when enabling response time edges in some cases.

mtho11:
** Describe the change **
There are some times when changing the graph edge label to **response time** causes a crash. This PR fixes that issue.

response-time-bug

** Issue reference **
fixes https://github.com/kiali/kiali/issues/2549

### DIFF
--- a/src/components/CytoscapeGraph/graphs/GraphStyles.ts
+++ b/src/components/CytoscapeGraph/graphs/GraphStyles.ts
@@ -331,8 +331,15 @@ export class GraphStyles {
           break;
         }
         case EdgeLabelMode.RESPONSE_TIME_95TH_PERCENTILE: {
-          const responseTime = edgeData.responseTime > 0 ? edgeData.responseTime : 0;
-          if (responseTime > 0) {
+          // todo: remove this logging once we figure out the strangeness going on with responseTime
+          let logResponseTime = edgeData.responseTime;
+          if (!isNaN(logResponseTime) && !Number.isInteger(logResponseTime)) {
+            console.log(`Unexpected string responseTime=|${logResponseTime}|`);
+          }
+          // hack to fix responseTime is sometimes a string during runtime even though its type is number
+          const responseTimeNumber = parseInt(String(edgeData.responseTime));
+          const responseTime = responseTimeNumber > 0 ? responseTimeNumber : 0;
+          if (responseTime && responseTime > 0) {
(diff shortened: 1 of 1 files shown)


src/components/CytoscapeGraph/graphs/GraphStyles.ts
```diff
@@ -331,8 +331,10 @@ export class GraphStyles {
           break;
         }
         case EdgeLabelMode.RESPONSE_TIME_95TH_PERCENTILE: {
-          const responseTime = edgeData.responseTime > 0 ? edgeData.responseTime : 0;
-          if (responseTime > 0) {
+          // hack to fix responseTime is sometimes a string during runtime even though its type is number
+          const responseTimeNumber = parseInt(String(edgeData.responseTime));
+          const responseTime = responseTimeNumber > 0 ? responseTimeNumber : 0;
+          if (responseTime && responseTime > 0) {
```
jshaughn: This works as a fix and we should probably merge in a fix just for safety in the immediate-term, but I'm concerned that we are somehow getting compromised or unexpected response time values and I'd like to know the root cause.  Can we add something that dumps out the string values when we hit this? Maybe like:

```
let responseTime = edgeData.responseTime;
if ( !Number.isInteger(responseTime) ) {
  console.log(`Unexpected string responseTime=|${responseTime}|`);
  responseTime = parseInt(responseTime);
}
```

I mean it's weird!  We could even dump out the entire graph json because I wonder what is getting returned.

mtho11: @jshaughn sure, good idea

mtho11: That ended up spewwing more in the logs than intended:

Kiali_Console

So I changed the boolean expression to ignore the NaN.

src/components/CytoscapeGraph/graphs/GraphStyles.ts
```diff
@@ -331,8 +331,15 @@ export class GraphStyles {
           break;
         }
         case EdgeLabelMode.RESPONSE_TIME_95TH_PERCENTILE: {
-          const responseTime = edgeData.responseTime > 0 ? edgeData.responseTime : 0;
-          if (responseTime > 0) {
+          // todo: remove this logging once we figure out the strangeness going on with responseTime
+          let logResponseTime = edgeData.responseTime;
+          if (!isNaN(logResponseTime) || !Number.isInteger(logResponseTime)) {
```
jshaughn: I thik this should be &&, no?

pbajjuri20:
LGFM.

PR #1718, opened 2020-03-31
Allow sparkline metrics on edge to service entries

jotak:
Fixes https://github.com/kiali/kiali/issues/2497

Capture d’écran de 2020-03-31 12-43-37

### DIFF
--- a/src/pages/Graph/SummaryPanelEdge.tsx
+++ b/src/pages/Graph/SummaryPanelEdge.tsx
@@ -237,6 +237,12 @@ export default class SummaryPanelEdge extends React.Component<SummaryPanelPropTy
     data: DecoratedGraphNodeData,
     isServiceEntry: boolean
   ) => {
+    if (isServiceEntry) {
+      // For service entries, metrics are grouped by destination_service_name and we need to match it per "data.destServices"
+      return getDatapoints(m, (metric: Metric) => {
+        return data.destServices && data.destServices.some(svc => svc.name === metric['destination_service_name']);
+      });
+    }
     let sourceLabel: string;
     let sourceValue: string | undefined;
     switch (sourceMetricType) {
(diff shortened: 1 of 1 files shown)


jotak:
@jshaughn I realized that the way metrics are queried is, as a first step, query metrics from an edge end grouped by appropriate label, and then filter / aggregate the result matching with the other edge end, that same algorithm actually works fine with service entries. Just some little tweaks are necessary to switch fetching initially from edge source instead of edge destination.

jotak:
Thanks!

jotak:
@hhovsepy or @mattmahoneyrh do you have some test scenario with service entries to test it with?
I have my Kafka scenario, but if you have others, that would increase coverage.

israel-hdez:
I've tried and I see the sparklines! This is indeed nice.
But I haven't done code review. If tomorrow you haven't merged, I'll do the deep review. You already have Jay's approval and he is more familiar than I with all metrics queries so, no need to wait for me if you want to merge.

jshaughn:
>  So I'd say it's out of the scope of this PR.

@jotak +1, this is for edges and that in itself is great to have.

src/pages/Graph/SummaryPanelEdge.tsx
```diff
@@ -252,12 +254,14 @@ export default class SummaryPanelEdge extends React.Component<SummaryPanelPropTy
...
-      return metric[sourceLabel] === sourceValue;
-    };
+    let comparator = (metric: Metric) => metric[sourceLabel] === sourceValue;
+    if (isServiceEntry) {
+      comparator = (metric: Metric) => {
+        return data.destServices && data.destServices.some(svc => svc.name === metric['destination_service_name']);
+      };
+    } else if (this.isSpecialServiceDest(destMetricType)) {
+      comparator = (metric: Metric) => metric[sourceLabel] === sourceValue && metric.destination_workload === UNKNOWN;
+    }
```
israel-hdez, line 264: So, if `isServiceEntry` is truthy, the evaluation of the `switch` is vain.

For code clarity, I'd prefer something like

```tsx
if (isServiceEntry) {
  /* all the old code here*/
} else {
  let comparator = (metric: Metric) => {
    return data.destServices && data.destServices.some(svc => svc.name === metric['destination_service_name']);
￼      return metric[sourceLabel] === sourceValue;
  };

  return getDatapoints(m, comparator);
}
```

It's just a small suggestion. It would be clearer for me, but for others it may be cristal clear as is. So, you choose to keep as is or to change.

What I really would like is that you add a comment on top of `getNodeDataPoints` definition. I see that when `isServiceEntry` arg is truthy, this method ignores `sourceMetricType` and `destMetricType` args (even if non-optional) and also assumes that `data` arg is the destination node and `m` are outbound metrics of the source node. On the other hand, if `isServiceEntry` arg is falsy, all args are used, but it's assumed that `data` arg is the source node and `m` are inbound metrics of the destination node. By not respecting these assumptions, this method will return garbage.

It took me a while to understand these assumptions, and the fact that all usages are passing `sourceMetricType` and `destMetricType` args is confusing for me.

jotak: I think I understand what you mean, I was just a bit confused because I think your code snippet should start with `if (!isServiceEntry) {` instead of `if (isServiceEntry) {` , right?

israel-hdez: Yes, that's what I meant.
Thanks for changing :)

PR #1720, opened 2020-04-01
Graph: Merge edge options into display dropdown to limit toolbar wrap

jshaughn:
The graph toolbar wraps even on relatively wide browser windows.  Merging it into the Display dropdown conserves quite a bit of space and helps limit wrapping.  It also consolidates display options since edge labeling is an option.

@lwrigh, @israel-hdez  The good news is that I was able to get the mixture of radio buttons and checkboxes into a single dropdown.  The bad news is that I had to toss out the use of the PF4 Select component.  The dropdown is now signifigantly long and the Select component did not handle scrolling well, and so for shorter windows the bottom options were clipped.  Note that this was true in the past as well, but the lesser number of options made it unlikely to happen.  As an alternative I reworked things in the model of what we do in the namespaces dropdown.   The fallout is the the look of the dropdown is no longer that of Select.  But we can likely tweak it if needed.

Before:
image

After:
image

With Scroll:
image

### DIFF
--- a/src/pages/Graph/GraphToolbar/GraphSettings.tsx
+++ b/src/pages/Graph/GraphToolbar/GraphSettings.tsx
@@ -15,6 +15,7 @@ import {
   PropertyType
 } from 'components/BoundingClientAwareComponent/BoundingClientAwareComponent';
 import { style } from 'typestyle';
+import { PfColors } from 'components/Pf/PfColors';
 
 type ReduxProps = Omit<GraphToolbarState, 'findValue' | 'hideValue' | 'showLegend' | 'showFindHelp'> & {
   setEdgeLabelMode: (edgeLabelMode: EdgeLabelMode) => void;
@@ -47,22 +48,24 @@ const containerStyle = style({
   overflow: 'auto'
 });
 
+// this emulates Select component .pf-c-select__menu
 const menuStyle = style({
   fontSize: '14px'
 });
 
-const firstTitleStyle = style({
-  margin: '0.5em 0.25em 0.5em 0.25em',
-  fontWeight: 'bold'
-});
-
+// this emulates Select component .pf-c-select__menu-group-title but with less bottom padding to conserve space
 const titleStyle = style({
-  margin: '1em 0.25em 0.5em 0.25em',
-  fontWeight: 'bold'
+  padding: '8px 16px 2px 16px',
+  fontWeight: 700,
+  color: PfColors.Black600
 });
 
-const checkboxStyle = style({
-  margin: '0.25em 0.5em 0.25em 0.5em'
+// this emulates Select component .pf-c-select__menu-item but with less vertical padding to conserve space
+const itemStyle = style({
+  alignItems: 'center',
+  whiteSpace: 'nowrap',
+  margin: 0,
+  padding: '6px 16px'
 });
 
 class GraphSettings extends React.PureComponent<GraphSettingsProps, GraphSettingsState> {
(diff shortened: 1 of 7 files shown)


pbajjuri20:
scroll


Tested and passed on Kiali UI 1.16.0 (531293ccf9ba902d24b440cb95eb285772e0c7af)

src/pages/Graph/GraphToolbar/GraphSettings.tsx
```diff
@@ -90,7 +138,34 @@ class GraphSettings extends React.PureComponent<GraphSettingsProps, GraphSetting
       toggleUnusedNodes
     } = this.props;
 
-    const visibilityLayers: VisibilityLayersType[] = [
+    const edgeLabelOptions: DisplayOptionType[] = [
+      {
+        id: EdgeLabelMode.NONE,
+        labelText: _.capitalize(_.startCase(EdgeLabelMode.NONE)),
```
josejulio, line 144: Nitpick, is the `_capitalize` over all here really needed?
`_.startCase('requestsPerSecond');` yields: `Requests Per Second`

src/pages/Graph/GraphToolbar/GraphSettings.tsx
```diff
@@ -151,47 +226,84 @@ class GraphSettings extends React.PureComponent<GraphSettingsProps, GraphSetting
...
-        placeholderText="Display"
-        onToggle={this.onToggle}
-        onSelect={() => undefined}
-        isExpanded={isOpen}
-        variant={SelectVariant.checkbox}
-        isGrouped={true}
-        selections={selection}
+      <BoundingClientAwareComponent
+        className={containerStyle}
+        maxHeight={{ type: PropertyType.VIEWPORT_HEIGHT_MINUS_TOP, margin: marginBottom }}
```
josejulio, line 232: I'm glad this thing has been useful :)

lwrigh:
It's looking good! I'm glad to see that you were able to figure out the radio button situation. Based off of the feedback from our review I would just add more padding to the dropdown menu so the title labels are not as close to the edge and I would try to left align the titles and toggles with one another. Ideally the same styles and spacing that are featured in the previous version can be added to the updated dropdown as well.

pbajjuri20:
Tested after changed the style to capitalize words " No edge labels" to "No Edge Labels"


label

src/pages/Graph/GraphToolbar/GraphSettings.tsx
jshaughn: smarty.  changed.

src/pages/Graph/GraphToolbar/GraphSettings.tsx
jshaughn: super useful! I still don't know how it works.

src/pages/Graph/GraphToolbar/GraphSettings.tsx
```diff
@@ -151,47 +229,81 @@ class GraphSettings extends React.PureComponent<GraphSettingsProps, GraphSetting
...
-              onClick={item.onChange}
-            />
+        <div className={menuStyle}>
+          <div className={titleStyle}>Show Edge Labels</div>
+          {edgeLabelOptions.map((item: DisplayOptionType) => (
+            <label className={itemStyle}>
+              <Radio
+                id={item.id}
+                name="edgeLabels"
+                isChecked={item.value}
```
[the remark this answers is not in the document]
jshaughn: @israel-hdez OK, I pushed the changes so that I can be cool too.  Please give it a quick look.

jshaughn:
@lwrigh I think we have the look where we want it now.  I did reduce the vertical spacing a big just because it was so tall.  I think if we need to tweak it further we can do it in another PR.  Here is what it looks like now:

image

src/pages/Graph/GraphToolbar/GraphSettings.tsx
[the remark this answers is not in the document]
israel-hdez: Nice :)

lwrigh:
> @lwrigh I think we have the look where we want it now. I did reduce the vertical spacing a big just because it was so tall. I think if we need to tweak it further we can do it in another PR. Here is what it looks like now:
> 
> image

It looks good to me!

PR #1724, opened 2020-04-07
Namespaces overview should differentiate idle from failure apps

aljesusg:
Fix https://github.com/kiali/kiali/issues/2049

The idea is show resources undesired replica
Screenshot from 2020-04-07 14-40-39

### DIFF
--- a/src/components/Health/__tests__/HealthIndicator.test.tsx
+++ b/src/components/Health/__tests__/HealthIndicator.test.tsx
@@ -4,7 +4,7 @@ import { shallowToJson } from 'enzyme-to-json';
 
 import { HealthIndicator, DisplayMode } from '../HealthIndicator';
 import { AppHealth } from '../../../types/Health';
-import { PFAlertColor } from 'components/Pf/PfColors';
+import { PFAlertColor, PfColors } from 'components/Pf/PfColors';
 
 describe('HealthIndicator', () => {
   it('renders when empty', () => {
@@ -96,12 +96,12 @@ describe('HealthIndicator', () => {
     // SMALL
     let wrapper = shallow(<HealthIndicator id="svc" health={health} mode={DisplayMode.SMALL} />);
     let html = wrapper.html();
-    expect(html).toContain(PFAlertColor.Danger);
+    expect(html).toContain(PfColors.Gray);
 
     // LARGE
     wrapper = shallow(<HealthIndicator id="svc" health={health} mode={DisplayMode.LARGE} />);
     html = wrapper.html();
-    expect(html).toContain(PFAlertColor.Danger);
+    expect(html).toContain(PfColors.Gray);
   });
 
   it('renders error rate failure', () => {
--- a/src/components/Health/__tests__/HealthIndicator.test.tsx
+++ b/src/components/Health/__tests__/HealthIndicator.test.tsx
@@ -4,7 +4,7 @@ import { shallowToJson } from 'enzyme-to-json';
 
 import { HealthIndicator, DisplayMode } from '../HealthIndicator';
 import { AppHealth } from '../../../types/Health';
-import { PFAlertColor, PfColors } from 'components/Pf/PfColors';
+import { PFAlertColor } from 'components/Pf/PfColors';
 
 describe('HealthIndicator', () => {
   it('renders when empty', () => {
@@ -39,7 +39,7 @@ describe('HealthIndicator', () => {
     expect(html).toContain(PFAlertColor.Success);
   });
 
-  it('renders workloads undesired', () => {
+  it('renders workloads defraded', () => {
     const health = new AppHealth(
       [
         { name: 'A', availableReplicas: 1, currentReplicas: 1, desiredReplicas: 10 },
@@ -52,12 +52,12 @@ describe('HealthIndicator', () => {
     // SMALL
     let wrapper = shallow(<HealthIndicator id="svc" health={health} mode={DisplayMode.SMALL} />);
     let html = wrapper.html();
-    expect(html).toContain(PFAlertColor.InfoBackground);
+    expect(html).toContain(PFAlertColor.Warning);
 
     // LARGE
     wrapper = shallow(<HealthIndicator id="svc" health={health} mode={DisplayMode.LARGE} />);
     html = wrapper.html();
-    expect(html).toContain(PFAlertColor.InfoBackground);
+    expect(html).toContain(PFAlertColor.Warning);
     expect(html).toContain('1 / 10');
   });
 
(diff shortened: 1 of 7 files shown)


aljesusg:
This is another case @lucasponce, The PR is about show undesired replicas. If there is a 0 pods by default was an error. I can add to be "this state" if pods are 0 if we think that this should be the result

Example:

I created a hello world in a hello namespace

```yaml 

apiVersion: v1
kind: Service
metadata:
  name: hello-kubernetes
spec:
  type: LoadBalancer
  ports:
  - port: 80
    targetPort: 8080
  selector:
    app: hello-kubernetes
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: hello-kubernetes
spec:
  replicas: 3
  selector:
    matchLabels:
      app: hello-kubernetes
  template:
    metadata:
      labels:
        app: hello-kubernetes
    spec:
      containers:
      - name: hello-kubernetes
        image: paulbouwer/hello-kubernetes:1.7
        ports:
        - containerPort: 8080
```

```bash
kubectl apply -f hello-world.yaml -n hello
```
Get the status

```bash
kubectl get replicaset -n hello
NAME                          DESIRED   CURRENT   READY   AGE
hello-kubernetes-6c444b8bd5   3         3         3       80s

```

Scale to 2
```bash
kubectl scale -n hello replicaset hello-kubernetes-6c444b8bd5 --replicas=2
```

lucasponce:
The PR should cover situations as described in the issue:

"I would like to see some differentiation between those apps that are really in some kind of failure state from those which are just idled (or scaled down to zero)."

So, a workload scaled down to zero by user shouldn't be marked as error.

I'm not sure if perhaps it needs more support from backend logic, but this PR is imcompleted.

aljesusg:
Don't need changes in backend but maybe I missed that sentence ;)

aljesusg:
Anyway @lucasponce I am going to check this case...it should be NA not failure

```ts
  if (desiredReplicas === 0 && currentReplicas === 0) {
    return NA;
  }
```

aljesusg:
Ok the point is that it was in NA status but we showed it like Failure in the UI.
So now if desired and current are 0 this will be unknown. (See default ns in image where we have 6 applications but only 5 health)

I think that if desired and current are 0 should be a UNDESIRED error , because there is no error from kubernetes or deployment, the error is in the configuration, maybe a new icon ? This will be solve if we let the user config that in backend
Screenshot from 2020-04-16 10-28-00

aljesusg:
Screenshot from 2020-04-21 16-32-16

src/components/Health/__tests__/HealthIndicator.test.tsx
```diff
@@ -39,7 +39,7 @@ describe('HealthIndicator', () => {
     expect(html).toContain(PFAlertColor.Success);
   });
 
-  it('renders workloads degraded', () => {
+  it('renders workloads defraded', () => {
```
lucasponce, line 42: typo "defraded" -> "degraded"

src/pages/Overview/OverviewCardContentExpanded.tsx
```diff
@@ -69,6 +73,15 @@ class OverviewCardContentExpanded extends React.Component<Props> {
         <OverviewCardBars status={this.props.status} />
         <div style={{ marginTop: -20, position: 'relative' }}>
           <Text component={TextVariants.h2} style={{ marginTop: 0 }}>
+            {status.inIdle.length > 0 && (
+              <OverviewStatus
+                id={name + '-undesired'}
```
lucasponce, line 78: Perhaps better `id={name + '-undesired'}` -> `id={name + '-iddle'}`
Undesired has not meaning on this context.

src/types/Health.ts
```diff
@@ -87,22 +101,67 @@ interface ThresholdStatus {
 const RATIO_NA = -1;
 
 export const ratioCheck = (availableReplicas: number, currentReplicas: number, desiredReplicas: number): Status => {
-  // No Pods returns No Health Info
-  if (desiredReplicas === 0 && currentReplicas === 0) {
-    return NA;
+  /*
+    IDLE STATE
+ */
+  // User performed the action
```
lucasponce, line 107: I think is better to put more context
```
// User has scaled down a workload, then desired replicas will be 0 and it's not an error condition
```

src/types/Health.ts
```diff
@@ -87,22 +101,67 @@ interface ThresholdStatus {
...
+  if (desiredReplicas === 0) {
+    return IDLE;
   }
-  // No available Pods when there are desired and current means a Failure
+
+  /*
+   DEGRADED STATE
```
lucasponce, line 113: In similar way, a comment like this will help to refresh/understand the logic (at least for me):
```
// When a workload has available pods but less than desired defined by user it should be marked as degraded
```

src/types/Health.ts
```diff
@@ -87,22 +101,67 @@ interface ThresholdStatus {
...
+    availableReplicas > 0 &&
+    (currentReplicas < desiredReplicas || availableReplicas < desiredReplicas)
+  ) {
+    return DEGRADED;
+  }
+
+  /*
+     FAILURE STATE
+  */
+  if (desiredReplicas > 0 && (currentReplicas === 0 || availableReplicas === 0)) {
```
lucasponce, line 127: Perhaps this an next case can be put together in a

```
// When availableReplicas is 0 but user has marked a desired > 0, that's an error condition
if (desiredReplicas > 0 && availableReplicas === 0)
```

src/types/Health.ts
```diff
@@ -87,22 +101,67 @@ interface ThresholdStatus {
...
+    return FAILURE;
+  }
+
+  // Some pods are up but not ready
   if (desiredReplicas > 0 && currentReplicas > 0 && availableReplicas === 0) {
     return FAILURE;
   }
+
+  if (desiredReplicas > 0 && currentReplicas < desiredReplicas) {
```
lucasponce, line 136: Perhaps this is not correct as it's mainly covered by previous one:

```
  if (
    desiredReplicas > 0 &&
    currentReplicas > 0 &&
    availableReplicas > 0 &&
    (currentReplicas < desiredReplicas || availableReplicas < desiredReplicas)
  ) {
    return DEGRADED;
  }
```

A `currentReplicas < desiredReplicas` may mean a degraded or error scenario, we would need the availableReplicas to check.

So I think this case could be removed from the logic.

src/types/Health.ts
```diff
@@ -87,22 +101,67 @@ interface ThresholdStatus {
...
+    return FAILURE;
+  }
+
+  if (desiredReplicas === currentReplicas && availableReplicas > currentReplicas) {
```
lucasponce, line 140: I think this is not going to happen: `currentReplicas > availableReplicas` always. As "availableReplicas" are those pods that are up and are responding, meanwhile "currentReplicas are all pods up".

It sounds a corner case that won't never happen (unless I'm missing something).

src/types/Health.ts
```diff
@@ -87,22 +101,67 @@ interface ThresholdStatus {
...
+
   // Pending Pods means problems
   if (desiredReplicas === availableReplicas && availableReplicas !== currentReplicas) {
     return FAILURE;
   }
-  // Health condition
-  if (desiredReplicas === currentReplicas && currentReplicas === availableReplicas) {
+
+  // No available Pods when there are desired and current means a Failure
+  if (desiredReplicas > 0 && currentReplicas > 0 && availableReplicas === 0) {
```
lucasponce, line 150: Note that you are covering this case in the previous

```
if (desiredReplicas > 0 && (currentReplicas === 0 || availableReplicas === 0)) {
```
So, as suggested in a previous comment, perhaps it can be consolidated.

aljesusg:
Fixed @lucasponce thanks for review it

PR #1725, opened 2020-04-07
Graph toolbar use custom secondary masthead

jshaughn:
A custom secondary masthead allows us to move some things around for a more balanced use of space and less overall width required.

Also includes:
- change to info icon for graph find help
- a couple of small tweaks to the find help popover to remove scrollbar in preface and arrow on tabs

Fixes https://github.com/kiali/kiali/issues/2620

### DIFF
--- a/src/components/Nav/SecondaryMasthead.tsx
+++ b/src/components/Nav/SecondaryMasthead.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import { style } from 'typestyle';
 
-const paddingStyle = style({
+const marginStyle = style({
   margin: '10px 20px 0 0'
 });
 const secondaryMastheadStyle = style({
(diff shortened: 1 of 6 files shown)


jshaughn:
@lwrigh A couple of options as discussed:

Left aligned:
image

Left aligned with vertical separator:
image

Vertical separator and right aligned Tour:
image

Let me know wheich you prefer, or if you'd like to see something else. Thanks!

lwrigh:
@jshaughn I think I prefer the third option with the separator and the tour link off to the righthand side.

jshaughn:
> I think I prefer the third option with the separator and the tour link off to the righthand side.

@lwrigh OK, that was my preference too.  I'll open up the PR for review...

jshaughn:
This is functionally approved but waiting for @aljesusg to comment on the approach used for a custom secondary masthead.  The approach was to create the custom GraphSecondaryMasthead, using the common SecondaryMasthead component.  Note that the custom secondary masthead is no longer referenced in Routes.tsx but moved to the graph page impl.  The downside is that Routes.tsx no longer indicates all routes that have a secondary masthead but the benefit is a much simpler impl of the custom component.  It is still quite easy to find secondary mastheads via search.

lucasponce:
Please, before to merge this one, I'd like to resolve the comment I've added in the issue.

jshaughn:
OK, from UX the decision is to have title/label on list views but not on non-list pages, preserving more page space for the content.  Also, Global Namespace dropdown will always be top left of the secondary masthead, local settings may follow to the right, with a vertical separator.  Page-level help will be on the secondary masthead, right adjusted.  Proceeding with merge.

PR #1726, opened 2020-04-08
Upgrade PF4 stuff to release 2020.04

aljesusg:
Related with https://github.com/kiali/kiali-ui/pull/1716

Fix https://github.com/kiali/kiali/issues/2660

Following release notes:

REACT => https://www.patternfly.org/v4/documentation/react/overview/release-notes
CORE => https://www.patternfly.org/v4/documentation/core/overview/release-notes

we are going to upgrade the PF4 stuff to versions released 2020-03-31 => version 2020.04 :

@patternfly/patternfly@v2.71.3
@patternfly/react-catalog-view-extension@1.4.48
@patternfly/react-charts@5.3.18
@patternfly/react-core@3.153.3
@patternfly/react-icons@3.15.15
@patternfly/react-inline-edit-extension@2.17.48
@patternfly/react-styles@3.7.12
@patternfly/react-table@2.28.29
@patternfly/react-tokens@2.8.12
@patternfly/react-topology@2.14.48
@patternfly/react-virtualized-extension@1.4.49

This fix

https://github.com/kiali/kiali-ui/pull/1728
https://github.com/kiali/kiali-ui/pull/1727
https://github.com/kiali/kiali-ui/pull/1716


aljesusg:
- I updated the tippy library, PF4 use version **5.1.2** when we used 3.4.1. Affect tippyProps props in **Popover component in graph tour**
- Changed route of light-border css of tippy
- Some Changes related with PF$

Opened bugs: 

- **Not important** https://github.com/patternfly/patternfly-react/issues/4037  


Things to review more than others

- [ ] Cytoscape , some changes in tippy library
- [ ] Graph Tour

Checking views...

src/components/CytoscapeGraph/CytoscapeContextMenu.tsx
```diff
@@ -153,10 +153,9 @@ export class CytoscapeContextMenuWrapper extends React.PureComponent<Props> {
         sticky: true,
         interactive: true,
         theme: 'light-border',
-        size: 'large',
         distance: this.tippyDistance(target)
       }
-    ).instances[0];
+    )[0];
```
aljesusg, line 159: Prop Size was removed from here. I am checking where we must to set this so I am learnign about tippy.

On the other hand tippy method now return ``` Instance<TProps> ``` so instances is not required.

src/components/Tour/TourStop.tsx
```diff
@@ -148,7 +148,7 @@ class TourStop extends React.PureComponent<TourStopProps> {
 
   render() {
     const offset: string = this.props.info.offset ? this.props.info.offset : '0, 0';
-    const tippyProps: Props = { offset: offset };
+    const tippyProps: Partial<Props> = { offset: offset };
     const isVisible = this.isVisible();
```
aljesusg, line 152: It's required now indicate that we are defining a partial of the props

src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
```diff
@@ -376,7 +375,7 @@ class IstioConfigDetailsPage extends React.Component<RouteComponentProps<IstioCo
 
     return (
       <div className="container-fluid container-cards-pf">
-        <Grid gutter={GutterSize.md}>
+        <Grid gutter={'md'}>
           <GridItem span={editorSpan}>
```
aljesusg, line 379: This error is a regresion from PF 

Opened Issue https://github.com/patternfly/patternfly-react/issues/4037

We can go ahead

src/pages/Login/LoginPage.tsx
```diff
@@ -259,7 +260,7 @@ export class LoginPage extends React.Component<LoginProps, LoginState> {
 
     return (
       <LoginNext
-        footerListVariants="inline"
+        footerListVariants={ListVariant.inline}
         brandImgSrc={kialiLogo}
```
aljesusg, line 264: On the other hand they create the variable for this ^^ I can't understand how they do to  create a new one and break others

jotak:
As a side task I guess we should also upgrade pf4 in k-charted

aljesusg:
Ok I found some problems with tippy so I need to research about

aljesusg:
@jshaughn Is there any neeed to use tippy version 3? PF4 use version 5 and I am seeing that actually they support 5.x and 6.x 

https://atomiks.github.io/tippyjs

jshaughn:
@aljesusg I don't think we have to use tippy3, as long as we can maintain a good user experience.  We need to move forward with supported versions of our dependencies, there is no way around it.  I know we have had to drill all the way into tippy props in a couple of places, but that's typically done as a "last resort" option.   I'd be happy to look at any problem area you have encountered.  Or we can get @israel-hdez or anyone else as well.

aljesusg:
@jshaughn The problems that I found are

-  In Graph tour, it seems that tippy optiosn in PopOver are used by props definedin tippy 5
- In Contextmenu
  - Tippy now return a Instances<Props> Type so .instances is not required
  - Prop size was removed (I saw taht now that code was moved to file types and you can find there some reference about it)
  - this.setCurrentContextMenu(tippyInstance) L170 now required another type.
(I didn't test more because I want to learn more, I know that you have mroe experience @jshaughn @israel-hdez, I'm going to spend sometime learning in my public holidays)
- Light-border css is now in tippy.js/themes/light-border.css instead tippy.js/dist/themes/light-border.css

jshaughn:
@aljesusg thanks for the detailed description.  I'm totally fine with an update of Tippy as long as we can resolve the issues with code changes and the user is not adversely affected.  As you saw, tippy is used in context menu. And Popover/tippy is used in GraphTour and Graph Find help.

hhovsepy:
I did run the full UI tests, there were several failures because of Xpath changes, but after fixing them all is passing.

jshaughn:
@aljesusg Just an FYI, I checked out the branch and can confirm a problem with graph tour, the popovers are not well-formed.  And also with graph node context menu which results in a crash.  The graph find help looks ok.   The solutions do not, in my opinion, need to produce something that is exactly the same as before, but just has to look fine and act reasonably.

aljesusg:
@jshaughn If you inspect code of the PopOver, they are generated inside a tippy-content, this element has a padding style, if we removed this css the popovers are well-formed, so for this case we can overwrite this style

aljesusg:
I continue working, I fixed some errors in console about keyProp.

Checking how upgrade this to version 5 from version 3

Docs: https://github.com/cytoscape/cytoscape.js-popper#usage-with-tippyjs

aljesusg:
Ok, to unblock this, I decided to do this.

I need more time for tippy, it seems that there are so changes from version 3.x to version 5.x so I'll keep the tippy version 3.x. (we should try to upgrade it by support not to be aligned with PF4 )

@jshaughn you can see that the graph is like before, no changes here, I did a change in the tour component to define the props like any to avoid the error version.

@hhovsepy please check if this is fine now.

Thanks  !!!

aljesusg:
> Only one place is failing.
> 
> 1. Create New Istio Config page,
> 2. Try to type in "Hosts" field.
>    After each keyup the focus/cursor is lost from that input field.
>    Screenshot from 2020-04-14 15-47-20
> 
> Happens in both FF and Chrome browsers.

Let me check if this in master or elated with PF4 because it's weird

hhovsepy:
It does not happen on master.

aljesusg:
> It does not happen on master.

verified, I am checking if is a bug to report to PF4 team

aljesusg:
Ok I am researching about that, anyway I opened an issue in PF4 repo (https://github.com/patternfly/patternfly-react/issues/4072) because other textInputs are working correctly.

lucasponce:
I'm reductant to merge with patternfly/patternfly-react#4072 open.
Is there a workaround to not leave master with a known bug in the forms ?

lucasponce:
Note, that perhaps it can be workaround it putting a FormGroup in that component, I haven't tried.
I can test it tomorrow morning to unblock this.

hhovsepy:
Regression run, the only problem is still the patternfly/patternfly-react#4072

aljesusg:
> Regression run, the only problem is still the [patternfly/patternfly-react#4072](https://github.com/patternfly/patternfly-react/issues/4072)

Yep I created the sample in codebox (https://codesandbox.io/s/texinput-table-ggx1l?file=/index.js) the problem is the renderer of the input inside a table

lucasponce:
Just to avoid different criterias.
If an upgrade of a library breaks a feature, a PR cannot be approved.

So, it will need a workaround, but a bump of a version can't justify a breaking feature.

And using the right terminology, that's a regression bug.

(We are on the same page what is the root cause, but I don't want that we merge things like this by mistake just because someone reads that regression passed. It is not.)

aljesusg:
> Just to avoid different criterias.
> If an upgrade of a library breaks a feature, a PR cannot be approved.
> 
> So, it will need a workaround, but a bump of a version can't justify a breaking feature.
> 
> And using the right terminology, that's a regression bug.
> 
> (We are on the same page what is the root cause, but I don't want that we merge things like this by mistake just because someone reads that regression passed. It is not.)

@lucasponce that is why I opened the issue with PF4 to find a solution, we are not going to merge this until kiali is at 100% correctly, I added last week "do not merge" for this

aljesusg:
@hhovsepy @lucasponce please review this. There is a workaround while PF4 teams work in the issue. I added a comment about that in the code. We are going to use textInputBase instead TextInput, in terms of code the comment that pf4 team says:

> The key implementation takeaway for TextInputBase vs TextInput is that TextInputBase is a normal React Component (that is typeof TextInputBase === 'function'), while TextInput is a special React object (that is typeof TextInput === 'object') with a .render prop.


More info in :https://github.com/patternfly/patternfly-react/issues/4072#issuecomment-616796398

lucasponce:
Thanks @aljesusg for the update.
I'll fetch it this morning and review it..

PR #1733, opened 2020-04-14
Add Istio component status into masthead

xeviknal:
Adding the Istio Component status tooltip:
When issues:
Screenshot of Kiali Console (39)

Screenshot of Kiali Console (42)

Whout issues:
Screenshot of Kiali Console (41)

### Testing notes
1. First, check out the list of components and its severity from https://github.com/kiali/kiali/pull/2598
2. This needs to be tested with 1.4.x and 1.5.x control planes.
3. To achieve the `not found` status, it is only needed to remove a deployment of a deployment.
4. To achieve the `unhealthy` status, it requires to be fast. Deleting a pod should be enough. Use the help of the `Interval duration` dropdown: pause it. Click the refresh button right after kill a pod.

Fixes https://github.com/kiali/kiali/issues/1377
Needs https://github.com/kiali/kiali/pull/2598 backend PR


lucasponce:
I'm going to start to review this one.
One preliminar comment that we had in the past: we can't use the Istio icon as far as I remember.
We used in the past when marked the sidecars pod and there was some mix of legal design issues.

I suggest to put some neutral one (i.e. and I from Istio in a circle) to represent that to avoid the same discussion.

xeviknal:
Good catch! I do propose to use the second option:
Screenshot of Kiali Console (31)

Screenshot of Kiali Console (32)

Moving to previous commit :)

xeviknal:
Rabased with master.

lucasponce:
I wonder if I prefer to view the icon only when there is a problem.
When all it's ok like this:
image

This is not giving me any new info.

But if there is a problem, I really would like to see this in the first place.

xeviknal:
For me to understand, do you say that we should have a message alerting the user that X,Y and Z modules are having problems instead of listing all the modules along its health?
i.e. " istiod, istio-ingress deployment are unhealthy. Grafana deployment not found"

lucasponce:
> For me to understand, do you say that we should have a message alerting the user that X,Y and Z modules are having problems instead of listing all the modules along its health?
i.e. " istiod, istio-ingress deployment are unhealthy. Grafana deployment not found"

Yes. I mean two things:
- The Istio status icon, only should appear when something goes wrong, and it should be the most left icon (in warning/red) to invite to take a first action on it.
- You are right: personally I think it's better to show which components have problems. Listing all of them perhaps it's adding additional info that doesn't help. I'd prefer to see "ingress is not found" rather to see all of them and identify which one is on red.

I mean, in 99% cases Istio should be up and running, it's a different health/status than an app pod, but if something fails, it should be great to have that info ASAP.

xeviknal:
Yep, I am working on it. Following the two steps you mentioned in your previous comment. Only showing the icon when issues are found. And in the message, showing only the affected components.
This message will be more in a sentence way instead of a list of component status as in the description.
Put the DNM label to warn QA that this is still evolving.

xeviknal:
Comments approached, the new version looks like this:

## mTLS not enabled

Screenshot of Kiali Console (38)

## With mTLS enabled

## Switching on and off mTLS

Screen recording (24)

jshaughn:
This is cool, just looking for the first time.  From the various options presented my thoughts:
- I like the left-most positioning.
- I wish we could use the Istio icon as opposed to bullet/circle thing.
- I actually like seeing it all times, green makes me feel good.
- I like the list format in the first shot, with the left-adjusted icons and aligned components. But I would limit to problem components.  I have a hard time parsing the more free formed tooltip.

xeviknal:
Besides the Istio logo icon, does this work for you guys?

Screenshot of Kiali Console (39)

Regarding the Istio logo, I found this page [1] mentioning scenarios where they allow to use it.

> Here are a few assets in case you want to show off your support for Istio, *integration to Istio*, or want to link back to us. 

[1] https://istio.io/about/media-resources/

lucasponce:
> I like the idea of keeping the icon all the time. It helps the user troubleshooting their apps. 

Please don't. It will just keep the masterhead populated.
It's similar as mTLS, it doesn't appear all the time.

Please, don't revisit the Istio logo discussion, it should be out of this PR.

Thanks

xeviknal:
Comments approached:

1. Icon moved to left
2. Icon only shown when there are issues.
3. Only shown problematic components.

When issues:
Screenshot of Kiali Console (39)

Screenshot of Kiali Console (42)

Whout issues:
Screenshot of Kiali Console (41)

xeviknal:
Ready to be reviewed! Thanks to all involved.

lucasponce:
@xeviknal would you mind to rebase ? I see some conflict in the branch.

PR #1741, opened 2020-04-15
Traces page: incremental refresh

aljesusg:
Fix https://github.com/kiali/kiali/issues/2221

Incremental Refresh like spans for traces

### DIFF
--- a/src/pages/ServiceDetails/ServiceDetailsPage.tsx
+++ b/src/pages/ServiceDetails/ServiceDetailsPage.tsx
@@ -37,6 +37,7 @@ import { PfColors } from '../../components/Pf/PfColors';
 import TimeRangeComponent from 'components/Time/TimeRangeComponent';
 import { serverConfig } from '../../config';
 import GraphDataSource from '../../services/GraphDataSource';
+import { URLParam } from '../../app/History';
 
 type ServiceDetailsState = {
   serviceDetailsInfo: ServiceDetailsInfo;
@@ -47,7 +48,6 @@ type ServiceDetailsState = {
   threeScaleInfo: ThreeScaleInfo;
   threeScaleServiceRule?: ThreeScaleServiceRule;
   currentTab: string;
-  traces: JaegerTrace[];
   errorTraces?: JaegerErrors[];
   selectedTrace?: JaegerTrace;
   errorSelectedTrace?: JaegerErrors[];
@@ -107,6 +107,8 @@ const tabIndex: { [tab: string]: number } = {
 
 class ServiceDetails extends React.Component<ServiceDetailsProps, ServiceDetailsState> {
   private promises = new PromisesRegistry();
+  private traces: JaegerTrace[] = [];
+  private lastFetchTraceMicros: number | undefined = undefined;
   private lastFetchTracesError = false;
   private graphDataSource: GraphDataSource;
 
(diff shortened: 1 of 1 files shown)


jotak:
It doesn't seem to work for me...

src/pages/ServiceDetails/ServiceDetailsPage.tsx
```diff
@@ -321,12 +322,16 @@ class ServiceDetails extends React.Component<ServiceDetailsProps, ServiceDetails
           }
         });
     } else {
-      fetchTraces(this.props.match.params.namespace, this.props.match.params.service, getQueryJaeger())
+      const params = getQueryJaeger();
+      fetchTraces(this.props.match.params.namespace, this.props.match.params.service, params)
         .then(traces => {
+          const appendTraces = this.lastFetchTraceMicros !== undefined;
```
jotak, line 328: I don't see how it would work, as `lastFetchTraceMicros` is never set.. am I missing something?

aljesusg: not you are not missing nothing ...I forgot updated that value let me open another PR :( I checked that there were new points but i didn't see that the news are the same

jotak: @aljesusg I see perhaps another issue: the traces fetched are never reset, which I think is a problem when query options change

jotak: keeping old data was the point of this task

aljesusg: Yes...I only tested refresh, because I calculate the from time that was the point...I'll reopen the issue

lucasponce:
Perhaps I tested too fast ?
I saw the update and the partial refresh, but it seems I missed details.

Please, add details, I'm interested to avoid test something wrong.

Perhaps we need to describe better that is the expected result to help reviewer.

jotak:
@lucasponce I'm not sure what you expected, or what isn't clear in task description, but the change here does absolutely nothing. Look at the code: `lastFetchTraceMicros` is undefined, so `appendTraces` is false, so state traces is *always* equal to `traces.data` ... which is exactly what it was before.

If the task is unclear or you don't understand what to test, you can leave a comment. To me it since quite clear, but perhaps I'm biased because I knew what to do. This PR was merged just three days ago, so just a week-end from me coming back from PTO ... The task was opened since early February, what's 3 days more in comparison? Also, perhaps if QE did test it they would have spot the issue.

lucasponce:
@jotak I merged this by error.

I'm asking author of the PR (this or other) perhaps a more clear description to help reviewer.

It seems when I tested locally I was seeing an incorrect behaviour and I thought it was the expected result.

That's what I'm asking.

Nothing wrong, this case was me, but tomorrow other reviewer and QE engineer may test something incorrectly, so author needs to help to explain better (even if it sounds obvious, I personally think that helps).

aljesusg:
Opened https://github.com/kiali/kiali-ui/pull/1745 to fix this.

Don't worry be happy

lucasponce:
> Don't worry be happy

Please, help QE or other reviewer explaning what should be the expected behaviour to avoid merge something by error as it happened.

Thanks.

jotak:
> Opened #1745 to fix this.
> 
> Don't worry be happy

Thanks! I'll look at it

PR #1742, opened 2020-04-15
Graph Hide can crash on graph refresh

jshaughn:
UPDATED DESCRIPTION (4/21): Graph hide (when using compressOnHide) keeps the hidden elements as "removed" elements in cy. The "removed" elements can later be restored when the hide-expression is changed or deleted.  It's only safe to remove/restore to the same set of graph elements, otherwise there can be ID clashes or other consistency problems (read: crash).  This PR fixes issues where the graph-hide logic could try to remove or restore elements when the graph had changed or was in the process of being updated.  There are several changes due to additional issues found while testing earlier revisions of this PR.  Ultimately we needed to make changes in how we utilized GraphDataSource in conjunction with react and redux.  See the comments added to GraphDataSource for more detail, when reviewing.

Test notes: See the issue for original reproduction steps, but there are other scenarios below.  Given the final number of changes, a full regression test should be performed as well as robust testing of hide and find in combination with display setting changes, namespace changes including clearing of namespaces), node graphs and mini-graphs.

Fixes https://github.com/kiali/kiali/issues/2656


pbajjuri20:
Tested and passed on Kiali UI 1.17.0 (ac84c220e61dded9796eca621d42b13e72241e49)

jshaughn:
@israel-hdez This issue is a major pain but I think now I have plugged the holes.  Good find with that last use case you reported.  Please try to break it one more time...  I'll reset the review request.

israel-hdez:
Nice changes. Aahh... but it's not perfect. Now it's broken in a different way. This is what I can do in **master** (I'm panning the graph to the right so behavior can be seen in the recording):

masterOK

and using this PR:

fixNotOK

I think you better delete all that code.
(just joking)

Ugh! I don't know why I had to hide node labels. I'd have approved & merged if I wouldn't tried that option.

jshaughn:
@israel-hdez OK, I think this is again ready for review/test.  At this point it's possible there are some changes that may not be necessary but that are now here because of the many iterations.  I have tried to trim it down but there are still a variety of changes.  As for GraphDataSource, I am certainly interested in your thoughts on the changes in that area.

In the end, the graphHide feature is delicate because it is processed as part of toolbar updates but extremely sensitive to changes in the cytoscape graph.  This PR maintains that approach and tries to make it all work.  But, if I were to start again I may have tried to move the actual find/hide logic into the core graph processing and limit the toolbar to only manage the expressions, storing that info in redux and then letting the graph code use it.  Most if not all of the problems result in the graph updating in one place and the hide logic updating and manipulating the graph in another.

src/pages/Graph/GraphToolbar/GraphFind.tsx
```diff
@@ -79,34 +82,41 @@ export class GraphFind extends React.PureComponent<GraphFindProps, GraphFindStat
...
+  // We only update on a change to the find/hide/compress values, or a graph change.  Although we use other props
+  // in processing (compressOnHide, layout, etc), a change to those settings will generate a graph change, so we
+  // wait for the graph change to do the update.
+  shouldComponentUpdate(nextProps: GraphFindProps) {
+    const cyChanged = this.props.cy !== nextProps.cy;
+    const findChanged = this.props.findValue !== nextProps.findValue;
+    const hideChanged = this.props.hideValue !== nextProps.hideValue;
+    const graphChanged = this.props.updateTime !== nextProps.updateTime;
+
+    return cyChanged || findChanged || hideChanged || graphChanged;
```
israel-hdez, line 94: So, this ignores `nextProps.showFindHelp`. This means that when you click the "Find/Hide help" button, this component doesn't re-renders and the help dialog is blocked until the next graph update happens and, suddenly, the help dialog makes its long awaited appearance. I think you need take into account that property, hoping that won't break all this good work.

jshaughn: @israel-hdez Another great catch, you could have a second job in QE :)  Fortunately, this is purely local to graphFind and I've added it back.  Should all be good now.  I'm going to release this to QE.  Thanks for all of your review.

pbajjuri20:
Tested and Passed on Kiali UI 1.17.0 (b76cc9b862fd6cdd84dd4ffd4fee797780fbacc5)

PR #1745, opened 2020-04-20
Incremental refresh in jaeger traces fix

aljesusg:
Fix https://github.com/kiali/kiali/issues/2221

### DIFF
--- a/src/pages/ServiceDetails/ServiceDetailsPage.tsx
+++ b/src/pages/ServiceDetails/ServiceDetailsPage.tsx
@@ -31,7 +31,7 @@ import { DurationDropdownContainer } from '../../components/DurationDropdown/Dur
 import RefreshButtonContainer from '../../components/Refresh/RefreshButton';
 import IstioWizardDropdown from '../../components/IstioWizards/IstioWizardDropdown';
 import { JaegerErrors, JaegerTrace, JaegerInfo } from '../../types/JaegerInfo';
-import { getQueryJaeger } from '../../components/JaegerIntegration/RouteHelper';
+import { getQueryJaeger, changeParams } from '../../components/JaegerIntegration/RouteHelper';
 import RefreshContainer from '../../components/Refresh/Refresh';
 import { PfColors } from '../../components/Pf/PfColors';
 import TimeRangeComponent from 'components/Time/TimeRangeComponent';
@@ -108,6 +108,7 @@ const tabIndex: { [tab: string]: number } = {
 class ServiceDetails extends React.Component<ServiceDetailsProps, ServiceDetailsState> {
   private promises = new PromisesRegistry();
   private traces: JaegerTrace[] = [];
+  private traceParams: any = {};
   private lastFetchTraceMicros: number | undefined = undefined;
   private lastFetchTracesError = false;
   private graphDataSource: GraphDataSource;
@@ -322,7 +323,13 @@ class ServiceDetails extends React.Component<ServiceDetailsProps, ServiceDetails
           }
         });
     } else {
-      const params = getQueryJaeger();
+      let params = changeParams(this.traceParams);
+      if (params) {
+        this.traceParams = params;
+        this.lastFetchTraceMicros = undefined;
+      } else {
+        params = getQueryJaeger();
+      }
       if (this.lastFetchTraceMicros) {
         params[URLParam.JAEGER_START_TIME] = this.lastFetchTraceMicros;
       }
@@ -333,7 +340,7 @@ class ServiceDetails extends React.Component<ServiceDetailsProps, ServiceDetails
           let myState = {};
           if (traces && traces.data) {
             this.traces = appendTraces
-              ? this.traces.filter(t => t.startTime >= params[URLParam.JAEGER_START_TIME]).concat(traces.data)
+              ? this.traces.filter(t => t.startTime <= params[URLParam.JAEGER_START_TIME]).concat(traces.data)
               : traces.data;
             if (traces.data.length === 0) {
               myState['selectedTrace'] = undefined;
--- a/src/pages/ServiceDetails/ServiceDetailsPage.tsx
+++ b/src/pages/ServiceDetails/ServiceDetailsPage.tsx
@@ -301,9 +301,16 @@ class ServiceDetails extends React.Component<ServiceDetailsProps, ServiceDetails
   };
 
   fetchTracesData = (cleanTrace: boolean = false, traceId?: string) => {
+    /*
+     Remove selected trace
+    */
     if (cleanTrace) {
       this.setState({ selectedTrace: undefined });
     }
+
+    /*
+      Fetch for a specific trace with ID traceId
+    */
     if (traceId) {
       fetchTrace(this.props.match.params.namespace, this.props.match.params.service, traceId)
         .then(trace => {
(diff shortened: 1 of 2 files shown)


jotak:
I think `lastFetchTraceMicros` must be reset (back to undefined) when query options change. Else, you would have old traces that don't correspond to the query performed.

jotak:
It still doesn't work :( New traces are still replacing old traces

I think it's because you modify `params[URLParam.JAEGER_START_TIME]` and that value is used for multiple purposes: it's not only used to pass as query params to jaeger query, but also used to define the time frame that Kiali is watching.

E.g. when you run `this.traces.filter(t => t.startTime >= params[URLParam.JAEGER_START_TIME])`, it will delete all previous data since you modified `params[URLParam.JAEGER_START_TIME]` a few lines above.

I think a solution could be to copy `params` into a new object, modify its `JAEGER_START_TIME`, and pass that copy to jaeger query.

aljesusg:
yeah the point is that I update all the params in a helper so...I am going to save params in another object and compare if query params are different

lucasponce:
Side comment.
Due the comments on the PRs and the issue around this topic, I suggest to add some comments explaning with more detail this in the code.

There was some confusion to understand and test this and probably if we need to review this logic later in some weeks it would be good if we can get some help in the code explaining it.

aljesusg:
Ok I think that now is fine, anyway @jotak @lucasponce 

Does it make sense to keep the limit when we are showing more traces in each update?

Maybe for the "first"

src/pages/ServiceDetails/ServiceDetailsPage.tsx
```diff
@@ -322,18 +323,29 @@ class ServiceDetails extends React.Component<ServiceDetailsProps, ServiceDetails
           }
         });
     } else {
-      const params = getQueryJaeger();
+      let params = changeParams(this.traceParams);
```
jotak, line 326: I had in mind something simpler with just a defensive copy of the params:

```javascript
const params = getQueryJaeger();
const fetchParams = { ...params }; // copy
if (this.lastFetchTraceMicros) {
  fetchParams[URLParam.JAEGER_START_TIME] = this.lastFetchTraceMicros;
}
fetchTraces(this.props.match.params.namespace, this.props.match.params.service, fetchParams)
```
This way when you modify fetchParam it doesn't affect the "global" params object.

aljesusg: yes but I can't check if the user change any option with this like tags or limit

src/pages/ServiceDetails/ServiceDetailsPage.tsx
```diff
@@ -322,18 +323,29 @@ class ServiceDetails extends React.Component<ServiceDetailsProps, ServiceDetails
...
+      }
       fetchTraces(this.props.match.params.namespace, this.props.match.params.service, params)
         .then(traces => {
           const appendTraces = this.lastFetchTraceMicros !== undefined;
           this.lastFetchTracesError = false;
           let myState = {};
           if (traces && traces.data) {
             this.traces = appendTraces
-              ? this.traces.filter(t => t.startTime >= params[URLParam.JAEGER_START_TIME]).concat(traces.data)
+              ? this.traces.filter(t => t.startTime <= params[URLParam.JAEGER_START_TIME]).concat(traces.data)
```
jotak, line 343: ok, this is the faulty check. It should be `>=` , but `params[URLParam.JAEGER_START_TIME]` should be here the start time in kiali time frame, not the one from the jaeger query (that's why I was mentioned doing a copy of params before changing it)

aljesusg: Sorry I agree with the copy now I understand.

But I can't understand the second step.

If I have the traces from (t-1) to t, and I perform the refresh to request traces from t to t+1.

I should filter traces to have all the traces <= t and concatenate them with the new traces than I am requesting from t to  t+1.

So I can't udnerstand why in https://github.com/kiali/kiali-ui/blob/master/src/components/Metrics/SpanOverlay.ts#L10-L38
(I add comments in code)
```ts
const nowMicros = new Date().getTime() * 1000;
    const frameStart = nowMicros - (duration + 60) * 1000000; // seconds to micros with 1min margin;
// frameStart is my params[URLParam.JAEGER_START_TIME]
// calculated in RouterHelper
    const opts: TracingQuery = { startMicros: this.lastFetchMicros || frameStart };
    API.getServiceSpans(namespace, service, opts)
      .then(res => {
        this.lastFetchError = false;
// Here you have in this.spans spans from t-1 to frameStart
// so why are you filtering by spans later than frameStart ? this should be zero
        this.spans = doAppend ? this.spans.filter(s => s.startTime >= frameStart).concat(res.data) : res.data;
        this.onChange(this.buildOverlay());
        // Update last fetch time only if we had some results
        // So that if Jaeger DB hadn't time to ingest data, it's still going to be fetched next time
        if (this.spans.length > 0) {
          this.lastFetchMicros = Math.max(...this.spans.map(s => s.startTime));
        }
      })
      .catch(err => {
        if (!this.lastFetchError) {
          AlertUtils.addError('Could not fetch spans.', err);
          this.lastFetchError = true;
        }
      });
```
Maybe I need to grab another coffee

jotak: `// Here you have in this.spans spans from t-1 to frameStart`

This isn't exactly correct. We must make the distinction between the frame start in Kiali (ie. time for whole graph) versus frame start for fetching data. Let's call them `kialiFrameStart` and `fetchFrameStart`.

So here, `this.spans` are spans before `fetchFrameStart`, and most of them are after `kialiFrameStart`. But because time flies, some of them can possibly be before `kialiFrameStart` and these are the ones we need to filter out, to keep only spans that fits in kiali time frame.

aljesusg:
ok now the timeframe works

hhovsepy:
Before:
Screenshot from 2020-04-21 15-41-52

After:
Screenshot from 2020-04-21 15-42-23

hhovsepy:
Better screenshots:
Before:
First load:
Screenshot from 2020-04-21 16-54-41
After some time period it deletes previous traces to keep Total=20
Screenshot from 2020-04-21 16-55-21

After:
Fist load:
Screenshot from 2020-04-21 16-55-38
After some time the previous loaded traces are kept:
Screenshot from 2020-04-21 17-00-27

jotak:
So, by copying params you fixed the issue frame start, but as you removed your previous changes, the issue of having `lastFetchTraceMicros` reset on query options change reappeared. You should combine your two last commits, having the params copying *and* the changed params detection.

aljesusg:
Ok I did the changes and I added some comments @jotak

lucasponce:
Please, I commented this before.
This issue for one reason or another has created some confusion.
That's ok, it's part of our job.

I ask before to merge to update the code with proper comments to help any external developer to jump on this.

It may obvious to well versed and familiar folks but goal should be that anyone could help to understand it.

I can't see this kind of comments in the current comments.

I know that I may sound like adding extra, but I'm also adding that practice to myself.

Thanks.

src/pages/ServiceDetails/ServiceDetailsPage.tsx
```diff
@@ -322,18 +323,30 @@ class ServiceDetails extends React.Component<ServiceDetailsProps, ServiceDetails
...
+      fetchTraces(this.props.match.params.namespace, this.props.match.params.service, fetchParams)
         .then(traces => {
           const appendTraces = this.lastFetchTraceMicros !== undefined;
           this.lastFetchTracesError = false;
           let myState = {};
           if (traces && traces.data) {
+            // We get new trace results so we need to concatenate them to the results that we have
             this.traces = appendTraces
-              ? this.traces.filter(t => t.startTime >= params[URLParam.JAEGER_START_TIME]).concat(traces.data)
+              ? this.traces.filter(t => t.startTime >= fetchParams[URLParam.JAEGER_START_TIME]).concat(traces.data)
```
jotak, line 344:
```suggestion
              ? this.traces.filter(t => t.startTime >= params[URLParam.JAEGER_START_TIME]).concat(traces.data)
```

jotak:
All of this is showing the flaws in the code architecture. I think we really need to think about refactoring some bits, to have a clean separation of concerns. Out of my head, I can think about the ServiceDetails components (too big and multi-purpose) and this params object (it's shared for different purposes). There's perhaps other areas to investigate.

aljesusg:
Ok i did the changes and I moved all of this to a new class to do some refactor in next steps @jotak to avoid make ServiceDetails bigger

aljesusg:
after verified that is working I'll add comments and clean it

hhovsepy:
When the Traces tab is open for the first time, it does not show traces, then updates after 10 seconds:
Screenshot from 2020-04-23 11-07-33

aljesusg:
mm it's not updating in the first time as I expected...I am going to redo this

lucasponce:
What is the status of this PR ?

I cannot see any comment related to document what was asked in 
https://github.com/kiali/kiali-ui/pull/1745#issuecomment-617756207

aljesusg:
@lucasponce I want to wait @jotak review to verified that it's working as expected, after that I'll add comments

aljesusg:
@lucasponce Added comments please check them.

lucasponce:
Thanks @aljesusg for the work.

Thanks @jotak and @hhovsepy for the review.

PR #1747, opened 2020-04-21
Add AuthorizationPolicy form

lucasponce:
This PRs provides a new form to create AuthorizationPolicy objects.

Note:
- IstioConfig new form only support the creation of the object, updates are done via yaml editor.
- AuthorizationPolicy is a complex object with multiple edges and corner cases, hard to focus just on some, so, the form supports all cases (in other objects as Gateways/Sidecars forms only support a subset of most direct scenarios).
- User requires to catch up with the terminology used by AuthorizationPolicy entity.

How to test:
- Best strategy would be to follow the Istio tasks tutorial https://istio.io/docs/tasks/security/authorization/ and check that all scenarios can be done via the form, checking that resulting yaml is valid and well formed.

Fixes https://github.com/kiali/kiali/issues/2041

It requires backend PR https://github.com/kiali/kiali/pull/2680

### DIFF
--- a/src/components/IstioWizards/IstioWizardActions.ts
+++ b/src/components/IstioWizards/IstioWizardActions.ts
@@ -6,6 +6,8 @@ import { SuspendedRoute } from './SuspendTraffic';
 import {
   AuthorizationPolicy,
   AuthorizationPolicyRule,
+  AuthorizationPolicyWorkloadSelector,
+  Condition,
   DestinationRule,
   DestinationRules,
   DestinationWeight,
@@ -18,8 +20,7 @@ import {
   Source,
   StringMatch,
   VirtualService,
-  VirtualServices,
-  WorkloadSelector
+  VirtualServices
 } from '../../types/IstioObjects';
 import { serverConfig } from '../../config';
 import { ThreeScaleServiceRule } from '../../types/ThreeScale';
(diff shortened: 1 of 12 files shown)


xeviknal:
I realized an small text (see the shot) that I don't know exactly if is an unwanted text or is part of the config:

Screenshot of Kiali Console (50)

It happens right after adding a when rule.

lucasponce:
>  realized an small text (see the shot) that I don't know exactly if is an unwanted text or is part of the config:

Good catch. I didn't notice, I'll take a look at it. Thanks.

xeviknal:
Second question: Rules field in an Authorization Policy is an array. Is this going to be covered in a second iteration?

lucasponce:
> Second question: Rules field in an Authorization Policy is an array. Is this going to be covered in a second iteration?

You mean, you can't enter multiple rules ?

xeviknal:
My bad, it is working. I can add multiple rules.
The problem I had it was that all looked like disabled: create rule button, add to, from, when. So I kind of thought it wasn't the option of adding another one.

skondkar:
Tested the Istio tasks tutorial:  https://istio.io/docs/tasks/security/authorization/.
I could add all the possible scenarios that can be done via the Add Authorization Policy form successfully. Looks good to me.

lucasponce:
@skondkar thanks ! good catch.

lucasponce:
@xeviknal @skondkar I've sent a commit that hopefully fixes the couple of issues commented.
Let me know any other comment/suggestion you may have.

This is a preliminary version, so I imagine this will evolver with future feedback.

skondkar:
Thanks @lucasponce 
The issue with 'Labels' field and 'state' text is gone now.

However, I am seeing an issue with all text boxes in 'Add From', 'Add To' and 'Add When' sections. I can enter first letter or number, then the text box loses focus and I then have to click in the text box again to enter another letter/number and then need to repeat for every letter/number to enter in the text boxes.
For ex: 'Add From'->Choose ipBlocks and then try to enter ip in 'Value' text box.

Kiali UI:     1.17.0 (a4608181890ec81ce4ae87de93210d5a69ed61b8)
Kiali Server:     v1.17.0-SNAPSHOT (22fb9d679154bbb4be42f915b7d1151b12415a58)

lucasponce:
@xeviknal @skondkar 
Ok, I've rebased the PR with master, I've applied the workaround to prevent the problem with patternfly tables, and I've also applied some improvements for a better handle of empty tables.

Let me know if it looks ok to you from your side.

Thanks, I know it's a large form and it takes time, but you are helping me a lot testing this one.

PR #1754, opened 2020-04-23
Fix issue zoom/pan/fit and graph-hide

jshaughn:
This fixes an issue with zoom/fit/pan and graph hide.  We only want to re-establish the zoom/pan settings if we believe the graph is unchanged (basically the same removed elements are in play as when we previously executed the hide logic).  The determination is not perfect but should work fairly well.
- also made a couple of minor tweaks to "safeFit()" on the off-chance it could help in general.

Fix related to https://github.com/kiali/kiali/issues/2514

In the _before_ video note that after applying the hide expression and toggling `Show Services` that we need to also click `Fit` in the cytoscape toolbar.  In the _after_ video we can see that the clicking `Fit `has no effect, the graph is already layed out correctly.

Before:
kiali#2514-before

After:
kiali#2514-after

After 2: this second _after_ video shows that we maintain the proper zoom/pan on a standard refresh (this is to prove no regression, the original code also works in this case):
kiali#2514-after-2

### DIFF
--- a/src/components/CytoscapeGraph/CytoscapeGraphUtils.ts
+++ b/src/components/CytoscapeGraph/CytoscapeGraphUtils.ts
@@ -67,14 +67,15 @@ export const CyNode = {
 };
 
 export const ZoomOptions = {
-  fitPadding: 25
+  fitPadding: 25,
+  maxZoom: 2.5
 };
 
 export const safeFit = (cy: Cy.Core, centerElements?: Cy.Collection) => {
   cy.fit(centerElements, ZoomOptions.fitPadding);
-  if (cy.zoom() > 2.5) {
-    cy.zoom(2.5);
-    cy.center(centerElements);
+  if (cy.zoom() > ZoomOptions.maxZoom) {
+    cy.zoom(ZoomOptions.maxZoom);
+    !!centerElements && !!centerElements.length ? cy.center(centerElements) : cy.center();
   }
 };
 
--- a/src/components/CytoscapeGraph/CytoscapeGraph.tsx
+++ b/src/components/CytoscapeGraph/CytoscapeGraph.tsx
@@ -536,12 +536,9 @@ export default class CytoscapeGraph extends React.Component<CytoscapeGraphProps>
 
     cy.endBatch();
 
+    // Run layout and fit outside of the batch operation for it to take effect on the new nodes
     if (updateLayout) {
       CytoscapeGraphUtils.runLayout(cy, this.props.layout);
-    }
-
-    // We need to fit outside of the batch operation for it to take effect on the new nodes
-    if (updateLayout) {
       this.safeFit(cy);
     }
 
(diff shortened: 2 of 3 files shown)


pbajjuri20:
I got an error, with the "zoom/pan/fit" is not properly fixed with "Node Names". It is perfectly fixed with "Service Nodes"

Namespace:  "Bookinfo"
Graph:  "Versioned App Graph"
Display:  "Node Names"
Hide: app=ratings

zoom

jshaughn:
> I got an error, with the "zoom/pan/fit" is not properly fixed with "Node Names

Hi @pbajjuri20 , so just to be clear, you did not actually get an error/crash but you do see a shift in position when toggling Show Node Names.  Is that correct?   Believe it or not, the node names option is fairly complicated, especially for "App" graphs that have the app boxes.  A problem here may actually be a separate issue.  I'll take a look and let you know if there is anything I can do, but if not, and as long as there is not an actual error/crash generated, we may need to make this a separate issue.  I'll let you know.

src/components/CytoscapeGraph/CytoscapeGraphUtils.ts
```diff
@@ -67,14 +67,15 @@ export const CyNode = {
...
 };
 
 export const safeFit = (cy: Cy.Core, centerElements?: Cy.Collection) => {
   cy.fit(centerElements, ZoomOptions.fitPadding);
-  if (cy.zoom() > 2.5) {
-    cy.zoom(2.5);
-    cy.center(centerElements);
+  if (cy.zoom() > ZoomOptions.maxZoom) {
+    cy.zoom(ZoomOptions.maxZoom);
+    !!centerElements && !!centerElements.length ? cy.center(centerElements) : cy.center();
```
israel-hdez, line 78: I think double bangs are not that elegant nowadays. There is a *Reasons for using double negation* and a *reasons agains using double negation* in [this sitepont article](https://www.sitepoint.com/javascript-double-negation-trick-trouble/). IMO, it's usage is a personal preference.

My preference is to use double negation only if a boolean value is strictly needed, instead of a truthy/falsy value. Here I'd prefer `centerElements && centerElements.length != 0 ? cy.center(centerElements) : cy.center();`. 

Again, it's a personal preference. If this is your preferred way, I'm fine :)

jshaughn: Maybe it is uglier but I like that it is explicit about being a boolean test for whatever the datatype may be.  I guess I'm subjectively for it :)

israel-hdez: Any code style is subjective :)

jshaughn:
> I saw the same issue as @pbajjuri20...

@israel-hdez I agree with you about the egraph not being particularly useful without labels and I'd go so far as to upvote the option removal.  I'd prefer one less option.

@israel-hdez @pbajjuri20 , I spent a while trying to deal with the node labels stuff but there is some odd behavior there, even in master and outside of graph hide.  I'll create another issue for that but as Edgar says, this is just too low priority to spend more time.  So, Praneeth, if you don't have any other issue please approve and I will merge as-is.

pbajjuri20:
@jshaughn, I don't have another issue. I'm approving this PR.

jshaughn:
@israel-hdez @pbajjuri20 Here is an example of other node label weirdness, I think it's not worth dealing much with this option unless it somehow causes an actual crash: https://github.com/kiali/kiali/issues/2696

israel-hdez:
Proposal to remove "Node Names" display option: https://github.com/kiali/kiali/issues/2699

PR #1759, opened 2020-04-29
Fix several CSS issues

lucasponce:
Fixes https://github.com/kiali/kiali/issues/2616
Fixes https://github.com/kiali/kiali/issues/2635
Fixes https://github.com/kiali/kiali/issues/2521
Fixes https://github.com/kiali/kiali/issues/2648

All steps and rationale for these CSS changes are described and documented under https://github.com/kiali/kiali/issues/2616

In short, early optimization in UI layer may create easily CSS problems. As soon as someone wants to put in a single component something that is used in several locations and at the end that shared component may have different styles, it can create easily problems that are hard to fix.

This PR has no intention to refactor the UI, don't propose "hey, this can be put together in a single component", all these problems started with a comment like that :-).

A little bit of duplication may help to fix CSS issues and to see different aspect/ratios across all pages.

To reviewers: please, try to check if the styles proposed in https://github.com/kiali/kiali/issues/2616 are consistent and nothing wrong is broken.

### DIFF
--- a/src/components/DefaultSecondaryMasthead/DefaultSecondaryMasthead.tsx
+++ b/src/components/DefaultSecondaryMasthead/DefaultSecondaryMasthead.tsx
@@ -28,7 +28,7 @@ export default class DefaultSecondaryMasthead extends React.Component {
         title = 'Create New Iter8 Experiment';
       }
       return (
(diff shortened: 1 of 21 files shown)


lucasponce:
Thanks @xeviknal and @prachiyadav for the review.
I will merge.
In any case, if there is anything left let me know to prepare a follow up patch.

PR #1760, opened 2020-04-29
Further attempts to eliminate graph drift on refresh when using graph hide

jshaughn:
Further attempts to eliminate graph drift on refresh when using graph hide. Perform a fit unless there is a significant zoom/pan change. Also, run the layout twice, it seems to help.

Possible fix for https://github.com/kiali/kiali/issues/2514


pbajjuri20:
Tested and passed on Kiali UI 1.18.0 (63b362bc2ff7b5184adec1ae38a57e189eddcee3)

PR #1763, opened 2020-04-30
Sync Health tooltip style with rest of Kiali

lucasponce:
Fixes https://github.com/kiali/kiali/issues/2706

Now health is consistent with other tooltips used in Kiali:
image

This out-of-sync is more visible in the same row that other tooltips are used:
image

image

Note, that this is consistent with other complex tooltips with similar structure (list of status with icons):

image


lucasponce:
Thanks for the review.

Please, let's not cross-post UX in PR, better to open a new issue:
https://github.com/kiali/kiali/issues/2725

Otherwise we are going to flood them with notifications hard to follow by them.

PR #1773, opened 2020-05-04
Iter8 issue2633

jadeyliu939:
** Describe the change **
1. Update iter8 experiment detail page
2. Fix font color issue in listing page's tooltip popup

** Changes depends on **
[pull request #2729](https://github.com/kiali/kiali/pull/2729)

** Issue reference **
[kiali#2633](https://github.com/kiali/kiali/issues/2633)

** Screenshot **
1. Tooltip font color fix
Screen Shot 2020-05-04 at 3 39 29 PM

2.  Detail page - main tab
Screen Shot 2020-05-04 at 3 39 43 PM

3.  Detail page - Criteria tab
Screen Shot 2020-05-04 at 3 39 53 PM


rhqci:

lucasponce:
Thanks @jadeyliu939.
I'm going to review it today.
Note that in master I did some style changes in list/details page that probably it may apply to this work (I see some old styles in the screenshots) but it will comment with more detail in the PR code.

prachiyadav:
Screenshot from 2020-05-06 12-26-35
Screenshot from 2020-05-06 12-25-30
Screenshot from 2020-05-06 12-24-26

prachiyadav:
Screenshot from 2020-05-06 12-30-10

jadeyliu939:
The `AnalyticsServiceError` is due to Iter8 release now default to support port 8080 instead of port 80. Kiali will need to be changed to support iter8 configmap for port info.

lucasponce:
> The AnalyticsServiceError is due to Iter8 release now default to support port 8080 instead of port 80. Kiali will need to be changed to support iter8 configmap for port info.

Hi @jadeyliu939 I've read your slack comment, you mean to update the counterpart backend PR of this one or where is that port mapped ?

package.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "@kiali/kiali-ui",
   "version": "1.18.0",
+  "proxy": "https://52.116.156.242:32263/kiali/",
```
lucasponce, line 4: Ops, we don't commit proxies in package.json, unless is a change on the libraries or scripts this file should not be commited.

jadeyliu939:
> > The AnalyticsServiceError is due to Iter8 release now default to support port 8080 instead of port 80. Kiali will need to be changed to support iter8 configmap for port info.
> 
> Hi @jadeyliu939 I've read your slack comment, you mean to update the counterpart backend PR of this one or where is that port mapped ?
> 
> Or in other words, that url should be accesible by which component ?
> 
> image

To be exact, The port for Iter8 Analytics services is configurable, user can configure to any port number they want. In earlier installation, It was default to 80, but in the latest installation, it was default to 8080. This is why if your Iter8 installation was done more than 2 weeks ago, you have no problem. In the Kiail Iter8 backend code, it uses the default port 80.  which make it incompatible with the latest installation yaml. In order for backward compatible, In release 0.0.1, The port can be found in the configmap : `iter8-analytics`.  So in the Kiali Iter8 backend code, we will check for port # specified in the configmap, if there is no configumap, we will use default port 80 ( meaning it is pre-0.01 release).

lucasponce:
>  In the Kiail Iter8 backend code, it uses the default port 80.

But I'm still missing one piece here, Kiali communicates with the k8s API but not directly with Iter8 service deployed in the pod.

Or where is that added ?

jadeyliu939:
Iter8 have two services - Iter8-Controller and Iter8-Analytic service. The port number is  for Iter8-analytics services, which is for the iter8-controller to talk to. Whenever an experiment is created ( through k8s API). the analytics services (uri and port number) needed to be specified in the CRD. The design thinking is so that user can specified different analytic service to be used by the controller. We did not expose that user setting in the Kiali UI. instead, we let Kiali to discover the analytics service in the environment.

jadeyliu939:
Close, New PR open - [#1784](https://github.com/kiali/kiali-ui/pull/1784)

PR #1776, opened 2020-05-06
[v1.12] OSSM-199 Fix showing empty error when jaeger is disabled

jotak:


hhovsepy:
@jotak  is this the error to be shown by this fix:
Screenshot from 2020-05-06 17-24-56

aljesusg:
It should show this error if jaeger is enabled in the configuration and kiali can't connect with Jaeger

hhovsepy:
With non existing Jaeger config:
Screenshot from 2020-05-06 17-54-07

jotak:
@hhovsepy , the error that you show previously (on jira) was triggered only when `tracing.enabled` is false in config, so I imagined this is what you meant when writing "Jaeger unavailable", and this is what I fixed.

What's the situation you have when you say you still see the issue? Disabled in config, or something else? (like Jaeger configured but pod not running)?
Also, what do you mean in your second comment by "With non existing Jaeger config:" ? (isn't it a configuration issue then, if there's no jaeger config?)

hhovsepy:
@jotak in my env the Jaeger pod fails to start with "Error: ImagePullBackOff", cannot pull 'ose-oauth-proxy' image.

PR #1777, opened 2020-05-07
Filters for different kind of labels are mixed up

aljesusg:
Fix https://github.com/kiali/kiali/issues/2559

### DIFF
--- a/src/pages/Overview/Filters.ts
+++ b/src/pages/Overview/Filters.ts
@@ -49,10 +49,10 @@ export const mtlsFilter: FilterTypeWithFilter<NamespaceInfo> = {
 };
 
 export const labelFilter: FilterTypeWithFilter<NamespaceInfo> = {
-  id: 'label',
-  title: 'Label',
+  id: 'nsLabel',
+  title: 'NsLabel',
   placeholder: 'Filter by Label',
-  filterType: FilterTypes.label,
+  filterType: FilterTypes.nsLabel,
   customComponent: LabelFilters,
   action: FILTER_ACTION_APPEND,
   filterValues: [],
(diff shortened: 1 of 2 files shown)


src/pages/Overview/Filters.ts
```diff
@@ -49,10 +49,10 @@ export const mtlsFilter: FilterTypeWithFilter<NamespaceInfo> = {
 };
 
 export const labelFilter: FilterTypeWithFilter<NamespaceInfo> = {
-  id: 'label',
-  title: 'Label',
+  id: 'nsLabel',
+  title: 'NsLabel',
```
jotak, line 53: I see that you couldn't use just "Label" here ... there must be something weird somewhere in our code because `title` should in theory be used only for display. So, +1 for this solution, perhaps it's just temporary and we should investigate the underlying issue later

However perhaps something more user-friendly than "NsLabel" should be set? Full words "Namespace label" ?

aljesusg: We assign the category to the title for example in Overview filters
```ts
af.category === availableFilter.title)
```
StatefulFilters L136
etc...

It should be the id IMO

jotak: Yes, I think we can open a new ticket about using `id` instead of `title` (it might trigger bigger refactoring & full regression tests needed so let's not mix up with this simple fix)

I'm testing locally with "Namespace label" as title, indeed it increases the width, but I find it very acceptable, no? See:

Capture d’écran de 2020-05-11 11-41-35

aljesusg: Yes...Ok I'll change it

PR #1780, opened 2020-05-08
Labels filtering: "all of" / "any of" switch

aljesusg:
Fix https://github.com/kiali/kiali/issues/2734

# Demo

- I had to add this like a new Filter to be able to move this param to other lists
- LabelOperation Dropdown is displayed when user go to Label filter selection or we have a Label filter applied
- LabelOperation not add a "chip" in filters because by default we'll apply "OR" operation and it make not sense to add it for me.
- Applied in Workload, Services and Apps.
- It make sense to have this in overview too when we are filtering by labels related with the namespace and not by the objects of the namespace?

## Or Operation (the results are the same that master)
Or

## This show results that match all label filters.
AND

## Other notes

The line https://github.com/aljesusg/kiali-ui/blob/fix_2734/src/helpers/LabelFilterHelper.ts#L14 throw TS error

```bash
TypeScript error in /home/aljesusg/Projects/kiali/kiali-ui/src/helpers/LabelFilterHelper.ts(14,59):
No overload matches this call.
  Overload 1 of 3, '(callbackfn: (previousValue: string, currentValue: string, currentIndex: number, array: string[]) => string, initialValue: string): string', gave the following error.
    Type '[string[], never[]] | [never[], string[]]' is not assignable to type 'string'.
      Type '[string[], never[]]' is not assignable to type 'string'.
  Overload 2 of 3, '(callbackfn: (previousValue: [never[], never[]], currentValue: string, currentIndex: number, array: string[]) => [never[], never[]], initialValue: [never[], never[]]): [never[], never[]]', gave the following error.
    Type '[string[], never[]] | [never[], string[]]' is not assignable to type '[never[], never[]]'.
      Type '[string[], never[]]' is not assignable to type '[never[], never[]]'.
        Type 'string[]' is not assignable to type 'never[]'.
          Type 'string' is not assignable to type 'never'.  TS2769

    12 |     TS Error but this works...
    13 |   */
  > 14 |   const [keys, keyValues] = filters.reduce(([p, f], e) => (!e.includes(':') ? [[...p, e], f] : [p, [...f, e]]), [[], []]);

```
In my opinion 

This is take more time
```ts
  //const keys = filters.filter(f => !f.includes(':'));
  //const keyValues = filters.filter(f => f.includes(':'));
```
than
```ts
const [keys, keyValues] = filters.reduce(([p, f], e) => (!e.includes(':') ? [[...p, e], f] : [p, [...f, e]]), [[], []]);
```
to separate the results.

cc @jotak @lucasponce

### DIFF
--- a/src/components/Filters/StatefulFilters.tsx
+++ b/src/components/Filters/StatefulFilters.tsx
@@ -226,7 +226,7 @@ export class StatefulFilters extends React.Component<StatefulFiltersProps, State
   renderInput() {
     const { currentFilterType, currentValue } = this.state;
 
-    if (!currentFilterType || currentFilterType.filterType === FilterTypes.opLabel) {
+    if (!currentFilterType) {
       return null;
     }
     if (currentFilterType.filterType === FilterTypes.typeAhead) {
--- a/src/components/Filters/StatefulFilters.tsx
+++ b/src/components/Filters/StatefulFilters.tsx
@@ -155,11 +155,6 @@ export class StatefulFilters extends React.Component<StatefulFiltersProps, State
     this.updateActiveFilters(activeFilters);
   };
 
-  updateOpSortLabel = (value: LabelOperation) => {
-    this.setState({ activeFilters: { filters: this.state.activeFilters.filters, op: value } });
-    this.updateActiveFilters({ filters: this.state.activeFilters.filters, op: value });
-  };
-
   selectFilterType = (value: string) => {
     const { currentFilterType } = this.state;
     const filterType = this.state.filterTypes.filter(filter => filter.id === value)[0];
@@ -367,7 +362,9 @@ export class StatefulFilters extends React.Component<StatefulFiltersProps, State
                 <span className={classNames(paddingStyle)}>Label Operation</span>
                 <FormSelect
                   value={activeFilters.op}
-                  onChange={value => this.updateOpSortLabel(value as LabelOperation)}
+                  onChange={value =>
+                    this.updateActiveFilters({ filters: this.state.activeFilters.filters, op: value as LabelOperation })
+                  }
                   aria-label="filter_select_value"
                   style={{ width: 'auto' }}
                 >
--- a/src/components/Filters/StatefulFilters.tsx
+++ b/src/components/Filters/StatefulFilters.tsx
@@ -15,7 +15,14 @@ import {
   ToolbarItem,
   ToolbarSection
 } from '@patternfly/react-core';
-import { ActiveFilter, FILTER_ACTION_UPDATE, FilterType, FilterTypes } from '../../types/Filters';
+import {
+  ActiveFilter,
+  FILTER_ACTION_UPDATE,
+  FilterType,
+  FilterTypes,
+  LabelFilter,
+  OpLabelFilter
+} from '../../types/Filters';
 import * as FilterHelper from '../FilterList/FilterHelper';
 import { PromisesRegistry } from '../../utils/CancelablePromises';
 import { style } from 'typestyle';
@@ -34,6 +41,7 @@ export interface StatefulFiltersState {
   activeFilters: ActiveFilter[];
   currentValue: string;
   isExpanded: boolean;
+  labelSortOperation: string;
 }
 
 export class FilterSelected {
@@ -51,6 +59,7 @@ export class FilterSelected {
     return FilterSelected.selectedFilters !== undefined;
   };
 }
+
 const rightToolbar = style({
   marginLeft: 'auto'
 });
@@ -63,7 +72,6 @@ export class StatefulFilters extends React.Component<StatefulFiltersProps, State
 
   constructor(props: StatefulFiltersProps) {
     super(props);
-
     let active = FilterSelected.getSelected();
     if (!FilterSelected.isInitialized()) {
       active = FilterHelper.getFiltersFromURL(this.props.initialFilters);
@@ -78,7 +86,8 @@ export class StatefulFilters extends React.Component<StatefulFiltersProps, State
       filterTypes: this.props.initialFilters,
       activeFilters: active,
       isExpanded: false,
-      currentValue: ''
+      currentValue: '',
+      labelSortOperation: OpLabelFilter.filterValues[0].id
     };
   }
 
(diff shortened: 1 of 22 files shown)


jotak:
I'm starting to review today - but just as a quick remark, when I enter a single label filter to begin with, it doesn't filter anything (items that don't match the filter are still displayed)

src/components/Filters/StatefulFilters.tsx
```diff
@@ -216,7 +237,7 @@ export class StatefulFilters extends React.Component<StatefulFiltersProps, State
   renderInput() {
     const { currentFilterType, currentValue } = this.state;
 
-    if (!currentFilterType) {
+    if (!currentFilterType || currentFilterType.filterType === FilterTypes.opLabel) {
```
jotak, line 240: Is there something blocking for *not* having this "op labels" as a standard filter? It sounds a bit odd and forces to make a lot of special case handling like on this line

aljesusg: For me it makes not sense has a filter if there is not labels selected. On the other hand we are going to apply the "Or" operation if the user not set "And" so...Should we add a filter when the list is loaded? When the user set the first label ? ...

For these reasons I thought about have this behavior. What do you think?

aljesusg:
> I'm starting to review today - but just as a quick remark, when I enter a single label filter to begin with, it doesn't filter anything (items that don't match the filter are still displayed)

mmm weird, let me check that

src/components/Filters/StatefulFilters.tsx
jotak: I mean, I'm questioning the idea to have `opLabel` as a `FilterTypes`. It isn't a filter, it's something different.

aljesusg: Understood...I added it to be able to pass the option to the others list, we can handle this in another way

> I had to add this like a new Filter to be able to move this param to other lists

On the other hand we'll need to think another way to pass it to the filter method
https://github.com/kiali/kiali-ui/pull/1780/files#diff-f1cc29563c6473251f2ce0e4043b27e6R318

Other solution is change the "LabelFilter" to let have more params like this operation

jotak: I'm trying to see how we could do differently, as having `opLabel` as a `FilterTypes` looks more like a hack.

Perhaps, instead of having that new `OpLabelFilter` object in `Filters.ts`, we would change how we use `ActiveFilter`. Something like that (in Filters.ts):

```typescript
// This already exists
export interface ActiveFilter {
  category: string;
  value: string;
}

// This is new
export type LabelOperation = 'and' | 'or';
export interface ActiveFiltersInfo {
  filters: ActiveFilter[];
  op: LabelOperation;
}
```

And then, in many place where we use `ActiveFilter[]`, we would use `ActiveFiltersInfo`.
For instance in `StatefulFilters` state we would have `activeFilters: ActiveFiltersInfo;`

That's a bigger refactoring.

aljesusg: yes I know,....but I expect more changes/filters in this line so maybe we must do it now or later ^^

aljesusg:
Ok refactor done @jotak pff I tested it lists and works, I need now where set the label operation dropdown in the overview page

jotak:
Thanks @aljesusg ! I'm reviewing...

src/components/Filters/StatefulFilters.tsx
```diff
@@ -129,21 +140,26 @@ export class StatefulFilters extends React.Component<StatefulFiltersProps, State
...
     } else {
-      activeFilters.push(activeFilter);
+      activeFilters.filters.push(activeFilter);
     }
 
     this.updateActiveFilters(activeFilters);
   };
 
+  updateOpSortLabel = (value: LabelOperation) => {
+    this.setState({ activeFilters: { filters: this.state.activeFilters.filters, op: value } });
```
jotak, line 159: this `setState` update is performed twice: first here, then within the `updateActiveFilters` call below. So I guess the first can be removed?

src/helpers/LabelFilterHelper.ts
```diff
@@ -4,36 +4,91 @@ import { ServiceListItem } from '../types/ServiceList';
...
-            ))
-        );
-    } else {
-      // Check if has Label
-      result = result.concat(
-        items.filter(item =>
-          Object.keys(item.labels).some(key => key.startsWith(filter.trim()) && !result.includes(item))
-        )
-      );
+const filterLabelByOp = (labels: { [key: string]: string }, filters: string[], op: string = 'or'): boolean => {
```
jotak, line 7: I think a couple of unit tests for this function and/or `filterByLabel` would be really nice

src/helpers/LabelFilterHelper.ts
```diff
@@ -4,36 +4,91 @@ import { ServiceListItem } from '../types/ServiceList';
...
+    TS Error but this works...
+  */
+  // const [keys, keyValues] = filters.reduce(([p, f], e) => (!e.includes(':') ? [[...p, e], f] : [p, [...f, e]]), [[], []]);
+  const keys = filters.filter(f => !f.includes(':'));
+  const keyValues = filters.filter(f => f.includes(':'));
+
+  // Get all keys of labels
+  const labelKeys = Object.keys(labels);
+  // Case OR operation
+  if (op === 'or') {
```
jotak, line 21: for easier reading it would be nice to split this into two different functions: one for the "and" case, one for the "or".

src/components/Filters/StatefulFilters.tsx
aljesusg: mmm you're right  I'll remove it

src/helpers/LabelFilterHelper.ts
```diff
@@ -4,36 +4,91 @@ import { ServiceListItem } from '../types/ServiceList';
...
+        filterOkForLabel = false;
+      }
+      return undefined;
+    });
+  }
+
+  return filterOkForLabel;
+};
+
+export const filterByLabel = (items: itemsType[], filter: string[], op: string = 'or'): itemsType[] => {
```
jotak, line 83: Hmm am I wrong or this whole function is equivalent to:
`  return filter.length === 0 ? items : items.filter(item => filterLabelByOp(item.labels, filter, op));`
?

aljesusg: totally :(

aljesusg:
I did the changes @jotak and I found the problem about comment #1780 (comment)

src/components/Filters/StatefulFilters.tsx
jotak: After your refactoring, I think this `opLabel` in `FilterTypes` isn't used anymore, and can be safely removed, right?

src/helpers/__tests__/LabelFilterHelper.test.ts
```diff
@@ -0,0 +1,226 @@
...
+    namespace: 'bookinfo',
+    healthPromise: new Promise(() => {}),
+    name: 'productpage',
+    istioSidecar: false,
+    labels: { app: 'productpage', service: 'productpage' },
+    validation: { name: 'productpage', objectType: 'service', valid: true, checks: [] }
+  }
+];
+
+describe('LabelFilter', () => {
```
jotak, line 135: Nice, thanks!

jotak:
adding @hhovsepy for testing

src/helpers/__tests__/LabelFilterHelper.test.ts
aljesusg: Removed, I removed it too in src/types/Filters.ts in the last commit...old code :)

lucasponce:
I remove myself of this one, I can't review it and I don't want to delay this.
If @hhovsepy and @jotak are ok, it's good to me.

PR #1783, opened 2020-05-09
Update Istio 1.6 types

lucasponce:
Related to 
https://github.com/kiali/kiali/issues/2750

Istio 1.6 types have changed and those changes need to be propagated in Kiali components.
UI maps Istio types for details, validations, wizards and config edition.

This is a draft, probably these changes need to be propagated in the Kiali backend validations and other logic that access to the type details.

@xeviknal I need to review backend, probably this PR is not completed without that counterpart.

### DIFF
--- a/src/types/IstioObjects.ts
+++ b/src/types/IstioObjects.ts
@@ -583,7 +583,7 @@ export interface ServiceEntrySpec {
   ports?: Port[];
   location?: string;
   resolution?: string;
-  endpoints?: WorkloadEntry[];
+  endpoints?: WorkloadEntrySpec[];
   exportTo?: string[];
   subjectAltNames?: string[];
 }
@@ -835,6 +835,10 @@ export interface ServiceRoleBindingSubject {
 }
 
 export interface PeerAuthentication extends IstioObject {
+  spec: PeerAuthenticationSpec;
+}
+
+export interface PeerAuthenticationSpec {
   selector?: PeerAuthenticationWorkloadSelector;
   mtls?: PeerAuthenticationMutualTls;
   portLevelMtls?: { [key: number]: PeerAuthenticationMutualTls };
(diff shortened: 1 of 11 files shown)


xeviknal:
@lucasponce is there any way I can see the changes in each object? It is a bit difficult to defect which validations are affected bc of this change.

lucasponce:
> @lucasponce is there any way I can see the changes in each object? It is a bit difficult to defect which validations are affected bc of this change.

I basically follow the ref doc of each object:
https://preliminary.istio.io/docs/reference/config/networking/

In the backend there is a dynamic mapping for secondary levels of each objects.

There is an old issue in the backlog proposing to add strong type in the backend as well, but well, it has always low priority.

src/types/IstioObjects.ts
```diff
@@ -722,6 +854,7 @@ export enum PeerAuthenticationMutualTLSMode {
   STRICT = 'STRICT'
 }
 
+// 1.6
 export interface WorkloadEntry extends IstioObject {
   address: string;
```
xeviknal, line 859: This is nitpicking right now since we are not using WorkloadEntry objects into the UI. So, not a blocker itself.

I think this should be something like:

```
export interface WorkloadEntry extends IstioObject {
  spec: WorkloadEntrySpec
}
```

and then, in ServiceEntrySpec:
```
export interface ServiceEntrySpec {
   endpoints?: WorkloadEntrySpec[];
}
```

this way we don't include kind, metadata and apiVersion.

lucasponce: It's not nitpicking, it's a good catch.
It was missed the "spec" field which could lead in a problem.
I'll fix it.

lucasponce: And it also happens to other objects I see.
Thatnks, you spotted several issues, that's good.

lucasponce: It happens same problem in PeerAuthentication, WorkloadEntry and RequestAuthentication.
I'll fix those 3 mappings as par of this PR.

PR #1787, opened 2020-05-14
Show broken locks into edges without mtls in mesh-wide mtls scanarios

xeviknal:
Fixes https://github.com/kiali/kiali/issues/2787

I've seen that when there has never been mtls traffic, isMTLS has a -1. It made the not getting in the if condition and therefore, not printing any badge.

My question is, should I protect the usage of `mtlsPercentage` or the system guarantees that there will be always a number?

Screenshot of Kiali Console (66)

Steps to test this PR in bookinfo: https://github.com/xeviknal/istio-lab/tree/master/meshwide-mtls/details-disabled

### DIFF
--- a/src/components/CytoscapeGraph/graphs/GraphStyles.ts
+++ b/src/components/CytoscapeGraph/graphs/GraphStyles.ts
@@ -367,7 +367,7 @@ export class GraphStyles {
       }
 
       const mtlsPercentage = edgeData.isMTLS;
-      if (cyGlobal.showSecurity && mtlsPercentage && mtlsPercentage >= 0) {
+      if (cyGlobal.showSecurity) {
         if (mtlsPercentage > 0 && !cyGlobal.mtlsEnabled) {
           content = `${EdgeIconMTLS} ${content}`;
         } else if (mtlsPercentage < 100 && cyGlobal.mtlsEnabled) {
--- a/src/components/CytoscapeGraph/graphs/GraphStyles.ts
+++ b/src/components/CytoscapeGraph/graphs/GraphStyles.ts
@@ -367,11 +367,15 @@ export class GraphStyles {
       }
 
       const mtlsPercentage = edgeData.isMTLS;
-      if (cyGlobal.showSecurity) {
-        if (mtlsPercentage > 0 && !cyGlobal.mtlsEnabled) {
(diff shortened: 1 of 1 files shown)


src/components/CytoscapeGraph/graphs/GraphStyles.ts
```diff
@@ -367,7 +367,7 @@ export class GraphStyles {
       }
 
       const mtlsPercentage = edgeData.isMTLS;
-      if (cyGlobal.showSecurity && mtlsPercentage && mtlsPercentage >= 0) {
+      if (cyGlobal.showSecurity) {
         if (mtlsPercentage > 0 && !cyGlobal.mtlsEnabled) {
           content = `${EdgeIconMTLS} ${content}`;
         } else if (mtlsPercentage < 100 && cyGlobal.mtlsEnabled) {
```
jshaughn, line 373: I agree with the change of the outer conditional, we should drop down when the Security display option is enabled.   I think after that for clarity we should have a conditional on !cyGlobal.mtlsEnabled because that determines how to handle the actual mtlsPercentage and should clear up any issue showing the open lock badge.  Maybe something like:

```
 if (cyGlobal.showSecurity) {
    if (cyGlobal.mtlsEnabled) {
         if (!mtlsPercentage || mtlsPercentage < 100) {
             content = `${EdgeIconDisabledMTLS} ${content}`;
         }
    } else {
        if (mtlsPercentage && mtlsPercentage > 0) {
          content = `${EdgeIconMTLS} ${content}`;
        } 
    }
}
```

Does that look good?

xeviknal: yep, code looks better this way.
There is one more problem we have to approach here: what when there is no traffic?

Screenshot of Kiali Console (67)

So my proposal is to not show any badge if there isn't traffic. The result is:

Screenshot of Kiali Console (68)

jshaughn: @xeviknal +1, the lock badging is supposed to indicate information about the traffic, so no traffic should mean no badging.

jshaughn:
Sorry @xeviknal , this didn't make it into 1.18.1.   Add the backport/v1.18 label if you want to make sure we include it into a 1.18.2 release.

PR #1790, opened 2020-05-15
Add a hide log lines feature(noise) when displaying workload logs

mtho11:
Add a hide log lines feature for eliminating noise when displaying workload logs. 

** Describe the change **
Many times there is so much noise in the logs that one must filter via `grep` or whatever in the CLI.  in the UI, we need the similar capability to filter the logs.

** Issue reference **
fixes: https://github.com/kiali/kiali/issues/2778

** Screenshot **

hide-log

### DIFF
--- a/src/pages/WorkloadDetails/WorkloadInfo/WorkloadPodLogs.tsx
+++ b/src/pages/WorkloadDetails/WorkloadInfo/WorkloadPodLogs.tsx
@@ -1,17 +1,21 @@
 import * as React from 'react';
 import {
+  Button,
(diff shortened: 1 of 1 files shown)


mtho11:
@jshaughn so I completed everything but the padding on the SBS switch as it is in the component. And vertical scroll as a separate non-blocking issue.

Kiali_Console

pbajjuri20:
Tested and Passed on Kiali UI 1.19.0 (9a05f34f2a5e1e94a4531595af8787c4e9445a78)

PR #1792, opened 2020-05-18
Use new patternfly spinner

jotak:
This is using the new spinner. Note that it's said to be experimental.
Fixes https://github.com/kiali/kiali/issues/2679

Capture d’écran de 2020-05-18 11-35-56

(See also https://www.patternfly.org/v4/documentation/core/components/spinner )

A couple of notes:

- Previously the spinner was positioned to the left via absolute position. For some reason, this style wasn't applied on the new spinner, so I just removed it. Our spinner is in the "masterhead" icons/controls, so it makes more sense IMO to display it to the right ... if it's not what we want, then we should reorganize the whole masthead I think. Let me know if it's not ok to have it to the right.

- The PF component is said to be experimental. While testing I've seen that sometimes it displays very short arcs (almost dots) instead of the long ones, I'm not sure if it's intentional or if it's related to it being  experimental.

- I haven't seen any big problem beside that.

### DIFF
--- a/src/components/Nav/Masthead/Masthead.tsx
+++ b/src/components/Nav/Masthead/Masthead.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
-
+import { style } from 'typestyle';
 import { Toolbar as ToolbarNext, ToolbarGroup, ToolbarItem } from '@patternfly/react-core';
+
 import { default as MeshMTLSStatus } from '../../../components/MTls/MeshMTLSStatus';
 import { default as IstioStatus } from '../../IstioStatus/IstioStatus';
 import PfSpinner from '../../PfSpinner';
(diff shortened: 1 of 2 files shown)


jotak:
Actually I can't find anymore where it was said to be experimental, so perhaps it's not anymore, after all ...

aljesusg:
It should work without problem

See codebox https://codesandbox.io/s/fdwrp

Maybe there  is any problem when render inside a masthead. I'll test it.

About the position maybe UX can say something here. We can apply the same styles to the icons too but maybe the masthead is overwriting them

aljesusg:
My concern is that if we add new icons in the right we are going to move the spinner to the left each time... :(

jotak:
I moved it back to the left

PR #1796, opened 2020-05-19
Link from traces to workload logs can be wrong

aljesusg:
Fix https://github.com/kiali/kiali/issues/2712

The proposal is /^([a-z0-9-\.]+)-[a-z0-9]+-[a-z0-9]+\.[a-z0-9-]+$/

We had ${srv}-v[0-9]*

I can't test if works because I have problems with me maistra cluster

Sample node to parse => sidecar~172.17.0.22~ai-visitors-779956ddf8-s78gr.default~default.svc.cluster.local

We need ai-visitors-779956ddf8-s78gr.default

In normal cases reviews-v2(.bookinfo)

### DIFF
--- a/src/components/JaegerIntegration/JaegerResults/SpanTable.tsx
+++ b/src/components/JaegerIntegration/JaegerResults/SpanTable.tsx
@@ -72,8 +72,8 @@ export class SpanTableC extends React.Component<SpanDetailProps, SpanDetailState
       node = sp.tags.filter(tag => tag.key === 'node_id')[0].value;
     }
     const srv = sp.process.serviceName.split('.')[0];
-    const regex = new RegExp(`${srv}-v[0-9]*`);
-    const result = node !== '' ? regex.exec(node) : null;
+    const regex = new RegExp(`([a-z0-9-\\.]+)-[a-z0-9]+-[a-z0-9]+\\.[a-z0-9-]+`);
+    const result = node !== '' ? regex.exec(node.split('~')[2]) : null;
     if (result) {
       return (
         <Tooltip content={<>View logs of workload {result[0]}</>}>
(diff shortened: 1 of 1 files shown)


src/components/JaegerIntegration/JaegerResults/SpanTable.tsx
```diff
@@ -72,8 +72,8 @@ export class SpanTableC extends React.Component<SpanDetailProps, SpanDetailState
       node = sp.tags.filter(tag => tag.key === 'node_id')[0].value;
     }
     const srv = sp.process.serviceName.split('.')[0];
-    const regex = new RegExp(`${srv}-v[0-9]*`);
-    const result = node !== '' ? regex.exec(node) : null;
+    const regex = new RegExp(`([a-z0-9-\\.]+)-[a-z0-9]+-[a-z0-9]+\\.[a-z0-9-]+`);
+    const result = node !== '' ? regex.exec(node.split('~')[2]) : null;
```
jotak, line 76: in the following lines 79/81/82, it's `result[1]` that must be used and not `result[0]`, else we get the full name (like `ai-visitors-6b86864b-bn6fh.default`) instead of just the extracted desired part (like `ai-visitors`)

aljesusg: mmm yes...now this change ...but Is the workload `ai-visitors-6b86864b-bn6fh` in youre environment and the route '/namespaces/deafult/workloads/ai-visitors-6b86864b-bn6fh?tab=logs' ? In that case we must remove the .default but we need to check in the expression if there are multiples `.<word>`.

Let me work on this

jotak: the regex you wrote is fine. There's the parenthesis that allow to extract the first part. You just need to use result[1], I've tested it, it works fine.

jotak: PS my workload is simply `ai-visitors`

aljesusg: yes so here is the problem because result[1] will be ai-visitors-6b86864b-bn6fh.default and the link to workloads should be /namespaces/deafult/workloads/ai-visitors?tab=logs . Am I right?

jotak: no, `result[0]` is `ai-visitors-6b86864b-bn6fh.default` and `result[1]` is `ai-visitors`, so `result[1]` is fine

PR #1797, opened 2020-05-19
Re-org namespaces in Create Istio New Config

lucasponce:
Changed the order of namespaces field:
image

Added "canCreate" validation on "isFormValid" to not enabled the create button if user has not rights.

Fixes https://github.com/kiali/kiali/issues/2762


lucasponce:
> I'd prefer more visibility of the notification, but this is all good anyway.
I think that's because it's an "info" notification.
We can move to warning and then it will show up as you comment.

hhovsepy:
Post merge verification:
Screenshot from 2020-05-20 16-15-56

PR #1798, opened 2020-05-19
Labels filtering: click on labels to filter

aljesusg:
Fixes https://github.com/kiali/kiali/issues/2735

## Demo
ezgif-4-4aee51cba4a5

I need to do it for OverviewPage too but I want to sent the PR for start reviewing it.
What do you think @jotak ? I didn't see another way to do it than use refs.

### DIFF
--- a/src/components/Filters/CommonFilters.ts
+++ b/src/components/Filters/CommonFilters.ts
@@ -4,7 +4,7 @@ import {
   FILTER_ACTION_UPDATE,
   FilterValue,
   ActiveFiltersInfo,
-  FilterTypes
+  FilterTypes,
 } from '../../types/Filters';
 import { HEALTHY, DEGRADED, FAILURE, NA, Health } from '../../types/Health';
 import { removeDuplicatesArray } from '../../utils/Common';
@@ -12,12 +12,12 @@ import { removeDuplicatesArray } from '../../utils/Common';
 export const presenceValues: FilterValue[] = [
   {
     id: 'present',
-    title: 'Present'
+    title: 'Present',
   },
   {
     id: 'notpresent',
-    title: 'Not Present'
-  }
+    title: 'Not Present',
+  },
 ];
 
 export const istioSidecarFilter: FilterType = {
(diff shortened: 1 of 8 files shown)


jotak:
That's really nice!
Do you think it would be easy to implement the reverse mechanism also? ie. when you click a label that is active, it's removed from filters?

aljesusg:
> That's really nice!
> Do you think it would be easy to implement the reverse mechanism also? ie. when you click a label that is active, it's removed from filters?
> If it's not trivial no problem - we could perhaps open a new task for enhancements.

Yes we can do it... I was thinking about to add a tooltip withj soemthing like "Remove filter" or "Add filter"

aljesusg:
> > That's really nice!
> > Do you think it would be easy to implement the reverse mechanism also? ie. when you click a label that is active, it's removed from filters?
> > If it's not trivial no problem - we could perhaps open a new task for enhancements.
> 
> Yes we can do it... I was thinking about to add a tooltip withj soemthing like "Remove filter" or "Add filter"

The problem here @jotak is that when user set label app:details this is going to match all labels "contain" details so if we have the label app:details-v1 activated and the user click I think that the behaviour should be something like...if there is a label app:details-v1 applied so the user can remove it, cursor change and there is a tooltip....otherwise no cursor, no tooltip.

What do you think?

aljesusg:
## behaviour
# Add Filter
Tooltip: Add label to Filters
Cursor: Pointer
add
# Remove Filter
Case: (The filter key:value is in filters)
Tooltip: Remove label from Filters
Cursor: Pointer
Remove
9379-ce8130c5ee2d.png)

What do you think @jotak ?

aljesusg:
🔍  Finding changed files since git revision f0ba580e.
🎯  Found 6 changed files.
🔍  Examining package.json.
🔍  Examining src/components/Filters/StatefulFilters.tsx.
🔍  Examining src/components/VirtualList/Config.ts.
🔍  Examining src/components/VirtualList/Renderers.tsx.
🔍  Examining src/components/VirtualList/VirtualItem.tsx.
🔍  Examining src/components/VirtualList/VirtualList.tsx.
✅  Everything is awesome!
Done in 0.91s.


:'(

aljesusg:
ok update dep...can we add some test for this ?

lucasponce:
Please, @aljesusg this requires rebase, thanks

aljesusg:
Rebased

jotak:
@aljesusg hmm I didn't realize that it wasn't exact matching. It's fine to do as you suggest.

PR #1800, opened 2020-05-20
Refactor Service|Workload|App Details

jotak:
Goal: better separation of concerns, smaller files.
Fetching backend is now performed by subcomponents. E.g. traces-related component will fetch traces. They also own their action bar.

Also:
- Updated prettier to 1.19.1 so that it works with optional chaining
- Renamed JaegerErrors => JaegerError
- Moved fetch-related code for tracing into a dedicated class
- Make wider use of private functions in components (it's not that important in terms of API since, anyway, component's methods are never directly called from outside; but it's valuable to spot unused methods)
- Thanks to the above, remove dead code
- Add some helper methods to fetch minigraph/traffic data per app/service/workload with some default params

Backend PR: https://github.com/kiali/kiali/pull/2822

Fixes https://github.com/kiali/kiali/issues/2732 and https://github.com/kiali/kiali/issues/2720


jotak:
I still need to update the unit tests, but it's ready for review / test

@aljesusg @xeviknal @israel-hdez I've put several reviewers as it touches several areas: Traffic details / Minigraph, Traces, App/Service/Workload info

@hhovsepy for testing, this is mostly for non-regression (related to the areas mentioned above). The only expected change would be this issue being fixed: https://github.com/kiali/kiali/issues/2732.

jotak:
@israel-hdez I'm also seeing it. Investigating ...

Reviewers: i've split in 2 commits to help review, the second one is exclusively about tests.

xeviknal:
LGTM! I did a couple of reviews into the code, however I might missed something. I did download it on my local and did some tests. Looks alright. We have couple of weeks to harden this one (if needed). Looks really good to see duplicated code going away. Necessary.

I'll add my approval after test goes green :+1:

jotak:
DNM to investigate issues reported by Edgar

jotak:
@xeviknal @israel-hdez fixed, perhaps not the cleanest fix, but it seems to work. I added the graph data as a state in minigraph, which acts as a flag for change detection (but the whole datasource in props is still the actual data holder). Let me know if you'd prefer another way.

jotak:
thank you @hhovsepy  @xeviknal @israel-hdez !

PR #1801, opened 2020-05-21
Validation reference link broken

xeviknal:
AuthorizationPolicy and ServiceMeshPolicy had a bad url definition.
The Istio Object list is using now same url than defined in the config file.


lucasponce:
@xeviknal worth to backport to 1.18 ?

src/components/VirtualList/Renderers.tsx
```diff
@@ -44,7 +44,7 @@ const getIstioLink = (item: TResource) => {
     const subtype = type === 'adapter' ? item['adapter']!.adapters : item['template']!.templates;
     to = to + '/' + istioType + '/' + subtype + '/' + name;
   } else {
-    to = to + '/' + dicIstioType[IstioTypes[type].name] + '/' + name;
+    to = to + '/' + IstioTypes[type].url + '/' + name;
```
lucasponce, line 47: Question, as this "to" field is used in other context, are we sure is not going to affect to other place ?
It sounds that the IstioTypes.url is doing the same of dicIstioType[] but just checking if there is anything missing here.

xeviknal: You are right. There are different approaches for rendering urls to istio object details page. In the latest commit, I unified this. There is one single source of truth. This way, it is less error-prone for future modifications/additions.

lucasponce:
@hhovsepy worth if the last changes on this one can be re-tested, as it affects to more pages.

xeviknal:
@lucasponce no need for backporting since it doesn't give much value to users.

PR #1805, opened 2020-05-21
Iter8 0.1.0.1

jadeyliu939:
** Describe the change **
1. Add Auto Refresh for Iter8 Detail page
2. Add Pause, Resume, Terminate with Success, Terminate with Failure for experiment action
3. Add Estimated experiment duration at the Experiment Create and Detail page
4. Add Shortcut to Traffic and Outbound metric from Detail page

** Dependency **
This PR depends on  [Kiali PR# 2815](https://github.com/kiali/kiali/pull/2815)

** Issue reference **
[Main issue #2007](https://github.com/kiali/kiali/issues/2007)
and 
[Issue #2633](https://github.com/kiali/kiali/issues/2633)

** Screenshot **
1. Add "Estimate Experiment duration" in Experiment Create and Detail page

Screen Shot 2020-05-21 at 3 39 44 PM

Screen Shot 2020-05-21 at 3 41 01 PM

2. Add New Actions

Screen Shot 2020-05-21 at 3 40 10 PM

3. ShortCut to Traffic and Outbound Metric view
Screen Shot 2020-05-21 at 4 30 45 PM


src/pages/extensions/iter8/Iter8ExperimentDetails/ExperimentCreatePage.tsx
```diff
@@ -546,22 +561,27 @@ class ExperimentCreatePage extends React.Component<Props, State> {
...
                     fieldId="candidate"
-                    label="Candidate"
+                    label="Select Candidate"
                     isRequired={true}
                     isValid={this.state.experiment.candidate !== ''}
                     helperText="The candidate deployment of the target service (i.e. reviews-v2)"
                     helperTextInvalid="Candidate deployment cannot be empty"
                   >
-                    <FormSelect
+                    <TextInput
```
lucasponce, line 570: Is this on purpose ?
It changes a little bit the both components:
image
Baseline is fixed to workloads but Candidate can enter any text.

I guess both should be use same model, if free format is included, I'd also add a validation on enter.

lucasponce: Note, that a side effect of this is:
image

When the candidate is populated, then it filters the list of possible workloads, I thought it was a bug but I saw the change of the component.

To see all workloads, user may delete some text then (I've tested with the reviews service).

lucasponce:
I have a side comment that I caught meanwhile testing this PR.

I've created an experiment:
image

But it seems iter8 would need to update to Istio 1.6 as metrics have changed.

image

Perhaps iter8 would need to detect the Istio backend ?

cc @kiali/iter8

lucasponce:
@jadeyliu939 question, when I terminate an experiment, what is the expected result in the phase ?

I'm having this result:

image

And I'm still seeing the experiment in "phase: In progress"

jadeyliu939:
When issuing Action - `pause, resume, terminate with ...`. It might take a while to reach the desired state. It the action did failed, an error should be shown. Did your terminate action eventually processed, or did the terminate just get ignored? If that is the case, it is a bug, I will look into it. If it is just taking time to process it, maybe we need to display some messages?

src/pages/extensions/iter8/Iter8ExperimentDetails/ExperimentCreatePage.tsx
jadeyliu939: Yes, the candidate is intended to be selectable from the existing workload version, or user can type-in non-existing one. But baseline does not make sense to have non-existing workload.

As for user need to delete the pre-filtered one to show the whole list,  I will change the behavior so that the field will be without default ( instead default to the first one in the list). 

what do you think?

lucasponce:
After installing the proper v2 I'm still seeing a problem when trying to define an experiment:
image

@jadeyliu939 @kimikowang any idea if I'm doing something wrong ?

jadeyliu939:
This is the difference between v1 and v2 for the Tolerance value. In the create page, you use the default 0.2 when adding assessment criteria, which is in unit second for v1. If you are using v2, you  will need to specify 200 ( unit : ms). For temp solution, we can add a checkbox to ask user which version they are using, but ultimately it would be better if Kiali can detect the version automatically, so the UI can use second as unit, and do the conversion based on v1 or v2.

lucasponce:
@jadeyliu939 Kiali already detects v1/v2, let me to make this info accessible in a standard way that extension can use and then we don't need to add that extra component and probably we can simplify the rendering.

kalantar:
@lucasponce can you point us at the way Kiali makes this determination? I don't think we have figured this out yet and it would be nice to apply to non-Kiali use cases as well.

lucasponce:
@kalantar sure
https://github.com/kiali/kiali/blob/master/status/versions.go#L335

I'm preparing a PR to expose this info to the UI, but in the backend that variable can be checked.

lucasponce:
Related 
https://github.com/kiali/kiali/pull/2846
https://github.com/kiali/kiali-ui/pull/1812

In backend, there is already a flag that can be queried, and with those PRs we publish that flag to the frontend to enable render logic (i.e. units s -> ms) or other similar logic.

lucasponce:
Referenced PRs with the telemetry v2 are already merged, so these ones can be rebased and use that to determine if v1/v2 and provide some optional rendenring for those scenarios.

prachiyadav:
Screenshot from 2020-06-04 15-06-58
Screenshot from 2020-06-04 15-05-51

prachiyadav:
Selected Resume not able to see the change in status.

jadeyliu939:
@prachiyadav You can select resume, but since the network problem is still there, it did retry but failed again. (Please check if the analytics services is up and running off port 8080 )

lucasponce:
Hi @jadeyliu939,

As it happens with the backend PR https://github.com/kiali/kiali/pull/2815 this UI PR has a lot of commits mixed.

I wonder if you rebase with master and create a PR with just the feature.

@prachiyadav and me we can give a final review and try to merge and consolidate on master.

If there is anything pending we can interate with minor PRs on master.

jadeyliu939:
Close this PR base on suggestion from @lucasponce  A new PR will be created soon.

jadeyliu939:
NEW PR #1829  created

PR #1808, opened 2020-05-26
Format with prettier after the upgrade :(

aljesusg:
I passed the prettier in all the project,

It's annoying for each PR pass prettier and only add the files that you changed...After that we need to reset the HEAD to be like master and not merge the files between branches.

**_Note: After upgrade lint or prettier run it please and update the files_**

Added rules

```json
  "printWidth": 120,
  "singleQuote": true,
  "trailingComma": "none"
```

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,4 +1,6 @@
 {
+  "arrowParens": "avoid",
   "printWidth": 120,
-  "singleQuote": true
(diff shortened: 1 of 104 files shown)


mtho11:
Something could be said about having the trailing comma, as a change at the end only affects one line instead of two. But no big deal.

PR #1810, opened 2020-05-27
Separate jaeger queries and kiali tracing queries

jotak:
Use TracingQuery class when querying kiali backend, so that kiali-ui is decoupled from jaeger query API.
+ couple of other adjustments:
- remove some url params, such as tags (tags are derived from other params)
- stricter typing on query params
- incremental fetch less ambiguous now that params are decoupled from jaeger fetch params

Fixes https://github.com/kiali/kiali/issues/2732

Backend PR required: https://github.com/kiali/kiali/pull/2835


jotak:
@hhovsepy this is a follow-up on https://github.com/kiali/kiali-ui/pull/1800 : here again, no functional change is expected. Impacted areas for non-regression are tracing pages and, to a lesser extent, spans overlay (on metrics page).

PR #1812, opened 2020-05-29
Expose Istio Telemetry Version to Public Config

lucasponce:
Counterpart of https://github.com/kiali/kiali/pull/2846 to have access to the canonicalMetrics flag to check if telemetry is based on v1/v2 metrics.

### DIFF
--- a/src/config/ServerConfig.ts
+++ b/src/config/ServerConfig.ts
@@ -61,7 +61,8 @@ let serverConfig: ComputedServerConfig = {
     globalScrapeInterval: 15,
     storageTsdbRetention: 21600
   },
-  durations: {}
+  durations: {},
+  canonicalMetrics: true
 };
 computeValidDurations(serverConfig);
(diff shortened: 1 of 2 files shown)


jshaughn:
> Counterpart of kiali/kiali#2846 to have access to the canonicalMetrics flag to check if telemetry is based on v1/v2 metrics.

@lucasponce This is not a safe indicator.   Telemetry V2 is available as early as Istio 1.4.  And Telemetry V1 can be used in V1.5 and V1.6.  Canonical fields were introduced in 1.5.  It is quite possible to have canonical fields and Telem V1.   It is better to use the methods we have in versions.go:

`IsMixerDisabled()` will tell you if it's telemetry V1 or V2
`AreCanonicalMetricsAvailable()` will tell you if canonical fields are available.

You may still need something in config, I'm not sure, but I just want to make sure the situation is well understood.

PR #1821, opened 2020-06-08
Fix bug in Hide logs when backspace or delete is used to reset 

mtho11:
Fix bug in Hide logs when `backspace` or `delete` is used to reset the hide filter. 
Note: using the 'x' already does the proper resetting of the hide filter.

** Issue reference **
fixes: https://github.com/kiali/kiali/issues/2878

** Screenshot **

hide-fix


mtho11:
@hhovsepy I fixed the #1 issue but ignored the #2 issue because this screen has already been redesigned (and will have two rows of toolbars). So that change will come with my next PR.

PR #1822, opened 2020-06-10
Remove subtypes. Add Instances/Handlers

lucasponce:
This PR removes the old "subtypes" used for adapter/templates.
In short, in Istio 1.0 there were some legacy mixer adapters that populated a lot of new interfaces.

We managed that in Kiali as logical "subtypes" of adapters/templates to simplify this view to the user.

Now, Istio manages those objects as standard handlers/instances, so we put them as regular object and clean some of the technical debt introduced in the first Kiali versions.

This PR requires the backend:
https://github.com/kiali/kiali/pull/2881

How to test this:
- Check regressions in all related Istio config existing features.
- Check that Handlers and Instances are showed up as normal Istio Config objects (3scale extension can be used for that easily).

### DIFF
--- a/src/components/Link/IstioObjectLink.tsx
+++ b/src/components/Link/IstioObjectLink.tsx
@@ -12,21 +12,11 @@ interface Props {
   query?: string;
 }
 
-export const GetIstioObjectUrl = (
-  name: string,
-  namespace: string,
-  type: string,
-  subType?: string,
-  query?: string
-): string => {
+export const GetIstioObjectUrl = (name: string, namespace: string, type: string, query?: string): string => {
   const istioType = IstioTypes[type];
   let to = '/namespaces/' + namespace + '/' + Paths.ISTIO;
 
-  if (!!subType) {
-    to = to + '/' + istioType + '/' + subType + '/' + name;
-  } else {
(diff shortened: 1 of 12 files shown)


lucasponce:
> Created 3scale handlers are not listed, neither in 3scale Config menu nor in Service Wizard.

You are absolutely right, I fixed locally but I forgot to push it.

src/components/Link/IstioObjectLink.tsx
```diff
@@ -12,21 +12,11 @@ interface Props {
...
+export const GetIstioObjectUrl = (name: string, namespace: string, type: string, query?: string): string => {
   const istioType = IstioTypes[type];
   let to = '/namespaces/' + namespace + '/' + Paths.ISTIO;
 
-  if (!!subType) {
-    to = to + '/' + istioType + '/' + subType + '/' + name;
-  } else {
-    to = to + '/' + istioType.url + '/' + name;
-  }
+  to = to + '/' + istioType.url + '/' + name;
```
jotak, line 19: minor, but now without the condition on subtype, the two lines above could be merged

hhovsepy:
Re-verified.

PR #1825, opened 2020-06-15
In-graph tracing - graph overlay

jotak:
WIP (early draft PR to allow iterative reviewing)

POC for task https://github.com/kiali/kiali/issues/2841


jotak:
This is a very early draft to just setup the basis of the work.
Service nodes that belong to a trace are displayed with a different color.

Capture d’écran de 2020-06-15 16-49-12
(Here, "ball" and "ui").

Missing stuff:
- Workload recognition
- Edges recognition
  => with the above two, we could have here the trace "ball => v1 => ui => v1" correctly highlighted
- Style definition (border color? background color? boldness?)
- How do we clear a trace from graph?

jotak:
@jshaughn I started with the smallest possible first step so that it's easy for a first review - especially to know if the changes hooked in graph code are correctly placed.

src/pages/Graph/GraphPage.tsx
```diff
@@ -294,6 +298,25 @@ export class GraphPage extends React.Component<GraphPageProps, GraphPageState> {
...
+            if (trace) {
+              this.setState({ trace: trace });
+            }
+          }
+        })
+        .catch(err => {
+          // TODO
+          console.log(err);
+        });
+    }
```
jshaughn, line 319: I think to be more consistent, if we are going to support a "traceId" query param in this class we would make it a redux property and let the URL setting override any redux setting, and we typically handle that in the constructor.   I would probably defer fetching the trace to componentDidUpdate.  You probably want the graph to be ready before fetching the trace, although I suppose it's ok to fetch the trace concurrently even if you have to wait to apply it to the graph.

jotak:
@jshaughn I'm fine to let you / team continue on this. At least I provided a starting point for how to deal with traces :-)

src/components/CytoscapeGraph/graphs/GraphStyles.ts
```diff
@@ -428,6 +428,9 @@ export class GraphStyles {
     };
 
     const getNodeBorderColor = (ele: Cy.NodeSingular): string => {
+      if (ele.data('spans')) {
+        return PfColors.Blue700;
+      }
```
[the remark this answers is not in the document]
jotak: yes, that was mainly for POCing, but styles need to be better defined

jotak:
Last commit: some reworking on how trace selection is managed, now with redux

jshaughn:
@jotak I think I have some time opening up, would it be useful if I were to help out on this one?

jotak:
@jshaughn sure. I'll probably update this poc once again today to include edges and workloads in the trace, but feel free to jump in. Do you want to start a fresh new PR, or continue based on this one?

PS: the commits list in this PR are mainly due to having rebased it on https://github.com/kiali/kiali-ui/pull/1837 ; at the time of writing, only the last commit is for this PoC.

jotak:
I think I can say the PoC is finished
It's taking shape:

Capture d’écran de 2020-07-08 12-07-53

What needs to be done:
- Better / proper styling
- Use cy selectors
- Eliminate the couple of TODOs remaining
- Whatever you think is missing...

src/components/CytoscapeGraph/CytoscapeTrace.ts
```diff
@@ -0,0 +1,42 @@
+import * as Cy from 'cytoscape';
+import { CyNode } from './CytoscapeGraphUtils';
+import { NodeType } from '../../types/Graph';
+import { JaegerTrace, Span } from 'types/JaegerInfo';
+
+export const showTrace = (cy: Cy.Core, trace: JaegerTrace) => {
```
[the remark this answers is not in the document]
jotak: +1
I guess that, with your suggested approach, we don't have to build these intermediate maps as I'm doing here. We can just iterate over the spans and apply the class to the relevant nodes and/or edges via selectors. So it should simplify this code a lot.
About using class or data: I had in mind to show the number of spans associated with nodes and/or edges, within the trace. It can be more than 1. If we do this, I guess we would have to store it in data. But we can do that later; having just the highlight would be a great first step.

jshaughn:
@jotak If it sounds OK I'm going to grab this and put together a PR against it with some of the suggestions I had above.

jotak:
+1 sure!

src/components/CytoscapeGraph/CytoscapeTrace.ts
```diff
@@ -0,0 +1,95 @@
...
+    return;
+  }
+
+  cy.startBatch();
+  hideTrace(cy);
+
+  trace.spans.forEach(span => {
+    const split = span.process.serviceName.split('.');
+    const service = split[0];
+    // TODO: add ns when missing
```
jotak, line 21: I think we can remove this "todo"; if namespace isn't part of trace's service name, we have no reliable way to guess it, so I think how you did is fine: select node only from service.
This should never happen with envoy-generated traces, but it may happen with user-defined traces

jshaughn: +1, I left it there but I didn't know what else to do to be able to guess a valid namespace.

jshaughn:
@jotak If you want to promote this from Draft it seems reasonable to me...

jotak:
@jshaughn so, this feature is still under a "xflags" umbrella and can be merged before the epic is done. Is it fine for merging? Or do you want another pair of eyes?

The next task is this one, I'll start it this week: https://github.com/kiali/kiali/issues/3010 - once there's a PR for that one, I think it will be ready for QE testing the whole and removing the "xflag".

src/components/JaegerIntegration/JaegerHelper.ts
```diff
@@ -40,3 +40,30 @@ export const getTimeRangeMicros = () => {
     to: boundsMillis.to ? boundsMillis.to * 1000 : undefined
   };
 };
+
+const workloadFromNodeRegex = new RegExp(`([a-z0-9-\\.]+)-[a-z0-9]+-[a-z0-9]+\\.([a-z0-9-]+)`);
```
jshaughn, line 44: Unchanging regexp is better as regexp literal:

`const workloadFromNodeRegex = /([a-z0-9-\\.]+)-[a-z0-9]+-[a-z0-9]+\\.([a-z0-9-]+)/;`

jotak:
>     * handle operation node issue

I'm not sure how we could handle that. Actually, I think the current situation is not too bad:

See here, the "Display" operation node is splitting the trace:
Capture d’écran de 2020-07-22 11-53-24

I can't see how to join the two parts in a reliable way [1]. But I think it's still possible to reason about that trace when seeing these parts. That's why I think it's currently not too bad.

[1] Perhaps we could read all EnvoyFilter configurations to "reverse engine" the operation node conception, and match that with the path that we've got in the trace info ... but I don't think it's worth it. It's complicated and perhaps not very reliable for small outcome.

jotak:
@jshaughn last commit changes:

- When an item is clicked on, after being fetched and stored in redux, corresponding item in list is refreshed (it may have grown in the meantime)
   => I did that because I was surprised that the trace info displayed in list didn't always match what I was seeing on graph... that's because traces are mutable in jaeger and they may grow over time.

- When the list of traces is reloaded, keep current selected trace accessible
   => so that a refresh doesn't make loosing context

- Adding a way to unselect trace, by clicking again on list item (like toggle button)
   => There was already the possibility to click wherever on graph, changing/clearing node selection etc. So this alternative way by clicking on list items adds up to that.

jotak:
>     * possibly revisit the "keep list" feature
> 

If you don't mind, for that one, I'd like to split up the work; task created: https://github.com/kiali/kiali/issues/3016

jshaughn:
> I can't see how to join the two parts in a reliable way [1]

You're right, it's not too bad and it can be resolved by just not displaying operation nodes.  So I agree, I think we can just ignore that for now, perhaps add a FAQ if necessary.

> If you don't mind, for that one, I'd like to split up the work; task created: kiali/kiali#3016

Sounds good.

> @jshaughn last commit changes:

All good, I think.

jotak:
Thanks @jshaughn for the help!

PR #1826, opened 2020-06-15
Navigation to other detail pages using mini-graph

israel-hdez:
Double clicks on mini-graph card is enabled. When double-clicking, the mini-graph will redirect the user to the detail page of the selected node. No redirection is done if the detail page of chosen node is the one already being displayed. Also, nodes with no detail page (like service entries and inaccessible nodes) are ignored.

A small refactor had to be done to the CytoscapeGraph component. The double-tap event handlers were pulled up to the GraphPage component in order to be able to perform a different action depending on the page hosting the CytoscapeGraph component.

Fixes kiali/kiali#2790.

### DIFF
--- a/src/components/CytoscapeGraph/CytoscapeGraph.tsx
+++ b/src/components/CytoscapeGraph/CytoscapeGraph.tsx
@@ -1,7 +1,6 @@
 import * as Cy from 'cytoscape';
 import * as React from 'react';
 import ReactResizeDetector from 'react-resize-detector';
-import history from '../../app/History';
 import Namespace from '../../types/Namespace';
 import { GraphHighlighter } from './graphs/GraphHighlighter';
 import TrafficRender from './TrafficAnimation/TrafficRenderer';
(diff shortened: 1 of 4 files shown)


lucasponce:
@israel-hdez thanks a lot, this is really great.
I have only one comment: In the minigraph I'd consider to use a single click instead to double click to navigate.

The double click on this area is not obvious and I think with a single click the experience would really good as nodes in the minigraph can be seen as regular links as it happens on these details pages.

jshaughn:
>  I'd consider to use a single click instead to double click...

@lucasponce I'm not against it if @israel-hdez agrees with you, but I'd point out that single-click never navigates away in the main graph, it only selects.  Double-click is the typical way to drill into a node, either to navigate to another namespace or to display the node-detail graph.

israel-hdez:
@lucasponce I was preferring double-click for consistency with the main graph. But if you think this will be too hidden, I can switch to single-click.

israel-hdez:
@lucasponce Just let me know so that I update the PR :)

lucasponce:
I may be agree if the minigraph also selects some parity with the main graph, but it's not the same.
On Minigraph context, the "select" operation is similar to navigate into the other detail page.
Having a double click behaviour when single click is not used looks not obvious to me.

To me, as it happens in the main graph, when I select I see the details of the specific node, and drill down is another operation on the node.

But anyway, I think it's something we can test how it works with a single click, to me it was the preferred way to navigate in this context as I see the minigraph as a set of links like the others.

israel-hdez:
@lucasponce to some extent, I agree.
OK, lets try single-click. I'll update the PR.

israel-hdez:
@lucasponce Updated and now using single-click.

I kept the refactor of the double-tap. I think it's worth to keep.

PR #1830, opened 2020-06-18
Actions on Overview page

lucasponce:
This PR is a preliminary refactoring of Overview page to introduce actions per Namespace level.

Before to add new actions we are going to move the existing links into "Show..." actions to maintain consistency with other cards and actions in Graph and details pages.

Also, when actions are introduced in cards, that have triggers some change sin the existing views:

image

image

image

Details of the discussion have been tracked on https://github.com/kiali/kiali/issues/2896.


lucasponce:
Note that with this refactoring, the Ui is prepared to discuss more actions like the main goal of kiali/kiali#2896

image

src/pages/Overview/OverviewPage.tsx
```diff
@@ -387,10 +413,81 @@ export class OverviewPage extends React.Component<OverviewProps, State> {
...
+        destination = `/${Paths.ISTIO}?namespaces=` + namespace;
+        break;
+      default:
+      // Nothing to do on default case
+    }
+    history.push(destination);
+  };
+
+  showActions = (): any[] => {
+    const namespaceActions: any[] = [
```
jotak, line 441: You created an exported type `OverviewNamespaceAction` somewhere else, isn't it this type that is used here ; and for  the other "any" here and there?

jotak:
A side note that is not strictly related to your change but I want to say it again : I don't like having the available links / actions changing depending on the currently selected entity type ; I think it would be "less surprising" to keep a consistent list of actions (workloads, apps, services) even if that same workloads/apps/services link is displayed elsewhere in the card. To me, the current behaviour is counter-intuitive when you look for actions in the menu: "why isn't it displayed?" "Oh yes, that's because it's at another place in this specific case!"

lucasponce:
@jotak +1 on your comment, I just moved the same icons behaviour to the links, but I agree that now we should put all to gain consistency.

src/components/VirtualList/VirtualItem.tsx
```diff
@@ -11,6 +12,7 @@ type VirtualItemProps = {
   index: number;
   config: Resource;
   statefulFilterProps?: React.RefObject<StatefulFilters>;
+  action?: any;
```
jotak, line 15: Actually, after looking more closely, here it seems it could be JSX.Element

src/components/VirtualList/VirtualList.tsx
```diff
@@ -22,6 +22,7 @@ type VirtualListProps<R> = {
   rows: R[];
   sort?: (sortField: SortField<NamespaceInfo>, isAscending: boolean) => void;
   statefulProps?: React.RefObject<StatefulFilters>;
+  actions?: any[];
```
jotak, line 25: ... and JSX.Element[]

src/components/VirtualList/VirtualItem.tsx
lucasponce: Yes, and probably I've left a couple of more `any` that I should change in the same way.

lucasponce:
@jotak Ok, I think I've addressed your comments, now actions will show all links, removing the conditional rendering:

image

Also, I think I've fixed the `any` types used.

lucasponce:
@jshaughn 

> maybe remove the duplicate links and just defer to the kebab

I don't think it hurts, I would find weird to not have an applications link. So, I'd keep the link.
Also, in labels there is a link that doesn't belong to the Kebap and it's consistent.

> capitalize "Labels"

+1

> maybe remove the leading numbers on this, so just:

I'm not sure, because we want to keep the link, if we move to the format 

```
Label: 1
Applications: 6
```

The link won't be clean, so I think `1 Label` or `6 Applications` are better for a link, also the icons show a split between state, at the moment you have 2 in Iddle, 2 healthy and 2 in error, you miss the whole `6 Applications` link.

So, I will keep the `6 Applications` format.

> move the lock and config validation below the namespace name (take from sparkline height if necessary)

Not sure about this, because it would make always the height variable. I'm +1 with your next comment to remove the `N/A` but keeping in the title sounds to me a better approach rather than increase a line. Othertiwise it affects to the sparkline aspect.

> instead of N/A just use a success icon

There is already a success icon, this is a different state that nothing is validated. But I'm fine to hide the `N/A` here on this case.

> Simplify the tooltip to just be.

+1 to simplify the success conditions

prachiyadav:
Screenshot from 2020-06-23 15-10-07
Screenshot from 2020-06-23 15-09-50
Screenshot from 2020-06-23 15-08-33

lucasponce:
I've tried to combine all feedback and finding a tradeoff:

image
image
image

In summary:
- Labels link is important because it allows user to get detail of all namespace labels and perform additional filtering.
- "X Applications" format is maintained, but removed the redundant link "Show ..." defined in actions.
- Specific filtering links for iddle/failured are maintained as there are not specific for those in actions.
- Some styling to align better icons.
- Simplfy wording on validation tooltip cc @xeviknal, it's minor, just changing text.

lucasponce:
Thanks, I think I've addressed your comments @jotak, let me know if something is pending and I can prepare an additional followup PR.

I'll consolidate this to avoid conflicts.

PR #1833, opened 2020-06-25
Workload detail actions + 3scale extension changes

lucasponce:
Fixes https://github.com/kiali/kiali/issues/2860
Fixes https://github.com/kiali/kiali/issues/2870

Requires https://github.com/kiali/kiali/pull/2936

In summary:
- Remove 3scale extension a from main navigation menu as 3scale is going to be configured using Istio Mixer resources: Handler, Instance, Rules.
- Create an specific action menu for 3scale config under Istio Config page.
- Introduce a 3scale config page that maps 3scale concepts into Istio Mixer config (3scale account and authorization).
- Move 3scale link action from Service to Workload details.
- Remove 3scale specific models.

How to test:

- When 3scale extension is disabled:
```
extensions:
  threescale:
    adapter_name: threescale
    adapter_port: 3333
    adapter_service: threescale-istio-adapter
    enabled: false
    template_name: threescale-authorization
  iter_8:
    enabled: false
```

There should not be a 3scale action in the Istio Config List -> Actions 
image

And Workload details shouldn't show any Action:
image

(Note, while writting this I see that this PR is not complete, it should disable/hide workload actions if there is none)

- When 3scale extension is enabled:
image

3scale navigation menu doesn't exist anymore, now it's implemented with a new Istio Contig List -> Actions -> Create 3scale Config form

A Create 3scale Config form allows to create 3scale Accounts mapped with a Istio Handler resource:
image

And 3scale Authorization mapped with a pair of Istio Instance+Rule:
image

Business rationale about this is detailed in the Issue https://github.com/kiali/kiali/issues/2860

At the end this form creates standar Istio Mixer objects wtith the 3scale configuration mapped into them:
image

A "3scale credential" is a convention name to refer a pair of Istio Instance+Rule previously created in the form.

Now, under the Workload Details there is a new action to link a workload with 3scale:
image

The user needs to introduce a "3scale service id" identifier which is previously generated in 3scale system and user needs to know:
image

When workload is "linked" with 3scale, additional labels will be added to the workload and propagated into Pods:
image

(Note that a re-start of Pods is expected)

When a workload is detected to be "linked" with 3scale (checking the 3scale labels), then the "unlink" operation appears:
image

Note that when modifying the labels and during the Pods restart, Kiali can fetch that situation:
image

So, this PRs refactor partially how Workload is refreshed.

One known defect/pending issue to address is to refactor the Logs tab to allow to refresh the workload at tab level and move the controls of refresh in the same location that others tab:

image

See, that the logs tab has the old pod container instead the new one, the workaround is to refresh the page.
As this PR is growing, perhaps that change can be done in a following PR, but it's a topic to comment.


xeviknal:
I've done a quick test to what you have so far. It works as expected.
A couple of thinks caught my eye:

1. Labels without value have an strange representation
Screenshot of Kiali Console (9)

2. You mentioned in your PR, the dropdown with empty list looks a bit weird now:
Screenshot of Kiali Console (7)

3. Istio handler form inputs: should we use placeholders instead of actual text?
Screenshot of Kiali Console (10)

Not necessary to be approached in this PR. Nice to have.

lucasponce:
Thanks for testing @xeviknal !
Yeah, the empty label and the empty menu are in the radar, I hope to fix them during the week and update the PR.

I'm less user about the placeholder, just thinking in other forms that we are using the same way (here I'm more following the same pattern used).

hhovsepy:
Minor comment in addition to what @xeviknal said:
Unlinking/linking remembers the previous service Id and auto-fills.
Screenshot from 2020-07-07 18-19-45

hhovsepy:
Create Handler, Rule, Instance.
List the Istio configs, filter by Handler, Rule and Instance. so it shows only 3 created obejcts.
Click on details and then Delete the Instance.
It redirects back to the objects list.
Then click on and delete the Rule.
Error is shown after deleting and returning to list.
Screenshot from 2020-07-13 17-08-41

hhovsepy:
Happens only in automations, with incognito browser opened by webdriver.
The select dropdown values for 3scale Account is not loaded, it seems while changing "3scale Config" select value some JS action or something is not triggered. Might this page have differences in select elements compared to other wizards?
Screenshot from 2020-07-14 13-05-19

lucasponce:
> Create Handler, Rule and Link to Workload.
> Remove the Handler and Rule.
> The link in Workload still remains and the labels are remaining on Workload and Pod.

@hhovsepy note this is expected, Link/Unlink operation is related to workload labels but it's not doing any modification in cascade. 

So, I create the Rule/Handler/Instance and next I update Workload labels. If I remove the previous Rule/Handler/Instance then the previous Workload lables would remain until those are unlinked.

Note that this was discussed in the issue with the 3scale team:

https://github.com/kiali/kiali/issues/2860

User has visibility that he/she is dealing with Mixer objects and workload labels.

Also, due Mixer is deprecated and probably this integration will be migrated into a new architecture I don't think is worth to invest more validations logic (i.e. one think could be check if a workload is linked but those objects don't exist, today it's not yet supported/included in this work).

lucasponce:
> Error is shown after deleting and returning to list.

Ok, that error shouldn't appear, perhaps there should return to list when operation is completed (in theory it should work in that way, I'll investigate).

lucasponce:
I can't reproduce this one in my local environment:
https://user-images.githubusercontent.com/604313/87321161-2e0c1c80-c52c-11ea-9c86-3d576810fb83.png

is the error consistent ?

hhovsepy:
> I can't reproduce this one in my local environment:
> https://user-images.githubusercontent.com/604313/87321161-2e0c1c80-c52c-11ea-9c86-3d576810fb83.png
> 
> is the error consistent ?

Cannot reproduce this anymore.

hhovsepy:
Facing an inconsistency.
Go to config list. Do not select any namespace.
Create New 3scale config.
Type the name for Handler.
Create button is not enabled, but the "istio-system" namespace is selected by default.
Screenshot from 2020-07-21 14-47-32

Refreshing the page makes all work.

lucasponce:
@hhovsepy thanks for the testing, I think I've addressed most of the comments.
@xeviknal Please, let me know if you find anything else, I'm going to prepare the operator PR.

PR #1834, opened 2020-06-25
Fix test to determine OpenId/OAuth preparation

israel-hdez:
OAuthLogin class is being used for both openshift and openid login strategies. OpenShift passes always the `access_token` parameter. OpenId providers *may* pass the access_token parameter, but id_token is warranted to be present (because of the asked response_type).

This is fixing a blocked login if the OpenId provider is not passing the access_token.

Fixes kiali/kiali#2925

### DIFF
--- a/src/services/Login.ts
+++ b/src/services/Login.ts
@@ -101,7 +101,8 @@ class OAuthLogin implements LoginStrategy<unknown> {
       return AuthResult.FAILURE;
     }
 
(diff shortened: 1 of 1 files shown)


israel-hdez:
@pbajjuri20 Can you please check that login in OpenShift is still working?

pbajjuri20:
Tested and Passed.

israel-hdez:
Thanks @cfcosta & @pbajjuri20

PR #1837, opened 2020-06-30
Traces list in side-panel

jotak:
- Split file SummaryPanelNode => (SummaryPanelNode + SummaryPanelNodeMetrics)
- When traces are enabled & service is selected, show a tabbed view Traces/Metrics below contextual node info

Fixes https://github.com/kiali/kiali/issues/2840

### DIFF
--- a/src/app/History.ts
+++ b/src/app/History.ts
@@ -31,7 +31,8 @@ export enum URLParam {
   SHOW_AVERAGE = 'avg',
   SORT = 'sort',
   TO = 'to',
-  UNUSED_NODES = 'unusedNodes'
+  UNUSED_NODES = 'unusedNodes',
+  EXPERIMENTAL_FLAGS = 'xflags'
 }
 
 export interface URLParamValue {
--- a/src/components/JaegerIntegration/JaegerResults/FormattedTraceInfo.ts
+++ b/src/components/JaegerIntegration/JaegerResults/FormattedTraceInfo.ts
@@ -1,0 +1,31 @@
+import { JaegerTrace } from '../../../types/JaegerInfo';
+import moment from 'moment';
+import { formatDuration, formatRelativeDate } from './transform';
+import { isErrorTag } from '../RouteHelper';
+
+export type FormattedTraceInfo = {
+  name: string;
+  spans: string;
+  duration?: string;
+  errors?: string;
+  relativeDate: string;
+  absTime: string;
+  fromNow: string;
+};
+
+export const getFormattedTraceInfo = (trace: JaegerTrace): FormattedTraceInfo => {
+  const { traceName, duration, spans, startTime } = trace;
+  const numSpans = spans.length;
+  const mDate = moment(startTime / 1000);
+  const timeStr = mDate.format('h:mm:ss a');
+  const numErredSpans = spans.filter(sp => sp.tags.some(isErrorTag)).length;
+  return {
+    name: traceName ? traceName : '<trace-without-root-span>',
+    spans: `${numSpans} Span${numSpans !== 1 ? 's' : ''}`,
+    duration: duration ? formatDuration(duration) : undefined,
+    errors: numErredSpans > 0 ? `${numErredSpans} Error${numErredSpans !== 1 ? 's' : ''}` : undefined,
+    relativeDate: formatRelativeDate(startTime / 1000),
+    absTime: timeStr.slice(0, -3) + ' ' + timeStr.slice(-2),
+    fromNow: mDate.fromNow()
+  };
+};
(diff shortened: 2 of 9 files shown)


jotak:
New Metrics / Traces tabs WIP: 

Capture d’écran de 2020-06-30 12-24-02

Capture d’écran de 2020-06-30 12-24-05

jotak:
@jshaughn do you validate the approach? Tabs placed within node side panel, below the contextual service information like labels/health. I think it's better to keep this info even in traces tab, so that people still have some context.

jshaughn:
@jotak Looks exactly like I was expecting, including splitting up the logic into separate files.  Only thing I'm not sure about is the naming.  Not sure if I'd prefer "Traffic" as opposed to "Metrics" but that is minor.

jotak:
@jshaughn ok, I've renamed that tab.

New commit: shows traces list:
Capture d’écran de 2020-07-01 16-32-18

Also, there's a button at the bottom for advanced search:
Capture d’écran de 2020-07-01 16-32-23

aljesusg:
> @jshaughn ok, I've renamed that tab.
> 
> New commit: shows traces list:
> Capture d’écran de 2020-07-01 16-32-18
> 
> Also, there's a button at the bottom for advanced search:
> Capture d’écran de 2020-07-01 16-32-23

I would trip the names with a tooltip with the full name.
Maybe in the same line we can show information about the spans with a tooltip when mouseover.

jotak:
Also I need to figure out what's the best way to deal with refreshes ... should it be coupled with graph refresh or not? Problem with that is the trace list is very "volatile", there's most chances that it gets completely renewed on refresh, which could be a pain for user experience.

jshaughn:
> Also I need to figure out what's the best way to deal with refreshes ... should it be coupled with graph refresh or not?

If it's not tied to graph refresh do we have a potential issue with the traces syncing well with the graph?  If it's not much of an issue then perhaps a manual refresh button on the tab could work.

jotak:
> If it's not tied to graph refresh do we have a potential issue with the traces syncing well with the graph? If it's not much of an issue then perhaps a manual refresh button on the tab could work.

Or I thought about a small check box saying "Do not refresh traces list", or something similar, within that traces tab. When checked, refresh requests would be ignored. Else when unchecked, it is tied to graph refresh.

jotak:
> I would trip the names with a tooltip with the full name.
> Maybe in the same line we can show information about the spans with a tooltip when mouseover.

Yes I need to improve the styles a little bit. For some reason when I tried to add "text-overflow: ellipsis" it didn't work... I need to try harder :)

jshaughn:
@jotak I saw in a separate issue you have some struggles with fonts in the side panel.  I wanted to mention that we did a decent amount of work on side-panel l+f consistency.  We use 12px throughout, so please also apply that sizing.  You will see a variety of places in the existing code where that is set.

jotak:
@jshaughn yes I've seen that there's a variable for side panel font-size and I think I'm using it as needed.

jotak:
hey @jshaughn @aljesusg I've added an URL-based switch for experimental features: https://github.com/kiali/kiali-ui/pull/1837/commits/4d0746b98afe8095edadee07c2e8c934d507778f
This will allow to safely merge PRs such as this one before the whole epic is complete (or at least before the minimal subset of tasks necessary for the feature to make sense is complete).

With this, I think this PR can be reviewed / merged. Is it fine with you?
For testing, you'd need to add `&xflags=igt` to the URL.

jotak:
@jshaughn I'll do the suggested style changes before it's merged.
About "keep list" versus "refresh", not sure to understand your suggestion, would it be something like a check box "Decouple from graph" (put better phrasing here) + a refresh button that is displayed / enabled only when the checkbox is checked? If so, I'm fine to do it, as long as keeping tying with graph refresh is still an option.

Also, for this "Keep list", I used different phrasing like "Do not refresh" or "Pin list" before ending on "Keep list", but any suggestion is appreciated.

jotak:
@jshaughn I've added some styles as suggested (ellipsis + bottom margin), squashed & pushed.
Also, I realized that when PassthroughCluster node is selected it was showing (empty) traces tab as well ... I figure out this is what is called an "Inaccessible" service, right? So I added `    !nodeData.isInaccessible &&` check to what would trigger traces display. Does it sound correct?

jshaughn:
>  I've added some styles as suggested (ellipsis + bottom margin), squashed & pushed.
Looks great with the ellipse/tooltip, and the margin.

> About "keep list" versus "refresh", not sure to understand your suggestion
The limitation with "keep list" is that you either get the graph refresh or no refresh.  I was thinking that when "keep list" is enabled a manual refresh of the traces could be useful.  To do this currenly you would need to also refresh the graph.  So, I had suggested changing "keep list" to "manual refresh", and when "manual refresh" was enabled, a local refresh button would also be enabled.

This X-Feature is mergeable as-is imo.

jotak:
thanks for review @jshaughn !

PR #1842, opened 2020-07-03
In kiosk mode，the main header should not be shown

aljesusg:
Fix https://github.com/kiali/kiali/issues/2946

Screenshot from 2020-07-03 10-57-45


lucasponce:
Remember that Graph can browse to other pages.
Is this being tested with rest of the pages ?

xeviknal:
Yes, this had been test before :+1:

PR #1843, opened 2020-07-06
Upgrade Cytoscape to the latest version

cfcosta:
Closes https://github.com/kiali/kiali/issues/1826.

This PR should fix the problems we have with big meshes because of cytoscape. Did some testing with small/medium/large graphs, everything seems to be ok.

QE can test it by following what Guilherme did on his issue.


aljesusg:
Please test tippy components we had problems with the last update of PF4 here because PF4 use tippy 5 and we have tippy 3, not supported and can be affected in cytoscape popper

cfcosta:
@aljesusg sure! Can you give me some pointers on what to test/verify?

jshaughn:
@cfcosta I'm not sure how much Tippy is used in the graph proper but check things like hover, badging on labels, label placement, single-click, double-click, right click menus, and although tangential, the graph tour.

> This PR should fix the problems we have with big meshes 

That's sort of unproven but we're hopeful that the past issues were releated to hash issues that may have been resolved.  Either way this update can close the issue and new issues can be created if/when a user encounters a problem.

cfcosta:
>@cfcosta I'm not sure how much Tippy is used in the graph proper but check
>things like hover, badging on labels, label placement, single-click,
>double-click, right click menus, and although tangential, the graph tour.

Ok, will take a look. I remember the right menus and hovers being ok, but let me verify for this.

>That's sort of unproven but we're hopeful that the past issues were releated to
>hash issues that may have been resolved. Either way this update can close the
>issue and new issues can be created if/when a user encounters a problem

Yeah, there's always the chance what I tested did not trigger an edge case for a
different mesh, but at least I could not find any problems.

cfcosta:
@jshaughn just verified it here. Hovers, badging, label placement, single click, double click, right click menus, graph tour are all exactly the same as before.

Just as a reference, this is before:
image
And this is after:
image

jshaughn:
@pbajjuri20 Make sure to also try the different graph layouts.  Also, make sure to try graph hide.

pbajjuri20:
I have tested on a larger mesh, the issue is fixed perfectly. 

LGTM! 

mesh

cfcosta:
Thanks @pbajjuri20!

PR #1850, opened 2020-07-15
Do not show tabs list until async data is fetched

jotak:
Fixes https://github.com/kiali/kiali/issues/2996


jotak:
Hey @aljesusg , if you could have a look at the issue I just created. I tried different things but I could not have the ParameterizedTabs component display this kind of data on page refresh. I'm not sure if I'm missing something; this PR fixes the symptoms but it's more a workaround, not fixing the root cause. What do you think?

jotak:
thanks @hhovsepy

PR #1851, opened 2020-07-15
Traffic wizards - handle PeerAuthentications

xeviknal:
This PR's handles the fact of creating/editing/removing a PeerAuthentication while user is performing any traffic management wizard from the service details page.

When the user switch on the 'Add PeerAuthentication', the wizard will create a PeerAuthentication object with the selected dropdown. In case, the user switch off the selector, the wizard will remove that PeerAuthentication object.

The PeerAuthentication object it always apply to the workload in the same namespace and applying to the same 'app' label.
There is not option to indicate the tls mode at port-level. IMO, this scenario is a bit advanced.

peerauthn-wizard

Fixes https://github.com/kiali/kiali/issues/1381

### DIFF
--- a/src/components/IstioWizards/ServiceWizard.tsx
+++ b/src/components/IstioWizards/ServiceWizard.tsx
@@ -8,9 +8,10 @@ import MatchingRouting from './MatchingRouting';
 import WeightedRouting, { WorkloadWeight } from './WeightedRouting';
 import TrafficPolicyContainer, {
   ConsistentHashType,
-  TrafficPolicyState
+  TrafficPolicyState,
+  UNSET
 } from '../../components/IstioWizards/TrafficPolicy';
-import { DISABLE, ROUND_ROBIN } from './TrafficPolicy';
+import { ROUND_ROBIN } from './TrafficPolicy';
 import SuspendTraffic, { SuspendedRoute } from './SuspendTraffic';
 import { Rule } from './MatchingRouting/Rules';
 import {
@@ -19,6 +20,7 @@ import {
   getInitGateway,
   getInitHosts,
   getInitLoadBalancer,
+  getInitPeerAuthentication,
   getInitRules,
   getInitSuspendedRoutes,
   getInitTlsMode,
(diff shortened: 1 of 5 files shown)


lucasponce:
I will use this PR to introduce a formal new EPIC https://github.com/kiali/kiali/issues/3015 which is related.

I think it's time to work on how we can improve the link of Istio config under the Workload and Service details.

I'm missing some tabs there, and specially when the wizard creates new config (Gateway, Security, etc) it would be nice to have those info with new tabs.

Anyway, it's nothing to do with this PR, but related.

src/components/IstioWizards/IstioWizardActions.ts
```diff
@@ -389,63 +392,101 @@ export const buildIstioConfig = (
...
+          },
+          mtls: {
+            mode: wState.trafficPolicy.peerAuthnSelector.mode
           }
         }
+      };
+
+      wizardDR.metadata.annotations = {};
+      wizardDR.metadata.annotations[KIALI_RELATED_LABEL] =
+        'PeerAuthentication/' + wProps.serviceName + '/' + wizardPA.spec.mtls!.mode;
```
lucasponce, line 435: I doubt about this one.
I like to link the DR with the PA and the serviceName, that makes sense.
But, if the user modifies the PA the status will fail because is coded there.

I think that PA should be fetched in the ServiceDetails page using the kiali_wizard label and this link with the service, but if I change the status in the Istio Config PA, that should be propagated in the wizard.

It happens today with VS/DR, if you edit them and go back to the service details and open the Wizard, this is updated with this new info.

We can review more in details in you like.

lucasponce:
One comment about the implementation of the PA, but changes in the wizard look ok from the user perspective.

xeviknal:
@skondkar @lucasponce approached your comments.
@skondkar good catch. It should be working now. Regarding your suggestion, leaving the Destination Rule TLS set to 'UNSET' doesn't mean that the user won't add a PeerAuthn. Then, you can have TLS set to 'UNSET' and 'Add PeerAuthentication' switched on.

PR #1855, opened 2020-07-21
Fix graph losing focused node context after first refresh

jotak:
Was due to improper array equality check in EmptyGraphLayout
Fixing at the same time some warning in dev console:
- Some missing key props in lists
- React complaining that GraphFind should not be PureComponent as it has 'shouldComponentUpdate'

Fixes https://github.com/kiali/kiali/issues/2957


pbajjuri20:
Tested and Passed on Kiali UI 1.22.0 (af87e29a5214069bd638865ff81260ef2930b676)

PR #1859, opened 2020-07-27
Graph traces list: add a separate refresh button

jotak:
Fixes kiali/kiali#3016

### DIFF
--- a/src/pages/Graph/SummaryPanelNodeTraces.tsx
+++ b/src/pages/Graph/SummaryPanelNodeTraces.tsx
@@ -1,7 +1,8 @@
 import * as React from 'react';
 import { connect } from 'react-redux';
 import { ThunkDispatch } from 'redux-thunk';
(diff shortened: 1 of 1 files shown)


jotak:
@jshaughn I changed a little bit the style of the button compared to our other refresh button, to make it more distinct from the top-level refresh. It also has smaller padding to fit better with the side-panel, which has smaller fonts:

This is with PF's "secondary" variant for buttons:
Capture d’écran de 2020-07-27 10-46-37

PS: I've changed the checkbox text, unlike in these screenshots, it's now "Use graph refresh"

How it would be with the main variant:
Capture d’écran de 2020-07-27 10-47-19

And the alternatives,
Tertiary variant:

Capture d’écran de 2020-07-27 10-47-43

"Control" variant:
Capture d’écran de 2020-07-27 10-48-12


Personally I prefer the "secondary" variant (the one used in this current PR).

jotak:
some little changes after feedback ... anyway we can also discuss this with UX tomorrow

PR #1860, opened 2020-07-28
In-graph tracing: show trace details

jotak:
- Add trace ID to URL
- Split side-panel with a bottom section to show trace info when a trace is selected
- Allow selecting other nodes from that trace, keeping that panel opened
- Show some trace information
- Show spans information for each span within the trace for the selected node
- Link to Trace Details
- Focus animation triggered to localize span other-end client/server

Fixes https://github.com/kiali/kiali/issues/3010

### DIFF
--- a/src/actions/JaegerThunkActions.ts
+++ b/src/actions/JaegerThunkActions.ts
@@ -6,9 +6,11 @@ import * as API from '../services/Api';
 import { KialiAppAction } from './KialiAppAction';
 import { transformTraceData } from 'components/JaegerIntegration/JaegerResults';
 import { JaegerActions } from './JaegerActions';
+import { setTraceId as setURLTraceId } from 'utils/SearchParamUtils';
 
 export const JaegerThunkActions = {
-  fetchTrace: (traceId?: string) => {
+  setTraceId: (traceId?: string) => {
+    setURLTraceId(traceId);
     return (dispatch: ThunkDispatch<KialiAppState, void, KialiAppAction>) => {
       if (traceId) {
         API.getJaegerTrace(traceId)
--- a/src/components/CytoscapeGraph/CytoscapeGraph.tsx
+++ b/src/components/CytoscapeGraph/CytoscapeGraph.tsx
@@ -55,6 +55,7 @@ type CytoscapeGraphProps = {
   refreshInterval: IntervalInMilliseconds;
   setActiveNamespaces?: (namespace: Namespace[]) => void;
   setNode?: (node?: NodeParamsType) => void;
+  setTraceId?: (traceId?: string) => void;
   setUpdateTime?: (val: TimeInMilliseconds) => void;
   showCircuitBreakers: boolean;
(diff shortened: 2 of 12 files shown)


jotak:
WIP:
- [ ] Clear trace when clicking around
- [ ] Link to external Jaeger

jotak:
Current state:
Capture d’écran de 2020-07-28 16-53-21

jotak:
@jshaughn this is ready for review, although we can also wait a little bit for UX feedbacks.
Here I'm removing the experimental flag!

jotak:
@jshaughn it can be confusing, but it's because the "9 spans" are on the whole trace, whereas the "1 span" is only on the selected node. Any suggestion to make it less surprising?

jotak:
@jshaughn I've added some precision about the spans: now it shows "span on xxx 1/n" to make it clearer it's only for the selected node. Is it better?

Also increased the size of trace details panel, but not as much as you suggest, because there's not so much data displayed, and increasing more than that would leave some big empty spaces.

lwrigh:
For the tracing visualization I would suggest having it look more similar to something like this: 
Artboard I would suggest making the purple highlight a bit more subdued and less segmented. By doing this it makes it easier to tell which shapes are in the foreground and makes it easier to tell the difference between the different layers of the visualization. 

Instead of having the additional panel pop up when the user clicks on the trace for more details I would suggest making the rows expandable under the 'Traces' tab.

jotak:
> For the tracing visualization I would suggest having it look more similar to something like this:
> Artboard I would suggest making the purple highlight a bit more subdued and less segmented. By doing this it makes it easier to tell which shapes are in the foreground and makes it easier to tell the difference between the different layers of the visualization.

I leave this part for @jshaughn who knows it better than me :-)

> Instead of having the additional panel pop up when the user clicks on the trace for more details I would suggest making the rows expandable under the 'Traces' tab.

But in this case we're not able to compare trace info with Traffic tab / metrics anymore.

jshaughn:
@jotak This PR has conflicts with master.  If you could rebase this against master then I can try and get my PR against this one to be reduced to just the proper changes.

jotak:
@jshaughn done

jotak:
@jshaughn I merged your changes and added a new commit if you want to review it

pbajjuri20:
Navigate to these steps for the issue:
the page was not loading

1) Click on service in the graph
2) Traces tab
3) Click on "Go to Tracing"

https://snipboard.io/vL5JR8.jpg

pbajjuri20:
@jotak, Thanks for fixing the issue! 

Tested and Passed on Kiali UI 1.23.0 (def411a5e85d50e461283a0861aaed1ff4e1d20f)

jshaughn:
@jotak Is this PR waiting on anything for merge?

jotak:
Thanks @jshaughn @pbajjuri20 🍾 🥂

PR #1861, opened 2020-07-28
remove login and ldap auth strategy

jmazzitelli:
https://github.com/kiali/kiali/issues/3017


jmazzitelli:
I am not going to wait for this to be reviewed before merging the other 3 PRs. The other 3 are ready to go. Once they are merged, even if this PR isn't merged, it won't matter because the server side will never tell the UI that it is configured for ldap or login so this is dead code anyway.

We can merge this PR as soon as possible.

PR #1866, opened 2020-07-31
Traces are being searched for by service name but apparently it's the…

aljesusg:
Fix https://github.com/kiali/kiali/issues/3045

Require PR: https://github.com/kiali/kiali/pull/3064

Screenshot from 2020-07-31 12-19-00

Now we have link to Inbound and Outbound metrics related with the APP

Tested with Istio 1.7.alpha2 and Travel-agency

### DIFF
--- a/src/components/JaegerIntegration/JaegerResults/SpanDetail.tsx
+++ b/src/components/JaegerIntegration/JaegerResults/SpanDetail.tsx
@@ -56,7 +56,7 @@ export class SpanDetailC extends React.Component<SpanDetailProps, SpanDetailStat
                     </GridItem>
                     <GridItem span={4} />
                     <GridItem span={2}>
-                      <Tooltip content={<>View traces of app {this.props.spans[0].process.serviceName}</>}>
+                      <Tooltip content={<>View traces of application {this.props.spans[0].process.serviceName}</>}>
                         <Link
                           to={this.goApp() + '?tab=traces'}
                           onClick={() => history.push(this.goApp() + '?tab=traces')}
--- a/src/components/JaegerIntegration/JaegerResults/SpanDetail.tsx
+++ b/src/components/JaegerIntegration/JaegerResults/SpanDetail.tsx
@@ -25,10 +25,10 @@ export class SpanDetailC extends React.Component<SpanDetailProps, SpanDetailStat
     this.state = { isModalOpen: false };
   }
 
-  goService = (service: string = this.props.spans[0].process.serviceName) => {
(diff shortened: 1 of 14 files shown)


src/components/JaegerIntegration/JaegerResults/SpanDetail.tsx
```diff
@@ -56,10 +56,10 @@ export class SpanDetailC extends React.Component<SpanDetailProps, SpanDetailStat
                     </GridItem>
                     <GridItem span={4} />
                     <GridItem span={2}>
-                      <Tooltip content={<>View traces of service {this.props.spans[0].process.serviceName}</>}>
+                      <Tooltip content={<>View traces of app {this.props.spans[0].process.serviceName}</>}>
```
jotak, line 59: I think we tend to not use "app" as a word in general (correct me if I'm wrong), we display the full word "application" instead.
Perhaps here we can just write "View traces of xxx".

aljesusg:
> Almost good but I've seen at least two places with broken links to traces:
> 
> * In Graph/summary panel actions
> * When clicking on spans in metrics view
> 
> I guess if you search for "tab=traces" in all code you can easily spot them.

totally I forgot them

src/components/JaegerIntegration/JaegerResults/SpanDetail.tsx
aljesusg: right and changed

lucasponce:
It looks good, but I wonder if we don't want to extend the traces in all app/workload/services.
Personally it's weird to have the service details metrics with span correlation and jump into the app one.

That's not obvious to me, but I'm fine to progress with this and add that later.

lucasponce:
Personally I'd prefer to have a traces tab under app/workload/service details, even if all of them use the app correlation.
I prefer that duplication, rather than jump from services metrics to application metrics or similar.

aljesusg:
> I have feeling that probably having traces under application, service and workload would help the user.
> In all cases, we have an app label to filter on.
> 
> And the tab can always point to the app in a first iteration, but removing a tab from service details is like "oh, tracing information is not longer linked with this" which is not correct.
> 
> I'd prefer to say, ok for review-v2 you will see traces for app=reviews rather than not showing anything, same for service=reviews.
> 
> In that way we can allow put the user in context in all dimensions (details, traffic, metrics and traces).

Jaeger is unable to query traces for multiple services, at least that kiali does...we can query all apps of service X and show traces in the tab.

hhovsepy:
Post merge verified.

PR #1867, opened 2020-07-31
Don't use message center alerts for regex errors

jshaughn:
use PF4 guideline and make it like graph find/hide.

Fixes https://github.com/kiali/kiali/issues/2973

image

### DIFF
--- a/src/pages/WorkloadDetails/WorkloadInfo/WorkloadPodLogs.tsx
+++ b/src/pages/WorkloadDetails/WorkloadInfo/WorkloadPodLogs.tsx
@@ -604,7 +604,7 @@ export default class WorkloadPodLogs extends React.Component<WorkloadPodLogsProp
     if (!!showValue) {
       if (this.state.useRegex) {
         try {
-          const regexp = new RegExp(showValue);
+          const regexp = RegExp(showValue);
           filteredLogs = filteredLogs.filter(l => regexp.test(l));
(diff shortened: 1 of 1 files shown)


jshaughn:
> This doesn't work on the OS/X browsers. The catch block never gets executed

I can't explain it, I've scoured the interwebs and there is no mention of the regexp constructor not throwing errors on webkit/safari.  To the contrary there are a lot of example of errors being thrown on safari, sometimes when they should not have been.  Safari supports both the regexp constructor and try/catch.  Something is wonky here.

Do me a favor, try something else bogus in a try catch on safari, like a div by 0 or something, anything.  what happens?  Or try a regexp compile(pattern).   It's deprecated in favor of the construtor but it too should throw an error.

There is, as far as I can see, no regexp validate function, so if this doesn't work on Safari I'm not sure we have an alternative other than to report a bug to safari - but I'd be shocked that we'd be the first to see it, something else must be going on here.

jshaughn:
@mtho11 I made one small change, maybe it somehow makes a difference. on Safari, I got rid of the extraneous "new" in "new RegExp".  Please try again.

mtho11:
Minor spacing issues when errors occur:

Kiali_Console

Kiali_Console

pbajjuri20:
Tested and Passed on Kiali UI 1.23.0 (3f048f913ef22f26bd014783d9371cb1aca44657)

PR #1870, opened 2020-08-04
Relax Warning AlertToasts on Istio Config validations

lucasponce:
Relax the popup warning for Istio Config validations.
Other notifications (AlertToasts) should remain as default behaviour.

Fixes https://github.com/kiali/kiali/issues/3074


xeviknal:
I see this behavior in the config list...
Screen recording (1) (1)

PR #1871, opened 2020-08-04
Find/Hide improvements:

jshaughn:
Fixes https://github.com/kiali/kiali/issues/3072

- For consistency between graph find/hide and logs show/hide, improve graph find handling to report independent errors for find and hide expressions.
- apply a UX-approved tweak to slightly reduce/lighten the find overlay to maintain better visibility of labels and nodes.
- immediately submit a find/hide value selected from the autocomplete dropdown   (or pasted).  Note that there is no native support for this, so we just assume any single input change > 1 character should be applied.  It seems to work well as a rule of thumb.

@mtho11 If you like the immediate submit on autocomplete select, we can add that to log show/hide as well.

### DIFF
--- a/src/components/CytoscapeGraph/graphs/GraphStyles.ts
+++ b/src/components/CytoscapeGraph/graphs/GraphStyles.ts
@@ -618,8 +618,8 @@ export class GraphStyles {
         selector: '*.find[^isGroup]',
         style: {
           'overlay-color': PfColors.Gold400,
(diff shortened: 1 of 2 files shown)


mtho11:
Kiali_Console

pbajjuri20:
Tested and passed on Kiali UI 1.23.0 (a6930eabce196f725a9f52a2b2fd28ef32bfb8b6)


find

PR #1874, opened 2020-08-05
Tooltip issues

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3082

When a CSS is built for production, there are some conflicts in the order of style for components compared with the developer one.


aljesusg:
it seems that we need to update snaps

lucasponce:
@aljesusg comments addressed

PR #1875, opened 2020-08-05
Add/Remove Auto Injection

lucasponce:
Fixes https://github.com/kiali/kiali/issues/2896
Needs https://github.com/kiali/kiali/pull/3085

In compact view:
image
image

In table views where all labels are explicit:
image
image

When this feature is disabled, via operator or direct config, the new action shouldn't be added in the menu:
image

### DIFF
--- a/src/pages/Overview/OverviewPage.tsx
+++ b/src/pages/Overview/OverviewPage.tsx
@@ -469,7 +469,7 @@ export class OverviewPage extends React.Component<OverviewProps, State> {
         action: (ns: string) => this.show(Show.ISTIO_CONFIG, ns, this.state.type)
       }
     ];
-    if (serverConfig.istioInjectionAction) {
+    if (serverConfig.kialiFeatureFlags.istioInjectionAction) {
       namespaceActions.push({
         isSeparator: true
       });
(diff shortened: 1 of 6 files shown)


src/pages/Overview/OverviewPage.tsx
```diff
@@ -478,6 +491,24 @@ export class OverviewPage extends React.Component<OverviewProps, State> {
       // Actions can be personalized by namespace
       // i.e. Add or Remove depending of presence of a flag/label
       const actions = this.showActions();
+      if (serverConfig.istioInjectionAction) {
```
xeviknal, line 494: Perhaps is personal style, but what about moving this `istioInjectionAction` into a specific method like `appendInjectionAction(actions, ns)`. The render method is really compact so far, it doesn't know about the implementation of the action list, so it looks a bit cleaner to have this logic in a different method.

lucasponce:
@xeviknal I followed your suggestion and I've located the actions in a method that it was designed for that.
Also fixing wording "Add/Remove" -> "Enable/Disable" as it's better accurate with Istio terminology here.

lucasponce:
Adding @jmazzitelli to validate a change triggered from the operator.

PR #1876, opened 2020-08-06
Honor preset URL after OAuth/OICD login

israel-hdez:
This makes sure that after an OAuth/OICD login, any pre-set URL is honored.

A "pre-built" URL could be a bookmarked page, or a link from other systems.

Fixes kiali/kiali#2131


israel-hdez:
@jshaughn Simplest implementation I could think of -- I wanted to avoid changes on both back-end and front-end.
Affects both OpenShift & OpenID auths.

pbajjuri20:
Tested and passed on Kiali UI 1.23.0 (6a1e5bd56214398d9012b0f6ac5e3eb4b833965b)

PR #1882, opened 2020-08-11
Support setting history mode by parameter

aimuz:
** Describe the change **

Read the base path set to WEB_ROOT, useful when using `kubectl proxy`.

Setting ``web_root` to `/api/v1/namespaces/istio-system/services/kiali:20001/` in the previous setup for `kiali` does not work as expected. Although the `env.js` setting is read in the previous logic, it doesn't seem to work properly with `subpath` in the working mode.

When I open `http://127.0.0.1:8001/api/v1/namespaces/istio-system/services/kiali:20001/` kiali will change the address to `http://127.0.0.1:8001/console/overview?pi=15000&duration=60`

I think the normal expectation is `http://127.0.0.1:8001/api/v1/namespaces/istio-system/services/kiali:20001/console/overview?pi=15000&duration=60`.

So, this code change is the solution to that problem. 

** Issue reference **

* If this is a fix for a JIRA, please put JIRA-reference in the PR title like _`KIALI-7 New icons`_ (Jira-id, one space, description)
* If this is a fix for a GH-issue, please link it here

** Backwards compatible? **

- [ ] Is your pull-request introducing changes in behaviour?

No changes will occur and the program will become more robust.

** Screenshot **

Add a screenshot of the UI after your changes.
image

** Documentation **

Please provide an update to documentation if appropriate. For configuration options, consider sending a PR to https://github.com/kiali/kiali/blob/master/README.adoc

### DIFF
--- a/src/app/History.ts
+++ b/src/app/History.ts
@@ -1,10 +1,15 @@
-import { createBrowserHistory, createMemoryHistory } from 'history';
+import { createBrowserHistory, createMemoryHistory, createHashHistory } from 'history';
 import { toValidDuration } from '../config/ServerConfig';
 import { BoundsInMilliseconds } from 'types/Common';
 
 const webRoot = (window as any).WEB_ROOT ? (window as any).WEB_ROOT : undefined;
 const baseName = webRoot && webRoot !== '/' ? webRoot + '/console' : '/console';
-const history = process.env.TEST_RUNNER ? createMemoryHistory() : createBrowserHistory({ basename: baseName });
+const historyMode = (window as any).HISTORY_MODE ? (window as any).HISTORY_MODE : 'browser';
+const history = process.env.TEST_RUNNER
+  ? createMemoryHistory()
+  : historyMode === 'browser'
+  ? createBrowserHistory({ basename: baseName })
+  : createHashHistory();
 
 export default history;
 
--- a/src/app/History.ts
+++ b/src/app/History.ts
@@ -7,9 +7,9 @@ const baseName = webRoot && webRoot !== '/' ? webRoot + '/console' : '/console';
 const historyMode = (window as any).HISTORY_MODE ? (window as any).HISTORY_MODE : 'browser';
 const history = process.env.TEST_RUNNER
   ? createMemoryHistory()
-  : historyMode === 'browser'
-  ? createBrowserHistory({ basename: baseName })
-  : createHashHistory();
+  : historyMode === 'hash'
+  ? createHashHistory()
+  : createBrowserHistory({ basename: baseName });
 
 export default history;
 


jmazzitelli (dismissed):
This requires a good amount of testing. We tried fixing this in the past and always hit cases where things broke.

These must pass:

1) Using OpenShift Route to access  Kiali
2) Using K8s Ingress to access Kiali
3) Using Istio Gateway/VirtualService/DestinationRule to access Kiali
4) Test 1 through 3 using server.web_root setting with value of "/"
5) Test 1 through 3 using server.web_root setting with value of "/kiali"
6) Test 1 through 3 using server.web_root setting with value of "/foo/bar"

I have no suggested changes, however, I am marking this as "request changes" only to block this PR from being merged until we verify those sets of tests pass.

aimuz:
I understand the reason for doing this, I'm just testing the usage scenario I'm in and I do think a full test needs to be done.

aimuz:
> Hi @aimuz .
> 
> This change has been proposed in the past and it's unsuitable. It's not enough to use the tag as a pivot to resolve the .`<base>``web_root`
> 
> This code works **only** if you access Kiali using the root path. In your example, it is likely to be . But once you navigate to other Kiali sections (for example ) if you hit the refresh button of the browser, you will land to a broken page. This basically means that you cannot "bookmark" a page of Kiali.`http://127.0.0.1:8001/api/v1/namespaces/istio-system/services/kiali:20001/``http://127.0.0.1:8001/api/v1/namespaces/istio-system/services/kiali:20001/services`
> 
> I suggest you to read this FAQ entry to learn other ways to access Kiali: https://kiali.io/documentation/latest/faq/#how-do-i-access-kiai.
> 
> If you really need to use you will need to set the right in the config.`kubectl proxy``web_root`

I think the way to do it once and for all is to change the routing mode to "hashhistory"`.

israel-hdez:
> I think the way to do it once and for all is to change the routing mode to "hashhistory"`.


Yes, but IMO that's a breaking change, because it will break existing bookmarks that users may already have set.

aimuz:
> > I think the way to do it once and for all is to change the routing mode to "hashhistory"`.
> 
> Yes, but IMO that's a breaking change, because it will break existing bookmarks that users may already have set.

So now change the routing mode to "hashhistory" , and merge it into master at the right time?

israel-hdez:
> So now change the routing mode to "hashhistory" , and merge it into master at the right time?

I'm not sure that will work, as I don't think there is a right time for merging this particular change.

Perhaps a better approach is to make this configurable and use the current behavior as the default, allowing people to switch to `hashhistory` if they prefer that one.

aimuz:
I'll change it to this way，Is this file the place for global configuration ?
```js
const historyMode = (window as any).HISTORY_MODE ? (window as any).HISTORY_MODE : 'browser';
const history = process.env.TEST_RUNNER
  ? createMemoryHistory()
  : historyMode === 'browser'
  ? createBrowserHistory({ basename: baseName })
  : createHashHistory();
```

israel-hdez:
> I'll change it to this way，Is this file the place for global configuration ?
> 
> ```js
> const historyMode = (window as any).HISTORY_MODE ? (window as any).HISTORY_MODE : 'browser';
> const history = process.env.TEST_RUNNER
>   ? createMemoryHistory()
>   : historyMode === 'browser'
>   ? createBrowserHistory({ basename: baseName })
>   : createHashHistory();
> ```

That code draft looks OK.

I'm not sure I would call it "global configuration", because there is this folder: https://github.com/kiali/kiali-ui/tree/master/src/config. You may notice there are several config values that are loaded "later" asynchronously.

What I can say is that the `env.js` file (and specially the WEB_ROOT) is configuration that we need loaded early. Doing a request to the back-end to fetch it would be too late.

aimuz:
I have passed the test locally and I don't know if such a change can be made. By default, it doesn't change the original behavior

Kiali has also been updated 
https://github.com/kiali/kiali/pull/3113

jmazzitelli:
> I have passed the test locally

When you say "passed the test" - does that mean all of these pass (from https://github.com/kiali/kiali-ui/pull/1882#pullrequestreview-465036368):

```
These must pass:

1. Using OpenShift Route to access Kiali
2. Using K8s Ingress to access Kiali
3. Using Istio Gateway/VirtualService/DestinationRule to access Kiali
4. Test 1 through 3 using server.web_root setting with value of "/"
5. Test 1 through 3 using server.web_root setting with value of "/kiali"
6. Test 1 through 3 using server.web_root setting with value of "/foo/bar"
```

aimuz:
I have only done a few tests. Based on 'ingress' and' kubectl proxy ', other tests may need help

israel-hdez:
Hi, @aimuz 
We've been somewhat busy. But if you are still interested, I can try your PR. Is this OK?

aimuz:
@israel-hdez 
I think we can try, we have used this change in production

src/app/History.ts
```diff
@@ -1,10 +1,15 @@
...
 import { toValidDuration } from '../config/ServerConfig';
 import { BoundsInMilliseconds } from 'types/Common';
 
 const webRoot = (window as any).WEB_ROOT ? (window as any).WEB_ROOT : undefined;
 const baseName = webRoot && webRoot !== '/' ? webRoot + '/console' : '/console';
-const history = process.env.TEST_RUNNER ? createMemoryHistory() : createBrowserHistory({ basename: baseName });
+const historyMode = (window as any).HISTORY_MODE ? (window as any).HISTORY_MODE : 'browser';
+const history = process.env.TEST_RUNNER
+  ? createMemoryHistory()
+  : historyMode === 'browser'
```
israel-hdez, line 10: I think this 'kind-of' is making `createHashHistory` to be the "standard" default, because an arbitrary string in the `web_history_mode` will have the effect to using hash-history.

Since `createBrowserHistory` is what Kiali has been using for a long time, I prefer to change this conditional so that Kiali will fallback to `createBrowserHistory` if the value of the history mode is unknown.

israel-hdez (changes requested):
It works, but I have a small comment.

src/app/History.ts
aimuz:
```js
const historyMode = (window as any).HISTORY_MODE ? (window as any).HISTORY_MODE : 'browser';
const history = process.env.TEST_RUNNER
  ? createMemoryHistory()
  : historyMode === 'hash'
  ? createHashHistory()
  : createBrowserHistory({ basename: baseName });
```
Sorry, I haven't paid attention for too long

Is this method feasible?

israel-hdez:
Looks OK

israel-hdez (approved):

israel-hdez:
A change is missing on Kiali operator in case you use the operator, but I think this is mergeable without it.

israel-hdez:
@jmazzitelli I think we can merge this. I see no conflict given our current behavior will remain as the default.

israel-hdez:
It's done @aimuz .
Feel free to do the missing [kiali-operator](https://github.com/kiali/kiali-operator) changes, in case you need them.

aimuz:
> It's done @aimuz .
> Feel free to do the missing [kiali-operator](https://github.com/kiali/kiali-operator) changes, in case you need them.

Yes, I will make changes and modify the help document

jshaughn:
@aimuz Is the follow-up work complete?

aimuz:
> @aimuz Is the follow-up work complete?

@jshaughn 
Sorry, I'm a little busy recently. I just created pr

jshaughn:
@aimuz No problem, I just wasn't sure where things had left off.  Thanks for your PR and contributions!

PR #1885, opened 2020-08-13
Join Istio Config under same tab in Service Details page

lucasponce:
Part of https://github.com/kiali/kiali/issues/3015

This is a previous step to:
- Put together all Istio config under a single tab.
- Unify the columns showed (there were some differences, coming from past designs), now it shows the same info.
- Allow to have a clean design if we decide to add more tabs for other domains (Iter8 can be a next one, which doesn't make sense to put it under Istio Config).
- Preliminary step to add more objects related to a Service (i.e. Gateways or security objects can be added in a next step).
- Unify the format and reuse component to be used under Workload Details too (potentially in App Details).

image

For testing:
- Check regressions in resources listed and validations.


hhovsepy:
Screenshot from 2020-08-13 14-53-08

lucasponce:
Thanks @hhovsepy I'm updating the tests, I missed some in failure state.

PR #1886, opened 2020-08-13
Enable/Disable/Remove Istio Auto Injection on Namespace/Workload

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3125
Fixes https://github.com/kiali/kiali/issues/3124
Requires https://github.com/kiali/kiali/pull/3130

@hhovsepy I've introduced a explicit "Disabled" vs "Removed" as this opens the possibility to annotate workloads individually.

Lucas

### DIFF
--- a/src/components/IstioWizards/WizardActions.ts
+++ b/src/components/IstioWizards/WizardActions.ts
@@ -49,6 +49,10 @@ export const WIZARD_SUSPEND_TRAFFIC = 'suspend_traffic';
 export const WIZARD_THREESCALE_LINK = '3scale_link';
 export const WIZARD_THREESCALE_UNLINK = '3scale_unlink';
 
+export const WIZARD_ENABLE_AUTO_INJECTION = 'enable_auto_injection';
+export const WIZARD_DISABLE_AUTO_INJECTION = 'disable_auto_injection';
+export const WIZARD_REMOVE_AUTO_INJECTION = 'remove_auto_injection';
+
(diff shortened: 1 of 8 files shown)


lucasponce:
Also, I think we should introduce the Enable/Remove/Disable label at workload level too.
As basically a "istio-injection: disabled" placed at namespace works together with a "sidecar.istio.io/inject: true" placed at workload level.

lucasponce:
Thanks @hhovsepy I'm going to hold this PR to add the same operation but at workload level, I think it makes sense.

lucasponce:
@hhovsepy I've added support to combine namespace and workload istio label + annotations.

image

Otherwise end-user should find odd to disable a namespace without possibility to annotate a specific workload or viceversa, to disable a specific workload when a namespace is annotated.

I'm going to review the travis in the CI, but I've added backend + operator PRs for this.

hhovsepy:
@lucasponce one note, from UX perspective:
1. Enable namespace.
2. disable workload.
3. disable namespace
4. enable namespace
Workload remains disabled.
Shouldn't it be better each time changing namespace it force overrides the child workloads?

lucasponce:
> Shouldn't it be better each time changing namespace it force overrides the child workloads?

Not sure, basically Kiali is performing these operations 
https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/#understanding-what-happened

So, manipulating namespace label and/or workload annotation but it's not adding more logic here.

It's seems that using CLI both (namespace, workload) are also independent, so, I guess in a first approach we can try go add them as well.

PR #1889, opened 2020-08-17
Refactor refresh log tab

mtho11:
** Describe the change **
Moves the refresh controls from inside the logging tab to outside the logging tab to match other tabs for consistency.

** Issue reference **
fixes #https://github.com/kiali/kiali/issues/3011

** Backwards compatible? **
Yes

** Screenshot **
Before:

Kiali_Console_and_Kiali_Console_and_kxr_ocp4_setup_upi_kvm__Script_to_Setup_an_OpenShift_4_UPI_Cluster_on_KVM__Based_on_this_guide__https___kxr_me_2019_08_17_openshift-4-upi-install-libvirt-kvm_

After:

Kiali_Console

### DIFF
--- a/src/pages/WorkloadDetails/WorkloadDetailsPage.tsx
+++ b/src/pages/WorkloadDetails/WorkloadDetailsPage.tsx
@@ -98,7 +98,11 @@ class WorkloadDetails extends React.Component<WorkloadDetailsPageProps, Workload
     const logTab = (
       <Tab title="Logs" eventKey={2} key={'Logs'}>
         {hasPods ? (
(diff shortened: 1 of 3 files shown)


mtho11:
Ok, I have assigned @lucasponce for review.

lucasponce:
Now that the controls are shared I see that the inbound/outbound metrics are not taking the shared duration (and they should).
Also I see a side effect that I introduced in the actions from my side, which is that every tab is responsible of refreshing the tab content, but the parent workload details refresh the workload.
This is correct, but with actions, the overview tab may modify the workload, so it needs a change in the way how workload is refreshed.

It has nothing to do with this PR, it's more a side effect of the actions work.

So, I'm +1 to merge this one and I will provide a followup PR to address those concerns.

PR #1890, opened 2020-08-18
Proper refresh Workload from WorkloadDetails

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3133

WorkloadInfo was refreshing workload when the proper way was delegating this into the parent component.

This causes side effect like run a Workload action and not having a fresh workload in other tabs (specially logs tab).

### DIFF
--- a/src/pages/WorkloadDetails/WorkloadDetailsPage.tsx
+++ b/src/pages/WorkloadDetails/WorkloadDetailsPage.tsx
@@ -78,9 +78,10 @@ class WorkloadDetails extends React.Component<WorkloadDetailsPageProps, Workload
     const overTab = (
       <Tab title="Overview" eventKey={0} key={'Overview'}>
         <WorkloadInfo
-          workloadName={this.props.match.params.workload}
+          workload={this.state.workload}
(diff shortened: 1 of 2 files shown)


hhovsepy:
I can still see the problem:
1. Enable auto injection on workload.
2. refresh the info tab. new pod is shown in Pods, status is  updated.
3. go to Logs tab, no refresh was done here. 
4. hit the refresh icon, no change (issue)
only after the webpage f5 refresh the new pod logs is there.

lucasponce:
@hhovsepy thanks for the feedback, let me re-test to check if I missed something in my local env

lucasponce:
@hhovsepy ok, finally addressed some scenarios that didn't work well, now that the parent component refreshes the workload, when it changes, the WorkloadInfo tab should also refresh the health as well.

hhovsepy (approved):
Thanks @lucasponce 
Verified now.

aljesusg:
This fix the problem that we had in the workloadHealth too. Now we can see the error rate in the health overview in details

PR #1891, opened 2020-08-18
Trivial: Remove the deprecated React.SFC and replace with FunctionComponent

mtho11:
** Describe the change **
Remove the deprecated `React.SFC` and replace with `React.FunctionComponent` as docs specify.
No behavioral changes just syntax change.

** Documentation **
kiali-ui_–_PfSpinner_tsx


mtho11:
With React version 17 coming out any day now I'm sure that the deprecated component `React.SFC` will even be valid in that version.

lucasponce (approved):

PR #1892, opened 2020-08-19
Adjust wording between traces in graph and details

lucasponce:
Two trivial changes about wording used in graph traces.

"Go to Tracing" -> "View App Traces"
"See trace in Jaeger" -> "View Trace in Tracing"

- Tracing is used in "Traces" Tab to refer to the external tracing tool, better to refer to "App Traces" on this case.
- The same external link is used in "traces" tab, so better to unify the same wording instead o introduce new one.

Place the external link at the bottom, when the app/service is too long, it's hardly differentiated from the trace title one.
image

I think with this little change it's more clear and it seems clean.

### DIFF
--- a/src/pages/Graph/SummaryPanelNodeTraces.tsx
+++ b/src/pages/Graph/SummaryPanelNodeTraces.tsx
@@ -174,7 +174,7 @@ class SummaryPanelNodeTraces extends React.Component<Props, State> {
           </SimpleList>
         )}
         <Button style={summaryFont} onClick={() => history.push(tracesDetailsURL)}>
-          View App Traces
+          Show Traces
         </Button>
       </div>
     );
--- a/src/pages/Graph/SummaryPanelNodeTraces.tsx
+++ b/src/pages/Graph/SummaryPanelNodeTraces.tsx
@@ -174,7 +174,7 @@ class SummaryPanelNodeTraces extends React.Component<Props, State> {
           </SimpleList>
         )}
         <Button style={summaryFont} onClick={() => history.push(tracesDetailsURL)}>
-          Go to Tracing
+          View App Traces
         </Button>
       </div>
     );
--- a/src/pages/Graph/SummaryPanelTraceDetails.tsx
+++ b/src/pages/Graph/SummaryPanelTraceDetails.tsx
@@ -166,7 +166,7 @@ class SummaryPanelTraceDetails extends React.Component<Props, State> {
             <>
               <br />
               <a href={jaegerTraceURL} target="_blank" rel="noopener noreferrer">
-                View Trace in Tracing <ExternalLinkAltIcon size="sm" />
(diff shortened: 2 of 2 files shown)


skondkar (approved):
Verified on Kiali UI:   1.23.0 (b70f9c9aaa4742a28a4206bcb1de3ae2217fd54b)
Verified both the links:

Screenshot_ViewAppTraces_link

Screenshot-Trace-ViewTrace

src/pages/Graph/SummaryPanelNodeTraces.tsx
```diff
@@ -174,7 +174,7 @@ class SummaryPanelNodeTraces extends React.Component<Props, State> {
           </SimpleList>
         )}
         <Button style={summaryFont} onClick={() => history.push(tracesDetailsURL)}>
-          Go to Tracing
+          View App Traces
```
jshaughn, line 177: The verb 'Show' is more consistent with the graph context and kebab menu options.  So I think `Show Traces`, `Show App Traces` or `Show Application Traces` would be better.

src/pages/Graph/SummaryPanelTraceDetails.tsx
```diff
@@ -170,6 +162,14 @@ class SummaryPanelTraceDetails extends React.Component<Props, State> {
                 this.renderSpan(nodeName + '.' + node.namespace, spans[this.state.selectedSpan])}
             </p>
           )}
+          {tracesDetailsURL && jaegerTraceURL && (
+            <>
+              <br />
+              <a href={jaegerTraceURL} target="_blank" rel="noopener noreferrer">
+                View Trace in Tracing <ExternalLinkAltIcon size="sm" />
```
jshaughn, line 169: I'm OK with the bottom-positioning but again, I'd use 'Show'.  Also, because we know we're looking at a trace I think we can be less wordy, so perhaps just `Show in Tracing`.

lucasponce:
@jshaughn ok, comments addressed.
I don't really care about the verb used, I'm fine moving to "Show" as it's also used in "Overview" and "Graph".

But I'd like to get consistency in the same actions/links being named in the same way.

Probably I'll revisit some of these wordings in the other pages and I'd try to unify things, those are little details that make that every page/tab may look different.

lucasponce:
> @lucasponce LGTM.
> Going forward, as you look for more consistency, we may want to look more closely at things after the move from Service to Application focus for traces. Currently on the graph you get the side-panel `Traces` tab when selecting a service. But you get the `Show Traces` context menu option only for an unversioned app node, or an app box. For an app node you get the same option in the side-panel kebab, but bnot for the app box. So the experience maybe could use some review.

+1

There are a pending Epic to enable Traces under Workload/Service but it's not yet finished. I agree that when Traces are shown on every entity then we should adjust correctly the link from/to Graph as well.

Thanks for reviewing.

PR #1893, opened 2020-08-21
Bump dot-prop and lodash

lucasponce:
Address a vulnerability detected by github in the 3rd party libraries.

I've addressed this in k-charted as well https://github.com/kiali/k-charted/pull/110.

I'd require more testing, smoke tests seem ok, but these changes may be good if are tested by anyone else as they can easily introduce side effects.

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "eventemitter3": "4.0.0",
     "js-yaml": "3.13.1",
     "json-beautify": "1.0.1",
-    "lodash": "4.17.19",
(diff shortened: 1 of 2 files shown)


lucasponce:
@hhovsepy @skondkar this PR should be safe to merge, but would you mind to perform some smoke test on it ?

Bumping dependencies is a way to introduce side effects, probably it's not necessary to run full regressions, just some smoke test in the features to see if there is something odd.

hhovsepy:
Smoke test pass.

PR #1894, opened 2020-08-25
Add info tooltip to Graph Security Display option.

jshaughn:
Related: https://github.com/kiali/kiali/issues/3152

### DIFF
--- a/src/pages/Graph/GraphToolbar/GraphSettings.tsx
+++ b/src/pages/Graph/GraphToolbar/GraphSettings.tsx
@@ -284,7 +284,8 @@ class GraphSettings extends React.PureComponent<GraphSettingsProps, GraphSetting
         tooltip: (
           <div style={{ textAlign: 'left' }}>
             <div>
-              Show closed or open lock icons on edges that differ from the global mTLS policy. Note that the global
+              Show closed or open lock icons on edges with traffic that differs from the global mTLS policy. The
+              percentage of mTLS traffic can be seen in the side-panel when selecting the edge. Note that the global
               masthead will show a lock icon when global mTLS is enabled.
             </div>
           </div>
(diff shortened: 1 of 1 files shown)


src/pages/Graph/GraphToolbar/GraphSettings.tsx
```diff
@@ -280,7 +280,15 @@ class GraphSettings extends React.PureComponent<GraphSettingsProps, GraphSetting
         id: 'filterSecurity',
         labelText: 'Security',
         isChecked: showSecurity,
-        onChange: toggleGraphSecurity
+        onChange: toggleGraphSecurity,
+        tooltip: (
+          <div style={{ textAlign: 'left' }}>
+            <div>
+              Show closed or open lock icons on edges that disagree with the global mTLS policy. Note that the global
```
jmazzitelli, line 287: "disagree" doesn't seem to be the right word here. Makes it seem like the value displayed is wrong or it should be something else.

Perhaps "contrasts" or "differs from"?

```suggestion
              Show closed or open lock icons on edges that differ from the global mTLS policy. Note that the global
```

src/pages/Graph/GraphToolbar/GraphSettings.tsx
```diff
@@ -280,7 +280,16 @@ class GraphSettings extends React.PureComponent<GraphSettingsProps, GraphSetting
         id: 'filterSecurity',
         labelText: 'Security',
         isChecked: showSecurity,
-        onChange: toggleGraphSecurity
+        onChange: toggleGraphSecurity,
+        tooltip: (
+          <div style={{ textAlign: 'left' }}>
```
mtho11, line 285: I personally think it is a little bit cleaner to declare the tooltip as a `const` rather than inline it.

PR #1895, opened 2020-08-25
Show login error message returned from back-end

israel-hdez:
Show in login screen the error that the back-end is outputting to better inform to the user what went wrong.

Now, we should see:

image

Fixes kiali/kiali#3150


israel-hdez:
@jmazzitelli You are omnipresent! :D

PR #1896, opened 2020-08-26
Add ability to drag nodes

jshaughn:
Introduce a new control on the cytoscape toolbar to toggle the ability to drag nodes (default=no drag).  This allows users to move nodes around to improve the layout manually for for visual inspection, screenshots, etc.

The manual repositioning is lost (and the toggle disabled) whenever there is a:

- change of graph type
- navigation away and back to the graph page (including node detail graphs)
- browser refresh

The manual repositioning is maintained on graph refresh unless a layout is required.  A layout change will not disable the toggle if it is currently enabled.   A layout is required when:

- the graph refreshes and nodes are lost or gained for the new time period.
- nodes are hidden via graph hide (and the Compress Hidden Display option is enabled)
- manual layout change via the cytoscape toolbar

Fixes https://github.com/kiali/kiali/issues/1342

### DIFF
--- a/src/components/CytoscapeGraph/CytoscapeReactWrapper.tsx
+++ b/src/components/CytoscapeGraph/CytoscapeReactWrapper.tsx
@@ -81,7 +81,6 @@ export class CytoscapeReactWrapper extends React.Component<CytoscapeReactWrapper
     const opts = {
       container: this.divParentRef.current,
       boxSelectionEnabled: false,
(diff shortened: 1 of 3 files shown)


pbajjuri20:
LGTM, I can drag nodes perfectly!


drag

lucasponce (approved):
Nice feature
image

As a user it performs 100% what I expected.

PR #1903, opened 2020-09-01
Enhance graph find/hide expression power

jshaughn:
Fixes https://github.com/kiali/kiali/issues/3165

Introduce the ability to combine AND and OR, as well as mix Node and Edge criteria.  For example, these expressions previously would generate an error, and now they work:

node=svc OR protocol=tcp

namespace=foo OR namespace=bar AND node=svc

cc @josejulio I think this is one you had wanted for a while (or maybe I am just dreaming)

### DIFF
--- a/src/pages/Graph/GraphHelpFind.tsx
+++ b/src/pages/Graph/GraphHelpFind.tsx
@@ -171,7 +171,8 @@ export default class GraphHelpFind extends React.Component<GraphHelpFindProps> {
       ['!traffic', 'edges with no traffic'],
       ['http > 0.5', `edges with http rate > 0.5 rps`],
       ['rt > 500', `edges with response time > 500ms. (requires response time edge labels)`],
-      ['%httptraffic >= 50.0', `edges with >= 50% of the outgoing http request traffic of the parent`]
+      ['%httptraffic >= 50.0', `edges with >= 50% of the outgoing http request traffic of the parent`],
+      ['node = svc and svc startswith det or !traffic', 'service node starting with "det" or edges with no traffic']
     ];
   };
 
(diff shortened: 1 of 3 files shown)


mtho11:
without parens what is the precedence?

jshaughn:
> without parens what is the precedence?

Although precedence doesn't often come into play, AND has precedence over OR.  Operators all have the same weight. 
Evaluation is basically:

   `expression1 OR expression2 OR ... expressionN`

Where each expression can be:

   `condition1 AND condition2 AND ... conditionN`

In rare cases you may have to "distribute" a condition verbosely.  Let's say I want to find **only** app nodes with incoming requestsPerSecond > 5 or outgoing rps > 10.  This won't get you what you want:

  `node=app AND httpin > 5 OR httpout > 10`

Because `node=app` will be considered only in expression1.  So you will find app nodes with `httpin > 5`, but also all nodes with `httpout > 10`.  More likely you'd want:

  `node=app AND httpin > 5 OR node=app AND httpout > 10`

mtho11:
@jshaughn thanks for the detailed explaination

josejulio:
> cc @josejulio I think this is one you had wanted for a while (or maybe I am just dreaming)

That's correct, we briefly talked about this capability before (i think) :smile:

josejulio:
Note that I *think* (just an idea) that this could be further improved in the future by evaluating each expression by separate and using cytoscape functions like [union](https://js.cytoscape.org/#eles.union) and [intersection](https://js.cytoscape.org/#eles.intersection) to evaluate the `or`s and `and`s. Grouping could be possible with a more complex syntax this way.

pbajjuri20:
Tested and Passed on Kiali UI 1.24.0 (5e6dc71bf21e6914589cfc811732ed34656050f0)

jshaughn:
@josejulio @jmazzitelli , I agree things could be more powerful but until community asks for more I think this should be good as it is a big step up from what we have had for quite some time.  As for precedence, I may add that example above as a FAQ.

PR #1904, opened 2020-09-02
Refactor IstioDetails Overview Tab

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3161

@hhovsepy I'm afraid this would impact the automation but I think it's a cleaner approach:

image
image

So, with this suggestion, Istio Config details get aligned and additional cards can be added per type to show some specific use case (i.e. in the future we may want to add some extra info in the security area).


xeviknal (approved):
LGTM! Less clicks and only keeping the useful information.

hhovsepy (approved):
Verified:
Screenshot from 2020-09-02 15-17-03
Screenshot from 2020-09-02 15-17-37
Screenshot from 2020-09-02 15-31-17
Screenshot from 2020-09-02 15-34-08

PR #1906, opened 2020-09-02
Improve graph find/hide autocomplete

jshaughn:
- fix issue when there are no completions
- support completion with ! operator
- support completion of more than just the first operand
- update help

Related to https://github.com/kiali/kiali/issues/1411

### DIFF
--- a/src/pages/Graph/GraphHelpFind.tsx
+++ b/src/pages/Graph/GraphHelpFind.tsx
@@ -210,7 +210,8 @@ export default class GraphHelpFind extends React.Component<GraphHelpFindProps> {
   };
   private noteRows = (): string[][] => {
     return [
-      ['Parentheses are not supported (or needed).'],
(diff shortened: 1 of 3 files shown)


jmazzitelli (approved):
This works, however, I have another nit that can be corrected in a future PR.

When you auto-correct (hit the TAB key) while at the end of the text field, the auto-corrected text shows, but the text field isn't scrolled to the right so you can't see the right-side of the new text that was automatically inserted. It would be nice to move the cursor to the end of the text field and have the text field scroll to the right so you can see the text on the far right that was just added.

See:

auto-complete

jshaughn:
@jmazzitelli , I think I fixed your issue.  Also fixed a separate issue.

pbajjuri20:
@jmazzitelli, Issue fixed perfectly.


fix find error

pbajjuri20 (approved):

PR #1913, opened 2020-09-07
Fault Injection Wizard

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3178

Part of https://github.com/kiali/kiali/issues/3174

The aim of this PR is to align Kiali Wizards with Istio Management scenarios described in the tasks:
image

In that way, the user will map same concepts to operate Istio.

image

The "Suspend Traffic" Wizard has been migrated to a direct "Fault Injection" wizard:

image

And the "Request Routing" Wizard it's becoming a more rich use case, allowing to add multiple routes combining "matching", "traffic shifting" and "fault injection" features:

image

Also, some minor styles:

image


xeviknal:
It happens in both creation and modification:
fault-wizard

Nitpick: the numeric input does a not natural behavior when replacing the %
The http code number works good here instead. When is not valid, disable the submit. When is back to work, it enables it.
fault-whole-wizard

lucasponce:
Thanks @xeviknal.

Comments addressed.

hhovsepy (approved):
Verified, now it adds auto completes  0% in case of wrong input, or 100% in case of >100%
Screenshot from 2020-09-08 14-51-06
Screenshot from 2020-09-08 14-54-47
Screenshot from 2020-09-08 14-55-39

hhovsepy:
Screenshot from 2020-09-08 15-03-17

PR #1914, opened 2020-09-08
Enable traces for app and workload nodes

jotak:
Fixes https://github.com/kiali/kiali/issues/3048

- Update links to point to new trace details pages (app/service/workload)
- Make it work for any graph type
- Did some experiments with service entry / TCP related traces, added a way to not break the trace when such node is reached (e.g. kafka)
(commented as experimental, probably more thoughts on this is necessary, but that's a starting point)
- Side-change: proper typing on destServices

### DIFF
--- a/src/components/CytoscapeGraph/CytoscapeGraph.tsx
+++ b/src/components/CytoscapeGraph/CytoscapeGraph.tsx
@@ -216,7 +216,7 @@ export default class CytoscapeGraph extends React.Component<CytoscapeGraphProps>
     }
 
     if (this.props.trace) {
-      showTrace(cy, this.props.trace);
+      showTrace(cy, this.props.graphData.fetchParams.graphType, this.props.trace);
     } else if (!this.props.trace && prevProps.trace) {
       hideTrace(cy);
     }
--- a/src/components/CytoscapeGraph/CytoscapeTrace.ts
+++ b/src/components/CytoscapeGraph/CytoscapeTrace.ts
@@ -3,57 +3,75 @@ import { CyNode } from './CytoscapeGraphUtils';
 import { JaegerTrace, Span } from 'types/JaegerInfo';
 import {
   getWorkloadFromSpan,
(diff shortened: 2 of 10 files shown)


jotak:
Some screenshots,

- App graph type, traces from app nodes:

Capture d’écran de 2020-09-08 12-28-07

- Workload graph type, traces from workload nodes:
Capture d’écran de 2020-09-08 12-28-23

- This third one showcases the experimental support of TCP service entries, here kafka
Capture d’écran de 2020-09-08 13-04-31

jshaughn (changes requested):
@jotak Looking good, getting very powerful.  Two things I think we need to change (as we discussed off-pr):

1. At startup, or when visiting a bookmark, when there is a traceId in the URL and that Trace can not be found we currently generate an Error with a toast:  "Could not fetch trace"/"unexpected end of JSON input".  I think in this scenario we should more silently handle it by cleaning the URL and generating a non-toast Info level message for the message center.
2. When changing namespaces, or graph types, the traceId is not cleared and the in-graph trace is maintained.  This can be powerful, especially to see the trace applied across graph types.  But we lose the "handle" on the active trace because the side-panel is re-focused on the full-graph.  Whenever there is an active traceId we should maintain the trace-detail in the side-panel, given the user a chance to close the trace.

jotak:
@jshaughn your "2" is done; "1" still wip

jshaughn:
> @jshaughn your "2" is done; "1" still wip

Tried it, looks good!

jotak:
@jshaughn and here's the fix for "1". By the way, this backend PR https://github.com/kiali/kiali/pull/3190 will make this error properly recognized as a "trace not found" issue rather than an obscure 503 / unmarshalling issue.

jshaughn (approved):
LGTM, good changes to AlertUtils.

mattmahoneyrh (approved):

PR #1915, opened 2020-09-08
Add Status field into Istio Objects

xeviknal:
Requires https://github.com/kiali/kiali/pull/3188
Fixes kiali/kiali#2236

There is a decoration: the status field is folded in order not to eat a lot of space and also avoid showing redundant information.

Demo:
analyzer-messages

### DIFF
--- a/src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
+++ b/src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
@@ -353,8 +353,8 @@ class IstioConfigDetailsPage extends React.Component<RouteComponentProps<IstioCo
 
   getStatusRange = (yaml: string | undefined): any => {
     let range = {
-      startRow: 0,
-      endRow: 0
+      startRow: -1,
+      endRow: -1
     };
 
     if (!!yaml) {
@@ -363,7 +363,7 @@ class IstioConfigDetailsPage extends React.Component<RouteComponentProps<IstioCo
         if (line.startsWith('status:')) {
           range.startRow = i;
         }
-        if (line.startsWith('spec:') && range.startRow !== 0) {
+        if (line.startsWith('spec:') && range.startRow !== -1) {
           range.endRow = i;
         }
       });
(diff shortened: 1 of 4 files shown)


lucasponce:
> There is a decoration: the status field is folded in order not to eat a lot of space and also avoid showing redundant information.

Ah, I was to comment that :-).
The status field is an uber copy of the whole object plus additional info, so a helper to avoid this is interesting.

Note that we need to test how it works with the patches, in theory it should be transparent but there may be some corner cases that may affect the create/update scenarios.

In any case, I'll review in detail too.

xeviknal:
I was about to place the status at the end of the yaml, however, it was easier like this bc of the backend part gets easier.

Regarding the patches, I've tried to update different objects and nothing wrong happened. It looks like safe.

lucasponce:
I'm having this experience:
image

All the fields collapsed by default.

Is that intentional ?

lucasponce (changes requested):
Can we check that only the status field get collapsed ?
Otherwise experience is degraded when access to the Istio Config detail.

src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
```diff
@@ -192,8 +193,16 @@ class IstioConfigDetailsPage extends React.Component<RouteComponentProps<IstioCo
     // Hack to force redisplay of annotations after update
     // See https://github.com/securingsincity/react-ace/issues/300
     if (this.aceEditorRef.current) {
+      const editor = this.aceEditorRef.current!['editor'];
+
       // tslint:disable-next-line
-      this.aceEditorRef.current!['editor'].onChangeAnnotation();
+      editor.onChangeAnnotation();
+
+      // Fold Status field
```
lucasponce, line 201: Ok, if the status is not present, this is failing in my case I guess I didn't test it with the status enabled.

xeviknal:
@lucasponce changes approached:

Without status field, there are no fold in place nor the analyzers title in the sidepanel:
Screenshot of Kiali Console (42)

With status and analyzer messages:
Screenshot of Kiali Console (43)

With status but no analyzer messages:
Screenshot of Kiali Console (44)

lucasponce (approved):

prachiyadav:
Screenshot from 2020-09-10 20-13-01

prachiyadav:
post merge verified

PR #1916, opened 2020-09-08
Request Timeouts Wizard

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3183

Add a new "Request Timeouts" Wizard to cover server timeout and retries.

image

Similar as "Fault Injection" one it creates a simplified scenario for a one-http-route case:

image

And it adds the "Request Timeouts" support to the "Request Routing" Wizard for complex scenarios:

image


xeviknal:
In the RequestRouting wizard, after applying a rule with both route, fault injection and time out, the tab with the fault injection is reseted whereas the other 2 have the same value. It sounds a bit inconsistent though. 

wizard-reqt

xeviknal (commented):
LGTM! There is one comment about initial values (see video) I'd like to hear of.

prachiyadav (approved):
Screenshot from 2020-09-10 17-18-38

Screenshot from 2020-09-10 17-13-23
Screenshot from 2020-09-10 17-18-28

Screenshot from 2020-09-10 17-05-15

lucasponce:
> LGTM! There is one comment about initial values (see video) I'd like to hear of.

Right, I'm working on them, and also in some improvements in the Matching tab as well.

lucasponce:
@xeviknal @prachiyadav I've send a commit to address Xavi's comment plus another issue I've found after more deep testing:
- Now the state in the builder is kept after a new rule addition, that will help to introduce more patterns.
- In the matching, there was some duplication on potential matching, that's also updated.

Thanks.

prachiyadav:
Screenshot from 2020-09-11 12-03-42
Screenshot from 2020-09-11 12-04-03
Screenshot from 2020-09-11 12-04-11

PR #1924, opened 2020-09-15
Traffic Policies actions on Namespace

lucasponce:
Fixes https://github.com/kiali/kiali/issues/2924

This PR introduces a new Namespace based action triggered from the Overview page.

Given a certain traffic:

image

Kiali can apply Traffic Policies (Istio AuthorizationPolicies) to lock in workloads with principals described in the telemetry.

The action is added into the Namespace menu:
image
image

And the list of AuthorizationPolicies can be listed from the "Istio Config" page:
image

Or individually per workload (when workload selectors are used):
image

There is a generic DENY_ALL rule per namespace to protect the workloads, with individual ALLOW rules per workload, like described in this one:

image

### DIFF
--- a/src/pages/Overview/OverviewPage.tsx
+++ b/src/pages/Overview/OverviewPage.tsx
@@ -215,13 +215,15 @@ export class OverviewPage extends React.Component<OverviewProps, State> {
           : OverviewDisplayMode.EXPAND;
         // Set state before actually fetching health
         this.setState(
-          {
-            type: type,
-            namespaces: Sorts.sortFunc(allNamespaces, sortField, isAscending),
-            displayMode: displayMode,
-            showConfirmModal: false,
-            nsTarget: '',
-            opTarget: ''
+          prevState => {
+            return {
+              type: type,
+              namespaces: Sorts.sortFunc(allNamespaces, sortField, isAscending),
+              displayMode: displayMode,
+              showConfirmModal: prevState.showConfirmModal,
+              nsTarget: prevState.nsTarget,
+              opTarget: prevState.opTarget
+            };
           },
           () => {
             this.fetchHealth(isAscending, sortField, type);
@@ -654,7 +656,9 @@ export class OverviewPage extends React.Component<OverviewProps, State> {
 
   hideConfirmModal = () => {
     this.setState({
-      showConfirmModal: false
+      showConfirmModal: false,
+      nsTarget: '',
+      opTarget: ''
     });
   };
 
@@ -667,11 +671,15 @@ export class OverviewPage extends React.Component<OverviewProps, State> {
         break;
       }
     }
+    const nsTarget = this.state.nsTarget;
+    const remove = this.state.opTarget === 'delete';
     this.setState(
       {
-        showConfirmModal: false
+        showConfirmModal: false,
+        nsTarget: '',
+        opTarget: ''
       },
-      () => this.onAddRemoveAuthorizationPolicy(this.state.nsTarget, aps, this.state.opTarget === 'delete')
+      () => this.onAddRemoveAuthorizationPolicy(nsTarget, aps, remove)
     );
   };
 
--- a/src/pages/Overview/OverviewPage.tsx
+++ b/src/pages/Overview/OverviewPage.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import {
+  Button,
   Card,
   CardActions,
   CardBody,
@@ -10,6 +11,7 @@ import {
   EmptyStateVariant,
   Grid,
   GridItem,
+  Modal,
   Title,
   Tooltip,
   TooltipPosition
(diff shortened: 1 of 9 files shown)


lwrigh:
Mini card actions dropdown

For the kebab menu in the card I would suggest using a header label for the display options and then using a divider to separate out the other actions.

lucasponce:
Thanks @lwrigh. I'm going to fill a separate issue for that recommendation as probably there are more places to apply same pattern.

src/pages/Overview/OverviewPage.tsx
```diff
@@ -376,12 +381,18 @@ export class OverviewPage extends React.Component<OverviewProps, State> {
   fetchValidationChunk(chunk: NamespaceInfo[]) {
     return Promise.all(
       chunk.map(nsInfo => {
-        return API.getNamespaceValidations(nsInfo.name).then(rs => ({ validations: rs.data, nsInfo: nsInfo }));
+        return Promise.all([
+          API.getNamespaceValidations(nsInfo.name),
+          API.getIstioConfig(nsInfo.name, ['authorizationpolicies'], false, '', '')
```
xeviknal, line 386: Should we add the `kiali_wizard=AuthorizationPolicy` in this call? This way, clicking into the `Delete Authz Policies` will only delete the AuthzPolicies created by this wizard.

xeviknal (commented):
@lucasponce is intended that when you click the `Delete traffic policies`, other Authz Policies non created by the wizard are deleted too? I'd expect to only remove the Authz Policies created by that wizard.

xeviknal:
One extra comment, does the update action considers the deletion of services? For example, if you remove the service reviews-v1 in the bookinfo, does the update remove the Authz Policy added for this reviews-v1 service?

src/pages/Overview/OverviewPage.tsx
lucasponce: I was unsure but now you bring this I guess we should keep consistency here and only delete the "managed" resources and let user delete other assets.

lucasponce:
> One extra comment, does the update action considers the deletion of services? For example, if you remove the service reviews-v1 in the bookinfo, does the update remove the Authz Policy added for this reviews-v1 service?

Nope, that's an uber task that I think it's outside Kiali scope for the moment.

I mean, Kiali can check existence of those services but linking this kind of operation in cascade is risky (and probably we would like to play it at webhooks level, i.e. when a service is deleted, internally a hook searches for other related resources and delete them).

Doing that only from Kiali sounds to open more corner cases and edges rather than help.

These actions we are adding are targetting for the "new" user not the "advanced" one.
So as a user I'd like to generate a base config that works and probably I want to modify and extend later manually when I have more knowledge of Istio.

src/pages/Overview/OverviewPage.tsx
xeviknal: It seems a bit of a toggle action: create a set of policies, remove that set of policies. Removing other policies would lead to some frustration, specially because there isn't any alert saying what is happening after clicking this. I can imagine people in production performing this operation and removing extra policies.

lucasponce: Actually we delete all info, not only the Kiali generated one, so I vote to keep this as it is at least in a first iteration.

xeviknal (approved):
LGTM!

lucasponce:
Discussed with @xeviknal, this feature misses a modal on "update/delete" action.

src/pages/Overview/OverviewPage.tsx
```diff
@@ -210,7 +218,10 @@ export class OverviewPage extends React.Component<OverviewProps, State> {
           {
             type: type,
             namespaces: Sorts.sortFunc(allNamespaces, sortField, isAscending),
-            displayMode: displayMode
+            displayMode: displayMode,
+            showConfirmModal: false,
```
xeviknal, line 222: I guess that this should be set to the previous value of this very same field. Otherwise, when the health is refreshed (via the drop-down (e.g. Every 10s) the modal is hidden.
Happen to me, if you are reading the modal more than 10s and the refresh is triggered, the modal is gone.

xeviknal (changes requested):
There is a little unexpected behavior in the modal.

src/pages/Overview/OverviewPage.tsx
lucasponce: > Happen to me, if you are reading the modal more than 10s and the refresh is triggered, the modal is gone.

Ah, true

lucasponce: @xeviknal updated

xeviknal (approved):
LGTM! Comments approached. tested locally.

hhovsepy:
Post merge verification.
Tested Create/Update/Delete Traffic Policies.
Update Traffic Policies will override any edited/changed Traffic Policies.
Screenshot from 2020-09-16 14-47-51
Screenshot from 2020-09-16 14-48-28
Screenshot from 2020-09-16 14-51-23

PR #1928, opened 2020-09-16
Group Show Actions

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3217

Overview page:
image
image
image

Graph:
image
image
image

cc @lwrigh

### DIFF
--- a/src/components/CytoscapeGraph/ContextMenu/NodeContextMenu.tsx
+++ b/src/components/CytoscapeGraph/ContextMenu/NodeContextMenu.tsx
@@ -23,6 +23,15 @@ const graphContextMenuTitleStyle = style({
   borderBottom: '1px solid black'
 });
 
+const graphContextMenuSubTitleStyle = style({
+  textAlign: 'left',
+  fontSize: '14px',
+  color: '#737679',
+  fontWeight: 700,
+  paddingTop: 2,
+  paddingBottom: 4
+});
+
 const graphContextMenuItemStyle = style({
   textAlign: 'left',
   fontSize: '12px',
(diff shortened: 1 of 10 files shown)


jshaughn (changes requested):
I like the idea of grouping the kebab and action menus but I think the design/PR still needs some work:
- I find this hard to parse: I think the options need to be indented under the title (not sure what Laura may think):
image

- The overview kebab is different than others I see in the graph, the title is centered:
image


- The Service actions should be consistent:
image

lucasponce:
> I find this hard to parse: I think the options need to be indented under the title (not sure what Laura may think):

I disagree. I don't want to deviate from PF standard as much as we can, otherwise we finish creating a custom UI where everyone has an opinion.

This is the model and it's what is recommended by UX why I stand.

image


> The overview kebab is different than others I see in the graph, the title is centered:

There is a new scss that needs to be built, how have you tested ? Perhaps there is missing css link there but it should be identical.

> The Service actions should be consistent:

Group is only on "Show" actions, "Create" are independent as they have state and may be individually "Create" or "Update".

Actions with state are not grouped.

jshaughn (approved):
OK, I looked for the PF patern for the group in a dropdown but was looking in the wrong place.   I don't like the lack of indentation and would prefer the verbose options.  But if that's the style I can't argue with it.   The centered group issue was due to my needing to restart yarn start.  And although it seems inconsistent I understand the need to leave the action menu as is.  So, I will approve.

lucasponce:
I'm going to test the group in services, perhaps in this case same pattern can be applied

lucasponce:
Ok, I think that Services Action can finally be :

image

Trick is that when there is one scenario in place, the user can only update that scenario, disabling others, then changing the group label is consistent with the pattern:

image

Other options are disabled unless user remove the scenario.

I guess on this case we can address your last comment @jshaughn

lucasponce:
Also, note that not all actions can be grouped, some of them with individual state will remain as it is, like this case:
image

xeviknal (approved):
LGTM! All dropdown look coherent to the original request.

jshaughn (commented):
re-approving after the latest changes.

lucasponce:
@prachiyadav testing this one is to validate that the "Show" actions in the Namespace are showed in a group as it's in the screenshot and same for "Create/Update" actions under Service.

PR #1930, opened 2020-09-18
Fix regression for RequestRouting Wizard

lucasponce:
During https://github.com/kiali/kiali/issues/3174 I've introduced a regression in the Request Routing Wizard.

Wizard allow users to create *multiple* Request Matching for a Route.

image

The case about having a complex *single* Request Matching is not covered by the Wizard and it's delegated to Istio Config yaml editor.


xeviknal (approved):
LGTM! Multiple uris can be appended

hhovsepy:
Post merge verified.

PR #1932, opened 2020-09-21
Refactor Istio Details: Bump Editor + Overview Drawer

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3196
image
image

Please @xeviknal @hhovsepy try to make extra tests on validations, I've bumped the ace editor dependency (it was pretty old) and I did some smoke tests but I'd like to confirm that nothing is broken with the features around the editor.

This is not a bug, but also thinking loud if we may want to cherrypick this to 1.24 branch to avoid future problems due the old editor dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
     "m-react-splitters": "1.2.0",
     "moment": "2.24.0",
     "react": "16.13.1",
-    "react-ace": "6.4.0",
+    "react-ace": "9.1.3",
     "react-copy-to-clipboard": "5.0.1",
     "react-datepicker": "2.14.0",
     "react-dom": "16.13.1",
--- a/src/components/IstioActions/IstioActionsButtons.tsx
+++ b/src/components/IstioActions/IstioActionsButtons.tsx
@@ -18,6 +18,9 @@ type Props = ReduxProps & {
   onCancel: () => void;
   onUpdate: () => void;
   onRefresh: () => void;
+  showOverview: boolean;
+  overview: boolean;
+  onOverview: () => void;
 };
 
 type State = {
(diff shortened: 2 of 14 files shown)


hhovsepy:
Now the red backgrounds are shown per section.
Before:
Screenshot from 2020-09-22 12-59-16
Screenshot from 2020-09-22 12-59-57

After:
Screenshot from 2020-09-22 13-00-32
Screenshot from 2020-09-22 13-01-13

lucasponce:
> Now the red backgrounds are shown per section.

Okey, I'll take a look, I think I changed this during the bump as there was not a 1to1 mapping with the config and I need to adapt it.

lucasponce:
@hhovsepy okey, fixing several issues due the bump of the editor, let me know how it looks with the regression tests.

Thanks !

hhovsepy (approved):
Verified, now it highlights only the wrong line, the whole line, also the Overview section works.
Screenshot from 2020-09-22 19-27-07
Screenshot from 2020-09-22 19-30-50

lucasponce:
Thanks @hhovsepy I missed to update some tests, I will update the PR.

package.json
```diff
@@ -80,7 +80,7 @@
     "m-react-splitters": "1.2.0",
     "moment": "2.24.0",
     "react": "16.13.1",
-    "react-ace": "6.4.0",
+    "react-ace": "9.1.3",
```
xeviknal, line 83: With this upgrade, have the foldings still work? I am thinking of the scenario where the status field is present in the yaml.

lucasponce: I haven't tested it, going to see if I can take some screenshots.

xeviknal:
It seems it also keep the folding working!
Screenshot of Kiali Console (47)

xeviknal (approved):
LGTM!

lucasponce:
> It seems it also keep the folding working!

Great, I know it's a big bump but we had a very old version and I think if no blockers it could be good to upgrade and also put this on 1.24.

PR #1933, opened 2020-09-22
Trace on graph: comply to opentracing standard

jotak:
Fixes https://github.com/kiali/kiali/issues/3226

Read some semantic conventions (https://github.com/opentracing/specification/blob/master/semantic_conventions.md) to extract span data:

- HTTP conventions
- peer / tcp conventions + messaging

Also read Envoy data which is a superset of OT http conventions

As a result, there's 4 different kind of displays depending on span kind:

- The usual Envoy:
Capture d’écran de 2020-09-22 19-14-50

- The non-Envoy HTTP:
Capture d’écran de 2020-09-22 19-14-39

- The TCP / messaging:
Capture d’écran de 2020-09-22 19-14-12

- The "unknown kind":
Capture d’écran de 2020-09-22 19-14-07

### DIFF
--- a/src/components/CytoscapeGraph/CytoscapeTrace.ts
+++ b/src/components/CytoscapeGraph/CytoscapeTrace.ts
@@ -3,9 +3,9 @@ import { CyNode } from './CytoscapeGraphUtils';
 import { JaegerTrace, Span } from 'types/JaegerInfo';
 import {
   getWorkloadFromSpan,
-  searchParentWorkload,
+  getAppFromSpan,
   searchParentApp,
-  getAppFromSpan
+  searchParentWorkload
 } from 'components/JaegerIntegration/JaegerHelper';
 import { NodeType, DestService, GraphType } from 'types/Graph';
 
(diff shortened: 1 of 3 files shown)


lucasponce:
I have a couple of comments:
image

- "From: <time>" and "From: <location>" it's a little confusing, but it's minor, I'm ok if you want to leave it as it is.

Also the trace details seems without format, if you don't want to use a grid / table, at least I suggest to follow a similar patter used in the details:

image

Just to separate with a bold the field vs value for better readability, as it is is harder to focus on the info, it needs some structure.

lucasponce:
I imagine that this logic will also apply to the refactoring of the trace details in the trace tab as well.

jotak:
> I have a couple of comments:
> image
> 
>     * "From: _time_" and "From: _location_" it's a little confusing, but it's minor, I'm ok if you want to leave it as it is.

Indeed, I didn't notice that

> 
> Also the trace details seems without format, if you don't want to use a grid / table, at least I suggest to follow a similar patter used in the details:
> 
> image
> 
> Just to separate with a bold the field vs value for better readability, as it is is harder to focus on the info, it needs some structure.

I think it needs to be a little bit condensed to avoid triggering a scrollbar. But yes, I agree, a little bit of styling should work well.

jotak:
> I imagine that this logic will also apply to the refactoring of the trace details in the trace tab as well.

Exactly

jotak:
Updated style:
Capture d’écran de 2020-09-23 14-36-33

jshaughn (approved):
I'm +1 to @lucasponce's comments and also to @jotak's response to the comments.  From/From -> Start/To is a good change.  And we definitely want to avoid scrolling in the trace detail, using bold labels is much more clear and I think satisfies the comment. This all LGTM.

lucasponce (approved):
Much better now:
image

PR #1943, opened 2020-09-23
Migrate detail view traffic tab to standard list look+feel

jshaughn:
Confusion with the Traffic tab's hierarchical view resulted in various UX proposals to change the view.  The resulting choice was to simplify to a single row approach, and to make it consistent with the standard list views we currently employ for workloads, applications, etc.  This also provides for sortable colums which have been requested by the community.

The Inbound/Outbound traffic lists are now manged by tabs.  List rows represent immediate inbound or outbound nodes (without service node injection).  Note that the "View Metrics" link has been lost in the migration, I'm not sure if we need it going forward.

Before:
image

After:
image

image

Fixes https://github.com/kiali/kiali/issues/1535
Fixes https://github.com/kiali/kiali/issues/3176


lucasponce (changes requested):
I wouldn't use the VirtualList for implementing this.
Instead I suggest to use a simple table like IstioConfigSubList or other lists used in subtabs.

VirtualList has a defect that means that you can't personalize as you'd like the columns without introducing regressions to other lists.

I think for this work, using a simple table is enough and would make the PR easier.

P.S.: Another good example that can help on this is the ExperimentListContainer, which also uses a Table with sorting.

@jshaughn I can also send a PR on top of your work if that helps, perhaps it's easier to me to help in that way.

jshaughn:
@lucasponce I obviously misinterpreted your recommendation to use a table format that was consistent with our other sortable table/list view.  After putting a decent amount of effort into this I'm not liking the idea of abandoning what seems to be working in favor of another impl.  Can you be more specific about the problem you foresee with using VirtualList?  I didn't really understand what you meant by, "VirtualList has a defect that means that you can't personalize as you'd like the columns without introducing regressions to other lists".  I don't see regressions in the other lists.

lucasponce:
> After putting a decent amount of effort into this I'm not liking the idea of abandoning what seems to be working in favor of another impl. 

I would be good if you could ask before to prepare the PR, then this may happen.

> Can you be more specific about the problem you foresee with using VirtualList?

VirtualList has some limitations in how can be personalized and it's easier to change something in one list and break other, or not be able to personalize the columns.

In short, it's some kind of technical debt that works for the main list that is used but it's in the TODO to refactor by a better model.

If you don't want to re-org the PR I understand, as I told you I will send a PR on top of this, but I won't approve to use the VirtualList because that component need to be changed and adding more logic on top it's in the opposite direction.

jshaughn:
> I would be good if you could ask before to prepare the PR, then this may happen.

Ask what, If I should use VirtualList?  Perhaps you should have been more clear that sortable table views that are consistent with our other pages *were not* to be done with VirtualList, the exact component used in our four main views of that type.  You wrote in the issue, " just align the current info using a table as part of the consistency work we are doing around that ... let's user to sort by columns when possible.".    You saw that I was assigned, if there was a huge impl note I should have known about I think it would have been nice for you to let me know in advance.

>  as I told you I will send a PR on top of this, but I won't approve to use the VirtualList because that component need to be changed and adding more logic on top it's in the opposite direction.

I was able to apply VirtualList, I think I can manage a different approach as well.  I will also add a LARGE NOTE in VirtualList that it is NOT TO BE USED.

lucasponce:
> You saw that I was assigned, if there was a huge impl note I should have known about I think it would have been nice for you to let me know in advance.

I didn't see you use this implementation until you send the PR, hence here I am discussing about this.

>  I will also add a LARGE NOTE in VirtualList that it is NOT TO BE USED.

VirtualList has its context and for secondary tables it's better to use the implementation alternative I mention.

Again, no need to touch your PR if you don't want, I can send my changes to this to also make this scenario more clear.

jshaughn:
@lucasponce @israel-hdez This is again ready for review after reverting use of VirtualList and applying the model of ExperimentListContainer.  Also, I re-introduces the "View metrics" in an Actions column to better maintain existing functionality.

src/components/VirtualList/VirtualList.tsx
```diff
@@ -14,6 +14,12 @@ import * as FilterHelper from '../FilterList/FilterHelper';
 import * as Sorts from '../../pages/Overview/Sorts';
 import { StatefulFilters } from '../Filters/StatefulFilters';
 
+// ******************************
+// VirtualList and its associated classes are deprecated and not to be used for new
```
lucasponce, line 18: I'd change this comment for something more technical with more info.

VirtualList and its associated classes are intented to be used for main list pages: Applications, Workloads, Services and Istio Config. They share common style and filter integration, so they have some limitations when you want to use it for other scenarios that may require different personalization (columns style, or layout).

For secondary list rendered inside a detail page consider to use the Table component as used in WorkloadServices, WorkloadPods, ServiceInfoWorkload, IstioConfigSubList, ExperimentListContainer or TrafficListComponent.

src/pages/ServiceDetails/ServiceDetailsPage.tsx
```diff
@@ -12,9 +12,9 @@ import { DurationInSeconds } from '../../types/Common';
 import { durationSelector } from '../../store/Selectors';
 import ParameterizedTabs, { activeTab } from '../../components/Tab/Tabs';
 import ServiceInfo from './ServiceInfo';
-import TrafficDetails from '../../components/Metrics/TrafficDetails';
 import TracesComponent from 'components/JaegerIntegration/TracesComponent';
 import { JaegerInfo } from 'types/JaegerInfo';
+import TrafficDetails from 'pages/TrafficList/TrafficDetails';
```
lucasponce, line 17: TrafficList is not a page, it's a component.
Specific components used for a page can go inside.
Shared components across all pages should go as a component.

src/pages/TrafficList/TrafficListComponent.tsx
```diff
@@ -0,0 +1,335 @@
...
+  {
+    title: 'Rate',
+    transforms: [sortable]
+  },
+  {
+    title: 'Percent',
+    transforms: [sortable, cellWidth(5) as any]
+  },
+  {
+    title: 'TrafficStatus',
```
lucasponce, line 67: In sublist with status (WorkloadPods, IstioConfigSubList) status is preffered to be the first column.

src/pages/TrafficList/TrafficListComponent.tsx
```diff
@@ -0,0 +1,335 @@
...
+  render() {
+    return (
+      <div className={containerPadding}>
+        <Table
+          aria-label="Sortable Table"
+          sortBy={this.state.sortBy}
+          cells={columns}
+          rows={this.rows()}
+          onSort={this.onSort}
+        >
```
lucasponce, line 119: In PF Table component, rows are generated from a structure passed into the component.
For sublist, consider adding the `className="table"` style to fix the vertical align, specially on icons,

src/pages/TrafficList/TrafficListComponent.tsx
```diff
@@ -0,0 +1,335 @@
...
+type TrafficListComponentProps = FilterComponent.Props<TrafficListItem> & {
+  direction: 'inbound' | 'outbound';
+  trafficItems: TrafficItem[];
+};
+
+type TrafficListComponentState = FilterComponent.State<TrafficListItem> & {
+  sortBy: ISortBy;
+};
+
+const columns = [
```
lucasponce, line 49: Now with this change of using directly Table instead of the previous component, you can adjust the columns to your needs.
I.e. putting status first with a short width and adding more space to those columns it may need more space, and reducing some fields that you need width would be short.

This kind of things can't be perform in the VirtualList component for a specific component, now it gives you more flexibility.

src/pages/TrafficList/TrafficListComponent.tsx
```diff
@@ -0,0 +1,335 @@
...
+        >
+          <TableHeader />
+          {this.state.listItems.length > 0 ? (
+            <TableBody />
+          ) : (
+            <tr>
+              <td colSpan={columns.length}>
+                <EmptyState variant={EmptyStateVariant.full}>
+                  <Title headingLevel="h5" size="lg">
+                    `No {this.props.direction} Traffic`
```
lucasponce, line 128: No need to add the `` quotes
image

lucasponce:
This is a comment more about the feature itself, not about implementation:

image

I wonder if it's clean just to mix both inbound and outbound with a column.

Now that it's sortable, it's easier to go from one to another and it may be handy to compare rates.

src/pages/TrafficList/TrafficListComponent.tsx
jshaughn: @lucasponce Do you recomment Traffic Status in the first column?  I was OK with the current column ordering but am open to changes.  I will look at the widths and make any adjustments I see fit.

src/pages/TrafficList/TrafficListComponent.tsx
jshaughn: Perhaps I should read all comments first :)  Ignore the question above...

src/pages/TrafficList/TrafficListComponent.tsx
jshaughn: @lucasponce I modeled this after ExperimentListContainer.  Shall I add the same recommended change in that class as part of this PR?

lucasponce: Yes, thanks. I've noticed it has same vertical alignment problem:
image

lucasponce:
Or if you want to keep two tables separated for some reason, I'd put it together for better visibility rather than two tabs.
Like in the old design:
image

But my preference would be to put it together in the same table with a sortable 'inbound/outbound' column.

jshaughn:
> I wonder if it's clean just to mix both inbound and outbound with a column.

I'm not sure, I think I'd find it confusing to see a mix of inbound and outbound traffic in the same list, even if it were sortable, as when not sorted on 'direction' it would be a mix.

Comparing inbound/outbound rates may be easier but I'd maybe suggest two tables on the same page, not unlike before.  But two tables increases the possibility of vertical scroll, which is possibly worse than tabs.  I'd like to hear from @israel-hdez or possibly UX before making a change from the tabs.   We often split inbound/outbound, like in the metrics tabs, so it's not inconsistent to have tabs used here in a similar way.

src/pages/TrafficList/TrafficListComponent.tsx
jshaughn: @lucasponce After our off-pr discussion we have decided that the className="table" is not necessary.  So I will ignore.

jshaughn:
@lucasponce I addressed your pr comments (not the feature comment, waitng on more input there).  Also, I worked on the sorting confusion I mentioned, inherited from ExperimentListContainer, which is currently sort of mixing two strategies.  I have a pattern working here that I think is pretty good.  It maintains use of FilterComponent for SortField definition and URL bookmarking, while also deferring to table for updating the current sort and direction.  Please re-review TrafficListComponent as there were a variety of changes.  Thanks.

israel-hdez (commented):
Partial review yet.
The only thing I see weird for now is the sorting behavior. Try this:
* Sort a column in descending mode (for example, the Rate column)
* Sort another column (for example, the name column. Just to change the sorting column, direction is not important)
* Go back to sort the previous column (for example, the Rate column).

You will see that the (Rate) column will be sorted in descending direction, but the sorting indicator in the header is telling the opposite direction. You will need to click the column twice to re-sort and fix the sorting.

I'm not sure if this is an issue in Kiali or if it's an issue in PF.

lucasponce:
> I'm not sure if this is an issue in Kiali or if it's an issue in PF.

Probably it's an edge case in "onSort" 

```
    console.log('TODELETE onSort ' + sortField.id);
    const isSortAscending = sortDirection === SortByDirection.asc;
    if (sortField.id !== this.state.currentSortField.id) {
      console.log('TODELETE updateSortField');
      this.updateSortField(sortField);
    } else if (isSortAscending !== this.state.isSortAscending) {
      console.log('TODELETE updateSortDirection');
      this.updateSortDirection();
    }
```

It assumes only sort once it's on the column, if I read the code correctly, it seems it won't change and column and change sorting on a single step.

lucasponce:
>  It maintains use of FilterComponent for SortField definition and URL bookmarking, while also deferring to table for updating the current sort and direction.

That's not 100% correct in your implementation because there is no tab url for inbound and outbound.

lucasponce:
And for feature perspective I don't see the two tabs inbound/outbound useful:
image

I'd prefer to have two tables instead of two tabs.

Otherwise we break a little bit the pattern used with Metrics as well (where we have two parent Inbound Metrics + Outbound Metrics).

jshaughn:
@israel-hdez I will look into the sort issue.

@lucasponce If we were to combine tabs and go back to the original approach, can we apply the same sort to both tables when selected, or bookmarked?

lucasponce:
> @lucasponce If we were to combine tabs and go back to the original approach, can we apply the same sort to both tables when selected, or bookmarked?

+1 it makes sense, probably that would be the default behaviour as I think it's reading the url parameters will be applied to both.

To me two tables in the same page makes more sense:
- It's aligned with same distribution used for logs, and when you are working in the details pages, it's odd this level of tabs at the beginning of the tab content. Subtabs are currently used in: a) inside a card, when you need to structure info (i.e. services detail). Or in the secondary sublist, when probably we are dealing with different entities (i.e. pods, services, workload, istio config, etc).

src/components/TrafficList/TrafficListComponent.tsx
```diff
@@ -0,0 +1,353 @@
...
+      trafficPercent: `${(100 - Number(percentError)).toFixed(1)}%`
+    };
+  };
+
+  private getProtocol = (traffic: ProtocolTraffic) => {
+    if (!traffic.protocol) {
+      return 'N/A';
+    }
+
+    return traffic.protocol.toUpperCase();
```
cfcosta, line 239:
```typescript
private getProtocol = (traffic) => (traffic.protocol || "N/A").toUpperCase();
```

src/components/TrafficList/TrafficListComponent.tsx
```diff
@@ -0,0 +1,353 @@
...
+        metrics = `${detail}?tab=${inverseMetricsDirection}`;
+        break;
+      default:
+        metrics = '';
+    }
+
+    return { detail: detail, metrics: metrics };
+  };
+
+  private nodeTypeToType = (type: NodeType, isURL?: boolean): string => {
```
cfcosta, line 341: This makes me feel that at some point we might need a proper inflector library to cover there kinds of string manipulation.

src/services/GraphDataSource.ts
```diff
@@ -216,32 +216,46 @@ export default class GraphDataSource {
 
   // Some helpers
 
-  public fetchForApp = (duration: DurationInSeconds, namespace: string, app: string) => {
+  public fetchForApp = (duration: DurationInSeconds, namespace: string, app: string, injectServiceNodes?: boolean) => {
     const params = GraphDataSource.defaultFetchParams(duration, namespace);
     params.graphType = GraphType.APP;
+    params.injectServiceNodes = !!injectServiceNodes;
```
cfcosta, line 222: Minor nitpick: injectServiceNodes is already boolean, no need to do the double negative ;)

src/services/GraphDataSource.ts
```diff
@@ -216,32 +216,46 @@ export default class GraphDataSource {
 
   // Some helpers
 
-  public fetchForApp = (duration: DurationInSeconds, namespace: string, app: string) => {
+  public fetchForApp = (duration: DurationInSeconds, namespace: string, app: string, injectServiceNodes?: boolean) => {
```
cfcosta, line 219: In general, I really dislike method flags, as they hide two methods/two different implementations. Not sure how to make this better, though, so just noting.

cfcosta (commented):
There are some things code-wise that I would change, but to be honest nothing is critical.

src/components/TrafficList/TrafficDetails.tsx
```diff
@@ -1,47 +1,73 @@
...
+
+export interface AppNode {
+  id: string;
+  type: NodeType.APP;
   namespace: string;
-  appName: string;
-};
+  name: string;
+  version: string;
+  isInaccessible: boolean;
```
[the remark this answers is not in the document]
jshaughn: Probably would be better but this now ties all the way back to the server code. So no change.

src/components/TrafficList/TrafficListComponent.tsx
```diff
@@ -0,0 +1,353 @@
...
+          sortBy={this.state.sortBy}
+          cells={columns}
+          rows={this.rows()}
+          onSort={this.onSort}
+        >
+          <TableHeader />
+          {this.state.listItems.length > 0 ? (
+            <TableBody />
+          ) : (
+            <tr>
```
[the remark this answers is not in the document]
jshaughn: It's been re-arranged

src/components/TrafficList/TrafficListComponent.tsx
```diff
@@ -0,0 +1,353 @@
...
+
+  private getTraffic = (traffic: ProtocolTraffic): { trafficRate; trafficPercent } => {
+    let rps = '0';
+    let percentError = '0';
+    let unit = 'rps';
+    if (hasProtocolTraffic(traffic)) {
+      switch (traffic.protocol) {
+        case 'http':
+          rps = traffic.rates.http;
+          percentError = traffic.rates.httpPercentErr || '0';
```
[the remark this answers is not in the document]
jshaughn: Probably.  But I love switch statements for the clarity.

src/components/TrafficList/TrafficListComponent.tsx
jshaughn: Yours is prettier, I admit.  I don't encourage unsolicited style reviews but I'll throw you a bone and apply this one.  I'll even do you one better and remove the single-use function completely.

src/components/TrafficList/TrafficListComponent.tsx
jshaughn: i refuse to replace a simple function with a proper library of a type I have never even heard of :)

src/services/GraphDataSource.ts
jshaughn: another nitpick, it's boolean | undefined.

src/services/GraphDataSource.ts
jshaughn: life is full of compromise.

src/components/TrafficList/TrafficListComponent.tsx
```diff
@@ -0,0 +1,353 @@
...
+  trafficRate: string;
+  trafficPercent: string; // percent success
+}
+
+type TrafficListComponentProps = FilterComponent.Props<TrafficListItem> & {
+  trafficItems: TrafficItem[];
+};
+
+type TrafficListComponentState = FilterComponent.State<TrafficListItem> & {
+  sortBy: ISortBy;
```
israel-hdez, line 40: I'm wondering whether it's hard to remove these two apparently redundant state fields ?
I know this component tends to not change a lot, but in some form I see these fields need to be in sync with the inherited state...

I'm thinking this could be a source of mistakes when modifying the code and forgetting to sync the state. But if it's hard, let it be.

src/components/TrafficList/TrafficListComponent.tsx
```diff
@@ -0,0 +1,353 @@
...
+  {
+    title: 'Name',
+    transforms: [sortable, cellWidth(30)]
+  },
+  {
+    title: 'Rate',
+    transforms: [sortable, cellWidth(12)]
+  },
+  {
+    title: 'Percent',
```
israel-hdez, line 57: I see this column was shortened from `Percent success` to only `Percent`.
As it is, I find it a little unclear what's the meaning of "Percent".

Any objection on going back to `Percent success`?

src/components/TrafficList/TrafficListComponent.tsx
```diff
@@ -0,0 +1,353 @@
...
+
+  private getLinks = (item: TrafficListItem) => {
+    if (item.node.isInaccessible) {
+      return { detail: '', metrics: '' };
+    }
+
+    const detail = `/namespaces/${item.node.namespace}/${this.nodeTypeToType(item.node.type, true)}/${item.node.name}`;
+
+    const metricsDirection = item.direction === 'inbound' ? 'in_metrics' : 'out_metrics';
+    let metrics = `${history.location.pathname}?tab=${metricsDirection}`;
```
israel-hdez, line 297: Looks like the behavior of the `View metrics` link is broken.
It switches to the metrics page, but it no longer pre-selects the right filters.

I think this has been broken for a while, as I see labels have changed (e.g.: `destination_app` is now `destination_canonical_service`). For this PR, I think it's OK to leave as is, because there is no regression. But, probably, we should open a bug.

israel-hdez (commented):
So, after all the work, I see this basically became a simplification, both in the feature itself and in the code.
I think it's quite fine. I'm have only a few very minor comments.

src/components/TrafficList/TrafficListComponent.tsx
lucasponce: +1 on this, as we have two percents (success or traffic distribution) I guess we should differentiate.

For example, if I come from graph, in the review service I may expect to see the split % instead of the percentage error.
image

lucasponce:
This PR is good
image

Only need to rebase the controls and address the "Percent" comment and it's ready to be merged.

Thanks @jshaughn for the discussion and the work !

lucasponce (approved):
It needs rebase and there is an open comment for the label of the "Percent" column but this PR is good.

src/components/TrafficList/TrafficListComponent.tsx
jshaughn: @israel-hdez I will look, it's possible it's an unnecessary holdover from the example impl I used originally.

src/components/TrafficList/TrafficListComponent.tsx
jshaughn: +1, will change back to Percent Success.

src/components/TrafficList/TrafficListComponent.tsx
jshaughn: I'll just fix it now.

jshaughn: OK, the 'View metrics' link for apps seems to be working.

src/components/TrafficList/TrafficListComponent.tsx
jshaughn: @israel-hdez Yes, I was able to remove the redundant sort state given the other refactoring already done in the PR.

jshaughn:
@lucasponce @israel-hdez Final changes pushed for your review, if you see fit.  Moving to QE.

pbajjuri20:
Tested and Passes on Kiali UI 1.25.0 (18ae209e9542d5dbbf3521e0611ecc745309d5c7)


traffic new

pbajjuri20 (approved):

israel-hdez (approved):
LGTM

PR #1945, opened 2020-09-24
Trace details page: change overall design

jotak:
- Switch from hierarchical view Traces => Trace => Span to a flat view with these 3 cards, to save space on screen and reduce DOM depth
- Move some elements (services buttons) from Trace Details to Spans Details. The content of the cards will change un following commits, Trace Details to show more high-level stats on the trace, and Span details to show the spans table + controls to manipulate that table

Fixes https://github.com/kiali/kiali/issues/3095


jotak:
This is WIP but here is already a view of the changes:

* **Before**, with a hierarchical view Traces => Trace => Spans:

Capture d’écran de 2020-09-24 10-45-04

* **After**, with a flat view:

Capture d’écran de 2020-09-24 10-44-35

lucasponce:
> After, with a flat view:

Thanks ! I suggest also to use simple tables as well to simplify the layout.
Perhaps I will try to provide a mockup with some ideas in case those may help on this.

jotak:
Current span table (still WIP) :

Capture d’écran de 2020-09-24 15-58-13

src/components/JaegerIntegration/JaegerResults/JaegerTraceTitle.tsx
```diff
@@ -18,12 +18,8 @@ export class JaegerTraceTitle extends React.Component<Props> {
...
         <Text component={TextVariants.h3} style={{ margin: 0, position: 'relative' }}>
-          {formattedTrace.name}
-          <Tooltip content={<>{traceID}</>}>
-            <span style={{ color: PfColors.Black600, paddingLeft: '10px', fontSize: '14px' }}>
-              {traceID.slice(0, 7)}
-            </span>
-          </Tooltip>
+          Trace details: {formattedTrace.name}
+          <span style={{ color: PfColors.Black600, paddingLeft: '10px', fontSize: '14px' }}>{traceID}</span>
           {this.props.onClickLink !== '' && (
```
aljesusg, line 23: Great one, not need slide now
Screenshot from 2020-09-28 13-59-28

aljesusg: What do you think about to have a link in the ID to the Trace in Jaeger too @jotak ? or use the id and remove the "View Trace in Tracing"

src/components/JaegerIntegration/JaegerResults/SpanTable.tsx
```diff
@@ -1,243 +1,250 @@
...
+          <strong>
+            {rqLabel}
+            {peerLink}:{' '}
+          </strong>
+          {info.method} {info.url}
+        </div>
+        <div>
+          <strong>Response status: </strong>
+          {rsDetails.join(', ')}
+        </div>
```
aljesusg, line 145: Should we have errorCodes with red like before? I don't know if it's useful for the user

src/components/JaegerIntegration/JaegerResults/SpanTable.tsx
```diff
@@ -1,243 +1,250 @@
...
-                  </>
-                }
-              >
-                <span>{span.operationName.slice(0, 40)}...</span>
-              </Tooltip>
+              <>
+                {formatDuration(span.duration)}
+                <br />
+                ... and more stats ...
+              </>
```
aljesusg, line 220: I don't understand "...and mroe stats..." I think that this is because is still wip maybe I should wait @lucasponce ^^

src/components/JaegerIntegration/JaegerResults/SpanTable.tsx
jotak: Yes, I wrote some TODO somewhere, I keep that in mind, +1

src/components/JaegerIntegration/JaegerResults/SpanTable.tsx
jotak: yes it's just a WIP thing :)

jotak:
I've added some traces stats:

Capture d’écran de 2020-09-28 18-39-25

It needs some styling, I didn't really take care of that yet, just taking care of the content / data at the moment.

jotak:
Now more colorful:

Capture d’écran de 2020-09-29 10-16-01

aljesusg:
> Now more colorful:
> 
> Capture d’écran de 2020-09-29 10-16-01

wow i like this comparison between other traces.

Are you comparing them with other traces with the same number of spans, aren't you?

jotak:
> wow i like this comparison between other traces.
> 
> Are you comparing them with other traces with the same number of spans, aren't you?

So yes and no, that's what the "similar" line is for, comparing with similar traces. Basically, "similar" mean traces with similar number of span and similar operations (it doesn't necessarily mean **exactly** the same, but "close to" to same). And the line below compares with every other traces displayed, even the very different ones.

aljesusg:
> > wow i like this comparison between other traces.
> > Are you comparing them with other traces with the same number of spans, aren't you?
> 
> So yes and no, that's what the "similar" line is for, comparing with similar traces. Basically, "similar" mean traces with similar number of span and similar operations (it doesn't necessarily mean **exactly** the same, but "close to" to same). And the line below compares with every other traces displayed, even the very different ones.

In that point I am with the tracing health, because if we want to warn the user when traces are x% more time than average I think that we should compare them with traces with the same number of spans to be fair.

I need to create bigger example to analyze this because is not fair compare a trace operation of get information that "generate" or pass by "N" services than other I think

jotak:
> In that point I am with the tracing health, because if we want to warn the user when traces are x% more time than average I think that we should compare them with traces with the same number of spans to be fair.
> 
> I need to create bigger example to analyze this because is not fair compare a trace operation of get information that "generate" or pass by "N" services than other I think

Yes, also it might be useful to compare the spans average duration rather than the whole trace duration in that case: then since it's a span average, if the number of spans is not exactly the same, then it doesn't matter so much

aljesusg:
> > In that point I am with the tracing health, because if we want to warn the user when traces are x% more time than average I think that we should compare them with traces with the same number of spans to be fair.
> > I need to create bigger example to analyze this because is not fair compare a trace operation of get information that "generate" or pass by "N" services than other I think
> 
> Yes, also it might be useful to compare the spans average duration rather than the whole trace duration in that case: then since it's a span average, if the number of spans is not exactly the same, then it doesn't matter so much

Yes, maybe we can show that in this view too, because all this information is in the UI. maybe in another PR :) Anyway I love this view Great job @jotak I take my hat off

jotak:
Thanks @aljesusg ! :-)

jshaughn:
Looks good.  In the screenshot above, what is the "Try get" link? "Try get" sounds maybe a little odd.

jotak:
> Looks good. In the screenshot above, what is the "Try get" link? "Try get" sounds maybe a little odd.

It's the operation name, it's supposed to tell what action is being performed in the trace / span. Kiali doesn't have control over that. This specific one is generated by my demo, so don't worry you won't see it except in my screenshots :)

jotak:
@aljesusg here's a screenshot with an error. I've put back the error style that was in place before on the row table.

Capture d’écran de 2020-09-29 17-05-01

jotak:
I think it's complete in terms of features. There's a couple of things that can be done in follow-up tasks, such as:

- more span filters, perhaps switching to our usual filtering toolbar styles - to allow filtering by operation name, error presence, component...
- compare span durations with trends based on metrics; e.g. compare versus avg/p95/p99 of last 2 hours, last 6 hours etc.

jotak:
(I'm going to do a cleanup pass, also make some stuff more homogene with the trace details in graph (e.g. show errors using the exclamation mark icon)

src/components/JaegerIntegration/JaegerResults/JaegerTraceTitle.tsx
jotak: I would prefer to keep external links very explicit.
But I didn't notice it was on the left, perhaps it should be moved next to "View on Graph".

lucasponce:
Some early feedback:
image

It seems there is an overlapping error there

lucasponce:
For consistency:
image
image

No need to label de column "Links" when using the table with kebap that element is without title.

When there are explicit "actions" we can label it as a column like this other example:
image

lucasponce:
Suggestion: after playing a little bit I guess this pattern:
image

Would benefit to put to subtabs: "Trace details" / "Span details", when reviewing the "spans" it typically scrolls so I don't need to have the "trace details" present and that would help to align with other subtabs pattern.

Also the pattern to use tabs to add more details is used in Iter8 details and I guess it would feet here as well.

lucasponce:
Not sure if it's related to this PR, probably not, but I see non consistent behaviour in the query when changing the interval, limit, etc:
image

Perhaps it's in the backend, but worth to review, from user perception, the expectation is that when you set back your previous criteria you get more or less the same previous result but you may have very different results.

lucasponce:
Perhaps "spans" and "services" labels may have different colors for better differentiation:
image

We introduced more colors in labels like in the wizard (aligned with the badges used for the same case), like in wizards:
image

Here I think it''s first time we differentiate services vs spans, so something you propose may work.

lucasponce:
Another UI suggestion:
image

First two blocks are more or less "fixed" and it would grow too much, but as you can see in the example it's easier in the third grid column that service names get large and then it misses the structure.

lucasponce:
Another suggestion about the filter in the spans, as it can get populated quite easily:
image

I see fit to also apply same pattern used in graph for multiple selection/unselection with checkboxes (show section):
image

lucasponce:
Another thought: one thing I really like is that Iter8 used the charts for comparison:
image

I wonder if this comparison between a trace an averages can be represented with charts:
image

Perhaps not with different charts but one single for the trace duration and other items compared, then tooltips can be used for absolute values then.

jotak:
> It seems there is an overlapping error there

(Was about Jaeger external link) => Fixed

> No need to label de column "Links" when using the table with kebap that element is without title.

Fixed

> Would benefit to put to subtabs: "Trace details" / "Span details", when reviewing the "spans" it typically scrolls so I don't need to have the "trace details" present and that would help to align with other subtabs pattern.
> 

+1, will do

> Not sure if it's related to this PR, probably not, but I see non consistent behaviour in the query when changing the interval, limit, etc:
> 
> Perhaps it's in the backend, but worth to review, from user perception, the expectation is that when you set back your previous criteria you get more or less the same previous result but you may have very different results.

Not impossible that I broke something, I will investigate

> Perhaps "spans" and "services" labels may have different colors for better differentiation:

Actually, I've renamed 'Services' to 'Apps' here which is more on page with our Kiali terminology (it *is* apps). In graph summary panel, apps/workloads/services labels are in light blue, so perhaps I should pick the same.

> First two blocks are more or less "fixed" and it would grow too much, but as you can see in the example it's easier in the third grid column that service names get large and then it misses the structure.

Fixed

> Another suggestion about the filter in the spans, as it can get populated quite easily [...]

That's one thing I mentioned in what I plan for follow-up task: I'd like to add more filters for this table, and as a consequence this should anyway go away to use a more standard filter toolbar as we have in other screens.

lucasponce:
Thanks ! I check the latest commit:
image

I know moving the two blocks into tabs in on the way, I'd say that in the traces tab all three links "View on Graph", "Compare in Tracing" and "View trace in Tracing" may go into a "kebap" menu similar as "spans" or a new columns called "actions" if you want to make them more explicit but kebap is being used as the main pattern to group this on similar places of Kiali.

jotak:
Last commit changes:
Capture d’écran de 2020-10-01 14-37-17

lucasponce:
It's good but I see a glitch with a transparent style in the tab title, perhaps something is missing, I'm going to fetch it locally and take a look.

lucasponce:
Okey,
image
it needs a minor addition in the _layout.scss

```
#service-tabs .pf-c-tabs__list {
  background-color: white;
}
```

To add a `trace-details` id there and also a proper id in the TracesComponent.tsx

jotak:
> > Perhaps "spans" and "services" labels may have different colors for better differentiation:
> 
> Actually, I've renamed 'Services' to 'Apps' here which is more on page with our Kiali terminology (it _is_ apps). In graph summary panel, apps/workloads/services labels are in light blue, so perhaps I should pick the same.

So I looked at those existing app things and it turns out they don't use an "official" PF color, I did a try anyway but it doesn't look great because white text on light blue isn't very readable. I tried a couple of other PF colors, but, how to say that, it looks like an arlequin-style mosaic. So I prefer to leave the standard blue and defer that for when we'll do a global design review next week. By the way, to clarify something, the green label next to it isn't randomly green, it's just using the PfAlert colors: green when there's 0 error, red when >0

What I've done however is move the blue labels on the first line and leave the second line to the alert-colored one, I think it looks better.

jotak:
@lucasponce the issue that you saw when setting the interval duration option should be fixed. I broke something indeed.

lucasponce:
Thanks @jotak I'm running a last review locally to your latest changes.

lucasponce:
I'm seeing the tabs problem described in https://github.com/kiali/kiali/issues/3284 but I think it can be addressed in a separated PR.

lucasponce (approved):

PR #1946, opened 2020-09-25
Remove ServiceMeshPolicy and SMRbacConfig for Maistra installations

xeviknal:
Maistra is no longer using those objects. This PR is meant to remove those references from the frontend.

requires https://github.com/kiali/kiali/pull/3256

PR #1949, opened 2020-09-29
IstioConfig: ValidationMessages are not mandatory

xeviknal:
This comes from this comment/fix in here: https://github.com/kiali/kiali-ui/pull/1946#pullrequestreview-498583490

It turns out that the `ValidationMessage[]` is not mandatory, therefore it might return null values.
I've changed it in the type and also reorganized two methods that are accessing this field. It makes it more consistent.

### DIFF
--- a/src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
+++ b/src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
@@ -376,7 +376,9 @@ class IstioConfigDetailsPage extends React.Component<RouteComponentProps<IstioCo
 
   getStatusMessages = (): ValidationMessage[] => {
     const istioObject = getIstioObject(this.state.istioObjectDetails);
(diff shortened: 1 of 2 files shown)


src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
```diff
@@ -414,23 +416,23 @@ class IstioConfigDetailsPage extends React.Component<RouteComponentProps<IstioCo
   };
 
   isExpanded = (istioConfigDetails?: IstioConfigDetails) => {
+    let isExpaned = false;
```
lucasponce, line 419: typo ? isExpaned -> isExpanded

xeviknal: sloppy. The typo is fixed now.

lucasponce (approved):

PR #1957, opened 2020-10-02
add 'healthy' operand to graph find/hind / ensure health for graph nodes is part of the overall graph fetch

jshaughn:
By ensuring health for graph nodes is part of the overall graph fetch we can then depend on it being available for find/hide.  Additionally it eliminates things like node flashing on graph refresh, due to late update of unhealthy nodes.

The downside is the additional delay while we gather health information.  See the issue discussion for more on that topic.

Technically this moves health-gen from CytoscapeGraph, where fetch was independent of graph-gen, to GraphDataSource, where it is now part of the overall graph fetch.

Fixes https://github.com/kiali/kiali/issues/3252


jshaughn:
@lucasponce @jotak @israel-hdez I added you all as reviewers although initially detailed code review is optional, more important is agreement about the overall approach.  Feel free to run the PR to experiment with responsiveness, etc.  There is some console logging still in place for some rough timing info.

src/services/GraphDataSource.ts
```diff
@@ -216,10 +234,21 @@ export default class GraphDataSource {
...
   // Some helpers
 
-  public fetchForApp = (duration: DurationInSeconds, namespace: string, app: string, injectServiceNodes?: boolean) => {
+  public fetchForApp = (
+    duration: DurationInSeconds,
+    namespace: string,
+    app: string,
+    includeHealth?: boolean,
+    injectServiceNodes?: boolean
+  ) => {
```
israel-hdez, line 243: Quite minor, but I would like to know if we can omit these changes.

I wouldn't like to see the an increased number of parameters on these helpers. At some point, I think it will be better to invoke the generic `fetchGraphData` which lets you fully customize the request.

israel-hdez (commented):
Thinking on the overall idea, I think it's quite fine to centralize the code.

Yet, I still don't like the need to fire a bunch of requests on every refresh. While this still works, I think it's fine as is. But, at some point, we may want to think into adding a health appender to get everything in the same request as from time to time, we hear from users and find that Kiali is bombarding prometheus, and that's hard to control knowing that the UI is also bombarding the back-end with several requests.

src/services/GraphDataSource.ts
jshaughn: I'll look at this, perhaps the helpers are no longer helpful.

lucasponce:
With all the discussion in the issue I was expecting something more visible from user perspective in terms of performance.

But it's not the case, I've just started to run some deep testing and I can share some results:

image

In that scenario with 6 namespaces quite intense in terms of traffic (and errors) the performance is okey.

Yes, refresh it may take a little bit time, but at least in my env it's inside of the expected "user response" time.

I'll continue commenting on that, but I wanted to share this to put everyone in the same context.

lucasponce:
I jumped directly to the performance discussion but that was more a side comment, the main feature seems to work as expected.

A screenshot of the previous scenario without the filtering:

image

lucasponce:
Thinking in more nice progress on this.
From the Overview page I can already filter by health:
image

It would be nice to have a link in the upper-left to say something like "Show in Graph" and then It could show all selected namespaces in the graph.

But anyway, this is a side comment, I'll fill a new issue for it.

lucasponce:
I'm ok if we want to merge this one for Kiali 1.25 but I'd like to address some of the scenarios discussed in 
https://github.com/kiali/kiali/issues/3252

specially related to the repeated queries found and possible to apply same strategy we have around caching but to these scenarios.

lucasponce:
I couldn't finish the testing of this PR and the other proposal, sorry, if we can wait I can do that on next week.
But this PR didn't present any major blocker from my testing (and you can see the screenshots and I guess I covered several high intense perf scenarios as well).

jshaughn:
@pbajjuri20 For testing this there are a couple of notes.  It's best but not required to have some unhealthy nodes.  You may or may not notice a slow down on the first rendering of your graph as the time is expected to be a bit slower.  But it may not be noticeable.  Subsequent refreshes should appear smooth and you may or may not notice that "flashing" goes away for unhealthy nodes.

And then there is the new feature, you can use the 'healthy' operand in find/hide.  It can also be negated as "!healthy".  It should also respond to autocomplete and have an entry in the help.

pbajjuri20:
Tested and Passed on Kiali UI 1.25.0 (49c32b1ead57944e77a9c26d277871a195752dec)


health hide

pbajjuri20 (approved):

israel-hdez (approved):
LG!

PR #1960, opened 2020-10-05
Sync VS/GW hosts on new GW

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3280

This is a convenient fix to make the user experience better when adding Gateways from the Wizards.

Wizards in the Advanced Options allow to define the VirtualService hosts; also in the "Advanced Options" is possible to create a "New Gateway".

This PR syncs GW + VH hosts in that case, when Wizard defines a new Gateway.

In https://github.com/kiali/kiali/issues/3280 was reported that if not sync-ed, user would always need to manually updated those config.

Note, only this is fixed in the scenario when a New Gateway is generated, when user selects an existing one, or updates, then Wizard won't sync the VS with the Gateway.

Basically Gateways are shared config and can't be always linked in 1-to-1 mapping with a VS so, trying to automate a sync here may introduce more weird situations.

### DIFF
--- a/src/components/IstioWizards/RequestRouting/Rules.tsx
+++ b/src/components/IstioWizards/RequestRouting/Rules.tsx
@@ -131,14 +131,18 @@ class Rules extends React.Component<Props> {
                   )}
                 </>,
                 <>
-                  {rule.workloadWeights.map((wk, i) => (
-                    <div key={'wk_' + i}>
-                      <Tooltip position={TooltipPosition.top} content={<>Workload</>}>
-                        <Badge className={'virtualitem_badge_definition'}>WS</Badge>
-                      </Tooltip>
-                      {wk.name} ({wk.weight} %)
-                    </div>
-                  ))}
+                  <div key={'ww_' + order}>
+                    {rule.workloadWeights.map((wk, i) => {
+                      return (
+                        <div key={'wk_' + order + '_' + wk.name + '_' + i}>
+                          <Tooltip position={TooltipPosition.top} content={<>Workload</>}>
+                            <Badge className={'virtualitem_badge_definition'}>WS</Badge>
+                          </Tooltip>
+                          {wk.name} ({wk.weight} %)
+                        </div>
+                      );
+                    })}
+                  </div>
                   {rule.delay && (
                     <div key={'delay_' + order}>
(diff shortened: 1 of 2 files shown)


xeviknal (approved):
LGTM!

hhovsepy (commented):
So the fix is overriding 'VirtualService Hosts' in Hosts tab by 'Gateway Hosts' from Gateways tab.
In Advanced options:
1. Hosts tab: VirtualService Hosts has some default value.
2. Switch to Gateways tab: Choose Create Gateway.  Set 'Gateway Hosts' value.
3. Switch to Hosts tab: VirtualService Hosts value is overridden by 'Gateway Hosts' value.
Submit the Wizard.
VS gets the hosts value the set one. Before the Gateway Hosts' value was ignored.
Screenshot from 2020-10-05 18-22-21

One potential confusion here: When user sets some value in 'Hosts tab: VirtualService Hosts' and then in next tab activates the "Create Gateway", the 'VirtualService Hosts' value from previous tab is changed by the dafault '*' of the 'Gateway Hosts' value.

lucasponce:
> One potential confusion here: When user sets some value in 'Hosts tab: VirtualService Hosts' and then in next tab activates the "Create Gateway", the 'VirtualService Hosts' value from previous tab is changed by the dafault '*' of the 'Gateway Hosts' value.

Yes, this is expected.

I can tell you why, typically the pre-populated value in the VirtualHost is internal FQDN only addressable inside the cluster.
When the Gateway is activated with an "all domains" wildcard `*` then Maistra will generate a OpenShift Route with the default public DNS, if we propagate the VirtualHost into the Gateway, then it would use the internal DNS as external DNS so it makes sense to move to by default the "all domains" wildcard in that step.

I know there are multiple corner cases that probably wizard can't handle (like to know in advanced the public DNS) but I guess resolving the default use case would help, but yes any other combination should be handled by the user.

hhovsepy (approved):
Verified, @lucasponce thanks for clarification.

hhovsepy:
Screenshot from 2020-10-06 08-56-50
Screenshot from 2020-10-06 08-57-01
Screenshot from 2020-10-05 18-22-21

PR #1964, opened 2020-10-06
Kiali#3252 concurrent

jshaughn:
This is actually a modification/alternativeto https://github.com/kiali/kiali-ui/pull/1957.  The approach here queries for namespace health concurrently with the graph-generation.  So instead of waiting for the graph and then fetching the necessary namespace health it basically grabs all of the health on the assumption that most of the time there is traffic for all of the nodes in the namespaces.  Unfortunately we still need to wait for the graph to then query for outsider node health, if any.

Without outsider nodes this approach (in my tests) saved about 40-50% as opposed to the original PR (in other words, almost no change from the existing graph-gen).  But depending on the number of outsiders the time then increases.

Note that this PR can replace #1957 , it's an alternate solution.


jshaughn:
I'm closing this PR although it could still be reopened as an alternative solution to our current choice, #1957.  Timing tests did show this to be a bit faster on average, but it also adds complexity and it a bit less deterministic as the perf varies with the number of outsider nodes.  Also, there is some indication that concurrent fetch of graph and health can bog down prom as it gets hit simultaneously.

PR #1965, opened 2020-10-07
Fix bad padding at bottom of the pages

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3273

I think I introduced this bug time ago when we aligned the style between pages making all of them look similar in terms of header size, no matter if they were located in Overview, Graph or Details.

Now in long pages like lists or trace span details scroll would be adjusted to the bottom of the browser size:

image

image

I think main testing here is to check there is no side effect/regression around this.

### DIFF
--- a/src/components/Nav/Page/RenderComponentScroll.tsx
+++ b/src/components/Nav/Page/RenderComponentScroll.tsx
@@ -1,10 +1,25 @@
 import React from 'react';
 
 // TOP_PADDING constant is used to adjust the height of the main div to allow scrolling in the inner container layer.
+// Overview page
+// 76px (header) + 76px (overview toolbar)
+const TOP_PADDING_OVERVIEW_NO_FILTER = 76 + 76;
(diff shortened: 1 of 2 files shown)


lucasponce:
cc @jotak in case you also saw that in the spans table, I'm addressing it in a generic way.

lucasponce:
It needs an additional commit for the overview page.

hhovsepy:
Besides the Overview, all looks good.

jotak (dismissed):
lgtm

lucasponce:
image
image

Also dealing a condition when filter is on / off that in case of the overpage it impacts in how the top-padding is calculated:

image
image

But know scroll of inner layer works from the header to the bottom of the window browser in all scenarios.

hhovsepy:
Verified.
Screenshot from 2020-10-08 17-59-32

hhovsepy (approved):

jotak (approved):
LGTM

PR #1966, opened 2020-10-07
Trace details page, spans table: add filtering toolbar

jotak:
Fixes https://github.com/kiali/kiali/issues/3278

Add sorting & filtering capabilities on the spans table.
reusing the StatefulFilters component

PR #1967, opened 2020-10-07
Refactor sort & filters

jotak:
This is a follow-up on https://github.com/kiali/kiali-ui/pull/1966 based on it; merge only after it.

Trying to simplify / mutualize / standardize filters
Changes:
- Remove the FilterComponent abstract layer that was used in list pages. Some parts were already obsolete since we introduced VirtualList. Now, the components inherits directly from React.Component. All the error handling functions have become useless for a while, as AlertUtils was growing up. State and Props types are still mutualized for type composition.

- Pretty much all filters now share the same structure with a "check" predicate, now having a single, mutualized way to apply filters on items, via the function 'runFilters' (or 'runPromiseFilters') in utils/Filters. This removed quite a lot of repeated patterns.
So basically, now in each component, instead of iterating over all filters and calling one by one ad-hoc filtering functions, we just need to define that "check" predicate on filters and call `runFilters`.

Other changes / renaming:
- components/FilterList/FilterHelper moved to utils/Filters
- components/Filters/CommonFilters split to utils/Filters and types/Filters
- 'FilterSelected' from StatefulFilters moved to 'GlobalFilters' in utils/Filters
- Some sorting functions are also simplified / mutualized, espacially Health comparison that was duplicated
- Fix a couple of warnings

Remarks:
Some pending questions in iter8 code that I haven't checked yet: some filters and sort functions seem perhaps  broken (?)


src/pages/extensions/iter8/Iter8ExperimentList/FiltersAndSorts.ts
```diff
@@ -36,113 +29,51 @@ export const sortFields: SortField<Iter8Experiment>[] = [
     id: 'baseline',
     title: 'Baseline',
     isNumeric: false,
-    param: 'is',
+    param: 'is', // !! same as phase?
```
jotak, line 32: I haven't tested yet the iter8 part, but it seems like there's two bugs spotted:
- this 'is' param is used for three different sortable fields, which means it probably won't work as expected with the URL params
- below, there's a "candidateFilter" but it is not tied to any filtering process, so I guess it won't work?

cc @lucasponce 
I would need to see how to setup iter8 for testing.

lucasponce: I'll help on that, this PR would need a lot of detailed review as it may have side effects.

lucasponce:
Also there are some pending refactorings in the VirtualList component (it's not virtual anymore but it maintained the name that may be misleading).

jotak:
Closing, this is old and would require a big rebase / conflict merging. We can still take it back later if we feel the need.

PR #1970, opened 2020-10-09
Try to disambiguate deployment-like and replicaset-like workload in spans

jotak:
(but this still has flaws, see code comments)

(Kind of) fixes https://github.com/kiali/kiali/issues/3306

### DIFF
--- a/src/components/JaegerIntegration/JaegerHelper.ts
+++ b/src/components/JaegerIntegration/JaegerHelper.ts
@@ -41,34 +41,69 @@ export const getTimeRangeMicros = () => {
   };
 };
 
-const workloadFromNodeRegex = /([a-z0-9-.]+)-[a-z0-9]+-[a-z0-9]+.([a-z0-9-]+)/;
-const workloadFromHostnameRegex = /([a-z0-9-.]+)-[a-z0-9]+-[a-z0-9]+/;
 type WorkloadAndNamespace = { pod: string; workload: string; namespace: string };
 export const getWorkloadFromSpan = (span: Span): WorkloadAndNamespace | undefined => {
   const nodeKV = span.tags.find(tag => tag.key === 'node_id');
-  if (!nodeKV) {
-    // Tag not found => try with 'hostname' in process' tags
-    const hostnameKV = span.process.tags.find(tag => tag.key === 'hostname');
-    if (hostnameKV) {
-      const result = workloadFromHostnameRegex.exec(hostnameKV.value);
-      if (result && result.length > 1) {
-        // As the namespace is not provided here, assume same as service's
-        const split = span.process.serviceName.split('.');
-        return { pod: hostnameKV.value, workload: result[1], namespace: split.length > 1 ? split[1] : '' };
-      }
+  if (nodeKV) {
+    // Example of node value:
+    // sidecar~172.17.0.20~ai-locals-6d8996bff-ztg6z.default~default.svc.cluster.local
+    const parts = nodeKV.value.split('~');
+    if (parts.length < 3) {
+      return undefined;
+    }
+    const podWithNamespace = parts[2];
+    const nsIdx = podWithNamespace.lastIndexOf('.');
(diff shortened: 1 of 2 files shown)


jotak:
This PR should fix what @hhovsepy reported, but, I'm not totally convinced we should merge it because it introduces another flaw:

If you consider the pod names:
- `kiali-traffic-generator-wqq8f`
- `kiali-traffic-g3n3r4t0r-wqq8f`
- `my-deployment-sdjvdfvmp-fg65d`

The first one will be seen as part of a replicaset. The second one part of a Deployment. And the third one, replicaset. Why? In this PR I'm setting some rules for disambiguation:
- If there's no dash in name after removing the pod part (without the '-wqq8f' in these examples), it's not a deployment
- If the second last part ('generator' / 'g3n3r4t0r') doesn't have exactly 9 characters, it's not a deployment
(too bad that 'generator' does have exactly 9 chars! :-( )
- If that part is only compound of letters, we assume it's human-made string hence it's not a deployment.

So, the first two rules are good IMO, but the third one is really flawed and could lead to false-positives, like with the 'my-deployment-sdjvdfvmp-fg65d' example because 'sdjvdfvmp' only has letters.
But unfortunately, without it I'm not fixing the corner case of 'kiali-traffic-generator-wqq8f'.

So, tell me what you think. I would prefer to remove the third rule even if it doesn't fix the reported bug. Having the first two rules is already a good tradeoff IMO. In any case, there still remains known issues.
Or come up with the ideal fix I couldn't think of :-)

jotak:
Or, a totally different solution could be to be able to load a "deployment" kind workload, from the backend, when we provide the replicaset name. @lucasponce do you think it could be feasible?

lucasponce:
> Or, a totally different solution could be to be able to load a "deployment" kind workload, from the backend, when we provide the replicaset name. @lucasponce do you think it could be feasible?

Probably is something we can review, I found this scenario when working on the workloads because situation can be also expand other controllers (ReplicationControllers, Jobs).

But I'll take a look.

A first short idea is that the pod has already the id generated from the controller -> replicaset -> pod id, so I think it's easier to extract the replicaset (or better, middle controller) from them to use it for this scenario (without repeating unnecessary queries).

But I'll investigate too.

jshaughn:
> If the second last part ('generator' / 'g3n3r4t0r') doesn't have exactly 9 characters, it's not a deployment

I don't have a lot to offer here but if you go with the "three-rule" approach above, I think you therre are a few things that may help:
- are you sure for deployments the second-last part is 9 chars?  It looks like 10 to me.
- the random strings use a limited char set to avoid bad words: "bcdfghjklmnpqrstvwxz2456789"
-- See https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/rand/rand.go#L83.
- the second-last part for replica-sets seems to use an even more limited set of chars, "bcdf2456789",  but i can't find where it's generated.

jotak:
Thanks @jshaughn , that helps
Indeed I'm not sure for the 9 characters, that's only empirical, and turning out to be wrong if you say 10 for you. Bad assumption, bad rule.
So, the limited char set is perhaps the only thing that can save this approach. I'd just want to make sure it's there for long enough (since which k8s version?) and hope it will stay long enough.

jotak:
I think it's sorted out now. Solution based on kube's charset. All tests pass, including the one with `kiali-traffic-generator`, and I've added a couple of new tests.
There's still a dead angle but it's been drastically reduced.

jshaughn (approved):
Visual check looks OK to me.

hhovsepy (approved):
Verified.

lucasponce (approved):
Nothing more to add.
I wonder if it's worth to add more context that can be used.
I mean, in the high level scenario, probably when this query is done there is a context of App/Workload/Service so, the replicaset (or intermediate controller as they can be others) Id should be already there and it could be passed as part of the context.

PR #1972, opened 2020-10-15
Fix validation layout on find/hide

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3275

I've just made a quick change and I was talking about this fix:
image

Then when "Find/Hide" text boxes shows a validations are aligned with previous "Display".


hhovsepy:
Fixed only on FF, still happens on Chrome 86.0.4240.75 
Screenshot from 2020-10-15 20-41-16
Screenshot from 2020-10-15 20-43-06

lucasponce:
> still happens on Chrome 86.0.4240.75

I'll check, that's version I used to test it.

jshaughn (approved):
Although I don't really think we need to change this, for me the change is fine on Chrome 81 and 86.  As well as FF 70.

lucasponce:
It's fine to not merge, I understood you didn't check if this change was good.

PR #1973, opened 2020-10-15
Add true fullscreen capability to logging tab

mtho11:
** Describe the change **
By utilizing the browser API to enable expand beyond the browser chrome to the actual full size of the monitor.
Part of the PoC was investigating various libraries that wrap the original web API (and support various browsers).

This uses the same library that OpenShift dev console uses for managing fullscreen mode - **screenfull**

** Issue reference **
fixes https://github.com/kiali/kiali/issues/3155

** Screenshot **

fullscreen


jshaughn (commented):
@mtho11 I'm not exactly sure what to expect, maybe you can shed some light on the expected behavior.  What exactly should go full-screen when I click the icon?  In some cases I see the entire Kiali go FS.  Sometimes I see just a logs window go FS.  I never see just the page view go FS, which is actually the view I expected to see go FS (by page view I mean the area that changes as I nav around; inside the left accordian and below the masthead).

Also, when I nav away from logs what should happen, should FS remain or go away?

mtho11:
@jshaughn that is really weird because in FS mode I can't even nav away from the logs because it's FS.

jshaughn:
Here is a video showing side-by-side tabs of OS console and Kiali, FS on console works but FS Kiali just expands the whole Kiali. I can't really explain when it happens, it seems inconsistent for me.

fullscreen

I have seen failures especially if the browser is already fullscreen when I start, but also at other times.  Both Win10 and F29.  Let's have @pbajjuri20 give it a try and see how it goes for him.

mtho11:
@jshaughn that is really weird. I have tested in both FF + Chrome on both OS/X and Fedora 30 and all work flawlessly.

We'll see what @pbajjuri20 turns up...

pbajjuri20:
@jshaughn @mtho11: when there are "no logs" it refecting that issue.

When it has the "logs" - we can see the "full screen"

pbajjuri20:
Kiali log - 1

mtho11:
@pbajjuri20 that situation has been remedied

pbajjuri20:
@mtho11: Full-screen button is disabled if there is "no log data" 


kiali log 3

pbajjuri20 (approved):

mtho11:
@lucasponce @jshaughn What do you guys think?

jshaughn:
> @lucasponce @jshaughn What do you guys think?

Take a look at my sub-PR, it may still have something useful.

lucasponce:
> @lucasponce @jshaughn What do you guys think?

I didn't review the feature/code, in case this feedback helps, I'd say that we are using "full screen" here with different context.
The "expand" verb can be more appropiate I guess (I also see in the videos that is the term used).

In terms of technically I guess user has a context and they want to expand a single fragment (container log or sidecar log) so I guess a full screen of the log (similar as openshift without the modal/framing and titles) would be nicer.

For standard Full Screen, the user can also just click the F11 and resize the screen, but entering in the "non distraction mode" would be an advanced feature.

jshaughn:
>  so I guess a full screen of the log (similar as openshift without the modal/framing and titles) would be nicer.

@mtho11 This was actually my next comment.  We had the modal because of the original "expand" approach.  Do we really need the modal anymore given the "screenfull" tooling?  Or can we just expand the target component?

mtho11:
Yes, getting rid of the modal makes sense to me as well now with the screenfull stuff.

mtho11:
Latest incarnation:
fullscreen3

mtho11:
Ok this should be finished now. I have removed the modal dialog and we just focus/fullscreen either the proxy/app log sections and apply the correct styles depending on if we are in fullscreen mode or not.

fullscreen4

src/pages/WorkloadDetails/WorkloadInfo/WorkloadPodLogs.tsx
```diff
@@ -155,20 +157,19 @@ export default class WorkloadPodLogs extends React.Component<WorkloadPodLogsProp
   private podOptions: string[] = [];
   private readonly appLogsRef: React.RefObject<any>;
   private readonly proxyLogsRef: React.RefObject<any>;
-  private readonly appFullScreenLogModalRef: React.RefObject<FullScreenLogModal>;
-  private readonly proxyFullScreenLogModalRef: React.RefObject<FullScreenLogModal>;
+  // private readonly appFullScreenLogModalRef: React.RefObject<FullScreenLogModal>;
+  // private readonly proxyFullScreenLogModalRef: React.RefObject<FullScreenLogModal>;
```
jshaughn, line 161: to be removed

jshaughn (approved):
@mtho11 Looks really good.  Some leftover comments to remove but that's not a blocker, we can get it on a subsequent PR.  Approving and sending to Praneeth for a re-check.

src/pages/WorkloadDetails/WorkloadInfo/WorkloadPodLogs.tsx
mtho11: Opps, forgot about that! Will remove.

pbajjuri20:
I have tested all the latest changes - LGTM.

pbajjuri20 (approved):

PR #1977, opened 2020-10-23
workload logs - use new LogEntry and Duration support

jshaughn:
Convert from using the logs blob string to the new structured LogEntry return supported by the backend.  With that support we also get support for Duration, so also add custome time range support.  And finally, add a switch to be able to toggle the k8s timestamps which take up a lot of space and may not be needed depending on the log message format.

Fixes https://github.com/kiali/kiali/issues/3067

REQUIRES SERVER PR: https://github.com/kiali/kiali/pull/3352

### DIFF
--- a/src/pages/WorkloadDetails/WorkloadDetailsPage.tsx
+++ b/src/pages/WorkloadDetails/WorkloadDetailsPage.tsx
@@ -103,11 +103,7 @@ class WorkloadDetails extends React.Component<WorkloadDetailsPageProps, Workload
     const logTab = (
       <Tab title="Logs" eventKey={2} key={'Logs'}>
         {hasPods ? (
-          <WorkloadPodLogs
-            namespace={this.props.match.params.namespace}
-            pods={this.state.workload!.pods}
-            duration={this.props.duration}
-          />
+          <WorkloadPodLogs namespace={this.props.match.params.namespace} pods={this.state.workload!.pods} />
         ) : (
           <EmptyState variant={EmptyStateVariant.full}>
             <Title headingLevel="h5" size="lg">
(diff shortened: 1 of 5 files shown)


lucasponce:
Some perf tests.

- `demos/travels` has a "normal" logs activity and Kiali doesn't show any perf stress:
image

- `demos/fraud-detection` has a high intense log activity, and Kiali shows some impact in term of perf stress:
image

But it's linear to the amount of lines fetched, so it's getting 40-50 Kb, but I think it's normal, just to keep an eye on future issues related to perf.

lucasponce (approved):
Code LGTM.
Local tests look much better now.

Backend PR needs some minot tweak to pass the CI.

jshaughn:
@pbajjuri20 , the QE for this is to ensure no log tab regressions and to try the new features:
- Custom time range (select "Custom" from the duration dropdown, and then a start and end time).
- New switch for displaying (or not) timestamps on the log entries (default is off, so a slight difference from previous behavior, this is to show more of the log message by default).
- New default of 100 lines (previously was 500, this was to speed up the initial fetch when just looking for the most recent logs)

Note that there is a server PR required.

pbajjuri20:
Tested the new features, it works perfectly.

custom duration:

custom



Changed to default of 100 lines :

duration


Timestamp:

timestamp

pbajjuri20 (approved):

jshaughn:
I'm going to merge the changes given approvals from @lucasponce  and QE.  Lucas, note that after your review I made a perf modfication that should make things faster when not using custom time range and as the number of tailed lines decreases.  @mtho11, if you have any feedback we can address it going forward.

PR #1978, opened 2020-10-26
Typescript compiler: use es2019

jotak:
Switching to es2019 (we're on es2016) would allow for instance to use `Array.flatMap` : https://www.ecma-international.org/ecma-262/10.0/index.html#sec-array.prototype.flatmap

This doesn't change the compilation target which is still es5.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -4,7 +4,7 @@
     "outDir": "build/dist",
     "module": "esnext",
     "target": "es5",
-    "lib": ["es2016", "dom"],
+    "lib": ["es2019", "dom"],
     "sourceMap": true,
     "allowJs": true,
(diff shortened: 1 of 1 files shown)


jotak:
I didn't see (and don't expect) any change neither in current build nor in runtime

mtho11 (approved):
Nice! I wasn't aware of this option. I tested everything compiled fine for me.

For reference: https://blog.logrocket.com/5-es2019-features-you-can-use-today/

israel-hdez (approved):
It builds OK.
I'm trusting the TS compiler will properly down-compile to es5 and no polyfill is needed.

jotak:
thanks for the review

PR #1901, opened 2020-08-28
Add IstioConfig under Workload Details

lucasponce:
Related to https://github.com/kiali/kiali/issues/3015

Requires https://github.com/kiali/kiali/pull/3157

It adds IstioConfig linked with a workload (resources with some workloadSelector that matches the workload labels).

image

For testers, I've used this (somehow artificial) example (you can modify it inside Kiali to check that validations are working):

```
apiVersion: security.istio.io/v1beta1
kind: PeerAuthentication
metadata:
  name: details-pa
  namespace: default
spec:
  selector:
    matchLabels:
      app: details
  mtls:
    mode: PERMISSIVE
---
apiVersion: security.istio.io/v1beta1
kind: RequestAuthentication
metadata:
  name: details-ra
  namespace: default
spec:
  selector:
    matchLabels:
      app: details
  jwtRules:
    - issuer: "issuer-foo"
      jwksUri: https://example.com/.well-known/jwks.json
---
apiVersion: security.istio.io/v1beta1
kind: AuthorizationPolicy
metadata:
  name: details-ap
  namespace: default
spec:
  selector:
    matchLabels:
      app: details
  rules:
    - from:
        - source:
            requestPrincipals: ["*"]
---
apiVersion: networking.istio.io/v1alpha3
kind: Sidecar
metadata:
  name: details-sc
  namespace: default
spec:
  workloadSelector:
    labels:
      app: details
  ingress:
    - port:
        number: 9080
        protocol: HTTP
        name: somename
      defaultEndpoint: unix:///var/run/someuds.sock
  egress:
    - port:
        number: 9080
        protocol: HTTP
        name: egresshttp
      hosts:
        - "prod-us1/*"
    - hosts:
        - "istio-system/*"
---
apiVersion: networking.istio.io/v1alpha3
kind: EnvoyFilter
metadata:
  name: details-ef
  namespace: default
spec:
  workloadSelector:
    labels:
      app: details
  configPatches:
    # The first patch adds the lua filter to the listener/http connection manager
    - applyTo: HTTP_FILTER
      match:
        context: SIDECAR_INBOUND
        listener:
          portNumber: 8080
          filterChain:
            filter:
              name: "envoy.http_connection_manager"
              subFilter:
                name: "envoy.router"
      patch:
        operation: INSERT_BEFORE
        value: # lua filter specification
          name: envoy.lua
          typed_config:
            "@type": "type.googleapis.com/envoy.config.filter.http.lua.v2.Lua"
            inlineCode: |
[...]

PR #1979, opened 2020-10-26
Remove no-labels validation from client format validation

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3336

It requires https://github.com/kiali/kiali/pull/3355

It was happening two things here:
- Kiali UI performed a format validation that showed a real validation, a DR has not labels in a subset.
- Kiali Backend hasn't this validation in place, so, when going into the Istio Details there was a mitmatch.

So, I've added that validation into the backend and remove the labels from the UI format validation to avoid duplication.


src/pages/ServiceDetails/ServiceInfo/types/SubsetValidator.ts
```diff
@@ -8,26 +8,13 @@ export default class SubsetValidator {
   }
 
   public isValid() {
-    return this.hasValidName() && this.hasValidLabels();
+    return this.hasValidName();
```
xeviknal, line 11: Does it make sense to keep the whole Subset validation and the DestinationRuleValidation into the UI?
I guess that it was put in place before Galley. Now, the case that created the motivation to put in place this UI validation is not valid anymore.

https://issues.redhat.com/browse/KIALI-2015

Besides that, the way we show DR and VS is no longer verbose (showing subsets, trafficpolicy, etc https://github.com/kiali/kiali-ui/pull/1901) and therefore we don't have problems accessing to the fields validated here.

Lastly, in case this validator or the DRValidator generates a message, the status column only shows the aggregated validation information (N objects validated, 1 warning found) and we would be at the same place: showing a warning in the list but not in the object details page.

So I guess that I would remove both SubsetValidator, DestinationRuleValidator and the  `addFormatValidation` from the ServiceInfo.

xeviknal (commented):
I've got a comment. Wdyt?

src/pages/ServiceDetails/ServiceInfo/types/SubsetValidator.ts
lucasponce: > Does it make sense to keep the whole Subset validation and the DestinationRuleValidation into the UI

Yes, +1, I was unsure, but thanks for confirming, I know this was something added before Galley was catching these errors.

I'll remove it.

hhovsepy:
Link to host from DR overview is removed in case of missing label, but it should remain, as the link to DR from Service stays.
Screenshot from 2020-10-26 14-25-29
Screenshot from 2020-10-26 14-25-13

hhovsepy (changes requested):
Warning is shown correctly, but the missing link to Service in Host is regression.
Screenshot from 2020-10-26 16-22-40
Screenshot from 2020-10-26 16-22-48

xeviknal:
I might be wrong, but I remember that links to services were disabled in case there were validation problems. Can you confirm that the link is present when there are no validations?

lucasponce:
> Warning is shown correctly, but the missing link to Service in Host is regression.

Yes, I've updated the backend one but still in WIP on this.

About validations as @xeviknal said, when there are validations problems, service links are disabled.
I'm going to see if perhaps we can refine this i.e. as we know which validations have invalidad services, perhaps it's a minor change we can add.

Let me look.

lucasponce:
I've removed the validations pointed by @xeviknal and @hhovsepy the current behaviour is consistent with other validations:
image

When there is a problem the link is disabled.

I tried to look how easy it could be to dig into the details but it's more complex as cross-namespace validation and/or service entries will bring cases of non-existence service that may be broken (i.e. a www.wikipedia.org can be a valid host if a service entry is declared).

So, perhaps I'd split the problem on this case and I'll continue disabling the link on this situation.
IstioConfig will always show that link, but without context bidirectional link (Service -> DR, DR -> Service) can't work in all cases.

lucasponce:
Oh, I realize that I guess I introduced some regression here when I worked in the side panel.

Please, don't merge yet, probably it needs more investigation.

lucasponce:
Also, looking into the VS:
image

Only when host is invalid the link is disabled:
image

So, yes, I'm +1 with @hhovsepy that probably it may be work in that way.

I'll continue working on this.

lucasponce:
Ok @hhovsepy there was a regression in the backend code.

On warnings:
image

On errors:
image

xeviknal (approved):
LGTM. The front-end validation has been removed preventing the scenario where users see a warning in the service details page but can't see any in the DR page.

hhovsepy (approved):
Verified, regression tested:
Screenshot from 2020-10-27 08-40-50
Screenshot from 2020-10-27 08-41-04
Screenshot from 2020-10-27 08-38-49

PR #1980, opened 2020-10-27
Fix wrong timestamp being use for traces query in graph

jotak:
Unlike in metrics, "queryTime" here is startTime, not endTime.

Fixes https://github.com/kiali/kiali/issues/3359


jshaughn (approved):
RIght, this is similar to querying for k8s logs, it's actually "sinceTime" as you are looking for data after the specified time. Whereas in Prometheus the queryTime is actually the endTime.

jotak:
Note, I merged with just Jay's approval, but the same was tested and approved on v1.24 https://github.com/kiali/kiali-ui/pull/1981

PR #1981, opened 2020-10-27
[v1.24] Fix wrong timestamp being use for traces query in graph

jotak:
Unlike in metrics, "queryTime" here is startTime, not endTime.

Fixes #kiali/kiali#3359


jshaughn (approved):
RIght, this is similar to querying for k8s logs, it's actually "sinceTime" as you are looking for data after the specified time.  Whereas in Prometheus the queryTime is actually the endTime.

mattmahoneyrh (approved):
Traces are displaying for all Bookinfo Services.

PR #1984, opened 2020-10-28
Mutual TLS actions

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3363

@xeviknal this PR is relatively simple but perhaps I missed something. I'd like if you could review the whole feature.

It simply adds a new Enable/Disable MutualTLS at namespace level to add a specific PeerAuthentication per namespace in STRICT mode:

image

But I see that tls info is different from the graph one.

In the Kiali tls endpoint I get this:

```
{"status":"MTLS_NOT_ENABLED"}
```

And feature works as expected:

image

But in the Graph I see a different semantic on the "Security" locks, which doesn't honor the status or Kiali is not collecting the right mutual TLS global status.

I don't know, perhaps I'm spotting a new issue.

Note the feature/business context was to help user to apprach this task [1]
Also reading [2] and wondering if the default mode perhaps it changes this and we don't need to add this action after all.
(That's also valid, this is a very minor PR)

[1] https://istio.io/latest/docs/tasks/security/authentication/mtls-migration/
[2] https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#MeshConfig

### DIFF
--- a/src/components/IstioWizards/WizardActions.ts
+++ b/src/components/IstioWizards/WizardActions.ts
@@ -1172,6 +1172,24 @@ export const buildPeerAuthentication = (
   return pa;
 };
 
+export const buildMutualTlsPeerAuthentication = (name: string, namespace: string): PeerAuthentication => {
+  const pa: PeerAuthentication = {
+    metadata: {
+      name: name,
+      namespace: namespace,
+      labels: {
+        [KIALI_WIZARD_LABEL]: 'PeerAuthentication'
+      }
+    },
+    spec: {
+      mtls: {
+        mode: PeerAuthenticationMutualTLSMode.STRICT
+      }
+    }
+  };
+  return pa;
+};
+
 export const buildRequestAuthentication = (
   name: string,
   namespace: string,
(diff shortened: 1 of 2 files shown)


xeviknal:
@lucasponce yep, default istio installations goes with a feature called `autoMtls` which means that:
1. by default, the mesh behaves as you got a mesh-wide Authn Policy with a Permissive mode, and
2. every client service tries to enable mTLS in each call.

As a result, the mesh has mTLS enabled in each communication unless you disable the mTLS explicitly though a AuthnPolicy in DISABLE mode.

WRT the feature, a couple of things:
1. Enabling mTLS for Kiali means that you create a AuthnPolicy strict (and in case of autoMtls disabled, you need to add a Destination Rule).
2. Disabling mTLS it doesn't necessarily mean that you remove the AuthnPolicy (+ DestRule), it means that you place a AuthnPolicy with DISABLE mode. This way, you ensure that the mesh doesn't have any mtls traffic. This is really useful for mesh-wide STRICT policies but you want to leave one NS with no mTLS at all.
2+1. Perhaps, it needs a different naming. Something like `Remove Strict mTLS`.

Regarding the graph, do you see the mTLS icon? What is happening exactly?

lucasponce:
Ok, discussed with @xeviknal, we're going to close this PR. With Istio AutoMtls mode this feature as it's coded is not adding real value and it's well covered with the Istio Config -> Create New Istio Config feature.

PR #1986, opened 2020-10-30
[POC] Allow custom health configuration to be deployed as annotations

aljesusg:
Fix https://github.com/kiali/kiali/issues/3332
Fix https://github.com/kiali/kiali/issues/3435

# Tasks

- [X] Support health views
- [x]  Support for health in graph

For workload and services

Require: https://github.com/kiali/kiali/pull/3362

### DIFF
--- a/src/components/TrafficList/TrafficDetails.tsx
+++ b/src/components/TrafficList/TrafficDetails.tsx
@@ -20,6 +20,8 @@ import { KialiAppState } from '../../store/Store';
 import { connect } from 'react-redux';
 import { durationSelector } from '../../store/Selectors';
 
+type HealthAnnotation = { [key: string]: string };
+
 export interface AppNode {
   id: string;
   type: NodeType.APP;
@@ -34,7 +36,7 @@ export interface WorkloadNode {
   type: NodeType.WORKLOAD;
   namespace: string;
   name: string;
-  healthAnnotation: any;
+  kialiAnnotation: HealthAnnotation;
   isInaccessible: boolean;
 }
 
@@ -43,7 +45,7 @@ export interface ServiceNode {
   type: NodeType.SERVICE;
   namespace: string;
   name: string;
-  healthAnnotation: any;
+  kialiAnnotation: HealthAnnotation;
   isServiceEntry?: string;
   isInaccessible: boolean;
   destServices?: DestService[];
@@ -196,7 +198,7 @@ class TrafficDetails extends React.Component<TrafficDetailsProps, TrafficDetails
           type: node.nodeType,
           namespace: node.namespace,
           name: node.service || 'unknown',
-          healthAnnotation: node.healthAnnotation,
+          kialiAnnotation: node.kialiAnnotation,
           isServiceEntry: node.isServiceEntry,
           isInaccessible: node.isInaccessible || false,
           destServices: node.destServices
@@ -207,7 +209,7 @@ class TrafficDetails extends React.Component<TrafficDetailsProps, TrafficDetails
           type: NodeType.WORKLOAD,
           namespace: node.namespace,
           name: node.workload || 'unknown',
-          healthAnnotation: node.healthAnnotation,
+          kialiAnnotation: node.kialiAnnotation,
           isInaccessible: node.isInaccessible || false
         };
     }
--- a/src/components/TrafficList/TrafficDetails.tsx
+++ b/src/components/TrafficList/TrafficDetails.tsx
@@ -36,7 +36,7 @@ export interface WorkloadNode {
   type: NodeType.WORKLOAD;
   namespace: string;
   name: string;
-  kialiAnnotation: HealthAnnotation;
+  healthAnnotation: HealthAnnotation;
   isInaccessible: boolean;
 }
 
@@ -45,7 +45,7 @@ export interface ServiceNode {
   type: NodeType.SERVICE;
   namespace: string;
   name: string;
-  kialiAnnotation: HealthAnnotation;
+  healthAnnotation: HealthAnnotation;
   isServiceEntry?: string;
   isInaccessible: boolean;
   destServices?: DestService[];
@@ -198,7 +198,7 @@ class TrafficDetails extends React.Component<TrafficDetailsProps, TrafficDetails
           type: node.nodeType,
           namespace: node.namespace,
           name: node.service || 'unknown',
-          kialiAnnotation: node.kialiAnnotation,
+          healthAnnotation: node.healthAnnotation,
           isServiceEntry: node.isServiceEntry,
           isInaccessible: node.isInaccessible || false,
           destServices: node.destServices
@@ -209,7 +209,7 @@ class TrafficDetails extends React.Component<TrafficDetailsProps, TrafficDetails
           type: NodeType.WORKLOAD,
           namespace: node.namespace,
           name: node.workload || 'unknown',
-          kialiAnnotation: node.kialiAnnotation,
+          healthAnnotation: node.healthAnnotation,
           isInaccessible: node.isInaccessible || false
         };
     }
--- a/src/components/TrafficList/TrafficDetails.tsx
+++ b/src/components/TrafficList/TrafficDetails.tsx
@@ -34,6 +34,7 @@ export interface WorkloadNode {
   type: NodeType.WORKLOAD;
   namespace: string;
   name: string;
+  healthAnnotation: any;
   isInaccessible: boolean;
 }
 
@@ -42,6 +43,7 @@ export interface ServiceNode {
   type: NodeType.SERVICE;
   namespace: string;
   name: string;
+  healthAnnotation: any;
   isServiceEntry?: string;
   isInaccessible: boolean;
   destServices?: DestService[];
@@ -194,6 +196,7 @@ class TrafficDetails extends React.Component<TrafficDetailsProps, TrafficDetails
           type: node.nodeType,
           namespace: node.namespace,
           name: node.service || 'unknown',
+          healthAnnotation: node.healthAnnotation,
           isServiceEntry: node.isServiceEntry,
           isInaccessible: node.isInaccessible || false,
           destServices: node.destServices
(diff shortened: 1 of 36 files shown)


lucasponce:
@aljesusg this also needs a rebase to proper testing
image

Thanks !

lucasponce:
Side comment but I guess we can address it in this PR as it's related: I think that health tooltip should be shown when healthy AND some error rate. Otherwise user needs to have in mind that there are some defaults. Perhaps not needed under absence of errors, but when I have a 3% I'd like to see that "it's green beacuse the degraded tolerance is 5%".

Note that now with the health config annotations in workload this would be even more necessary.

src/components/TrafficList/TrafficDetails.tsx
```diff
@@ -34,6 +34,7 @@ export interface WorkloadNode {
   type: NodeType.WORKLOAD;
   namespace: string;
   name: string;
+  healthAnnotation: any;
```
lucasponce, line 37: Note that I put some comments in the backend that may affect to the UI code as well.
I'll try to add some early comments that I guess make sense skipping those related to naming described in

https://github.com/kiali/kiali/pull/3362

For this one, I guess this type can be a string instead of any.

src/pages/ServiceDetails/ServiceInfo/ServiceInfoDescription.tsx
```diff
@@ -151,6 +151,7 @@ class ServiceInfoDescription extends React.Component<ServiceInfoDescriptionProps
                       );
                     })}
                   </Stack>
+                  33
```
lucasponce, line 154: Typo

src/pages/ServiceDetails/ServiceInfo/__tests__/__snapshots__/ServiceInfoDescription.test.tsx.snap
```diff
@@ -137,6 +137,7 @@ exports[`#ServiceInfoDescription render correctly with data should render servic
                 />
               </StackItem>
             </Stack>
+            33
```
lucasponce, line 140: Snatshop typo.

I think most of our snapshots tests are a little bit outdated and the only test they perform is "it renders with some basic info".

But nothing to do with this PR, just this is an example about the "outdated" status of these ones.

src/types/ErrorRate/__tests__/utils.test.ts
```diff
@@ -0,0 +1,77 @@
...
+  describe('Parse Annotation', () => {
+    it('should return tolerances with defaults if only code', () => {
+      var tol = defaultTolerance();
+      tol.code = new RegExp('[45]');
+      expect(U.parseAnnotationTEST('[45]')).toEqual(tol);
+    });
+
+    it('should replace x or X by d', () => {
+      var tol = defaultTolerance();
+      tol.code = new RegExp('[45]d[2]');
```
lucasponce, line 24: This RegExp:
```
tol.code = new RegExp('[45]d[2]');
```
won't match any "442" I guess it should be 

```
tol.code = new RegExp('[45]\\d[2]');
```

Instead, and from here propagate this in places where used, I'm not finished with the review but I was stopped by this one.

(Unless I'm missing any other transformation out of the scope of that test).

src/types/ErrorRate/__tests__/utils.test.ts
```diff
@@ -0,0 +1,77 @@
...
+      expect(U.parseAnnotationTEST('[45]x[2]')).toEqual(tol);
+      expect(U.parseAnnotationTEST('[45]X[2]')).toEqual(tol);
```
lucasponce, line 26: duplicate line ? or would you expect add some test, as commented, I thik that the replace of x by d is broken.
An easy way to validate it could be to add a test of a matching with the resulting parsed annotation.

(That would help any other maintainers to understand it; nothing wrong, just thinking loud minor changes that may help in that sense).

src/types/ErrorRate/utils.ts
```diff
@@ -115,3 +121,62 @@ export const aggregate = (
...
+    }
+    if (len > 0) {
+      // Defined Protocol or direction
+      if (len > 1) {
+        direction = new RegExp(cads[1]);
+      }
+      protocol = new RegExp(cads[0]);
+    }
+  }
+  const regex = code.toLowerCase().replace('x', 'd');
```
lucasponce, line 174: Commented in the test, I think this miss the `\\d` format.
Locally fetched it seems to not work as expected.

src/types/ErrorRate/utils.ts
```diff
@@ -115,3 +121,62 @@ export const aggregate = (
...
+  }
+  return getRateHealthConfigBy(ns, name, kind).tolerance;
+};
+
+const isNumber = (value: string | number): boolean => {
+  return value != null && value !== '' && !isNaN(Number(value.toString()));
+};
+
+export const parseAnnotation = (tolerance: string): ToleranceConfig => {
+  const cads = tolerance.split(',');
```
lucasponce, line 142: Perhaps a guard here to return the default in case  split is empty.

aljesusg:
# Update
so time just an update to start feedback

## Tab to see annotation with markers and errors (If the user configure this out the wizard with errors)
tab

## Option in menu to add health annotation
menu

## Annotation builder
builder
builder2


The sliders have some conditions, If failure value is lower than degraded degraded bar is going to move at the same time for example.
We can add or remove annotations.

Next Step: show annotations error to remove them in the wizard

lucasponce:
@aljesusg thanks for the update but I think you should have discussed the feature and design proposed.

I'll start to review and trying to understand the feature proposed and implementation

aljesusg:
> @aljesusg thanks for the update but I think you should have discussed the feature and design proposed.
> 
The feature was opened here https://github.com/kiali/kiali/issues/3435 with some way to show alerts in annotations and a way to edit them. About the design this is why I updated this to open discussions about it. Anyway is a small wizard, not much options
> I'll start to review and trying to understand the feature proposed and implementation

lucasponce:
> The feature was opened here kiali/kiali#3435 with some way to show alerts in annotations and a way to edit them. About the design this is why I updated this to open discussions about it. Anyway is a small wizard, not much options

As discussed in https://github.com/kiali/kiali/issues/3332#issuecomment-712656093 the health config feature is dynamic and it will evolve.

This PR is positive as it proposes an approach (and I understand that seeing a first end-to-end flow is good to review) but in the other hand, https://github.com/kiali/kiali/issues/3435 was pretty open and it would benefit to more short iterations.

In a first thought I think that first we need to focus to show the user the health config that is applied (but merging conceptually all, the config applied for the entire mesh, specific config defined for a namespace, and particular config overwritten for a service/workload), and next, allow to edit them.

Also, from user perspective, Health config defaults should be "good" for 90% of the cases, another 5-9% may overwrite the config for a particular namespace, and a 1-5% of the cases may need to overwrite the health config for a particular service/workload.

In that context, probably this feature make sense to place it close to the health card, but that's just an early thought.

I'm going to start reviewing this idea, I take it as a first prototype (again, it's good to have it), but we'll iterate on it.

aljesusg:
> > The feature was opened here [kiali/kiali#3435](https://github.com/kiali/kiali/issues/3435) with some way to show alerts in annotations and a way to edit them. About the design this is why I updated this to open discussions about it. Anyway is a small wizard, not much options
> 
> As discussed in [kiali/kiali#3332 (comment)](https://github.com/kiali/kiali/issues/3332#issuecomment-712656093) the health config feature is dynamic and it will evolve.
> 
> This PR is positive as it proposes an approach (and I understand that seeing a first end-to-end flow is good to review) but in the other hand, [kiali/kiali#3435](https://github.com/kiali/kiali/issues/3435) was pretty open and it would benefit to more short iterations.
> 
> In a first thought I think that first we need to focus to show the user the health config that is applied (but merging conceptually all, the config applied for the entire mesh, specific config defined for a namespace, and particular config overwritten for a service/workload), and next, allow to edit them.
> 
> Also, from user perspective, Health config defaults should be "good" for 90% of the cases, another 5-9% may overwrite the config for a particular namespace, and a 1-5% of the cases may need to overwrite the health config for a particular service/workload.
> 
> In that context, probably this feature make sense to place it close to the health card, but that's just an early thought.
> 
> I'm going to start reviewing this idea, I take it as a first prototype (again, it's good to have it), but we'll iterate on it.

Ok, so let me separate this in 2 PR to address this.

1- Let apply the new annotation
2- Let modify it.

What do you think

lucasponce:
> Ok, so let me separate this in 2 PR to address this.

> 1- Let apply the new annotation
> 2- Let modify it.

> What do you think

It sounds good, I would propose you something similar, (1) it's more straight forward and I think most of the backend work is still valid and most of them discussed (the annotation feature needs to be tested in an end-to-end way).

And (2) Show the health config applied and modify it, it's an open area in terms of feature (I don't see the annotation as something user want to edit from Kiali, the annotation is how Kiali implement it, so that may change some of the UI ideas proposed here).

So, I'm +1 to focus on (1), as that's clear and we can progress, I'd suggest to keep this PR as the prototype that we can review for (2) as probably it may have more iterations (from user view).

abonas:
Regarding the tab named "Kiali annotation" - the naming itself perhaps is a bit misleading?

aljesusg:
> Regarding the tab named "Kiali annotation" - the naming itself perhaps is a bit misleading?

Yes, It should be Health Annotation

abonas:
> > Regarding the tab named "Kiali annotation" - the naming itself perhaps is a bit misleading?
> 
> Yes, It should be Health Annotation

annotation or is it plural "annotations"?

lucasponce:
> Yes, It should be Health Annotation

Note that this feature is still quite active it may iterate.

Probably from user's perspective we wouldn't like to present this tab as it's but propose a different approach about what is the "health config" applied for a given element, without showing the low level details.

Annotation on this context is more a implementation detail.

Again, we'll have more iterations around this.

lucasponce:
This POC is good but it will require more planning and discussion in next Epics.
Short term actions are handled on #2056.

I'm going to close it to not merge it by mistake.

PR #1987, opened 2020-10-30
Compare metrics-based stats for every envoy spans, using a heatmap

jotak:
It uses new endpoint `api/stats/metrics`

Fixes https://github.com/kiali/kiali/issues/3293

Required backend PR: https://github.com/kiali/kiali/pull/3347

### DIFF
--- a/src/components/JaegerIntegration/JaegerResults/StatsComparison.tsx
+++ b/src/components/JaegerIntegration/JaegerResults/StatsComparison.tsx
@@ -14,7 +14,7 @@ import { Button, ButtonVariant, Tooltip } from '@patternfly/react-core';
 // TODO / follow-up: these variables are to be removed from here and managed from the toolbar
 const statsQuantiles = ['0.5', '0.9', '0.99'];
 const statsAvgWithQuantiles = ['avg', ...statsQuantiles];
-const statsIntervals = ['6h', '60m', '10m'];
+const statsIntervals = ['10m', '60m', '6h'];
 const statsPerPeer = false;
 let statsCompareKind: 'app' | 'workload' = 'workload';
 
@@ -107,8 +107,8 @@ const renderHeatMap = (item: SpanTableItem, stats: StatsWithIntervalIndex[]) => 
   const data = buildStatsData(item, stats);
   return (
     <HeatMap
-      xLabels={statsIntervals}
-      yLabels={statsAvgWithQuantiles.map(s => statToText[s]?.short || s)}
+      xLabels={statsAvgWithQuantiles.map(s => statToText[s]?.short || s)}
+      yLabels={statsIntervals}
       data={data}
       colorMap={HeatMap.HealthColorMap}
       dataRange={{ from: -10, to: 10 }}
--- a/src/components/JaegerIntegration/JaegerResults/StatsComparison.tsx
+++ b/src/components/JaegerIntegration/JaegerResults/StatsComparison.tsx
@@ -3,7 +3,7 @@ import { InfoAltIcon } from '@patternfly/react-icons';
 import _round from 'lodash/round';
 
 import * as API from 'services/Api';
-import { Color, HeatMap } from 'components/HeatMap/HeatMap';
+import { HeatMap } from 'components/HeatMap/HeatMap';
 import { MetricsStatsQuery, Target } from 'types/MetricsOptions';
 import { EnvoySpanInfo } from '../JaegerHelper';
 import { SpanTableItem } from './SpanTableItem';
@@ -12,12 +12,10 @@ import { PfColors } from 'components/Pf/PfColors';
 import { Button, ButtonVariant, Tooltip } from '@patternfly/react-core';
 
 // TODO / follow-up: these variables are to be removed from here and managed from the toolbar
-const statsQuantiles = ['0.5', '0.75', '0.9', '0.99'];
+const statsQuantiles = ['0.5', '0.9', '0.99'];
 const statsAvgWithQuantiles = ['avg', ...statsQuantiles];
 const statsIntervals = ['6h', '60m', '10m'];
 const statsPerPeer = false;
-const statsRed: Color = { r: 1, g: 0, b: 0 };
-const statsGreen: Color = { r: 0, g: 1, b: 0 };
 let statsCompareKind: 'app' | 'workload' = 'workload';
 
 const statToText = {
@@ -112,8 +110,8 @@ const renderHeatMap = (item: SpanTableItem, stats: StatsWithIntervalIndex[]) => 
       xLabels={statsIntervals}
       yLabels={statsAvgWithQuantiles.map(s => statToText[s]?.short || s)}
       data={data}
-      colorRange={{ min: statsGreen, max: statsRed }}
-      dataRange={{ min: -10, max: 10 }}
+      colorMap={HeatMap.HealthColorMap}
+      dataRange={{ from: -10, to: 10 }}
       colorUndefined={PfColors.Black200}
       valueFormat={v => (v > 0 ? '+' : '') + _round(v, 1)}
       tooltip={(x, y, v) => {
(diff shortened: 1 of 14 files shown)


jotak:
In traces details page / spans list:

Capture d’écran de 2020-10-30 11-49-06

src/components/JaegerIntegration/JaegerResults/StatsComparison.tsx
```diff
@@ -0,0 +1,181 @@
...
+import * as API from 'services/Api';
+import { Color, HeatMap } from 'components/HeatMap/HeatMap';
+import { MetricsStatsQuery, Target } from 'types/MetricsOptions';
+import { EnvoySpanInfo } from '../JaegerHelper';
+import { SpanTableItem } from './SpanTableItem';
+import { MetricsStats } from 'types/Metrics';
+import { PfColors } from 'components/Pf/PfColors';
+import { Button, ButtonVariant, Tooltip } from '@patternfly/react-core';
+
+// TODO / follow-up: these variables are to be removed from here and managed from the toolbar
```
jotak, line 14: About this TODO: I created a task: https://github.com/kiali/kiali/issues/3379

lucasponce:
Adding @lwrigh.

@jotak conceptually the heatmap in the Span looks good, I like it, but I think it's taking too much space in the row and personally I see it weird.

I suggest to add a topic in the UX agenda for next Tuesday and see if Laura can help.

to @lwrigh, feel free also to add questions and/or commens even to catch up with the context.

jotak:
@lucasponce yes, I wanted to see it more horizontal than vertical as well, so I did several tries, like switching the axis (but it's weird to have durations on Y) or changing default stats to show more durations and less percentiles (but IMO this is less interesting).

Also note that, if we do the follow-up task that I'm planning ( https://github.com/kiali/kiali/issues/3379 ), percentiles and durations to show could be changed, which means that the shape of the heatmap will change as well.

jotak:
(probably we can remove the p75 line by default, it will still be vertical, but not as much as now)

lwrigh:
PatternFly 4 does not currently have a heat map component but should be a component that's available in the future. PatternFly 3 has a heat map component if you want to use that instead so it's more consistent with the rest of the UI: https://www.patternfly.org/v3/pattern-library/data-visualization/heat-map/index.html 

If you don't want to change the current heat map component that is being used I would suggest to update the colors to be more consistent with PatternFly at least. PatternFly colors for charts: https://www.patternfly.org/v4/guidelines/colors-for-charts

We can talk further about the different options in the next UX meeting.

jotak:
Thanks @lwrigh

The problem about the colors is that because the nature of the heatmap, these are gradient computed mathematically, so they won't strictly match pre-predefined colors, except perhaps at the boundaries. Having a heatmap built from a limited set of colors (like in PF3) removes all the tones that I think are important for this kind of dataviz. If we take a look at grafana for instance, heatmaps show gradiant colors: https://www.google.com/search?q=grafana+heatmap

We can discuss it in the next meeting

jotak:
I'm still thinking about it .. trying to see if we could map pf4 color families to a scale that goes, more or less linearly, from plain red (worst) to plain green (best). If there is a decent number of tones, it worth it to try. I'm annoyed with the "green" family, I think it only have 3 usable tones out of the 5 proposed.

lwrigh:
> I'm still thinking about it .. trying to see if we could map pf4 color families to a scale that goes, more or less linearly, from plain red (worst) to plain green (best). If there is a decent number of tones, it worth it to try. I'm annoyed with the "green" family, I think it only have 3 usable tones out of the 5 proposed.

I would also take a closer look at how accessible some of the color combinations are too. The contrast of the text against the square background color doesn't look like there's enough contrast for some of them.

hhovsepy:
@jotak can these comparison maps be expandable per row?
Screenshot from 2020-11-02 11-29-53

jotak:
@hhovsepy not sure what you mean, you'd like them to not be displayed by default, but only on demand?

hhovsepy:
I mean hide these column content by default, and instead show 'show more info' link, and expand the table only when clicking 'show more info' link. Just my idea from UX perspective, like we do for labels in lists.

lucasponce:
> I mean hide these column content by default, and instead show 'show more info' link, and expand the table only when clicking 'show more info' link. Just my idea from UX perspective, like we do for labels in lists.

I'm also inclined to think that we are accumulating a lot of good info per row, so perhaps a table/expandable pattern can suit on it http://patternfly-react.surge.sh/components/table#expandable

About the heatmap, I'm undecided I see several drawbacks:
- It uses a lot of space.
- Also, it delegates to the user to but the absolute value of the span inside of the heatmap to compare.

So, my overall suggestion:
- Use a table/expandable pattern.
- Try to see if some custom heatmap or minigraph suits, where it's clear there is the span request time compared with the stats.

Some UI suggestions, but here I delegate on @lwrigh comments, I personally see very clean a similar pattern from Github:
image

I'd also included in a different contrast where the span is located inside, and I'd use more tooltip to get details, instead to put too much explicit legends.

Goal is to get the attention of the user in the span that it may require further investigation.

jotak:
Here is another try, with a couple of changes:

- Although it's still uses gradient / color interpolation, I'm using a bunch of patternfly colors as "key frames", so the result looks more "patternflyish".
- Remove p75 to make it smaller
- Removed cell borders

Capture d’écran de 2020-11-04 11-12-38

mtho11:
This is quite cool! Perhaps in the future it could even be made more interactive, such that the time period (duration) is adjustable via a slider. This way you could see it change in time by just moving the slider. Similar to our existing **Graph Replay** technology. IMO it would make a really cool demo feature.

jotak:
Here's a new screenshot:

Capture d’écran de 2020-11-09 13-47-45

Changes are:
- Swapped axis (to make it horizontal by default)
- margin & border radius on cells
- reversed interval order (10m to 6h)
- increased text contrast

@lwrigh I'm not sure if the patternfly table component can have sticky headers as we discussed, I couldn't find anything like that ( https://www.patternfly.org/2020.04/documentation/react/components/table )

jotak:
> This is quite cool! Perhaps in the future it could even be made more interactive, such that the time period (duration) is adjustable via a slider. This way you could see it change in time by just moving the slider. Similar to our existing **Graph Replay** technology. IMO it would make a really cool demo feature.

Thanks, I've mentioned it in this follow-up task, because it wouldn't be for that PR: https://github.com/kiali/kiali/issues/3379

jotak:
Another follow-up tasks entered to the backlog: https://github.com/kiali/kiali/issues/3420 => for expandable mode

jotak:
BTW, personally I prefer the version with plain squares & no margin, wdyt?

lucasponce:
I've fetched locally and started to review.

A first comment, it seems that it needs a wide resolution, in 1500px x 1200px I got this displacement in the titles of the heatmap:

image

lucasponce:
Note, probably for a new issue, but a "Killer" feature would be to put a heatmap in a floating layer when clicking a trace, like:
image

jotak:
> Note, probably for a new issue, but a "Killer" feature would be to put a heatmap in a floating layer when clicking a trace

Good idea. But since the comparisons are done per span, not per trace, we need to aggregate all compared spans stats (like: take the worst of each stat? the average?). Anyway it's worth thinking about it !

lucasponce (approved):
LGTM
If @hhovsepy can give a final test we can consolidate.

I have a naming comment in the backend code, but easy, I think.

lucasponce:
> Good idea. But since the comparisons are done per span, not per trace, we need to aggregate all compared spans stats (like: take the worst of each stat? the average?). Anyway it's worth thinking about it !

Perhaps it's simpler for a first step, look at the chart:

image

As a user I have an average of "25ms", which is seems slow compared with others, but the heatmap would tell me if that "25ms" is very bad or something okey based on the context.

Obviously analytics can go deep as we can, but I was refering using the current data (trace avg time used to place it in the chart, and heatmap calculated from stats).

Anyway, my thinking has always same topic: how to locate the span I'm interested on.

jotak:
> A first comment, it seems that it needs a wide resolution, in 1500px x 1200px I got this displacement in the titles of the heatmap:

Fixed, adding `nowrap` to css

hhovsepy (approved):
Verified in all traces tabs:
Screenshot from 2020-11-12 13-03-42

jotak:
Thanks!

PR #1992, opened 2020-11-03
Refactor TimeControl in List+Details pages

lucasponce:
As a previous step to address https://github.com/kiali/kiali/issues/3350 the TimeControl component belongs to the page instead to the subcomponent used.

To properly resolve the https://github.com/kiali/kiali/issues/3350 issue is necessary to apply this refactoring, otherwise we will continue using some hacks to address instead to solve it in a good way.

Also, as a side effect of this work, I'm trying to keep consisntency between pages, like:

image


lucasponce:
Some details about the work in progress on this:
- Discussed with @jshaughn to move the time controls to the upper place in all pages for consistency.
- Moved time controls to page scope instead of tab.
- Every tab will keep it's logic, main change is how state is being handled.

Then, look and feel show settle for list:

image

And for details:

image

Note, when page is in a narrow state, time controls won't overlap tabs:
image

As in the previous implementation (keep it in the previous screenshot just to give more context to this comment).

Again, this is not a refactoring of the logic, this PR won't change that, it's just a previous refactoring of these components.

It's still in WIP, most of these changes are not yet updated in the PR.

lucasponce:
Side comment, patternfly has a now a formal datepicker, in the context of this task I'm going to test if it improves the current component that we are using as it has some issues with the layout and style.

lucasponce:
Some WIP in details page, as a solution for placing actions:

image

Aligned with list pages as well:

image

image

lucasponce:
I think I've addressed some pure style / alignment problems and we may have a consolidated style for common headers.

This is how it looks like when this PR is applied.

Overview

image

image

Graph

image

App/Workload/Service List

image

Istio Config List (with Actions)

image

Extensions / Iter8 List (combined time control + actions)

image

Details without Actions

image

Details with custom Time (with and without overlap)

image

image

Details with Actions

image

image

image

Istio Details (No time, but manual refresh + actions)

image

Extensios / Iter8 details (auto-refresh + actions)

image

Some things are still broken but you'll get an idea about the changes.

I'm going to continue fixing the current broken parts and addressing latest comments, but any early comment about these changes is welcome, it took time to consolidate this and I'd like to have the PR prepared soon.

lucasponce:
Ok.
I've moved this PR to review mode.

It's not 100% finished, I haven't removed the debug lines yet, but as it touches a wide number of different files the probability that something is broken is close to 1.

This PR basically refactors timecontrols for two reasons: they don't belong to the tab but to the upper page. With that goal in mind, it refactors the overview/list/details pages.

To reviewers: two main goals
- Regression testing of existing features.
- PR is not finished and I have pending actions to do, it won't refactor more things than planned, the idea is to leave in a status that can be easy to address other concerns.

src/components/BreadcrumbView/BreadcrumbView.tsx
```diff
@@ -78,13 +89,22 @@ export class BreadcrumbView extends React.Component<BreadCumbViewProps, BreadCum
     return this.state.pathItem === 'istio';
   };
 
+  isIter8 = () => {
+    return this.state.pathItem === 'iter8';
+  };
```
jshaughn, line 94: I guess breadcrumb support for extensions is currently explicit?    How would a dev know to add this support, is it documented?

src/types/Common.ts
```diff
@@ -68,3 +68,12 @@ export const durationToBounds = (duration: DurationInSeconds): BoundsInMilliseco
     from: new Date().getTime() - duration * 1000
   };
 };
+
+export const isEqualTimeRange = (t1: TimeRange, t2: TimeRange): boolean => {
+  if ((t1 as BoundsInMilliseconds).from && (t2 as BoundsInMilliseconds).from) {
+    const t1b = t1 as BoundsInMilliseconds;
+    const t2b = t2 as BoundsInMilliseconds;
+    return t1b.from === t2b.from && t1b.to === t2b.to;
```
jshaughn, line 76: .to? is optional but I think this still looks good.

src/components/JaegerIntegration/TracesComponent.tsx
```diff
@@ -41,6 +38,7 @@ interface TracesProps {
   namespaceSelector: boolean;
   showErrors: boolean;
   duration: number;
+  lastRefresh: number;
```
jshaughn, line 41: These should likely both be typed using out existing definitions, like DurationinSeconds and TimeInMilliseconds (those are examples, not sure if those units are correct).  This comment can likely be applied in multiple places.

jshaughn (commented):
@lucasponce Things behave pretty well, I tried things on Chrome and FF.  Below are some comments, not approving or requesting changes at this point as I know the PR is not finished.  Just making comments...

- Without looking too deeply I see a lot of 'lastRefresh' prop/state var additions.  I just want to make sure that this is necessary as we already have 'lastRefreshAt' in redux state, and any container can be updated automatically by declaring this in redux state.  lastRefreshAt is updated by the Refresh container.

- This doesn't necessarily have to be fixed but narrowing in the graph is somewhat degraded.  It jumps from 2 lines to 4 or 5, in sort of a weird way. As soon as TimeRange wraps we also see a wrap with GraphFind and 'graph tour'.
Just before wrap:
image
Just after wrap:
image
Just after wrap (master):
image

- This may just be personal preference, I think I mentioned before, but the placement of 'Actions' bothers me.  It's in-between rows.  I think I'd pefer it to either have its own row (moving the tabs down a bit) or on the same horizontal as the tabs.

image


- There is some strange behavior with the left menu when reducing and increasing the window width, but this is not new, I can get the something like this to happen on master as well:

image

src/components/Health/HealthDetails.tsx
```diff
@@ -52,7 +52,7 @@ export class HealthDetails extends React.PureComponent<Props, {}> {
         )}
       </div>
     ) : (
-      <></>
+      <div key={idx}></div>
```
jotak, line 55: If this is just to avoid react warning about keys, iirc you can do instead: `<React.Fragment key={idx}/>` . That way we don't "pollute" the DOM, if i'm correct.

src/pages/AppDetails/AppDetailsPage.tsx
```diff
@@ -164,22 +166,15 @@ class AppDetails extends React.Component<AppDetailsProps, AppDetailsState> {
...
-          );
         tabsArray.push(
-          <Tab eventKey={4} style={{ textAlign: 'center' }} title={jaegerTitle} key={tracesTabName}>
+          <Tab eventKey={4} style={{ textAlign: 'center' }} title={'Traces'} key={tracesTabName}>
             <TracesComponent
               namespace={this.props.match.params.namespace}
               target={this.props.match.params.app}
               targetKind={'app'}
-              showErrors={this.state.nbErrorTraces > 0}
+              showErrors={false}
```
jotak, line 175: This is a functional change, we won't restrict the traces view to show only errors where there are some, is that the intent?
I'm +1 for this change anyway.
If so, I guess this `showErrors` props isn't used anymore and could go away.

jotak:
>     * Without looking too deeply I see a lot of 'lastRefresh' prop/state var additions.  I just want to make sure that this is necessary as we already have 'lastRefreshAt' in redux state, and any container can be updated automatically by declaring this in redux state.  lastRefreshAt is updated by the Refresh container.
> 

Also reacting to this `lastRefresh`, I'd like to understand why it is necessary and if we could avoid it. Do these changes generate more component updates that need to be avoided?

lucasponce:
Thanks @jshaughn for your comments.

About the "Actions": we are not implementing a new pattern, but an existing one. "Actions" where placed in that location for quite some time, i.e. "Istio Config Detail":

image

Also, in terms of CSS align, now all pages are having the same height in the "toolbar" + "main area".
I attach a working screenshot used to check that:

image

So, pushing the tabs down into the main area would break the main height consistency across pages.

Also, actions won't belong to the tab content, because those are related to the main element.

In a non "narrow" scenario it works ok:

image

image

image

image

Also, "Actions" are in same level of distance that second line of controllers introduced in the Graph page.

So, main issue here is that the "tabs" component takes space of the upper component in terms of visualization and that would trigger a major change, so, at least for the scope of the PR I wouldn't change that.

jotak:
Hint for reviewers: if you don't want to be mad 😱 , check the "hide whitespaces" box in files diff! 😉

Capture d’écran de 2020-11-16 10-54-07

lucasponce:
> Also reacting to this lastRefresh, I'd like to understand why it is necessary and if we could avoid it. Do these changes generate more component updates that need to be ignored?

Rationale for this was that now the tab is not responsible to initiate a refresh but it's the page upper component.

A refresh fetches all backend but not in all tabs, so that variable was used to force a change of state and a new render.

But I'm reviewing @jshaughn comments as I didn't see this variable was collected by the refresh component as it's available in redux, so I'm going to work on that.

src/pages/AppDetails/AppDetailsPage.tsx
lucasponce: Yes, sorry, I should leave a note.
Also, this was not consistency as error icons only happen on "app/traces" details tab, not on "workload/traces" nor "service/traces".

Also, with all new logic, I guess it was good to reconsider that icon, because in the same way, it should appear also in the metrics and log tabs (as error would happen there).

So, yes, I guess that for this moment, we can remove that icon in the tab, and delegate in the filter options used on every tab.

src/components/Health/HealthDetails.tsx
lucasponce: I'll see, I didn't know that trick.
Yes, this was to avoid an annoying warning for lack of keys.

src/components/JaegerIntegration/TracesComponent.tsx
lucasponce: Good point, I didn't see that this property was injected from the refresh container.
Then perhaps it's not necessary in children components.
I need to work on it.

src/components/BreadcrumbView/BreadcrumbView.tsx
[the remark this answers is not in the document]
lucasponce: This question is ambiguous in terms of time and space :-).
The answer is yes, it's documented:
https://kiali.io/documentation/latest/extensions/#_kiali_front_end_guideline

And also doc needs to be updated once things are consolidated.

For extensions, in the first approach, I tried to avoid tight couple with other shared components, but for these scenarios, perhaps I can create (in a follow up PR) some better extensibility model. I.e. a "ExtensionBreadcrumbView" which basically uses the "BreadcrumbView" but adding custom logic like this.

But I don't want to mix these concerns in the current PR as there are places that it still require explicit support (i.e. adding a new page).

xeviknal:
One thing I see in the metrics pages is the refreshing effect. I reload the tab content instead of rerendering each of the charts in it

reloading-metrics-tab-refresh

Compared to master and to traces page, the whole tab content is not refreshed, only the charts are re-rendered.

lucasponce:
@xeviknal thanks.
Yes, there is something broken on the refresh, I'm actively working on that now.

hhovsepy:
Filter name and value are not aligned:
Screenshot from 2020-11-16 14-10-56

lucasponce:
@hhovsepy 

> Filter name and value are not aligned:

Not sure if I understand the comment.

This is a screenshot of this PR:

image

And this is master:

image

I see a minor CSS gap in the padding in both, but about the filter/value I need help to understand what is not aligned (there's a 1px gap between filter and text box, perhaps you refer to that ? But I see it in both PR and master).

aljesusg:
@lucasponce Did you see this? I don't know if is a behaviour related with browser or PF4. It's only happening in input

## Demo
input

lucasponce:
> @lucasponce Did you see this? I don't know if is a behaviour related with browser or PF4. It's only happening in input, I think that we can go ahead

Yeah, this is a known issue, it's a mix between PF and a third party datepicker component.

I'd like to address in a followup PR.

lucasponce:
With a minor change in the label "Show Health for" -> "Health for", overview page may support 1024x800 screens without breaking the default layout.

I don't think we can't do more without thinking in more deep changes.

image

xeviknal:
In the metrics page, the traces are disabled after every refresh:

Screen recording (15) (1)

And for me, the reload still reloads the whole tab not only the graph objects.

lucasponce:
> And for me, the reload still reloads the whole tab not only the graph objects.

Not sure if I understand this comment.
Can you see which queries to the backend performs and what is expected compared with master ?

xeviknal:
Master doesn't reload the whole tab, it reload only the content within each chart. It gives the illusion of new datapoints being added to the chart.

Screen recording (16) (1)

This PR:
Screen recording (17) (1)

lucasponce:
> Master doesn't reload the whole tab, it reload only the content within each chart. It gives the illusion of new datapoints being added to the chart.

Okey, thanks.

Yes, now page is the component who starts a refresh, I'm going to see if we can get the previous behaviour with the current design.

jotak:
I've finished some partial code review, will test tomorrow.

lucasponce:
> I've finished some partial code review, will test tomorrow.

Thanks.
Note that I'm addressing on some comments from @xeviknal and also working in some changes with the duration/timeRange to align both components.

So probably there will be another update tomorrow.

(Sorry if that may force to re-review, I hope the scope of these changes will be the detail pages only).

jotak:
I think there's an issue with the sync'ing of duration between pages. Consider this scenario:

1. From a workload overview tab, set a duration (e.g. 10 min)
2. Switch to the metrics page and set a different duration (e.g. 1 hour)
3. Switch to overview tab, the duration is the previous overview's one (e.g. 10 min)
4. Switch back to the metrics page, the duration is still the overview one, not the metrics one

So, depending on what we want, there's a problem either in step 3 or 4. Til now we have been using two decoupled durations, one for short period data representation (like graph) and one for large periods of data (like metrics). If it hasn't changed, then the problem is on step 4: metrics page should keep its configured duration. If we want the behaviour to change, then the problem is on step 3: overview page should keep what was set in metrics page.

jotak:
> So, depending on what we want, there's a problem either in step 3 or 4. Til now we have been using two decoupled durations, one for short period data representation (like graph) and one for large periods of data (like metrics). If it hasn't changed, then the problem is on step 4: metrics page should keep its configured duration. If we want the behaviour to change, then the problem is on step 3: overview page should keep what was set in metrics page.

I think the current changes, that put the time controls look more global than how it was before, should involve some changes in behaviours too. I've always been advocating decoupling time between graph and metrics because the use cases are different and I still think it is better in terms of page context, however now, it is harder to advocate because the time controls *looks* unique.

Hence I'm wondering if we could imagine a solution to make it visible, somehow, that the time controls are referring two different settings, one being for historical data sets and one for "snapshot-like" data sets. Not sure yet how we could do that.

jotak:
In Istio/Custom Metrics page, I think there's a missing redux-dispatch to notify time interval change, in function `onDomainChange`. This function is called when you use the "brush" selector on charts (remember, you can drag on chart to select a custom period).

lucasponce:
Thanks @jotak for the review.
Yes, yesterday I realized in the linking issues, I'm working on it.

The strategy I'm using is to align TimeRange control with other TimeControl and store it in Redux, and connect to the components that use it.

In local it seems to work and address most of the concerns.

As this PR doesn't pretend to change the behaviour, it should store two different of "time data sets", once used for regular Time selection, and the other used for the control used in metrics+logs that enables "Custom" intervals.

lucasponce:
Also, another minor detail to address is that same web param "duration" is used for both context, so to avoid some clash here I'm going to rename this for the TimeRange use cases, then it would keep the same behaviour described (or that's the intention).

hhovsepy:
Yes I mean that 1px gap.
> I see a minor CSS gap in the padding in both, but about the filter/value I need help to understand what is not aligned (there's a 1px gap between filter and text box, perhaps you refer to that ? But I see it in both PR and master).

lucasponce:
> Yes I mean that 1px gap.

Ok, thanks for clarification, as I see in master, probably I'll address it in a follow up task, this PR is getting much more bigger than I expected and I'm going to focus to consolidate all the features and no regressions first.

hhovsepy:
In custom dashboards, the button and dropdowns is doubled:
Screenshot from 2020-11-18 14-24-49

jshaughn:
> Hence I'm wondering if we could imagine a solution to make it visible, somehow, that the time controls are referring two different settings, one being for historical data sets and one for "snapshot-like" data sets. Not sure yet how we could do that

Today: Overview, Graph, Detail-Overview, and Detail-Traffic are tied to timeRange-1 and the other Detail-* tabs are tied to timeRange-2.

timeRange-1 has a default of 1m and does not allow custom ranges.
timeRange-2 has a default of 10m and allows custom ranges.

It's confusing but it makes some sense as typically Overview and Graph are looking for near-term issues and Metrics are interested in longer, trend-oriented views.  Logs could go either way but get tied to metrics because, I think. locality and the need for a custom range.

To be perfectly honest, I'm not sure users have noticed very much and may simply change it for the page/tab they are on based on need.  As suchm  I'm not really advocating a change.  But if we were to make a change I's suggest maybe doing the following:
- Give Overview its own time range (maybe change default to 5m).
- Give Graph its own time range (keep default at 1m).
- Give Detail its own time range, allowing custom, and apply it to every tab (keep default of 10m).
- Fix the tooltips for each to indicate more specifically its scope.

If nothing else, I think with the current time controls we could update the tooltips to  be more specific to the current scopes.  This could/should be done in this PR.

lucasponce:
> In custom dashboards, the button and dropdowns is doubled:

Right, noted. I'll fix this in the next bunch of changes.

lucasponce:
> To be perfectly honest, I'm not sure users have noticed very much and may simply change it for the page/tab they are on based on need. As suchm I'm not really advocating a change. But if we were to make a change I's suggest maybe doing the following:
> 
> * Give Overview its own time range (maybe change default to 5m).
> * Give Graph its own time range (keep default at 1m).
> * Give Detail its own time range, allowing custom, and apply it to every tab (keep default of 10m).
> * Fix the tooltips for each to indicate more specifically its scope.
> 
> If nothing else, I think with the current time controls we could update the tooltips to be more specific to the current scopes. This could/should be done in this PR.

I didn't want to bring that discussion on this PR.

To me, the goal of this work is to try to consolidate with the less possible changes.

About this discussion, as I commented in previous issues, I don't like and I don't see that user needs to keep two different states. 

Sorry, but not, if there is a common component, the main reasoning is to keep that state in all possible scenarios.

But, we have two different controls here, one that admits custom time ranges, and other that only support duration.

So, it's logical that custom time ranges can be propagated.

In any case, we are discussing if the user needs to perform one extra click or not (i.e. I select 5 minutes in the graph, I move to a metrics page and I need to change to 10 minutes).

I don't see a good reason why duration can't be shared here.

Yes, semantically perhaps in that tab I want to see a long duration (or not).

But again, in this PR I hope to put the refactoring in a stable point, additional more specific changes around this can be addressed later.

hhovsepy (approved):
LGTM, as agreed comments will be resolved in next PRs.

lucasponce:
Ok, I guess it's in a stable state now, with all/major comments fixed.

@jotak I think I've addressed all your comments and previous behaviour is implemented, from here, potential changes in one or another direction should be easier to deal with.

@aljesusg the style in the custom datepicker is now fixed.

@xeviknal the refresh behaviour should work as normal.

@jshaughn I think all your comments were addressed as well.

@hhovsepy only a couple of CSS comments are pending, but hopefully those can be addressed in a next PR.

Thanks for all the reviews and discussion, let me know if you miss anything else on this.

lucasponce:
Ok, in a self review I saw that list pages have something broken on the sorting headers.

Let me know if you find anything else, thanks

lucasponce:
> Ok, in a self review I saw that list pages have something broken on the sorting headers.

I'll address this one on https://github.com/kiali/kiali/issues/3437 work.

src/components/Time/MainTimeControl.tsx
```diff
@@ -0,0 +1,23 @@
+import * as React from 'react';
+import TimeControlsContainer from './TimeControls';
+import TimeRangeComponent from './TimeRangeComponent';
+import RefreshContainer from '../Refresh/Refresh';
+
+type Props = {
+  customDuration: boolean;
+};
+
+export default class MainTimeControl extends React.Component<Props> {
```
jshaughn, line 10: Naming is always hard but I do find it a little confusing to have MainTimeControl and also TimeControls.   Maybe this could just be "TimeControl" if we were to rename TimeControls to be more in line with TimeRangeComponent.  TimeRangeComponent is well named, it indicates that you can set a complete time range (aka custom time range).  TimeControls is the component that is a little weird, maybe it should be DurationComponent to make the difference more explicit.  I realize that it also includes the Refresh whereas TimeRangeComponent does not.  Anyway, just an observation, not a blocker.

jshaughn (approved):
@lucasponce Looking really good for me, no blockers found in code or testing.  Nice job getting rid of the drill-down time props and deferring to redux.  A couple of code comments and testing observations for your consideration, but no blockers.

Possible issues (or just suggestions):
- Now that the time range is always just below the primary masthead I suggest setting the tooltip position to bottom to avoid the black on black.  Also, I think the tooltip text should be updated to better distinguish between the time range settings:
  - "Sets common time range for Overview and graph pages, as well as Detail page Overview and Traffic tabs.
  - "Sets common time range for Detail page Metric and Log tabs. Custom time ranges are supported".
Non-Blocking issues (changed behavior but OK):
- Graph toolbar wrapping is not as good as before, but only happens on very narrow, unlikely widths.

Non-Blocking issues (shows same on master):
- Overview list view kebab click is sometimes immediately dismissed (menu comes and goes).
- Overview list view Status sort seems not to work.
- Overiew small cards do not have uniform height.
- Overview list view traffic sparkline can spill into next row.

Approving, any changes are at your option.

jotak:
I think there is a bug in the Traces page: changing the duration there seem to have no effect

lucasponce:
Thanks @jshaughn some of your comments will be implemented in kiali/kiali#3437 but I'm going to consolidate this first.
We can iterate on the other non-blocking comments.

lucasponce:
> I think there is a bug in the Traces page: changing the duration there seem to have no effect

ah !

jotak (changes requested):
Looks like the Tracing page is broken, changing the duration there has no effect, the chart keeps showing always the same interval of time

jotak:
Also it seems like the default metrics duration has been changed to 1m, before it was 10m if i'm correct

src/app/History.ts
```diff
@@ -32,6 +32,7 @@ export enum URLParam {
   OPERATION_NODES = 'operationNodes',
   OVERVIEW_TYPE = 'otype',
   QUANTILES = 'quantiles',
+  RANGE_DURATION = 'rangeDuration',
```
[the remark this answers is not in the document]
lucasponce: It may create some conflicts when navigating from one tab to another as there is a process to set the URL to state and viceversa.

I found some race on that logic, so keeping this separated I guess it helps.

src/components/Time/MainTimeControl.tsx
lucasponce: +1
TimeControl, and TimeDuration(Component) and TimeRange(Component) sounds like a good way.

lucasponce:
Ok, @jotak traces tab was getting the duration instead rangeDuration, good catch.

@jshaughn I've changed the tooltip position, and some of the suggestions will go in a separate PR were consolidating the empty list pages.

jadeyliu939 (commented):
Look nice :-)

jotak (approved):
LGTM, nice work!

lucasponce:
Thanks for all the reviewers. Merging.

PR #1994, opened 2020-11-04
Make sure API logout completes before submitting openshift logout form

jshaughn:
This is related to, and possibly the root cause fix for https://github.com/kiali/kiali/issues/3313.   The original PR was also needed but the problem resurfaced during testing for https://github.com/kiali/kiali/issues/2905.  So this is being included as part of the fix for that issue.

SERVER PR: https://github.com/kiali/kiali/pull/3397

Testing this issue is somewhat tedious but a series of successful login/logouts is what you are looking for, with OpenshiftAuth.  The fault was more frequent with a backend load, especially high numbers of oath sessions coming and going.

This fix was a joint effort with @israel-hdez .


israel-hdez (approved):
LGTM.

It removes the race condition.

mattmahoneyrh (approved):

PR #2001, opened 2020-11-12
Revise namespace selector (add select all)

jshaughn:
Fixes https://github.com/kiali/kiali/issues/3173

This PR revises the namespace selector in a few ways:

- adds select/deselect all
  - this is the primary RFE addition for 3173
  - note that select/deselect all applies only to the filtered set of namespaces.  Without a name filter it is a "true" all.  This allows the user to easily modify the state of a subset of namespaces, without affecting prior selections.
- batches changes
  - an important behavioral change.  prior to this PR each individual namespace select/deselect generated a state change.  this was distracting and inefficient as even changing from one namespace to another would generate 2 changes and cause pages to update while the user was still working with the dropdown.  now all changes are applied when the dropdown is closed (by clicking the button or outside the dropdown, or via the Escape key (I still wonder whether escape should act as cancel...)).
- changes button label
  - apply badging pattern to indicate multiple namespace selection

See the issue for screenshots/videos.

### DIFF
--- a/src/components/NamespaceDropdown.tsx
+++ b/src/components/NamespaceDropdown.tsx
@@ -3,7 +3,16 @@ import { connect } from 'react-redux';
 import { ThunkDispatch } from 'redux-thunk';
 import _ from 'lodash';
 import { style } from 'typestyle';
-import { Button, ButtonVariant, Dropdown, DropdownToggle, TextInput, Tooltip } from '@patternfly/react-core';
+import {
+  Button,
+  Dropdown,
+  DropdownToggle,
+  TextInput,
+  Tooltip,
+  // DropdownItem,
+  // DropdownPosition,
+  DropdownToggleCheckbox
+} from '@patternfly/react-core';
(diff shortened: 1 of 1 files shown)


lucasponce (approved):
Thanks.
Tested locally, it works ok.

jshaughn:
@lucasponce thanks for the review.  Although, going forward it would be better not to merge, just approve.  On Saturn we often wait to add QE until after dev approval.  This PR has not yet seen QE approval.  Also, when squash merging I often fix the commit message to avoid a bunch of appended messages for the various commits. Still, glad you liked the PR enough to merge :)

lucasponce:
Ok, but I was the only reviewer, I didn't see QE on the PR this morning and there was changes located into a single file.
Just to sync both teams, add them in the PR otherwise we'll repeat similar scenarios.

pbajjuri20:
LGTM!

I have validated from here https://quay.io/repository/kiali/kiali?tab=tags because it is already merged to Master.

PR #2003, opened 2020-11-16
OSSM-310 Fix lost settings on metrics page

jotak:
'combineLabelsSettings' should not mutate input labels

(to be backported)

### DIFF
--- a/src/components/Metrics/Helper.ts
+++ b/src/components/Metrics/Helper.ts
@@ -22,23 +22,22 @@ export const combineLabelsSettings = (newSettings: LabelsSettings, stateSettings
   // This is allowed because the labels filters state is managed only from this component,
   // so we can override them in props from state
   // LabelsSettings received from props contains the names of the filters with only a default on/off flag.
+  const result: LabelsSettings = new Map();
   newSettings.forEach((lblObj, promLabel) => {
+    const resultValues: SingleLabelValues = {};
     const stateObj = stateSettings.get(promLabel);
+    Object.entries(lblObj.values).forEach(e => {
+      resultValues[e[0]] = stateObj && stateObj.defaultValue === false ? false : e[1];
+    });
     if (stateObj) {
       lblObj.checked = stateObj.checked;
-      if (stateObj.defaultValue === false) {
-        // 1st pass: override default filters (this case only happens when filters are defined from URL)
-        Object.keys(lblObj.values).forEach(k => {
-          lblObj.values[k] = false;
-        });
-      }
-      // 2nd pass: retrieve previous filters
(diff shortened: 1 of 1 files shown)


src/components/Metrics/Helper.ts
```diff
@@ -22,23 +22,22 @@ export const combineLabelsSettings = (newSettings: LabelsSettings, stateSettings
   // This is allowed because the labels filters state is managed only from this component,
   // so we can override them in props from state
   // LabelsSettings received from props contains the names of the filters with only a default on/off flag.
+  const result: LabelsSettings = new Map();
   newSettings.forEach((lblObj, promLabel) => {
+    const resultValues: SingleLabelValues = {};
     const stateObj = stateSettings.get(promLabel);
+    Object.entries(lblObj.values).forEach(e => {
+      resultValues[e[0]] = stateObj && stateObj.defaultValue === false ? false : e[1];
```
xeviknal, line 30: It is a personal style, I guess, but I like to have the positive scenario first and then the negative. And also this scenario bugs me a bit to have a comparison between a boolean variable and a static boolean to generate a boolean expression. It is quite redundant. 
I see that `defaultValue` is a boolean field and also mandatory in the `LabelSettings` types. So I guess that is quite safe to use it as a boolean.

```
stateObj && stateObj.defaultValue ? e[1] : false
```

xeviknal (commented):
It makes sense. There is one comment I'd like to hear from @jotak

hhovsepy (approved):
Verified

src/components/Metrics/Helper.ts
jotak: What you write is not exactly equivalent: if `stateObj` is undefined (which can happen) then we must return `e[1]`, not false. So the correct expression would be: 

```
!stateObj || stateObj.defaultValue ? e[1] : false
```
But this is less easy to read I think.

Perhaps the behaviour can be changed and simplified, because it is actually a bit odd as it is. Something like this would make sense at first sight:

```
resultValues[e[0]] = stateObj ? stateObj.defaultValue : e[1];
```

Perhaps this would work, it's not strictly equivalent, I haven't tried it to play safe and keep exactly the same behaviour as before (and ok perhaps also because i am lazy here :) )

xeviknal (approved):
LGTM!

src/components/Metrics/Helper.ts
xeviknal: Oopsie! :facepalm:

jotak: > Oopsie! facepalm

😄
But I agree it's odd and catches attention

PR #2006, opened 2020-11-18
Add envoy proxy config into pod level (workload page)

xeviknal:
Requires: https://github.com/kiali/kiali/pull/3400

The idea is to show a modal where you can both see different envoy proxy config (both raw and digested data).
In this first step, it shows the raw information:

Screen recording (18) (1)

### DIFF
--- a/src/components/Envoy/EnvoyModal.tsx
+++ b/src/components/Envoy/EnvoyModal.tsx
@@ -9,7 +9,7 @@ import AceEditor from 'react-ace';
 import { aceOptions } from '../../types/IstioConfigDetails';
 import { style } from 'typestyle';
 
-const resources: string[] = ['All', 'Clusters', 'Listeners', 'Routes'];
+const resources: string[] = ['all', 'clusters', 'routes'];
 
 const displayFlex = style({
   display: 'flex'
@@ -53,10 +53,20 @@ class EnvoyDetail extends React.Component<EnvoyDetailProps, EnvoyDetailState> {
     this.aceEditorRef = React.createRef();
     this.state = {
       config: 'loading...',
-      resource: 'All',
+      resource: 'all',
       fetch: false,
       loading: true,
       pod: this.sortedPods()[0],
+    }
+  }
+
+  componentDidMount() {
+    this.fetchContent();
+  }
+
+  componentDidUpdate() {
+    if (this.state.fetch) {
+      this.fetchContent();
     }
   }
 
@@ -73,14 +83,23 @@ class EnvoyDetail extends React.Component<EnvoyDetailProps, EnvoyDetailState> {
   };
 
   setResource = (resource: string) => {
+    const resourceIdx: number = +resource;
     this.setState({
-      resource: resource,
+      fetch: true,
+      resource: resources[resourceIdx],
     });
+  };
+
+  fetchContent = () => {
+    if(this.state.resource === "all") {
+      this.fetchEnvoyProxy()
+    } else {
+      this.fetchEnvoyProxyResourceEntries()
+    }
   };
 
   fetchEnvoyProxy = () => {
     API.getPodEnvoyProxy(this.props.namespace, this.state.pod.name).then(resultEnvoyProxy => {
-      console.log(`Update proxy status: ${this.props.namespace}, ${this.state.pod.name}`);
       this.setState({
         config: resultEnvoyProxy.data,
         fetch: false,
@@ -91,16 +110,17 @@ class EnvoyDetail extends React.Component<EnvoyDetailProps, EnvoyDetailState> {
     });
   };
 
-  componentDidMount() {
-    this.fetchEnvoyProxy();
-  }
-
-  componentDidUpdate() {
-    if (this.state.fetch) {
-      console.log("update fetch!");
-     this.fetchEnvoyProxy();
-    }
-  }
+  fetchEnvoyProxyResourceEntries = () => {
+    API.getPodEnvoyProxyResourceEntries(this.props.namespace, this.state.pod.name, this.state.resource).then(resultEnvoyProxy => {
+      this.setState({
+        config: resultEnvoyProxy.data,
+        fetch: false,
+        loading: false,
+      });
+    }).catch(error => {
+      AlertUtils.addError(`Could not fetch envoy config ${this.state.resource} entries for ${this.state.pod.name}.`, error);
+    });
+  };
 
   render() {
     return (
--- a/src/components/Envoy/EnvoyModal.tsx
+++ b/src/components/Envoy/EnvoyModal.tsx
@@ -23,7 +23,7 @@ import ToolbarDropdown from '../ToolbarDropdown/ToolbarDropdown';
 import AceEditor from 'react-ace';
 import { aceOptions } from '../../types/IstioConfigDetails';
 import { style } from 'typestyle';
-import { SummaryWriterBuilder } from './writers/BaseWriter';
+import { SummaryTableBuilder } from './writers/BaseTable';
 
 // Enables the search box for the ACEeditor
 require('ace-builds/src-noconflict/ext-searchbox');
@@ -173,7 +173,7 @@ class EnvoyDetail extends React.Component<EnvoyDetailProps, EnvoyDetailState> {
   };
 
   render() {
-    const builder = SummaryWriterBuilder(this.state.resource, this.state.config);
+    const builder = SummaryTableBuilder(this.state.resource, this.state.config);
     const SummaryWriterComp = builder[0];
     const summaryWriter = builder[1];
     return (
--- a/src/components/Envoy/EnvoyModal.tsx
+++ b/src/components/Envoy/EnvoyModal.tsx
@@ -25,7 +25,7 @@ import { style } from 'typestyle';
 import { SummaryWriterBuilder } from './writers/BaseWriter';
 
 // Enables the search box for the ACEeditor
-require("ace-builds/src-noconflict/ext-searchbox");
+require('ace-builds/src-noconflict/ext-searchbox');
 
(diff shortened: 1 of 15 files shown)


lucasponce:
One first comment.
I'd add a 'Cancel' buttom in the Envoy modal, it's same pattern used in the Wizards:

image

And it will help to organize when it will add additional steps in the future.

lucasponce:
Minor, "Show Envoy details" -> "Show Envoy Details".
image

lucasponce:
I'dd add an easy sorting in the table headers for:
image
image

(Note: I understand the scrolling but it looks weird, I wonder if using some padding in the inner table would make a difference).

lucasponce:
From experience, "listeners" are also commonly used for debugging, so I've add them in the dropdown feature:
image

lucasponce:
If no sidecar or exceptions, it should show the menu option disabled (otherwise it will fail).
Example: grafana workload.

image

lucasponce:
Another thought, on errors like this:
image

I guess the modal should close and an error should be shown in the AlertManager.

xeviknal:
Comments approached. This is ready for another round of reviews. Thanks

src/types/IstioObjects.ts
```diff
@@ -139,6 +139,10 @@ export interface PodLogs {
   entries: LogEntry[];
 }
 
+export interface EnvoyProxy {
```
lucasponce, line 142: I'd add in the backend a model.EnvoyProxy { Configs: interfaces{} }, just to have this mapped.

I understand that any type would be necessary if we can't map the nested types here as discussed.

src/services/Api.ts
```diff
@@ -495,6 +495,14 @@ export const getPodLogs = (
   return newRequest<PodLogs>(HTTP_VERBS.GET, urls.podLogs(namespace, name), params, {});
 };
 
+export const getPodEnvoyProxy = (namespace: string, pod: string) => {
+  return newRequest<EnvoyProxy>(HTTP_VERBS.GET, urls.podEnvoyProxy(namespace, pod), {}, {});
+};
+
+export const getPodEnvoyProxyResourceEntries = (namespace: string, pod: string, resource: string) => {
+  return newRequest<EnvoyProxy>(HTTP_VERBS.GET, urls.podEnvoyProxyResourceEntries(namespace, pod, resource), {}, {});
```
lucasponce, line 503: Perhaps I missed to review this in the backend, so both "getPodEnvoyProxy" and "getPodEnvoyProxyResourceEntries" return a generic EnvoyProxy ? I thought the second may return a model.Clusters, model.Routes, etc. Perhaps we could take a look if it's working due dynamic parsing but we are missing some type here.

xeviknal: You are right. This is could be one of `BootstrapSummary | ClusterSummary | ListenerSummary | RouteSummary` interfaces.

src/types/IstioObjects.ts
```diff
@@ -139,6 +139,39 @@ export interface PodLogs {
   entries: LogEntry[];
 }
 
+export interface EnvoyProxyDump {
+  configs: any[];
+}
+
+export type EnvoyProxyResourceEntries = BootstrapSummary | ClusterSummary | ListenerSummary | RouteSummary;
```
lucasponce, line 146: I have a comment here.
These optional types look good on paper but in practice it's hard to check in runtime which type is coming from as there are not strong types.

So I guess if we can put a variable for any of these types and check if defined to see if the EnvoyProxyResourceEntries refers to a BootstrapSummary or a ClusterSummary.

Because this comes from an invocation

```
export const getPodEnvoyProxyResourceEntries = (namespace: string, pod: string, resource: string) => {
  return newRequest<EnvoyProxyResourceEntries>(
    HTTP_VERBS.GET,
    urls.podEnvoyProxyResourceEntries(namespace, pod, resource),
    {},
    {}
  );
};
```

So, perhaps this is missing in my backend PR review but instead to have this type pretty open:

```
type ResourceDump interface {
	Parse(dump *kubernetes.ConfigDump) error
}
```

I'd suggest to have an entry for the one of the "clusters", "routes", "boostrap", "listeners",

```
type ResourceDump interface {
         Clusters *models.Clusters,
         Routes *models.Routes,
         Bootstrap *models.Bootstrap,
         Listeners *models.Listeners,
	Parse(dump *kubernetes.ConfigDump) error
}
```
In that way, the caller would know the returning type just checking if that field is populated, not just relying on the parameter asked, I think.

src/components/Envoy/writers/BaseWriter.tsx
```diff
@@ -0,0 +1,78 @@
...
+        >
+          <TableHeader />
+          <TableBody />
+        </Table>
+      );
+    }
+  };
+}
+
+export const SummaryWriterBuilder = (resource: string, config: any) => {
```
lucasponce, line 60: I know naming is hard and opinionated :).
Coming from Java background, I read "Writer" very confusing as normally it's used for an abstraction of streams data.

If this is just a table and we are in the UI, a suggestion, "SummaryTable", "ListenersTable" or using "*List" like "SummaryList", "ListenerLists".

I think that's the way we are naming these patterns in other components and just to follow it to not introduce a new one.

src/components/Envoy/EnvoyModal.tsx
```diff
@@ -0,0 +1,234 @@
...
+};
+
+type EnvoyDetailState = {
+  config: any;
+  resource: string;
+  fetch: boolean;
+  pod: Pod;
+};
+
+export const EnvoyDetailsModal = ({ namespace, workload, isOpen, onClose }: EnvoyDetailModalProps) => (
```
lucasponce, line 61: Perhaps you may want to reuse here same pattern we are using for the Wizards.

There is a skeleton used called "WorkloadWizard" which basically performs same steps this is doing:
- Passing the workload, namespace, and which sub-wizard you want to open.
- Taking care of common controls: Create/Cancel buttons, title, etc.

So, all EnvoyDetail class would suit inside the WorkloadWizard and leave as it is, but then we can share and modify aspect of the modal for same category (Workload, Services, etc).

src/components/Envoy/EnvoyModal.tsx
```diff
@@ -0,0 +1,234 @@
...
+        );
+      });
+  };
+
+  render() {
+    const builder = SummaryWriterBuilder(this.state.resource, this.state.config);
+    const SummaryWriterComp = builder[0];
+    const summaryWriter = builder[1];
+    return (
+      <Stack gutter={GutterSize.sm}>
```
lucasponce, line 180: Also, using the WorkloadWizard would help to not need the Stack component.
I think I remove it (or delegate it) to any of the subwizards to render it.

Perhaps you can take a look to the Service Wizards as well.

src/components/Envoy/EnvoyModal.tsx
xeviknal: I don't know if I fully understand this one.

This is not exactly a wizard, this is a regular modal, isn't it?. Perhaps, the modal you use in the `WorkloadWizard` could be extracted to a common component that everyone could use. 

On the other hand, I don't know why we should use a `sub-wizard` prop in this case. There is a drowpdown with multiple values, but they are not coming from the outside component neither the initial selection.

Regarding the common controls, this scenario doesn't need any "Create"/"Update", it is just informative actions you can do in this modal. There isn't any impact in the backend after finishing the task in the modal.

One thing I could do to fit more the "pattern" is not to have two components one for the modal and another for the content of the modal.

src/components/Envoy/EnvoyModal.tsx
xeviknal: I think we need the stack anyway since every modal content needs a grid to distribute the content.
I see that in the WorkloadWizard there is nothing in the content just a plain text.
Perhaps I missed something. I could use a clarification here.

xeviknal:
@lucasponce I've approached some comments in the latest commits. I've also applied the changes in the backend to fit the data structure according the mentioned comments.

src/types/IstioObjects.ts
```diff
@@ -139,6 +139,45 @@ export interface PodLogs {
   entries: LogEntry[];
 }
 
+export interface EnvoyProxyDump {
```
lucasponce, line 142: +1

src/components/IstioWizards/WorkloadWizardDropdown.tsx
```diff
@@ -153,7 +174,17 @@ class WorkloadWizardDropdown extends React.Component<Props, State> {
...
     );
     // TODO WorkloadWizard component contains only 3scale actions but in the future we may need to bring it back
-    return <>{dropdown}</>;
+    return (
+      <>
+        {dropdown}
+        <EnvoyDetailsModal
+          namespace={this.props.namespace}
+          workload={this.props.workload}
+          show={this.state.showWizard}
```
lucasponce, line 183: Not a blocker, because today we have only "one" modal in the Workload Details.
But in previous Kiali version we used to have other modals for other extensions operations.
So, as the WorkloadWizardDropdown manages the state about when to show or not the modal, it's good to have a wrapper "WorkloadWizard" to manage this in a single place.

Otherwise when we brought back more modals in this section (others are on the way) we'll need to adjust this.

So basically the idea open the WorkloadWizard and inside the workload wizard modal says "show Envoy".

That's how it works in the Service Details.

Not a blocker, I suggest I'll send a followup PR with that, it wouldn't change anything relevant just a couple of refactoring about how to structure components.

lucasponce (approved):
LGTM

I have one comment that probably the WorkloadWizardDropdown shoudln't invoke directly the EnvoyModal but use the WorkloadWizard as parent "modal" wrapper to manage better the state when there are more "wizards" under the Workload details.

I'm ok to move forward and I'll send a followup PR with a couple of structure changes that won't affect into the feature.

PR #2007, opened 2020-11-20
Align select-all namespaces in List pages

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3437

Also fixes the sorting on table from main list pages.

Fetching all items for all namespaces is now unified using the "select-all" option in the Namespace Selector.

Note that Overview page is out of the scope of this work, that page have not namespace selection on purpose, but just filtering.

image

image

image

image

image

image

### DIFF
--- a/src/pages/IstioConfigList/IstioConfigListPage.tsx
+++ b/src/pages/IstioConfigList/IstioConfigListPage.tsx
@@ -30,8 +30,7 @@ import DefaultSecondaryMasthead from '../../components/DefaultSecondaryMasthead/
 
 interface IstioConfigListPageState extends FilterComponent.State<IstioConfigItem> {}
 interface IstioConfigListPageProps extends FilterComponent.Props<IstioConfigItem> {
-  // We keep this as Optional because it does not come from the params
-  activeNamespaces?: Namespace[];
+  activeNamespaces: Namespace[];
 }
 
 class IstioConfigListPageComponent extends FilterComponent.Component<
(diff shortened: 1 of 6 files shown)


src/pages/IstioConfigList/IstioConfigListPage.tsx
```diff
@@ -60,8 +59,11 @@ class IstioConfigListPageComponent extends FilterComponent.Component<
   componentDidUpdate(prevProps: IstioConfigListPageProps, _prevState: IstioConfigListPageState, _snapshot: any) {
     const prevCurrentSortField = FilterHelper.currentSortField(IstioConfigListFilters.sortFields);
     const prevIsSortAscending = FilterHelper.isCurrentSortAscending();
-    const [paramsSynced] = this.paramsAreSynced(prevProps);
```
xeviknal, line 63: Following the idea of not using the `paramsAreSynced` method, should this method be removed in this component? I see you remove the usage of the method but not the definition.

On the other hand, I see that the exact same logic is being needed in all List pages. Do you think it could be a good idea to move this logic into a shared space and use it from all the list pages?

lucasponce: > On the other hand, I see that the exact same logic is being needed in all List pages. Do you think it could be a good idea to move this logic into a shared space and use it from all the list pages?

No for the moment.

And I'll tell you why: this method sounds generic, but it's not

```
paramsAreSynced = (prevProps: IstioConfigListPageProps): [boolean, boolean]
```

It depends on the page props, on this case, some pages may have some props than others, on this case, the IstioConfigListPage has not the duration prop.

In the other hand, in the React lifecycle it recommmend to make pretty explicit the conditions used in the `componentDidUpdate`, so having a explicit condition with all the conditions involved:

```
    if (
      this.props.match.params.namespace !== prevProps.match.params.namespace ||
      this.props.match.params.app !== prevProps.match.params.app ||
      this.props.lastRefreshAt !== prevProps.lastRefreshAt ||
      currentTab !== this.state.currentTab
    ) {
}
```

helped me to understand bettter what the page was expecting, to move some of this logic to a third party function will work, but I can tell you it introduces some level of missing what is expected, as all expression are at the same level.

xeviknal: > In the other hand, in the React lifecycle it recommmend to make pretty explicit the conditions used in the componentDidUpdate, so having a explicit condition with all the conditions involved:

You are right. never thought from this point of view. It makes sense. 

I think it is still necessary to remove the definition of this `paramsAreSynced` method in this component. You removed it in the other components, but not in this one.

lucasponce: Yeah, your suggestion is right, I've removed from that class.

xeviknal (approved):
LGTM!

prachiyadav:
Screenshot from 2020-11-20 20-10-38
Screenshot from 2020-11-20 20-10-15

prachiyadav:
verified

PR #2008, opened 2020-11-20
Spans table: fix expand state being lost

jotak:
There was a bug when there are expanded rows in the spans table, and some actions are done that trigger table refresh, such as ordering table by columns or opening the links menu. The expanded rows were back to collapsed state.

Now they are kept expanded.


aljesusg (approved):
LGFM

hhovsepy (approved):
Verified

PR #2013, opened 2020-11-25
Fix guard null values in ContextMenu

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3465

### DIFF
--- a/src/components/CytoscapeGraph/CytoscapeContextMenu.tsx
+++ b/src/components/CytoscapeGraph/CytoscapeContextMenu.tsx
@@ -79,7 +79,7 @@ export class CytoscapeContextMenuWrapper extends React.PureComponent<Props> {
       event.preventDefault();
       if (event.target) {
         const currentContextMenu = this.getCurrentContextMenu();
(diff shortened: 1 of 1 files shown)


src/components/CytoscapeGraph/CytoscapeContextMenu.tsx
```diff
@@ -79,7 +79,7 @@ export class CytoscapeContextMenuWrapper extends React.PureComponent<Props> {
       event.preventDefault();
       if (event.target) {
         const currentContextMenu = this.getCurrentContextMenu();
-        if (currentContextMenu) {
+        if (!!currentContextMenu) {
```
israel-hdez, line 82: Nitpick: I think this double negation is not needed.

Anyway, this same condition is on other places. If this is really contributing to the fix, I think the other places should also be changed.

israel-hdez (approved):
I couldn't replicate. So, I'm only doing code review.
It's OK. Leaving one minor comment, for your consideration.

jotak (approved):
+1

PR #2015, opened 2020-11-26
Iter8 v1.0.0 ratiostatistics - add traffic control, min/max in assessment and warning msg in terminate 

jadeyliu939:
** Describe the change **
1. Add Traffic Control Frame in Experiment Detail page
2. Add Min/Max credible interval in Experiment Assessment tab when the values are available. also add Legend for the assessment chart
3. Check Traffic Split Percentage in Terminate, display warning message when the "Terminate" button is pressed if the total percentage is not 100

** Screenshot **
1) Traffic Control pane 
Screen Shot 2020-11-25 at 4 58 17 PM

2) Min/Max credoble interval in Assessment
Screen Shot 2020-11-25 at 4 58 37 PM

3) Warning message on Terminate Experiment 
Add a screenshot of the UI after your changes.
Screen Shot 2020-11-25 at 4 59 09 PM

### DIFF
--- a/src/pages/extensions/iter8/Iter8ExperimentDetails/AssessmentInfoDescription.tsx
+++ b/src/pages/extensions/iter8/Iter8ExperimentDetails/AssessmentInfoDescription.tsx
@@ -10,9 +10,15 @@ import {
   Title,
   Tooltip
 } from '@patternfly/react-core';
-import { ChartBullet, ChartThemeColor } from '@patternfly/react-charts';
+import { ChartBullet, ChartLegend, ChartThemeColor, ChartThemeVariant } from '@patternfly/react-charts';
 
-import { emptyExperimentItem, Iter8Experiment, Iter8Info, MetricProgressInfo } from '../../../../types/Iter8';
+import {
+  emptyExperimentItem,
+  Iter8Experiment,
+  Iter8Info,
+  MetricProgressInfo,
+  Statistics
+} from '../../../../types/Iter8';
 import {
   cellWidth,
   expandable,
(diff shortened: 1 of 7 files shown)


lucasponce:
The "traffic control" tab doesn't follow 100% the pattern for subtasks:

image

Look at the service detail as example:

image

lucasponce:
I'm not familiar with this termination flow:
image

But I'd like to reuse the TrafficShift component, I'd see weird that use specify some traffic split using a component in one scenarios and a different in others.

Note, perhaps it's not for this PR, but something we want to unify.

lucasponce:
About the Matching Rules I have some doubts, I mean, we have a similar functionality already implemented, with a specific style:
image

It makes sense that if the component can't be reused, but at least it can be seen in a similar way.

I know this sounds too vague, so, I'm going to study the PR closely to suggest something more actionable.

jadeyliu939:
I c. So instead of a full listing of match rules. you suggest we change the following 
Screen Shot 2020-11-30 at 4 23 41 PM
to something like (probably without the `Route to`
Screen Shot 2020-11-30 at 4 20 12 PM

Also regarding `The "traffic control" tab doesn't follow 100% the pattern for subtasks:`. Maybe it should not in the `subtasks` tab, and should be in the top tab (just like Assessment tab)?

lucasponce:
For the "traffic control" subtab I mean:

image

Vs

image

It seems there is something broken in the component used.

For the "Match Rules", yes, you are right, something in that direction.

Also I may have other minor UI enhancements that I want to collect, now that these pages are in a more stable point, it's good if we try to align the cards with same structure of the other details.

Probably not for this PR but I'll work to collect all these suggestions in isssues and/or PR, nothing to change in terms of features but I'd see if we can reuse the same "pattern" for presenting similar information used in other places.

jadeyliu939:
As for Terminate with Traffic split, I will reuse the Traffic Shifting component

jadeyliu939:
Close this PR, a new PR has been created - https://github.com/kiali/kiali-ui/pull/2020

PR #2016, opened 2020-11-27
Use heatmap in trace details

jotak:
Fixes https://github.com/kiali/kiali/issues/3279

- Reduxify stats metrics for comparison with traces
Same data will be use in several places, so redux is a convenient
solution here

- Refactor heatmap to use css grid layout instead of flex (flex doesn't
  work so well for two-axis grids). Heatmap style & structure is well
simplified

- Rename SpanItemData -> RichSpanData and make it computed directly in
  transformTraceData (sooner than before)

- Create one heatmap from Metrics Stats as an aggregation of all
  span-based stats...

- ...and another heatmap to replace the textual information we had,
  comparing trace stats with other traces displayed on screen


jotak:
Initial layout, with two heatmaps:
Capture d’écran de 2020-11-25 16-10-14

Alternative layout, showing the two heatmaps side by side, and reorganizing the grid a little bit:
Capture d’écran de 2020-11-27 15-44-03

lucasponce:
One earlier comment: is there any case where similar traces will be different than the selected one ?

image

I mean, the list of the right can be seen clear with an array of boxes with the differences, in that case it may save space and have a more uniform visualization with the heatmap.

And tooltips can be used to show (if needed) the service url, and they can be clickable as well.

jotak:
@lucasponce although most of the time we will see the same name here, it is not entirely guaranteed. That name is actually the operation name of the root span, and the similarity algo allows for slight differences in that regard, or different spans ordering. So, two traces with all spans identical except the first, for instance, would appear as similar and have different names.

lucasponce:
> That name is actually the operation name of the root span, and the similarity algo allows for slight differences in that regard, or different spans ordering. So, two traces with all spans identical except the first, for instance, would appear as similar and have different names.

I understand, but as user perhaps the interesting here is to see the differences in time, and I guess that can be represented as a list of squares, more close than a heatmap rather than a list.

My suggestion would be that tooltip can be used for showing the name, and probably that may help to navigate across them.

jotak:
@lucasponce is it something like this that you're thinking about?
(quick mockup)

Capture d’écran de 2020-11-27 15-44-03

Here I put the 2x2 heatmap merged at the bottom of the "similar traces" Nx2 heatmap, since it can take the same columns.

lucasponce:
> is it something like this that you're thinking about?

Yes thanks, much better that a list of the full operation name.

In my mind I was imaging an horizontal layout instead for better alignment with the heatmap, but yes in esence I think this info invites to use a representation in this way.

jotak:
If we want an horizontal layout we need to manage column header labels differently I guess, because the text is longer than the cell width (that's why I kept it vertical). Could rotate 45° the label maybe, i don't know ... Or keep it vertical.

jotak:
So, next iteration, here's the new proposal:

Capture d’écran de 2020-11-30 12-37-38

jotak:
I've added a tooltip / info icon for the similarity algo

Screenshot with the irish flag :)

Capture d’écran de 2020-11-30 13-42-23

(now turning as ready for review)

hhovsepy:
Verified, the similar traces links and the duration calculations.
Screenshot from 2020-11-30 16-08-28
Screenshot from 2020-11-30 16-08-41

aljesusg (approved):
LGFM

src/components/JaegerIntegration/JaegerHelper.ts
```diff
@@ -281,3 +263,7 @@ export const extractSpanInfo = (span: Span) => {
       : extractOpenTracingBaseInfo(span);
   return { type: type, info: info };
 };
+
+export const sameSpans = (a: Span[], b: Span[]): boolean => {
+  return a.map(s => s.spanID).join() === b.map(s => s.spanID).join();
```
lucasponce, line 268: Mmm is this working as intended, is checking deep equality ?

src/components/JaegerIntegration/JaegerResults/SpanTableItem.tsx
```diff
@@ -33,47 +25,12 @@ const linkStyle = style({
...
-};
-
-type RowProps = SpanItemData & {
+type RowProps = RichSpanData & {
   toggledLinks?: string;
   setToggledLinks: (key: string) => void;
   externalURL?: string;
   onClickFetchStats: () => void;
-  metricsStats: { [key: string]: MetricsStats };
+  metricsStats: Map<string, MetricsStats>;
```
lucasponce, line 33: I think this Map doesn't come from an api call, it's built internally

jotak: Yes that's correct, is that a problem?

lucasponce:
One UI consideration, we are starting to "fix" or at least to mitigate scenarios on narrow screens (something between 1024 -> 1280 for width).

I wonder if it could have some change to at least not overflow out of the subtask card
image

lucasponce (approved):
LGTM

I've tested locally and it works as expected.
Code may need some additional testing as there are helpers to fetch, reduce and build the heatmap data that is harsh to follow and decide if it can be optimized or not.

Also a comment about narrow screens that perhaps can be solved just reducing the number of similar traces from 8 -> 6 (if there is no an easier alternative).

But not a blocker, ok to merge if you like @jotak.

Most of my comments can fit on this or a following PR.

jotak:
@lucasponce good points I will at least add a couple of unit tests in this PR
For the responsiveness issue, I need to investigate more on the css grid layout in that regard.

jotak:
@lucasponce I think your comments are addressed:

- The heatmaps width can now decrease for smaller resolutions:
Capture d’écran de 2020-12-02 13-08-56

Note that if you reduce the width to an extreme, the design is still screwed up, but it's not just the heatmaps, it's the whole panel content. I'm not fixing it right now, it's probably a broader task to do.

- Added some tests

jshaughn:
I think very narrow widths are out of scope for Kiali.  Perhaps we should also document our min resolution requirements.

lucasponce:
Thanks @jotak I'll test, yes as @jshaughn says, I guess we have a non written agreement to make it work in width (1024 - 1280) as minimal.

I agree that below that, IMO the only responsability is that components are not broken.

At least that was the criteria for the last changes in that direction in the main headers and other commons elements.

lucasponce:
Width 1024:
image

It looks perfect.

src/utils/__tests__/TraceStats.test.ts
```diff
@@ -87,3 +111,86 @@ describe('TraceStats.isSimilarTrace', () => {
     expect(isSimilarTrace(traceDifferentOperations, traceBase)).toBe(false);
   });
 });
+
+describe('TraceStats.reduceMetricsStats', () => {
```
lucasponce, line 115: thanks

lucasponce (approved):

PR #2017, opened 2020-11-27
Support for Mirror configuration in Kiali Wizards

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3476

This PR add supports for mirrored workloads in the Traffic Shifting and Request Routing wizards.

The user may select one workload that will be used as "mirror" (in the Istio VS terminology)

image

And the Traffic Weight for a mirror workload represents the % of traffic mirrored from the normal routing rule.

The rest of the workload distributes the regular 100% of the traffic as a normal rule but with one workload left.

For QE:

- Test the feature from Traffic Shifting / Request Routing scenarios.
- Validate no regressions and also that VS can be edited after creation and reflected in the Wizard.

There may be some comments from @lwrigh about style of the icon and color used for mirrored but feature is completed to test.

Note that mirror traffic is a concept available on HTTP scenarios, so TCP Traffic Shifting won't show this capability.

### DIFF
--- a/src/components/IstioWizards/RequestRouting.tsx
+++ b/src/components/IstioWizards/RequestRouting.tsx
@@ -3,7 +3,7 @@ import { WorkloadOverview } from '../../types/ServiceInfo';
 import Rules, { MOVE_TYPE, Rule } from './RequestRouting/Rules';
 import RuleBuilder from './RequestRouting/RuleBuilder';
 import { ANYTHING, EXACT, HEADERS, PRESENCE, REGEX } from './RequestRouting/MatchBuilder';
-import { WorkloadWeight } from './TrafficShifting';
+import { MSG_WEIGHTS_NOT_VALID, WorkloadWeight } from './TrafficShifting';
 import { getDefaultWeights } from './WizardActions';
 import { FaultInjectionRoute } from './FaultInjection';
 import { TimeoutRetryRoute } from './RequestTimeouts';
@@ -143,7 +143,8 @@ class RequestRouting extends React.Component<Props, State> {
             name: ww.name,
             weight: ww.weight,
             locked: ww.locked,
-            maxWeight: ww.maxWeight
+            maxWeight: ww.maxWeight,
+            mirrored: ww.mirrored
           })
         );
         const newRule: Rule = {
@@ -244,9 +245,10 @@ class RequestRouting extends React.Component<Props, State> {
     });
   };
 
-  onSelectWeights = (_valid: boolean, workloads: WorkloadWeight[]) => {
+  onSelectWeights = (valid: boolean, workloads: WorkloadWeight[]) => {
     this.setState({
-      workloadWeights: workloads
+      workloadWeights: workloads,
+      validationMsg: !valid ? MSG_WEIGHTS_NOT_VALID : ''
     });
   };
 
--- a/src/components/IstioWizards/RequestRouting/RuleBuilder.tsx
+++ b/src/components/IstioWizards/RequestRouting/RuleBuilder.tsx
@@ -90,8 +90,10 @@ class RuleBuilder extends React.Component<Props, State> {
           </Tab>
           <Tab eventKey={1} title={'Route To'}>
             <TrafficShifting
+              showValid={false}
               workloads={this.props.workloads}
               initWeights={this.props.weights}
+              showMirror={true}
               onChange={this.props.onSelectWeights}
             />
           </Tab>
--- a/src/components/IstioWizards/RequestRouting/Rules.tsx
+++ b/src/components/IstioWizards/RequestRouting/Rules.tsx
@@ -132,16 +132,34 @@ class Rules extends React.Component<Props> {
                 </>,
                 <>
                   <div key={'ww_' + order}>
-                    {rule.workloadWeights.map((wk, i) => {
-                      return (
-                        <div key={'wk_' + order + '_' + wk.name + '_' + i}>
-                          <Tooltip position={TooltipPosition.top} content={<>Workload</>}>
-                            <Badge className={'virtualitem_badge_definition'}>WS</Badge>
-                          </Tooltip>
-                          {wk.name} ({wk.weight} %)
-                        </div>
-                      );
-                    })}
+                    {rule.workloadWeights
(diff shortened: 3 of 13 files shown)


abonas:
Hi Lucas, can you please elaborate a bit on the mirroring feature supposed to work? 
- If user clicks on the mirror button/icon, then all the traffic from the rest of the workloads should be copied to that selected workload?
- If there was a different traffic distribution that includes that same workload (say, 3 workloads and each gets ~33%), what happens then? For example like in the screenshot above, it says 65% on the mirrored workload, I'm not sure how it is supposed to behave and what traffic does that number reflect.
- I would consider choosing a different icon for mirroring, something that looks closer to "copy" icon, was the icon selection discussed with UX?
thanks

lucasponce:
@abonas note that this PR has a "do not merge" label which means it's still WIP.

> If user clicks on the mirror button/icon, then all the traffic from the rest of the workloads should be copied to that selected workload?

Yes

> If there was a different traffic distribution that includes that same workload (say, 3 workloads and each gets ~33%), what happens then? For example like in the screenshot above, it says 65% on the mirrored workload, I'm not sure how it is supposed to behave and what traffic does that number reflect.

That's not supported. At least from the Wizard, the idea is to choose one workload to be "the mirror". If you have 3 workloads and select one as a mirror, then the regular traffic routes over those two workloads with one to the mirror.

> I would consider choosing a different icon for mirroring, something that looks closer to "copy" icon, was the icon selection discussed with UX?

We'll ask for feedback from UX. Any specific suggestion about the icon you like is welcome.

abonas:
> @abonas note that this PR has a "do not merge" label which means it's still WIP.
absolutely clear

> 
> > If user clicks on the mirror button/icon, then all the traffic from the rest of the workloads should be copied to that selected workload?
> 
> Yes
> 
> > If there was a different traffic distribution that includes that same workload (say, 3 workloads and each gets ~33%), what happens then? For example like in the screenshot above, it says 65% on the mirrored workload, I'm not sure how it is supposed to behave and what traffic does that number reflect.
> 
> That's not supported. At least from the Wizard, the idea is to choose one workload to be "the mirror". If you have 3 workloads and select one as a mirror, then the regular traffic routes over those two workloads with one to the mirror.

yes, that's what I thought, thanks.
so the workload will have that percentage part disabled if a user clicks on mirror, correct?
> 
> > I would consider choosing a different icon for mirroring, something that looks closer to "copy" icon, was the icon selection discussed with UX?
> 
> We'll ask for feedback from UX. Any specific suggestion about the icon you like is welcome.

yes, as I mentioned above, perhaps something that looks like "copy" will reflect the action of mirroring, because it creates a copy of the traffic. But UX will know best.

lucasponce:
> so the workload will have that percentage part disabled if a user clicks on mirror, correct?

No exactly, the percentage will indicate the % of the traffic mirrored for that case.

abonas:
> > so the workload will have that percentage part disabled if a user clicks on mirror, correct?
> 
> No exactly, the percentage will indicate the % of the traffic mirrored for that case.

sounds good.

xeviknal (approved):
LGTM!

hhovsepy:
When making one workload mirrored, can the rest of traffic be automatically evenly distributed between other workloads? because it makes confusing and any touch +/- buttons traffic causes jumping distribution.
Also the error notification about 100% sum will not be displayed.
Screenshot from 2020-12-01 16-27-12

lucasponce:
> When making one workload mirrored, can the rest of traffic be automatically evenly distributed between other workloads? because it makes confusing and any touch +/- buttons traffic causes jumping distribution.
Also the error notification about 100% sum will not be displayed.

+1 

I have doubts if to automatically change them or keep it for a first step, but yes, it makes sense, I'll add the change.

hhovsepy:
@lucasponce 2 minor thins to mention:
1. The format of VS generated via Wizard in case of mirror is a bit different that what we see in istio documentation. but still is correct.
Generated one:
`
spec:
  hosts:
    - ratings.bookinfo.svc.cluster.local
  http:
    - mirror:
        host: ratings.bookinfo.svc.cluster.local
        subset: v2
      mirrorPercentage:
        value: 100
      route:
        - destination:
            host: ratings.bookinfo.svc.cluster.local
            subset: v1
          weight: 50
        - destination:
            host: ratings.bookinfo.svc.cluster.local
            subset: v2-mysql
          weight: 50
`
 Documented one:
`
kind: VirtualService
metadata:
  name: httpbin
spec:
  hosts:
    - httpbin
  http:
  - route:
    - destination:
        host: httpbin
        subset: v1
      weight: 100
    mirror:
      host: httpbin
      subset: v2
    mirror_percent: 100
`
2. There should be follow up PR to replicate traffic mirroring functionality in graph right?

lucasponce:
I've added the "do not merge" label to work in some suggestion from the UX meeting.
Basically, when mirroring, to place the workload at the end of the list with some separation to make more clear that it's a different type.

lucasponce:
@lwrigh I've addressed the comments from past UX to separate the mirrored workload from the route workloads

image
image

Also changed the icon as suggested.

@hhovsepy now traffic is redistributed on mirror/unmirror operations.

About the doc comment in [1], yes, it's using a deprecated but correct way [2]. Wizard is using the new one.

[1] https://istio.io/latest/docs/tasks/traffic-management/mirroring/#mirroring-traffic-to-v2
[2] https://github.com/istio/api/blob/master/networking/v1alpha3/virtual_service.proto#L628

hhovsepy:
Can't there be mirroring in case of 2 workloads?
Screenshot from 2020-12-02 09-00-41

lucasponce:
Yes.
Mirroring over 2 workloads should be supported.

It's not included for TCP TrafficShifting but it seems a bug from my latest
changes.

Thanks

lucasponce:
@hhovsepy you caught a good one, I tested from 1 workload to 3 workloads but the 2 case has some specific logic for the locks that it as left for the mirror case.

image

So, now it should work as expected.

Very good testing.
Thanks again

hhovsepy (approved):
Verified , thanks @lucasponce those 2 issues are fixed.

PR #2018, opened 2020-11-30
Enable the use of node.js 14.x for building/testing kiali-ui.

mtho11:
** Describe the change **
Allow building/testing the UI with the latest LTS version of node.js (14.x)

** Issue reference **
fixes: https://github.com/kiali/kiali/issues/3484

** Backwards compatible? **
Yes


FilipB (approved):
UI tests are passing.
Tested following on Fedora 33:
```
yarn install
yarn build
```
There are a lot of warnings during the build probably not related to this change. Does it make sense to create new issue for those?
Also is it possible to backport this to v1.24?

jshaughn:
>  is it possible to backport this to v1.24?

@FilipB It should be possible but is it necessary?  Unless it's solving a bug/cve we typically don't backport.

FilipB:
> > is it possible to backport this to v1.24?
> 
> @FilipB It should be possible but is it necessary? Unless it's solving a bug/cve we typically don't backport.
It's not necessary, it's just that we will have to maintain two jenkins workers one with nodejs 12 to build v1.24 and one with nodejs 14 for master. Or keep one and somehow switch between nodejs versions. But that's fine.

PR #2019, opened 2020-12-01
Do not try to fetch traces for a service entry

jotak:
Fixes https://github.com/kiali/kiali/issues/3461


jotak:
I'm just removing here the "traces" side-panel for service entries. The reason is that, it's not only that we can't find any trace, it's worse than that: we are trying to fetch the k8s service object to get the associated "app" prior to running jaeger query, but with service entries it just doesn't make sense. We can eliminate that whole use case.

hhovsepy (approved):
Verified:
Screenshot from 2020-12-01 11-52-37

jshaughn (approved):
Totally makes sense.

aljesusg (approved):
LGFM

PR #2021, opened 2020-12-03
Fix from time parameters on tracing tabs

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3503

There was a bug in the TracesComponent using a wrong duration parameter that may fall into the default (10m).
This bug was introduced in latest changes when refactoring the TimeControl component.

I think it's fixed now @israel-hdez, would you mind to test ?
I did some quick tests, but let me know if there is anything else to fix and I'll continue tomorrow my time.


israel-hdez (approved):
Works correctly.

PR #2023, opened 2020-12-07
Fix the rebase issues and update Mesh page to conform to Lucas' lates…

jshaughn:


israel-hdez (approved):
I think refresh button is not needed, as data in this page will rarely change. Anyway, let's merge it.

Screenshot, for reference:

image

lucasponce:
I'd suggest to consolidate this work in master and add a config flag similar as we have for extensions.
Otherwise this split of work in different branches would cause more rebasing and potential conflicts.

If changes are tied to this new page it seems doable.

PR #2024, opened 2020-12-08
Checkpoint WIP for https://github.com/kiali/kiali/issues/3439

jshaughn:
- move from 'Group' to 'Box' terminology to consolidate terminology and avoid overused 'group'
- initial pass at handling different box types
- initial support for cluster field
- remove unnecessary selector clause in CytoscapeTrace as box is now its own node type
- add support for boxing options
- remove Show Node Labels option to simplify code and reduce size of display menu
  (in anticipation on new options being added)
- run group-compound-layout for each level of boxing
- add some colors to distinguish mesh boxing


src/components/CytoscapeGraph/Layout/GroupCompoundLayout.ts
```diff
@@ -34,6 +34,9 @@
     b. For every child set the relative position to its parent
  */
 
+import { CyNode } from '../CytoscapeGraphUtils';
```
josejulio, line 37: I haven't looked in depth, but wondering why this file had to change? I only remember using the concept of compound and not the group/app?

jshaughn: Uh Oh, @josejulio is watching what I'm doing.  I would actually love for you to come back for a while and help with this stuff :)   The short story is that to help visualize multi-cluster stuff we are looking to have up to 3-levels of boxing.  So no longer can we assume that boxes are just for apps.  But now there could be up to 3 levels, where app boxes could have a namespace (or cluster) box as a parent, and a namespace box could have cluster box as a parent.  So "group-compound-layout" was basically assuming just app boxing, now it is starting to handle the nested boxing.

josejulio: > Uh Oh, @josejulio is watching what I'm doing

:eyes: 

> 3-levels of boxing

Ah, it makes sense, I never covered that case. I'll take a look at the file and leave comments. But I already forgot most of what is done here :)

mtho11:
@jshaughn can we get some screenshots?

src/components/CytoscapeGraph/Layout/GroupCompoundLayout.ts
jshaughn: I'm not in love with the results, although not terrible, in general I don't understand why the layouts often want to "stack" the boxes as opposed to having a more horizontal result.  Actually, I'm not sure that's really the fault of our custom layout, but rather the "real" layouts.

josejulio: Could you show some screenshots of the results that you don't love?

jshaughn:
> @jshaughn can we get some screenshots?

It's just a WIP and not a PR for master, just the multi-cluster branch.  So not really looking for any feedback yet.  But here is one just to give you an idea:

image

mtho11:
Don't know if this Cytoscape plugin could be useful: https://apps.cytoscape.org/apps/boundarylayout

src/components/CytoscapeGraph/Layout/GroupCompoundLayout.ts
josejulio: > I'm not in love with the results, although not terrible, in general I don't understand why the layouts often want to "stack" the boxes as opposed to having a more horizontal result. Actually, I'm not sure that's really the fault of our custom layout, but rather the "real" layouts.

I recall I did a stack-layout (or maybe because I used dagre?) somewhere in that same file so all the contents of the compounds are dagre-layouts. Maybe you can try to use the compound (if not already) in the inner boxes?

src/components/CytoscapeGraph/Layout/GroupCompoundLayout.ts
```diff
@@ -107,117 +110,119 @@ export default class GroupCompoundLayout {
...
 
-      // see https://github.com/cytoscape/cytoscape.js/issues/2402
-      const boundingBox = parent.boundingBox();
+      // (1.a) Prepare parents by assigning a size and running the compound layout
+      parents.each(parent => {
+        const children = parent.children();
+        const targetElements = children.add(children.edgesTo(children));
+
+        // We expect a discrete layout here
+        const compoundLayout = targetElements.layout(compoundLayoutOptions);
```
josejulio, line 124: This is `dagre` by default IIRC. So we are running that as the layout for any child.

src/components/CytoscapeGraph/Layout/GroupCompoundLayout.ts
```diff
@@ -107,117 +110,119 @@ export default class GroupCompoundLayout {
...
+      const children = parents.children();
 
-      // see https://github.com/cytoscape/cytoscape.js/issues/2402
-      const boundingBox = parent.boundingBox();
+      // (1.a) Prepare parents by assigning a size and running the compound layout
+      parents.each(parent => {
+        const children = parent.children();
+        const targetElements = children.add(children.edgesTo(children));
+
+        // We expect a discrete layout here
```
josejulio, line 123: The reason we expect a discrete layout is because we expect them to be finished after the "run" call. 
This could be extended to allow any type of non discrete layout that don't take too long to finish.

src/components/CytoscapeGraph/Layout/GroupCompoundLayout.ts
```diff
@@ -107,117 +110,119 @@ export default class GroupCompoundLayout {
...
+
+      // Ensure we only touch the requested elements and not the whole graph.
+      const layoutElements = this.cy.collection().add(this.elements).subtract(removedElements).add(syntheticEdges);
+
+      // Before running the layout, reset the elements positions.
+      // This is not absolutely necessary, but without this we have seen some problems with
+      //  `cola` + firefox + a particular mesh
+      layoutElements.position({ x: 0, y: 0 });
+
+      const layout = this.cy.layout({
```
josejulio, line 179: This is done for the top-level nodes IIRC, so we are only using the main layout with the top elements.

src/components/CytoscapeGraph/Layout/GroupCompoundLayout.ts
josejulio: Yes, it is `dagre` https://github.com/kiali/kiali-ui/blob/master/src/components/CytoscapeGraph/CytoscapeGraphUtils.ts#L111

josejulio (commented):
The change to the compound layout seems fine. The strategy to run it on the App -> Namespace -> Cluster seems fine, and I think it works well because the layout does not reset the positions but uses them. 
Remember a compound doesn't really have any position or size it self, but is driven by the content.

src/components/CytoscapeGraph/Layout/GroupCompoundLayout.ts
jshaughn: Right, I think using Dagre inside the boxes is a good idea.  The "real" layout would apply to the final layout of the compound nodes, and nodes with no parent, right?

src/components/CytoscapeGraph/Layout/GroupCompoundLayout.ts
jshaughn: I think it's fine as is, and we would want to use dagre I think.

src/components/CytoscapeGraph/Layout/GroupCompoundLayout.ts
josejulio: Right, to every "top" node.

israel-hdez:
When trying this, I'm seeing the following graph in _ClusterA_:

image

and _ClusterB_ shows this:

image


Quite fine IMO.

jshaughn:
OK, merging this checkpoint and we can address other issues going forward, at least this way we have something to play with.  @israel-hdez Feel free to further review the code changes at your leisure.

src/pages/Graph/SummaryPanel.tsx
```diff
@@ -109,7 +109,8 @@ class SummaryPanel extends React.Component<MainSummaryPanelPropType, SummaryPane
               rateInterval={this.props.rateInterval}
             />
           ) : null}
-          {this.props.data.summaryType === 'group' ? (
+          {this.props.data.summaryType === 'box' ? (
+            // TODO, how to handle non-app box?
```
israel-hdez, line 113: Hmm... oh!
You should create a GitHub issue for this... or at least, list it as a bullet in the Epic just to not forget it.

israel-hdez (commented):
Post merge review:

I don't understand the changes to GroupCompountLayout :)
Besides that, LGTM.

I dropped one comment to address later.

jshaughn:
> I don't understand the changes to GroupCompountLayout :)

So you've probably noticed that nodes in our app boxes are consistently layed out using Dagre (our default layout for the main graph).  This tends to look good and makes the graph look much more consistent.  Outside of the app boxes, for the "top-level nodes", we apply the selected layout set in the cytoscape toolbar, and the app box is basically treated as a new, top-level node within that layout.   Now that we have introduced nested boxing, we may have  [ ClusterBox [ NamespaceBox [AppBox] ] ].  The changes in GroupCompoundLayout are so that we basically apply the same Dagre layout at each level of nesting, again to provide a more consistent layout.  And again, the outmost box will be considered a top-level node when we apply the final, user-selected layout.

src/pages/Graph/SummaryPanel.tsx
jshaughn: Added subtask to epic https://github.com/kiali/kiali/issues/1833

PR #2026, opened 2020-12-08
Add tooltip info to the "Requests Per Second" edge label option

jshaughn:
Add tooltip info to the "Requests Per Second" edge label option to better explain the resulting edge labeling.

Also, take care of some other improvements in the Display menu:
- rename "Requests Per Second" to "Request Rate"
- rename "Requests Percentage" to "Request Distribution"
- use consistent plurals in the edge label options
- add tooltip info for all non-obvious options

Fixes https://github.com/kiali/kiali/issues/3477

image

### DIFF
--- a/src/components/CytoscapeGraph/graphs/GraphStyles.ts
+++ b/src/components/CytoscapeGraph/graphs/GraphStyles.ts
@@ -310,7 +310,7 @@ export class GraphStyles {
       const edgeData = decoratedEdgeData(ele);
 
       switch (edgeLabelMode) {
-        case EdgeLabelMode.REQUESTS_PER_SECOND: {
+        case EdgeLabelMode.REQUEST_RATE: {
           let rate = 0;
           let pErr = 0;
           if (edgeData.http > 0) {
@@ -348,7 +348,7 @@ export class GraphStyles {
           }
           break;
         }
-        case EdgeLabelMode.REQUESTS_PERCENTAGE: {
+        case EdgeLabelMode.REQUEST_DISTRIBUTION: {
           let pReq;
           if (edgeData.httpPercentReq > 0) {
             pReq = edgeData.httpPercentReq;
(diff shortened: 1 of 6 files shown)


pbajjuri20:
"Tooltip" is not in the same row with  "Request Distribution"

RD bug
 "Request Distribution"

pbajjuri20:
"Firefox" looks good, I can see this issue in the Chrome

pbajjuri20:
I have tested with new changes but still, I can see the bug. I'm using Chrome Version 86.0.4240.75, Screen Resolution - 1920 * 1080.

pbajjuri20:
I have tested on Chrome version - 74 and 87, there is no bug. On my laptop the chrome version - 86, it is showing a bug.

LGTM on different Chrome version except - 86 version.


Screenshot of Chrome -87 version, It looks good.

tool1

pbajjuri20 (approved):

pbajjuri20:
I have tested on a different laptop with chrome version - Version 86.0.4240.198 (Official Build) - I don't see this bug, It seems that it is not a chrome version problem.  there is an issue with my laptop chrome.

lucasponce:
I'm catching up with the review.
Hope to give a try this today, sorry for my delay, I missed some notifications.

lucasponce (approved):
"Request Per Second" - > "Request Rate" sounds ok.
"Request Percentage" -> "Request Distribution" sounds somehow ok, but the former was clear I think.

In any case, it's good to review and change. I'm ok with the changes.

jshaughn:
@lucasponce Thanks for the review.   The problem with "Request Percentage" is that "Percentage" is really just a unit.  It leaves the question, "Percentage of what?".   I'm going to leave the change, hopefully it's OK.  Of course either way there is now an info tooltip  to make it clear.

PR #2031, opened 2020-12-09
Don't fetch the config dump before opening the modal

xeviknal:
fixes https://github.com/kiali/kiali/issues/3515

Fixing error message when visiting a workload page that doesn't have a sidecar.
There is no call performed under the hood before opening the modal.

proxy-details-grafana-err

### DIFF
--- a/src/components/Envoy/EnvoyModal.tsx
+++ b/src/components/Envoy/EnvoyModal.tsx
@@ -75,9 +75,12 @@ class EnvoyDetailsModal extends React.Component<EnvoyDetailProps, EnvoyDetailSta
     };
   }
 
-  componentDidUpdate(prevProps: EnvoyDetailProps) {
-    // Fetch data when the modal has just opened or some action request a fetch.
-    if ( (!prevProps.show && this.props.show) || this.state.fetch) {
+  componentDidMount() {
+    this.fetchContent();
+  }
+
+  componentDidUpdate() {
+    if (this.state.fetch) {
       this.fetchContent();
     }
   }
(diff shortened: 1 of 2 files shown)


hhovsepy (approved):
Verified:
Screenshot from 2020-12-09 18-18-51

jotak:
In the initial issue I noticed that the error was appearing like in an infinite loop, and it seems this problem isn't fixed.
If you find another way to trigger an error while loading the config_dump (example: kill the pods while the popup is opened, and change some dropdown value) : an error will be displayed (which is normal), but it's going to be triggered in loop:

Capture d’écran de 2020-12-10 12-09-56

src/components/Envoy/EnvoyModal.tsx
```diff
@@ -70,17 +70,14 @@ class EnvoyDetailsModal extends React.Component<EnvoyDetailProps, EnvoyDetailSta
...
 
-  componentDidMount() {
-    this.fetchContent();
-  }
-
-  componentDidUpdate() {
-    if (this.state.fetch) {
+  componentDidUpdate(prevProps: EnvoyDetailProps) {
+    // Fetch data when the modal has just opened or some action request a fetch.
+    if ( (!prevProps.show && this.props.show) || this.state.fetch) {
```
jotak, line 80: just a suggestion, that perhaps would make the things simpler: if you remove this `show` props, and in the parent component `WorkloadWizardDropdown` you only render this component when it has to be shown, then i guess it makes everything easier and you can go back having the backend call in `componentDidMount`, no?

xeviknal:
@jotak yes, you are right. It is caught in an infinite loop due to the close of the component (it changes the prop, so triggers a re-rendering). Applied your suggestion. Make it way easier to manage.

hhovsepy (approved):
Re-verified.
Screenshot from 2020-12-10 17-15-31

jotak (approved):
LGTM

lucasponce (approved):
As commented with @xeviknal we talked past week to adjust this logic with the pattern used for the Wizard to put in wrap this modal, that it's also responsible of part of this logic.

PR #2032, opened 2020-12-09
Add display options for boxing.

jshaughn:
Part of https://github.com/kiali/kiali/issues/3439

This adds two Display options to control Namespace and Cluster boxing, both off by default.


israel-hdez:
Not related to this PR, but FYI:

I'm seeing this graph when using only _Show Namespace boxes_:

image

Notice the two Bookinfo namespace boxes. I guess it's two boxes because each box is hosted on a different cluster.

From an Istio philosophy, these two bookinfo boxes should be one because the are logically treated as the same NS. I'm not sure how much Maistra will agree with this, but it's probably something to discuss.

israel-hdez (approved):
Feature works and code is straightforward.
LG!

PR #2033, opened 2020-12-10
Add Github action to run install and run CI tests for PRs.

mtho11:
** Describe the change **
A github action to run CI: install/build/test. Similar to what Travis CI is doing today.

** Issue reference **
https://github.com/kiali/kiali/issues/3502

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [ '12' ]
+        node: [ '14' ]
     name: Node ${{ matrix.node }}
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,32 @@
+# This is a basic workflow to help you get started with Actions
+
+name: Kiali-UI CI
+
+# Controls when the action will run.
+on:
+  # Triggers the workflow on push or pull request events but only for the master branch
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node: [ '12' ]
+    name: Node ${{ matrix.node }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup node
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node }}
+      - run: yarn install
+      - run: yarn pretty-quick --check --verbose --branch origin/master
+      - run: yarn build:dev
+      - run: yarn test


lucasponce:
Note that this PR doesn't close the issue as it's also necessary a similar action for the backend repo.

mtho11:
@lucasponce So everything is deploying properly and tests being run properly. The issue that exists is one (un)related to tests failing because of ES2019 syntax of `xx.flatMap()` https://github.com/kiali/kiali-ui/pull/1978. You can see item #3 is the new flatMap addition: https://blog.logrocket.com/5-es2019-features-you-can-use-today/

So to test this stuff you need the action already deployed. So what I have done is copy the kiali-ui repo as a fresh copy for myself that I can use just to test gh actions. That way, I can install the gh action then add PRs to see if it is working properly.

This is my test repo PR that is failing for me: https://github.com/mtho11/mykiali/pull/1/checks?check_run_id=1532286116

mtho11:
We probably need to run in parallel for a while with gh-actions and Travis CI until we determine the champion. And fully, understand the pros/cons of each.

lucasponce:
Thanks @mtho11 I'll test it.
As a first thought perhaps we need to personalize something related to the yarn version used, because in travis we are installing a specific one on each job.

But anyway, as I have a pending github action in parallel I'll run this one in my local repo and give you feedback.

mtho11:
@lucasponce this should be good to go!

.github/workflows/ci.yml
```diff
@@ -0,0 +1,32 @@
...
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node: [ '12' ]
```
lucasponce, line 21: I guess travis is using 14 now.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,32 @@
...
+    name: Node ${{ matrix.node }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup node
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node }}
+      - run: yarn install
```
lucasponce, line 29: I don't know if there is a way to personalize which version of yarn to use.
If the default one is good enough, it's ok, but I remember that this created problems in the past and travis installs a specific version.

lucasponce: Note, this command is not installing yarn, I guess yarn comes from the node image

lucasponce: Node 14 image uses:
```
Run yarn install
yarn install v1.22.5
```
And travis installs

```
--version 1.22.4
```

So, perhaps it's ok.

lucasponce (changes requested):
Let's adjust the node version to 14 but I'm ok to give it a try.
I'm quite sure that we'll tweak the file as we go, I'm curious if we hit the memory problems that travis had and it may need to add more settings, but I'm ok to adjust that on the go.

.github/workflows/ci.yml
mtho11: Evidently, there is a yarn version implicit to the image and that seems to work. I started out using a yarn action because I thought that I had to and that caused various issues in running the tests (related to ES2019 compatibility). I spent over a day trying to figure that one out.

mtho11:
@lucasponce Ok I have changed the node version from 12 to 14 as requested.

.github/workflows/ci.yml
mtho11: 12 now.

lucasponce (approved):
@mtho11 let's merge it to see how it goes and what we need to tweak.

I wouldn't add them yet as part of the branches security to have some days of testing and when we think it's good enough then we can link the PR workflows with the actions.

I guess for the backend repo will have something similar as well, to put some actions first to see how it goes and we can give some days to see how it works.

PR #2034, opened 2020-12-10
Fix time control not working on custom dashboards

jotak:
I found a bug with custom dashboards, time selection doesn't work anymore. So here's a quick fix.
cc @lucasponce 

This is not ideal, I think it would be preferable that the child component (metrics, traces, whatever) decides which time component to use, instead of the parent deciding on behalf on its child. But not sure how to do it in a nice way, leaving that perhaps for later...

### DIFF
--- a/src/components/Metrics/CustomMetrics.tsx
+++ b/src/components/Metrics/CustomMetrics.tsx
@@ -18,7 +18,6 @@ import MetricsRawAggregation from '../MetricsOptions/MetricsRawAggregation';
 import { GrafanaLinks } from './GrafanaLinks';
 import { MetricsObjectTypes } from 'types/Metrics';
 import { SpanOverlay, JaegerLineInfo } from './SpanOverlay';
-import { retrieveTimeRange } from 'components/Time/TimeRangeHelper';
 import { DashboardModel, ExternalLink } from 'types/Dashboards';
 import { Overlay } from 'types/Overlay';
 import { Aggregator, DashboardQuery } from 'types/MetricsOptions';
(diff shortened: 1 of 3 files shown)


lucasponce:
> This is not ideal, I think it would be preferable that the child component (metrics, traces, whatever) decides which time component to use, instead of the parent deciding on behalf on its child. But not sure how to do it in a nice way, leaving that perhaps for later...

This is not true anymore, the time component doesn't belong to the child and parent (page) is responsible to decide which one to use until there can be a single component that work for all scenarios which would be the ideal. IMO.

jotak:
Well maybe yes the problem will be different if there was a single component. What I'm saying is just that with these two components that we currently have, it's easy to make it wrong because the logic of which component to use is dispatched between child and parent. I started to fix the child component, it built correctly so I thought that was ok, but the timerange props wasn't actually being updated... until I figured out some code had to be changed in the parent too.

Anyway, regardless this discussion, this fix should be ok.

lucasponce:
> Well maybe yes the problem will be different if there was a single component. What I'm saying is just that with these two components that we currently have, it's easy to make it wrong because the logic of which component to use is dispatched between child and parent. I started to fix the child component, it built correctly so I thought that was ok, but the timerange props wasn't actually being updated... until I figured out some code had to be changed in the parent too.
> 
> Anyway, regardless this discussion, this fix should be ok.

Yes, sorry, I was in the middle of testing it, I needed to recreate my cluster:
image

But I wanted to run it to see if I leaft any other side effect.

lucasponce (approved):

PR #2035, opened 2020-12-10
[v1.26] backport of fix for empty pod logs

jshaughn:
Fixes https://github.com/kiali/kiali/issues/3504

I don't understand how this bug manged it's way into 1.26.  I remember fixing it in the dev phase, but somehow that fix must have been lost.  Anyway, this problem is already fixed in 1.27+ of Kiali, but because Istio 1.8 is tied to 1.26 we should backport and make a patch release, so that Istio 1.8 users that install Kiali as an addon do not hit this issue.

Verification Steps:
Without the fix the following should show the reported error:

1) Navigate to Workload Details for istio-system, grafana workload
2) Select Logs tab

With the fix you should not see a crash and the two log windows should show messages about no logs.


jmazzitelli:
Why do we care about the v1.26 branch?

mtho11 (approved):
Verified the bug and the fix. Good to go!

PR #2037, opened 2020-12-11
Add graph display option to optionally show idle edges

jshaughn:
Add graph display option to optionally show idle edges (edges with notraffic for the time period, but with existing time-series (has had traffic at some other time period).  Default is false.

NOTE - this is a behavioral change as in the past Kiali always included these edges.  The new default is to hopefully improve the user experience.

**SERVER PR**: https://github.com/kiali/kiali/pull/3522

Fixes https://github.com/kiali/kiali/issues/3519

kiali#3519


lucasponce:
I'm starting to test it:
image
image

I don't expect to solve the discussed problem with mirroring traffic, but I suggest to put some comment in the tooltip:
image

Like "Include edges that have no request traffic for the time period **(including mirrored workloads)**" 

Perhaps wording can be better re-phrased but just a note to explain that situation to a potential user (without enter in details).

lucasponce:
Suggestion:
image

"Iddle edges", "Iddle nodes", to have two terms to refer same concept sounds odd.

jshaughn:
> I don't expect to solve the discussed problem with mirroring traffic, but I suggest to put some comment in the tooltip:
image  ...  Like "Include edges that have no request traffic for the time period (including mirrored workloads)"

As we talked about last week, the mirrored edge issue and the idle edges, are orthogonal.  The fact that you have an idle edge in your graph is due to previous, non-mirrored traffic to music-v3.  If mirrored traffic was reported by Istio in source telemetry then this would not be an issue, as the mirrored traffic edge would appear.

But, if you like I could make a reference, how about:

`Include edges that have no request traffic for the time period. Disabled by default to provide cleaner graphs. Enable to help detect unexpected traffic omissions, or to confirm expected edges with no traffic (due to routing, mirroring, etc).`

I'll make that change with the hope it resolves your comment.

pbajjuri20:
Tested and passed on Kiali UI 1.29.0 (0f4318a859e8769d9c4a4bc61869c8386b936ec4).


idel_edge

pbajjuri20 (approved):

lucasponce:
> Include edges that have no request traffic for the time period. Disabled by default to provide cleaner graphs. Enable to help detect unexpected traffic omissions, or to confirm expected edges with no traffic (due to routing, mirroring, etc).

> I'll make that change with the hope it resolves your comment.

+1, my point is to at lest have a reference in the app in case a user wonders for the omission.

> I thought about changing "Unused Nodes" to "Idle Nodes", but they are not really the same thing. "Idle" edges are edges that have seen traffic but have no activity for the current time period. "Unused" nodes have never seen traffic and would never show up in the graph unless the option is enabled. They are not the source or destination nodes for idle edges. So, I think I prefer to keep it as is.

But from user's perspective those are similar, so I guess we should try to align (here on in a followup) PR.

I mean, if we introduce the "iddle", personally I don't care if it has never show traffic or it doesn't have traffic for past 1h, I think as a user I'd like to see them both together.

jshaughn:
> But from user's perspective those are similar, so I guess we should try to align (here on in a followup) PR.
I mean, if we introduce the "iddle", personally I don't care if it has never show traffic or it doesn't have traffic for past 1h, I think as a user I'd like to see them both together.

For me it's more confusing to align the language, as it infers that the two are related, and they are not (imho).   As an aside "idle" has a single 'd'.

Anyway, I would defer to UX.  @lwrigh, do you have a comment based on the above discussion, otherwise we can ask in the UX meeting. Thanks.

lucasponce:
> Anyway, I would defer to UX. @lwrigh, do you have a comment based on the above discussion, otherwise we can ask in the UX meeting. Thanks.

I'm not sure if this is a UX issue but a feature.

You are supposing that the user has the different between Kiali's definition of "iddle" and "unused".

That's an expectation that is not right in my opinion.

As a user I don't know the difference (and I don't mind) it's same concept: no traffic for workloads, the difference between if they had ever traffic sometime stored in prometheus vs no prometheus traffic but I know that exist a workload because I correlate with other source should be transparent to the user.

Hence, same concepts should have same wording (to be honest, I don't care if those are iddle or unusued).

jshaughn:
@lucasponce Considering you are not convinced by my argument that the two are fundamentally different, I will make the change because I'd like this to progress.  The users can defer to the info tooltip for details about the options.   This gives us idle pods, idle edges and idle nodes, all somewhat different imo but there is commonality in the idea that it represents a lack of activity.

pbajjuri20:
Deployed and tested the new changes on "IDLE EDGES" and "IDLE NODES".


Idell_node

lucasponce (approved):

PR #2040, opened 2020-12-16
Fetch Namespace if they arent present while mounting the Istio Status feat

xeviknal:
Trying to approach [this](https://github.com/kiali/kiali/issues/3508#issuecomment-742617476) I have realized that the Istio component status feature in the masthead had a minor/little bug.

The component relies in the namespace list (from redux) just for reducing the severity of alerts for that component. This is specially important for multi-tenancy scenarios where users haven't given access to any namespace. Then, we don't want Kiali show an error every other refresh/page navigation.

I have found that since the `Istio Status` feature is rendered in the masthead and the `Namespace Dropdown` is fetch in the second secondary masthead, the first one won't have any namespace rendered. It will only have namespaces right after the `Namespace Dropdown` has been successfully rendered.

This should fix it. Pretty tiny, but the behavior was a bit odd.


mtho11 (approved):
LGFM. Tested ok.

PR #2041, opened 2020-12-17
For Github actions CI add caching to reduce build time (a bunch)

mtho11:
** Describe the change **
This is is a build CI optimization for Github Actions to reduce the build time by > 50% by using cached libraries that refresh only when the `yarn.lock` changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,10 @@ jobs:
     name: Node ${{ matrix.node }}
     steps:
       - uses: actions/checkout@v2
+      - uses: actions/cache@v2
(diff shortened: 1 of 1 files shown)


mtho11:
My testing on my repo has shown this enhancement to greatly improve performance

**New**:
Add_caching_of_node_modules_·_mtho11_mykiali_d0a219b

lucasponce (approved):
Certainly it looks ok.
I was to ask how to manually refresh the caches (it's not an usual, but in Travis sometimes it's necessary to start "fresh") but it seems that detecting changes the yarn.lock is a good way.

PR #2042, opened 2020-12-17
Add a copy to clipboard feat into the proxy-config

xeviknal:
Requested in the last sprint meeting: adding a copy to clipboard function in the proxy-config modal (only for json options: all, bootstrap).

proxy-details-copyclipboard

@jshaughn and @lucasponce suggested this improvement. 
Lucas, you suggested to copy partial selections, I don't know if having the button up there for that is helpful since the Ctrl+C work pretty good.

Another thing it could be to have an advanced copy-to-clipboard that copies specific parts of the json, e.g. cluster section.


jshaughn (changes requested):
For me the test failed.  Clicking the copy icon had no effect.  As a sanity check I tried the copy feature on the logs tab and it worked as expected.  Not sure what the issue is.   As for partial copy, standard c+p is fine in my opinion.

hhovsepy:
Works on Chrome, but on FF the copy is not working properly for 'all', works only after clicking several time on different places in copy icon.

jshaughn:
> Works on Chrome,

Actually, it failed on my chrome, which is a bit old on F29.

xeviknal:
@hhovsepy @jshaughn my bad. Didn't remember about the clipboard of logs and thought the PF clipboard would work.
I've changed the feat to be equal to what we have in the logs. 
Thanks!

hhovsepy (approved):
Verified both on FF and Chrome.

jshaughn (approved):
LG!

PR #2045, opened 2021-01-04
[multi-cluster] graph side-panel support for app/cluster/namespace box nodes

jshaughn:
Updates to side-panel support for multi-cluster.  Renamed SummaryPanelGroup to SummaryPanelAppBox and updated as needed.  Added SummaryPanelClusterBox and SummaryPanelNamespaceBox support; each is modeled roughly like a constrained full-graph side-panel.

Closes https://github.com/kiali/kiali/issues/3545


israel-hdez:
I'll review tomorrow at my start of day.

src/pages/Graph/SummaryPanelClusterBox.tsx
```diff
@@ -0,0 +1,245 @@
...
+
+    const numSvc = boxed.filter(`node[nodeType = "${NodeType.SERVICE}"]`).size();
+    const numWorkloads = boxed.filter(`node[nodeType = "${NodeType.WORKLOAD}"]`).size();
+    const { numApps, numVersions } = this.countApps(boxed);
+    const numEdges = boxed.edges().size();
+    // when getting accumulated traffic rates don't count requests from injected service nodes
+    const nonServiceEdges = boxed.filter(`node[nodeType != "${NodeType.SERVICE}"][!isBox]`).edgesTo('*');
+    const totalRateGrpc = getAccumulatedTrafficRateGrpc(nonServiceEdges);
+    const totalRateHttp = getAccumulatedTrafficRateHttp(nonServiceEdges);
+    const incomingEdges = clusterBox.cy().nodes(`[${CyNode.cluster} != "${cluster}"]`).edgesTo(boxed);
```
israel-hdez, line 60: I guess the difference in how edges are resolved may be the issue with the data mismatches between Total and In/Outcoming.

israel-hdez (commented):
The new NS and Cluster side panels are good for the "source" cluster. But for the destination, there are missing charts. For NS panel, _Total_ tab is populated, but _Incoming_ and _Outgoing_ are not. For Cluster panel, there is data in _Incoming_ tab, but nothing in _Outgoing_ (which is OK) and also nothing in _Total_ (which is not OK).

IDK, p8s metrics could be wrong, or maybe it's the panels:

SidePanelIssues

Besides that, all else looks OK. Code is fine.

Some side comments that are probably unrelated to this PR:
* Probably, for now, we should disable drill-down for cluster and ns boxes. The drilled down graph is not the expected one IMO.
* The already mentioned _highlighted but grayed-out nodes_ issue

jshaughn:
@israel-hdez I pushed a commit to fix issues with drill-down of app and app-box nodes, and to disable double-click for namespace-box and cluster-box nodes.  I'll take another look at highlighting now..

israel-hdez:
@jshaughn I'm happy with the fixes.
The side panel issue is, probably, tricky to solve without a proper env for testing. So, do you want to move this forward?

I can spend some time finding how to share my env so that you (and others) can test. In the meantime, I could try to fix that specific charts issue.

jshaughn:
> I can spend some time finding how to share my env so that you (and others) can test. In the meantime, I could try to fix that specific charts issue.

@israel-hdez contributions accepted :)   Also, I am working on improving the layout so please look into the side panel issue if you can.

src/components/CytoscapeGraph/Layout/BoxLayout.ts
```diff
@@ -0,0 +1,265 @@
...
+        const boxLayout = boxedElements.layout(boxLayoutOptions);
+
+        boxLayout.on('layoutstart layoutready layoutstop', _evt => {
+          // Avoid propagating any local layout events up to cy, this would yield a global operation before the nodes are ready.
+          return false;
+        });
+
+        // synch layouts (dagre) stop before run() returns, async layouts (cose,cola) don't, so
+        // wait for the async layouts to stop before continuing.
+        this.waitForLayout(boxLayout);
```
josejulio, line 157: Does this work? AFAIK it is not really waiting outside of the `waitForLayout`.
`waitForLayout` is an `async function`, meaning that it is really returning a promise which you need to `await` here so it "waits". The async will bubble up to the top and this layout would need to be an async layout too.

jshaughn: No, it does not work :)  But I fixed it...

src/components/CytoscapeGraph/Layout/BoxLayout.ts
```diff
@@ -0,0 +1,269 @@
...
+
+      // (1.a) Prepare box node by assigning a size and running the compound layout
+      boxNodes.each(boxNode => {
+        const boxedNodes = boxNode.children();
+        const boxedElements = boxedNodes.add(boxedNodes.edgesTo(boxedNodes));
+        const boxLayout = boxedElements.layout(boxLayoutOptions);
+
+        // discrete layouts (dagre) stop before run() returns, continuous layouts (cose,cola) don't,
+        // force discrete behavior because we need each box to layout before continuing.
+        this.runDiscrete(boxLayout);
```
josejulio, line 152: This is not really waiting here. Is that OK?

jshaughn: Should be fixed.

jshaughn:
@israel-hdez I believe all of your comments are now addressed.  In the latest commit I fixed one more issue in the side panels (edge counts) and tweaked your fixes for the traffic accumulations (your fix was good, I just changed the logic a little to maybe be more tolerant of future changes).

I do have one more test that I need to perform locally to ensure it's tolerant of operation nodes, but first I have to update the script for injecting operation nodes into the travel demo...

src/pages/Graph/SummaryPanelNamespaceBox.tsx
```diff
@@ -119,19 +119,20 @@ export default class SummaryPanelNamespaceBox extends React.Component<
...
     const { numApps, numVersions } = this.countApps(boxed);
-    const numEdges = boxed.edges().size();
-    // when getting accumulated traffic rates don't count requests from injected service nodes
-    const nonServiceEdges = boxed.filter(`node[nodeType != "${NodeType.SERVICE}"][!isBox]`).edgesTo('*');
+    const numEdges = boxed.connectedEdges().size();
+    // incoming edges are from a different namespace or a different cluster
     const incomingEdges = namespaceBox
       .cy()
-      .nodes(`[${CyNode.namespace} != "${namespace}"], [${CyNode.cluster} != "${cluster}"], [?${CyNode.isRoot}]`)
+      .nodes(`[${CyNode.namespace} != "${namespace}"],[${CyNode.cluster} != "${cluster}"]`)
```
israel-hdez, line 126: I see you removed root nodes for incoming edges both in NS and Cluster boxes.
But this have the side effect that the incoming tab is not populated for gateways:

image

I though we wanted that, because the `SummaryPanelGraph` counts edges of root nodes as incoming traffic and it is more consistent with the "Total" tab. Anyway, I'm not sure if we want to count this for NS and Cluster boxes and I see it may be slightly confusing as the box may give the idea that traffic is not really "going in". So it's your choice.

src/components/CytoscapeGraph/Layout/BoxLayout.ts
```diff
@@ -0,0 +1,269 @@
...
+      }
+
+      // (1.a) Prepare box node by assigning a size and running the compound layout
+      boxNodes.each(boxNode => {
+        const boxedNodes = boxNode.children();
+        const boxedElements = boxedNodes.add(boxedNodes.edgesTo(boxedNodes));
+        const boxLayout = boxedElements.layout(boxLayoutOptions);
+
+        // discrete layouts (dagre) stop before run() returns, continuous layouts (cose,cola) don't,
+        // force discrete behavior because we need each box to layout before continuing.
```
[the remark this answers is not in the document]
josejulio: > I think @josejulio comment still stands, because `runDiscrete` can return while the layout is still running.

You are correct there.

> Anyway, I'm not seeing any issue. I guess it's because it's always dagre inside the boxes, right?

If dagre is used in children, then it's fine, but I understood that you wanted to use the continuos layouts. If they work a raw guess is that It could also be that the continuous layouts do a first pass on their "run" method and that is enough for the purposes of this algorithm. Hard to tell without looking at their code :).

src/pages/Graph/SummaryPanelNamespaceBox.tsx
jshaughn: @israel-hdez I was going back and forth on this and so I asked @jmazzitelli for his opinion.  His initial reaction was to keep it as is, but then he too started to wonder whether it would be more consistent to include the boxed root nodes.  After some playing around we liked the idea of including the boxed root nodes (aka traffic source nodes) but also making them more obvious in the graph, so that users could more easily see the nodes contributing to incoming traffic.  So, we added a new badge.  We both liked it, and in chat you also were +1.  So, I have reverted things and so incoming traffic will include the boxed root nodes (as well as any edges entering the box), as well as be badged:

image

src/components/CytoscapeGraph/Layout/BoxLayout.ts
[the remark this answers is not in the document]
jshaughn: In my testing, the current code waits as desired for the layout to stop, regardless of which layout is selected.  We force dagre only for app boxes, namespace and cluster boxing use the selected layout for the boxed nodes, as does the final layout for the full graph.  If you are not seeing this behavior let me know.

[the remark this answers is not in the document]
jshaughn: @israel-hdez @josejulio OK, after talking to Edgar and doing some more reading I think I now have a better handle on what is needed to support a continuous layout for boxes.  I had been hoping for something more simple but have now accepted that it's hard.  :)  I have restructured the code and I'm fairly sure it now performs correctly.

src/components/CytoscapeGraph/Layout/BoxLayout.ts
```diff
@@ -245,25 +207,120 @@ export default class BoxLayout {
...
+        }
+        // (2.d) Remove all child nodes from parents (and their edges).
+        removedElements = removedElements.add(this.cy.remove(boxedNodes));
+
+        resolve({ boxNodes: boxNodes, syntheticEdges: syntheticEdges, removedElements: removedElements });
+      });
+    });
+  }
+
+  async runLayout(layoutName, layout): Promise<any> {
```
israel-hdez, line 306: minor: since you are not using `await` inside this func, you can remove the `async` keyword.

src/components/CytoscapeGraph/Layout/BoxLayout.ts
```diff
@@ -245,25 +207,120 @@ export default class BoxLayout {
...
       });
     });
+
     layout.run();
   }
 
-  async runDiscrete(layout) {
-    layout.on('layoutstart layoutready layoutstop', _evt => {
-      // Avoid propagating any local layout events up to cy, this would yield a global operation before the nodes are ready.
+  async layoutBoxType(boxByType: BoxByType): Promise<LayoutBoxTypeResult> {
```
israel-hdez, line 213: ídem: minor: since you are not using `await` inside this func, you can remove the `async` keyword.

israel-hdez (approved):
All good!
It's only two quite minor comments.

src/components/CytoscapeGraph/Layout/BoxLayout.ts
jshaughn: The async keyword is still useful because it ensures a promise is returned.   Even if the function returns a non-promise it is wrapped in a resolved promise.  I know I needed this in at least one of the functions, but either way it doesn't really hurt anything.  I'll keep it for now.

src/components/CytoscapeGraph/Layout/BoxLayout.ts
jshaughn: see above comment.

PR #2047, opened 2021-01-06
Graph: navigate in trace

jotak:
Redesign the trace details panel in graph to allow navigating in a given
trace, from span to span, following parent/children relationships

Fixes kiali/kiali#3361

Prior refactoring (1st commit): move tracing utility functions to 'utils/'

### DIFF
--- a/src/components/SimplerSelect.tsx
+++ b/src/components/SimplerSelect.tsx
@@ -1,0 +1,33 @@
+import * as React from 'react';
+import { Select, SelectOptionObject, SelectProps } from '@patternfly/react-core';
+
+type Props = Omit<Omit<Omit<SelectProps, 'isExpanded'>, 'onSelect'>, 'onToggle'> & {
+  onSelect?: (selection: string | SelectOptionObject) => void;
+  onToggle?: (isExpanded: boolean) => void;
+};
+
+const SimplerSelect = (props: Props) => {
+  const [isExpanded, setExpanded] = React.useState(false);
+  return (
+    <Select
+      {...props}
+      onToggle={isExpanded => {
+        if (props.onToggle) {
+          props.onToggle(isExpanded);
+        }
+        setExpanded(isExpanded);
+      }}
+      onSelect={(_, selection, isPlaceholder) => {
+        setExpanded(false);
+        if (!isPlaceholder && props.onSelect) {
+          props.onSelect(selection);
+        }
+      }}
+      isExpanded={isExpanded}
+    >
+      {props.children}
+    </Select>
+  );
+};
+
+export default SimplerSelect;
--- a/src/pages/Graph/SummaryPanelTraceDetails.tsx
+++ b/src/pages/Graph/SummaryPanelTraceDetails.tsx
@@ -3,42 +3,37 @@ import { Link } from 'react-router-dom';
 import { connect } from 'react-redux';
 import { ThunkDispatch } from 'redux-thunk';
 import { style } from 'typestyle';
-import { Tooltip, Button, ButtonVariant, pluralize } from '@patternfly/react-core';
-import {
-  CloseIcon,
-  AngleLeftIcon,
-  AngleRightIcon,
-  ExternalLinkAltIcon,
-  ExclamationCircleIcon
-} from '@patternfly/react-icons';
+import { Tooltip, Button, ButtonVariant, pluralize, SelectOption } from '@patternfly/react-core';
+import { CloseIcon, ExternalLinkAltIcon, ExclamationCircleIcon, MapMarkerIcon } from '@patternfly/react-icons';
 
 import { URLParam } from '../../app/History';
-import { JaegerTrace, Span } from 'types/JaegerInfo';
+import { JaegerTrace, RichSpanData, EnvoySpanInfo, OpenTracingHTTPInfo, OpenTracingTCPInfo } from 'types/JaegerInfo';
 import { KialiAppState } from 'store/Store';
 import { KialiAppAction } from 'actions/KialiAppAction';
 import { JaegerThunkActions } from 'actions/JaegerThunkActions';
+import { GraphActions } from 'actions/GraphActions';
 import { PFAlertColor } from 'components/Pf/PfColors';
-import {
-  extractEnvoySpanInfo,
-  extractOpenTracingHTTPInfo,
-  extractOpenTracingTCPInfo,
-  getSpanType
-} from 'components/JaegerIntegration/JaegerHelper';
-import { formatDuration } from 'components/JaegerIntegration/JaegerResults/transform';
+import { findChildren, findParent, formatDuration } from 'utils/tracing/TracingHelper';
 import { CytoscapeGraphSelectorBuilder } from 'components/CytoscapeGraph/CytoscapeGraphSelector';
 import { decoratedNodeData } from 'components/CytoscapeGraph/CytoscapeGraphUtils';
 import FocusAnimation from 'components/CytoscapeGraph/FocusAnimation';
 import { FormattedTraceInfo, shortIDStyle } from 'components/JaegerIntegration/JaegerResults/FormattedTraceInfo';
+import SimplerSelect from 'components/SimplerSelect';
+import { summaryFont } from './SummaryPanelCommon';
+import { NodeParamsType, GraphType } from 'types/Graph';
+import { bindActionCreators } from 'redux';
 
 type Props = {
   trace: JaegerTrace;
   node: any;
+  graphType: GraphType;
   jaegerURL?: string;
   close: () => void;
+  setNode: (node?: NodeParamsType) => void;
 };
 
 type State = {
-  selectedSpan: number;
+  selectedSpanID: string | undefined;
 };
 
 const textHeaderStyle = style({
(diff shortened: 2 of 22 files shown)


jotak:
Draft PR, as not all kind of graphs are ok
- [x] Version app graph
- [x] Workload graph
- [ ] Service graph
- [ ] App graph

jotak:
Before going further we can discuss if the design is good or not.
Example:

Capture d’écran de 2021-01-06 15-26-53

Main changes are:
- I removed the arrow-based span selection, replaced with a dropdown. Reason is the arrows are misleading IMO, they tend to suggest we're going to see the next or previous spans in trace, whereas it's other spans _for the node_, without necessarily any direct relationship.

- Added these new "Previous" and "Next" items, which _do_ stand for the parent/children relationships. In these items, we have:
  - Zero or one parent
  - Zero to N children
  - Each comes with a link to select the span, eventually re-focusing the relevant graph node, and an icon to locate on graph (triggers the focus animation).

jshaughn:
> Main changes are:
> I removed the arrow-based span selection, replaced with a dropdown. Reason is the arrows are misleading IMO, they tend to suggest we're going to see the next or previous spans in trace, whereas it's other spans for the node, without necessarily any direct relationship.

+1, in fact I think I was one of the mislead users :)  

>Added these new "Previous" and "Next" items, which do stand for the parent/children relationships. In these items, we have:
Zero or one parent
Zero to N children
Each comes with a link to select the span, eventually re-focusing the relevant graph node, and an icon to locate on graph (triggers the focus animation).

I think this is definitely more clear and more useful.   Although, I am finding the density of the display a little difficult.  I don't really like the "N Spans on this Node" dropdown.  Maybe it would be combined with the Previous/Next in some way, like maybe an expandable table that allowed a single expansion to show the details of a span.  

Spans for the Node:
```
> From    productpage-v1
v To      details-v1
    Started after: 15.77ms
    Duration: 161.12ms
    Inbound request: GET http://istio-ingressgateway-istio-system.apps.jshaughn.lab.upshift.rdu2.redhat.com/productpage
    Response: code 200
> To      reviews-v1
```

israel-hdez:
> * Reason is the arrows are misleading IMO, they tend to suggest we're going to see the next or previous spans in trace, whereas it's other spans _for the node_

That's interesting. So, as Jay, I also though I was changing spans within the selected trace.

I'd suggest reaching UX, because the way the side panel works (first click a trace to be able to see details and move to other spans) still suggests some hierarchy. Without you mentioning this, even with the dropdown I would still think it's spans for the chosen trace AND "this" node.

I think this will be a tricky one to properly solve for UX.

> I think this is definitely more clear and more useful. Although, I am finding the density of the display a little difficult. I don't really like the "N Spans on this Node" dropdown. Maybe it would be combined with the Previous/Next in some way, like maybe an expandable table that allowed a single expansion to show the details of a span.

I support this idea from Jay.
I'd say that we shouldn't force all this info to be in the side panel because space is quite limited, so I think it would be quite fine to open a modal dialog (or a draggable dialog) in this case, because both horizontal and vertical space is really needed.

We have been avoiding dialogs in the graph and we have succeeded. But this info is somewhat detailed to live in the side panel and, IMHO, it deserved a dedicated space.

The table that Jay suggests is good, but you get extra cookies if you can show Jaegers Gantt. And more cookies if by clicking over the Gantt chart you do the graph re-focus as you want. Anyway, this is probably my "dream" and won't fit the sprint ;-) So, it's a suggestion for "later" in case it's doable.

jotak:
> > * Reason is the arrows are misleading IMO, they tend to suggest we're going to see the next or previous spans in trace, whereas it's other spans _for the node_
> 
> That's interesting. So, as Jay, I also though I was changing spans within the selected trace.
> 
> I'd suggest reaching UX, because the way the side panel works (first click a trace to be able to see details and move to other spans) still suggests some hierarchy. Without you mentioning this, even with the dropdown I would still think it's spans for the chosen trace AND "this" node.

Wait, I think I wasn't clear enough, again :-)
This is still constrained into the selected trace, not moving to other traces. It's as you say: spans for the node AND for the trace. But what I mean is, arrows suggest a navigation that follows parental relationship between spans. And that is not the case. The "next" span may not be directly related to the current one. I think it was misleading, and is more clear with a dropdown.

israel-hdez:
> This is still constrained into the selected trace, not moving to other traces. It's as you say: spans for the node AND for the trace. But what I mean is, arrows suggest a navigation that follows parental relationship between spans. And that is not the case. The "next" span may not be directly related to the current one (though it has indirect relations, but there may be other spans between, attached to other nodes ... or they may be "cousins", belonging to different subtrees). I think it was misleading, and is more clear with a dropdown.

Oh! That makes much more sense :-)
So, UX doesn't need to be involved.

jotak:
> I think this is definitely more clear and more useful. Although, I am finding the density of the display a little difficult. I don't really like the "N Spans on this Node" dropdown. Maybe it would be combined with the Previous/Next in some way, like maybe an expandable table that allowed a single expansion to show the details of a span.

But the items shown in the dropdown might be different from the items available through navigation (previous/next), they are different things, it's for that purpose that I want to show them apart. To me, it's the whole point of that PR to have this disambiguation.


> 
> Spans for the Node:
> 
> ```
> > From    productpage-v1
> v To      details-v1
>     Started after: 15.77ms
>     Duration: 161.12ms
>     Inbound request: GET http://istio-ingressgateway-istio-system.apps.jshaughn.lab.upshift.rdu2.redhat.com/productpage
>     Response: code 200
> > To      reviews-v1
> ```

I'm not sure if I understand correctly this example, but here you don't see all the spans that are available for the focused node.

jotak:
Regarding the density of information, I agree. Perhaps instead of listing all children span by their operation names, I could just write "child 1", "child 2" etc. instead, and more info in tooltip?
Or as you suggested, have an expandable section (but keeping the dropdown above as per my previous comment)

jotak:
@jshaughn @israel-hdez next try, with a single list of related + tooltips, wdyt?
Capture d’écran de 2021-01-08 14-33-30

israel-hdez:
> next try, with a single list of related + tooltips, wdyt?

Much cleaner, IMO!
But then, why not use the shortname (like `ai-visitors`) and use the tooltip for the long name?
In your previous trials it was always the long names plus the short names.

Anyway, if you keep this one, it's OK. Looks quite good.

jotak:
> > next try, with a single list of related + tooltips, wdyt?
> 
> Much cleaner, IMO!
> But then, why not use the shortname (like `ai-visitors`) and use the tooltip for the long name?
> In your previous trials it was always the long names plus the short names.

It wasn't really a long name plus a short name, what you see as short name is the workload where the span is created, and the long name is the operation name, which in the case of envoy spans, announces the destination of a request.

If you look at my previous screenshot (for my first attempt), you'll notice that the short name is always "travels-v1" and long names are "insurance...", "flights..." etc. Because these are spans created for a client connection, by the client. That would be different for server-side spans. And that would be different again with spans not created by envoy. So ... at the moment I haven't found a nice way to name spans with meaningful short names, that work in every situation. We can keep thinking about it :)

jotak:
this is ready for test/review

pbajjuri20:
I have tested with the latest changes

Trace_Graph1

src/components/SimplerSelect.tsx
```diff
@@ -0,0 +1,33 @@
+import * as React from 'react';
+import { Select, SelectOptionObject, SelectProps } from '@patternfly/react-core';
+
+type Props = Omit<Omit<Omit<SelectProps, 'isExpanded'>, 'onSelect'>, 'onToggle'> & {
+  onSelect?: (selection: string | SelectOptionObject) => void;
+  onToggle?: (isExpanded: boolean) => void;
+};
+
+const SimplerSelect = (props: Props) => {
+  const [isExpanded, setExpanded] = React.useState(false);
```
israel-hdez, line 10: Oh! you are trying react hooks :)

src/pages/Graph/SummaryPanelTraceDetails.tsx
```diff
@@ -155,42 +147,33 @@ class SummaryPanelTraceDetails extends React.Component<Props, State> {
...
                   }}
                 >
-                  <AngleRightIcon />
-                </Button>
+                  {spans.map(s => {
+                    return (
+                      <SelectOption
+                        key={s.spanID}
+                        value={s.spanID}
+                        onClick={() => this.setState({ selectedSpanID: s.spanID })}
```
israel-hdez, line 166: Isn't this a duplicate of `onSelect` event?

israel-hdez (approved):
It's interesting to see this live.
Also, quite interesting to see that "child" spans can have an operation name that are similar to other workloads yet it's a span of the same node.

I'm good with this. I only have very minor comments.

src/components/SimplerSelect.tsx
jotak: yes it's actually the second time, it's a lot less boilerplate when the situation allows it!

src/pages/Graph/SummaryPanelTraceDetails.tsx
jotak: looks like it is, good catch!

jotak: fixed

pbajjuri20 (approved):

israel-hdez:
Thanks @jotak. Nice improvement.

PR #2049, opened 2021-01-11
Fix service-type graph not showing service-to-service trace edges

jotak:
Fixes https://github.com/kiali/kiali/issues/3572

**Before**
Capture d’écran de 2021-01-11 12-09-05

**After**
Capture d’écran de 2021-01-11 12-05-51

### DIFF
--- a/src/components/CytoscapeGraph/CytoscapeTrace.ts
+++ b/src/components/CytoscapeGraph/CytoscapeTrace.ts
@@ -27,7 +27,25 @@ const showSpanSubtrace = (cy: Cy.Core, graphType: GraphType, span: Span) => {
   // From upstream to downstream: Parent app or workload, Inbound Service Entry, Service, App or Workload, Outbound Service Entry
   let lastSelection: Cy.NodeCollection | undefined = undefined;
 
-  if (graphType === GraphType.APP) {
+  if (graphType === GraphType.SERVICE) {
+    // In service graph type, parent can be a Service or a Workload (e.g. when it initiates the transaction)
+    const sourceAppNs = searchParentApp(span);
+    if (sourceAppNs) {
+      let selector = `node[!${CyNode.isGroup}][${CyNode.nodeType}="${NodeType.SERVICE}"][${CyNode.app}="${sourceAppNs.app}"][${CyNode.namespace}="${sourceAppNs.namespace}"]`;
+      let parent = cy.elements(selector);
+      if (!parent || parent.length === 0) {
+        // Try workload
+        const sourceWlNs = searchParentWorkload(span);
+        if (sourceWlNs) {
+          selector = `node[${CyNode.workload}="${sourceWlNs.workload}"][${CyNode.namespace}="${sourceWlNs.namespace}"]`;
+          parent = cy.elements(selector);
+        }
+      }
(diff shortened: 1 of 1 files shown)


jotak:
Note, for seeing this kind of traces with both workload-to-service and service-to-service edges while testing, you can use demo-mesh-arena with traces enabled, see quickstart https://github.com/jotak/demo-mesh-arena/#quick-start

src/components/CytoscapeGraph/CytoscapeTrace.ts
```diff
@@ -27,7 +27,25 @@ const showSpanSubtrace = (cy: Cy.Core, graphType: GraphType, span: Span) => {
...
+    if (sourceAppNs) {
+      let selector = `node[!${CyNode.isGroup}][${CyNode.nodeType}="${NodeType.SERVICE}"][${CyNode.app}="${sourceAppNs.app}"][${CyNode.namespace}="${sourceAppNs.namespace}"]`;
+      let parent = cy.elements(selector);
+      if (!parent || parent.length === 0) {
+        // Try workload
+        const sourceWlNs = searchParentWorkload(span);
+        if (sourceWlNs) {
+          selector = `node[${CyNode.workload}="${sourceWlNs.workload}"][${CyNode.namespace}="${sourceWlNs.namespace}"]`;
+          parent = cy.elements(selector);
+        }
```
aljesusg, line 42: What do you think about this and avoid another if confition?

```ts
const sourceWlNs = searchParentWorkload(span);
if ((!parent || parent.length === 0) && sourceWlNs) {
        // Try workload
          selector = `node[${CyNode.workload}="${sourceWlNs.workload}"][${CyNode.namespace}="${sourceWlNs.namespace}"]`;
          parent = cy.elements(selector);
}
```
do you make sense to you?

aljesusg (approved):
Not block, just a comment LGFM anyway

src/components/CytoscapeGraph/CytoscapeTrace.ts
aljesusg: Note: I didn't check searchParentWorkload(span) , maybe it's a lot of complexity in that method

jotak: Thanks for the review.
`searchParentWorkload` is neither very complex nor very trivial, but still, I'd prefer not to do unnecessary call here.

aljesusg: understood so we can merge this one

jshaughn: post-merge review :-)  The logic looks OK to me, I think it should work as expected.  Style-wise I may have used a switch statement but that's just personal pref.

aljesusg: OMG Sorry my fault @jshaughn :'( I didn't see the others reviewers assign :(

jshaughn: Not a problem, @aljesusg.   I usually wait for the PR developer to perform the merge, but it's not a strict process on our team.  Anyway, no big deal!

PR #2050, opened 2021-01-13
Make idle priority above healthy but below degraded

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/3493

Note that there appears to be a bug here regardless of what the priority ordering should be between Idle, Degraded, Failure, and Healthy. So we need to fix that no matter what. The bug is seen here in the original code where "Idle" and "Failure" BOTH have the same value of "3" - this causes the [map.reduce](https://github.com/kiali/kiali-ui/blob/v1.28.0/src/types/Health.ts#L273-L275) to potentially produce random results (depending on which Idle and Failure map items are in which order).

What this PR also does (aside from making all the priority values unique from 0 through 4) is places "Idle" ahead of Healthy but before Degraded. This means a global status "Idle" will show if one or more items are Idle (with the rest Healthy). It also means if any items are degraded (or worse), that higher priority status (degraded or worse) will show as the global status.

It is arguable if that priority ordering is what we want. But it seems on the surface as the best ordering of priority out of all the other possible combinations. We can discuss in this PR or the issue regarding the priority ordering. But, as I say, IMO at minimum we have to change this so that Idle and Failure are NOT the same value ("3").


jshaughn:
The proposed fix/change makes sense to me, although I'll defer to @lucasponce and @aljesusg.

jmazzitelli:
not ready to merge... need to fix unit tests

jmazzitelli:
added @israel-hdez to reviewers list since he was the one who wrote the original issue this PR will fix.

lucasponce:
The fix sounds ok, I've added also @hhovsepy and QE to make some regression as they have tested more corner cases scenarios.

aljesusg (approved):
LGFM

hhovsepy (approved):
Regression test passed.

hhovsepy:
Screenshot from 2021-01-14 11-24-26
Screenshot from 2021-01-14 11-24-41

israel-hdez (approved):
I see all is fixed (I found another bug, but I'll open a new issue for that one).

Personally, I don't like showing Idle when there are both Idle and Healthy pods because I that confuses me. Anyway, the code is fine. The original issue is fixed. And consensus says statuses shown are good enough. So, I'm approving.

PR #2051, opened 2021-01-13
Adding filters into envoy dump tables

xeviknal:
belongs to https://github.com/kiali/kiali/issues/3346

Adding filters to help explore the envoy dump tables for each pod.

proxy-filtering

### DIFF
--- a/src/components/Envoy/EnvoyModal.tsx
+++ b/src/components/Envoy/EnvoyModal.tsx
@@ -101,10 +101,11 @@ class EnvoyDetailsModal extends React.Component<EnvoyDetailProps, EnvoyDetailSta
   }
 
   componentDidUpdate(_prevProps: EnvoyDetailProps, prevState: EnvoyDetailState) {
-    if(this.state.pod.name !== prevState.pod.name || this.state.resource !== prevState.resource) {
-      if (this.state.fetch) {
-        this.fetchContent();
-      }
+    if (
+      this.state.fetch &&
+      (this.state.pod.name !== prevState.pod.name || this.state.resource !== prevState.resource)
+    ) {
+      this.fetchContent();
     }
   }
 
--- a/src/components/Envoy/EnvoyModal.tsx
+++ b/src/components/Envoy/EnvoyModal.tsx
@@ -230,7 +230,7 @@ class EnvoyDetailsModal extends React.Component<EnvoyDetailProps, EnvoyDetailSta
             </Toolbar>
           </StackItem>
           <StackItem>
-            <Card style={{ height: '400px' }}>
+            <Card style={{ height: '600px' }}>
               {this.isLoadingConfig() ? (
                 <Loading />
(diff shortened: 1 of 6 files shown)


skondkar (changes requested):
Tested on Kiali UI :     1.29.0 (55fa282bdcde6138f98a863b5c0ed5da59dc463e)

Looks good in chrome, but some issue in firefox ( version 70 and 68.8)
When selecting values from 'resources' dropdown, from 'all' to 'clusters', it works, and then from 'clusters' to 'listners', it throws some error in UI:

Could not fetch envoy config routes entries for productpage-v1-6b746f74dc-qmsq5.
Details shows: Unable to listen on any of the requested ports: [{15000 15000}]

Screenshot-PortError2

skondkar:
One more observation:
When we select resources as 'Clusters' and then 'Direction' from the dropdown, the next dropdown shows values 'Direction, inbound, outbound and a hyphen as seen in the screenshot
The values should be inbound and outbound.

Screenshot-DirectionDropdown

xeviknal:
@skondkar both changes approached. thanks

skondkar (approved):
Thanks @xeviknal 
Verified on Kiali UI :  1.29.0 (1c2bce8d305546e82f2036ee9b0e7e8c9528f31b)
LGTM

src/components/Envoy/EnvoyModal.tsx
```diff
@@ -74,17 +78,33 @@ class EnvoyDetailsModal extends React.Component<EnvoyDetailProps, EnvoyDetailSta
...
   }
 
-  componentDidUpdate() {
-    if (this.state.fetch) {
-      this.fetchContent();
+  componentDidUpdate(_prevProps: EnvoyDetailProps, prevState: EnvoyDetailState) {
+    if(this.state.pod.name !== prevState.pod.name || this.state.resource !== prevState.resource) {
+      if (this.state.fetch) {
+        this.fetchContent();
+      }
```
aljesusg, line 107: what do you think about
```ts
 if(this.state.fetch && (this.state.pod.name !== prevState.pod.name || this.state.resource !== prevState.resource)) {
```

aljesusg (approved):
onlye one minor comment. no block, LGFM

lucasponce:
I guess I may ask a change in the feature.

Filter is single value:
image

But as it happens with app/workload/service lists, multi-value would help a lot:
image

For example, I'd like to filter by port 8080 or 9090:
image

Otherwise I'd see the filter a little bit limited (I guess it's a minor change in the filter code, component is prepared for this).

lucasponce (changes requested):
A small change in the behaviour of the filter.

xeviknal:
@lucasponce comments approached. The behavior now is like the workloads/apps/services/istio lists. Applies OR for the same filter and AND between different filters.

lucasponce (approved):

PR #2053, opened 2021-01-14
Support Wizards with one workload without version label

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3483

Kiali Wizards expect workloads labeled with "app" and "version" labels.

image

It seems that "version" label tends to not be used in scenarios where there is a single workload for a service, then Fault Injection or Request Timeout scenarios can't be used.

This PR enables that use case.

For QE:
- Test potential regression in Service -> Actions scenario.
- Check that a Workload with "app" but NO "version" can be supported.

Thanks !


hhovsepy (approved):
Verified, regression tested.

PR #2054, opened 2021-01-18
Fix ini resolution dep

lucasponce:
Fixes the ini version but adding it on the "resolution" package.json section.

This takes precended over https://github.com/kiali/kiali-ui/pull/2036

For QE:
- Just a simple regression test

### DIFF
--- a/package.json
+++ b/package.json
@@ -131,6 +131,7 @@
     "ajv": "6.12.3",
     "dot-prop": "4.2.1",
     "hoist-non-react-statics": "3.3.0",
+    "ini": "1.3.7",
     "kind-of": "6.0.3",
     "victory": "34.3.12",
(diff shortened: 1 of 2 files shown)


jotak:
just out of curiosity, why is it better to add it to resolutions? the bot's PR would work as well, no?

jotak (approved):

lucasponce:
> just out of curiosity, why is it better to add it to resolutions? the bot's PR would work as well, no?

I learnt that it gives more flexibility if needed [1], for this case probably both works, but last time I bumped something like this I was recommended to use it on this way.

[1] https://classic.yarnpkg.com/en/docs/selective-version-resolutions/

prachiyadav (approved):

PR #2056, opened 2021-01-19
Allow custom health configuration to be deployed as annotations

aljesusg:
Fix https://github.com/kiali/kiali/issues/3332

# Tasks

- [X] Support health views
- [X]  Support for health in graph

For workload and services

Require: https://github.com/kiali/kiali/pull/3362

Step 1 of https://github.com/kiali/kiali/issues/3332

### DIFF
--- a/src/config/HealthConfig.ts
+++ b/src/config/HealthConfig.ts
@@ -38,3 +38,4 @@ const replaceXCode = (value: string): string => {
 */
 export const allMatchTEST = allMatch;
 export const getExprTEST = getExpr;
+export const replaceXCodeTEST = replaceXCode;
--- a/src/config/HealthConfig.ts
+++ b/src/config/HealthConfig.ts
@@ -2,6 +2,9 @@ import { HealthConfig, RegexConfig } from '../types/ServerConfig';
 
 const allMatch = new RegExp('.*');
 
+/*
+ Parse configuration from backend format to regex expression
+*/
 export const parseHealthConfig = (healthConfig: HealthConfig) => {
   for (let [key, r] of Object.entries(healthConfig.rate)) {
     healthConfig.rate[key].namespace = getExpr(healthConfig.rate[key].namespace);
@@ -16,11 +19,14 @@ export const parseHealthConfig = (healthConfig: HealthConfig) => {
   return healthConfig;
 };
 
-export const getExpr = (value: RegexConfig | undefined, code: boolean = false): RegExp => {
+/*
+  Convert the string to regex, if isCode is true then call to replaceXCode to change the X|x in code expression to \d
+*/
+export const getExpr = (value: RegexConfig | undefined, isCode: boolean = false): RegExp => {
   if (value) {
     if (typeof value === 'string' && value !== '') {
       const v = value.replace('\\\\', '\\');
-      return new RegExp(code ? replaceXCode(v) : v);
+      return new RegExp(isCode ? replaceXCode(v) : v);
     }
     if (typeof value === 'object' && value.toString() !== '/(?:)/') {
       return value;
@@ -29,6 +35,9 @@ export const getExpr = (value: RegexConfig | undefined, code: boolean = false): 
   return allMatch;
 };
 
+/* Replace x|X by the regular expression
+   Example: 4XX or 5XX to 4\d\d 5\d\d
+*/
 const replaceXCode = (value: string): string => {
   return value.replace(/x|X/g, '\\d');
 };
--- a/src/config/HealthConfig.ts
+++ b/src/config/HealthConfig.ts
@@ -8,7 +8,7 @@ export const parseHealthConfig = (healthConfig: HealthConfig) => {
     healthConfig.rate[key].name = getExpr(healthConfig.rate[key].name);
     healthConfig.rate[key].kind = getExpr(healthConfig.rate[key].kind);
     for (let t of Object.values(r.tolerance)) {
-      t.code = getExpr(t.code);
+      t.code = getExpr(t.code, true);
       t.direction = getExpr(t.direction);
       t.protocol = getExpr(t.protocol);
     }
@@ -16,16 +16,21 @@ export const parseHealthConfig = (healthConfig: HealthConfig) => {
   return healthConfig;
 };
 
-export const getExpr = (value: RegexConfig | undefined): RegExp => {
+export const getExpr = (value: RegexConfig | undefined, code: boolean = false): RegExp => {
   if (value) {
     if (typeof value === 'string' && value !== '') {
-      return new RegExp(value.replace('\\\\', '\\'));
+      const v = value.replace('\\\\', '\\');
+      return new RegExp(code ? replaceXCode(v) : v);
     }
     if (typeof value === 'object' && value.toString() !== '/(?:)/') {
       return value;
     }
   }
   return allMatch;
+};
+
+const replaceXCode = (value: string): string => {
+  return value.replace(/x|X/g, '\\d');
 };
 
 /*
--- a/src/types/ErrorRate/__tests__/GraphEdgeStatus.test.ts
+++ b/src/types/ErrorRate/__tests__/GraphEdgeStatus.test.ts
@@ -1,0 +1,19 @@
+import { setServerConfig } from '../../../config/ServerConfig';
+import { serverRateConfig } from '../__testData__/ErrorRateConfig';
+import { getTotalRequest } from '../GraphEdgeStatus';
+import { Responses } from '../../Graph';
+
+describe('getRateHealthConfig', () => {
+  beforeAll(() => {
+    setServerConfig(serverRateConfig);
+  });
+
+  it('getTotalRequest', () => {
+    // Check with annotation to set 400
+    const resp: Responses = {
+      '200': { hosts: {}, flags: { '-': '3', XX: '4', YYY: '2' } },
+      '400': { hosts: {}, flags: { '-': '1', XX: '1', YYY: '6' } }
+    };
+    expect(getTotalRequest(resp)).toBe(17);
+  });
+});
--- a/src/types/ErrorRate/__tests__/GraphEdgeStatus.test.ts
+++ b/src/types/ErrorRate/__tests__/GraphEdgeStatus.test.ts
@@ -1,7 +1,26 @@
 import { setServerConfig } from '../../../config/ServerConfig';
 import { serverRateConfig } from '../__testData__/ErrorRateConfig';
-import { getTotalRequest } from '../GraphEdgeStatus';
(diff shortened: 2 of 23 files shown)


aljesusg:
One open question @lucasponce 

Example Alpha namespace item x-server.

I have the configuration

```yaml
- namespace: "alpha"
      tolerance:
        - code: "[45]\\d[^\\D4]"
          protocol: "http"
          failure: 40
```

And I add a annotation to x-server service

```bash
curl --request PATCH http://localhost:8000/api/namespaces/alpha/services/x-server --data '{"metadata":{"annotations":{"health.kiali.io/rate":"4XX,20,60,http,inbound|outbound"}}}'
```

**So now for 4XX codes is degraded when rate is >=20 and failure in >=60 and the result is the next image**

Screenshot from 2021-01-19 17-42-29

The point here is the edge calculation where we get the configuration of source and target, in this case b-client has the _default_ and x-serve the _new annotation_ so:

-  in source the status is failure by default configuration
-  in target the status is degraded by the new annotation

That means that the edge is failure by the source configuration...in this case where source and target is degraded this could be confusing even though one thing is the status of the node and another of the traffic.

lucasponce:
Thanks for the PR @aljesusg, I'll fetch it locally today to provide feedback.

The scenario you describe is interesting, perhaps that triggers more corner cases and edges about the feature itself (i.e. does it make sense that a Service and a Workload has different health config settings, or if when it's present in the service, that should apply to the workloads as well?).

Again, as I commented in the previous PRs, it's "health config" default should be good for most of the cases, and I think that we won't be able to deal with all cornercases, but the user should know to understand that a particular configuration is intentional.

In any case, that was just a general comment, I'll fetch and test and I'll ellaborate a better/deep set of comments around this.

aljesusg:
> Thanks for the PR @aljesusg, I'll fetch it locally today to provide feedback.
> 
> The scenario you describe is interesting, perhaps that triggers more corner cases and edges about the feature itself (i.e. does it make sense that a Service and a Workload has different health config settings, or if when it's present in the service, that should apply to the workloads as well?).

on the other hand, the source node status is calculate in this case by the outbound to the x-server and the other server...ant the status is degraded...this is different to the edge status who is calculating the traffic from this source to x-server only. So it could make sense. I think that is right the calculation


> 
> Again, as I commented in the previous PRs, it's "health config" default should be good for most of the cases, and I think that we won't be able to deal with all cornercases, but the user should know to understand that a particular configuration is intentional.
> 
> In any case, that was just a general comment, I'll fetch and test and I'll ellaborate a better/deep set of comments around this.

lucasponce:
A first error:
image

The edges doesn't honor the health config.

For that example I didn't test the annotations or I didn't have any specific on my health config, just with defaults, but edges seem incorrect nor consistent.

lucasponce:
I can't see tests for health annotation processing.

Not for the user, but for other developers, it would be beneficial to add them to help what is the expected format.

lucasponce:
A fist example doesn't work:

```
curl --request PATCH http://localhost:8000/api/namespaces/alpha/services/w-server --data '{"metadata":{"annotations":{"health.kiali.io/rate":"[45]XX,60,80,http,inbound|outbound"}}}'
```

image

lucasponce (changes requested):
In order to progress this and the backend https://github.com/kiali/kiali/pull/3362 one I'd suggest:

- Add better testing on the regex used in the annotations, some clear tests about what expressions are expected are good for the reviewer, there is a lack of doc in that area.
- A couple of expressions didn't work but I tested others and I saw that the normal flow worked as expected.

In order to reduce the scope the problem, if @jshaughn and @hhovsepy can test this and the backend together my plan is to consolidate and work in the next high level scenario, which is improve how the user knows which health config is applied.

For that I'll propose a brainstorming session, as we need to tackle scenario globally.

lucasponce:
Another problem or scenario to test:
image

If I change the health in the workloads v-server and w-server, I understand that the edge shouldn't take the default namespace one.

jshaughn (approved):
Worked for my test case.  As mentioned in the other PR, we need some updated docs to go with this nice feature.

lucasponce:
What is the status of this work ?

aljesusg:
Adding tests but it should work now with all cases, just adding test

lucasponce:
> Adding tests but it should work now with all cases, just adding test

Are you sure ? there is no new commit after my last comment reporting a scenario that it doesn't work.

aljesusg:
I have changes in local let me finish the tests please and let me analyze all of them. I couldn't replicate your scenario the other day, I'll try it today

aljesusg:
ok @lucasponce I found the error in the calculation for the edge thanks for the patience

lucasponce:
>  I found the error in the calculation for the edge thanks for the patience

I don't see the edges fixed:

image

image

image

So in all graph types services and workloads' health seem correct, but edges are calculated in a different way, so it's not consistent.

aljesusg:
> > I found the error in the calculation for the edge thanks for the patience
> 
> I don't see the edges fixed:
> 
> image
> 
> image
> 
> image
> 
> So in all graph types services and workloads' health seem correct, but edges are calculated in a different way, so it's not consistent.

Please share with me the annotations that you set in the source/dear nodes

lucasponce:
Updated on the original comment.

P.S.: I've edited previous one to not repeat the images on the response.

aljesusg:
I did some changes in the code I was working with a old backend build after last changes in naming

> > I found the error in the calculation for the edge thanks for the patience
> 
> I don't see the edges fixed:
> 
> image
> 

For the first case. Services:  

v-server use default configuration

The status for requests outbound from v-server is FAILURE

w-server use the annotation {health.kiali.io/rate: "[45]XX,60,80,http,inbound|outbound"}

The status for requests inbound from w-server is HEALTHY


So the edge is FAILURE...Another thing is why is not with FAILURE status. I need to check it

> image
> 

In the workload view (Now is green after last changes)

v-server and w-server use {health.kiali.io/rate: "5XX,60,80,http,inbound|outbound;4XX,60,80,http,inbound|outbound"}

v-server outbound is HEALTHY and w-server inbound is HEALTHY

> image
> 

For this 

They use the defaults.

Outbound v-server and  Inbound w-server are FAILURE both so edge is correctly.

Here again is seems that there is a bug in the node health I am going to check but the edge calculation seems correct.


If you check the side panel information requests there are 400 error codes and should be failure

Screenshot from 2021-02-09 17-11-41


> So in all graph types services and workloads' health seem correct, but edges are calculated in a different way, so it's not consistent.
> 
> Update, annotations used:
> 
> ```
> curl --request PATCH http://localhost:8000/api/namespaces/alpha/services/w-server --data '{"metadata":{"annotations":{"health.kiali.io/rate":"[45]XX,60,80,http,inbound|outbound"}}}'
> 
> curl --request PATCH http://localhost:8000/api/namespaces/alpha/workloads/w-server --data '{"metadata":{"annotations":{"health.kiali.io/rate":"5XX,60,80,http,inbound|outbound;4XX,60,80,http,inbound|outbound"}}}'
> 
> curl --request PATCH http://localhost:8000/api/namespaces/alpha/workloads/v-server --data '{"metadata":{"annotations":{"health.kiali.io/rate":"5XX,60,80,http,inbound|outbound;4XX,60,80,http,inbound|outbound"}}}'
> ```



> > I found the error in the calculation for the edge thanks for the patience
> 
> I don't see the edges fixed:
> 
> image
> 
> image
> 
> image
> 
> So in all graph types services and workloads' health seem correct, but edges are calculated in a different way, so it's not consistent.
> 
> Update, annotations used:
> 
> ```
> curl --request PATCH http://localhost:8000/api/namespaces/alpha/services/w-server --data '{"metadata":{"annotations":{"health.kiali.io/rate":"[45]XX,60,80,http,inbound|outbound"}}}'
> 
> curl --request PATCH http://localhost:8000/api/namespaces/alpha/workloads/w-server --data '{"metadata":{"annotations":{"health.kiali.io/rate":"5XX,60,80,http,inbound|outbound;4XX,60,80,http,inbound|outbound"}}}'
> 
> curl --request PATCH http://localhost:8000/api/namespaces/alpha/workloads/v-server --data '{"metadata":{"annotations":{"health.kiali.io/rate":"5XX,60,80,http,inbound|outbound;4XX,60,80,http,inbound|outbound"}}}'
> ```

src/config/HealthConfig.ts
```diff
@@ -8,18 +8,19 @@ export const parseHealthConfig = (healthConfig: HealthConfig) => {
...
+      t.code = getExpr(t.code, true);
       t.direction = getExpr(t.direction);
       t.protocol = getExpr(t.protocol);
     }
   }
   return healthConfig;
 };
 
-export const getExpr = (value: RegexConfig | undefined): RegExp => {
+export const getExpr = (value: RegexConfig | undefined, code: boolean = false): RegExp => {
```
lucasponce, line 19: I'm thinking if I need to maintain this function, I'd appreciate a litle bit of more context.

Suggestion:
```
export const getExpr = (value: RegexConfig | undefined, code: boolean = false)
```
->
```
export const getHealthRegExp = (value: RegexConfig | undefined, isCode: boolean = false)
```

Not a blocker, but it helped me to follow the UI changes.

src/config/HealthConfig.ts
```diff
@@ -28,8 +29,13 @@ export const getExpr = (value: RegexConfig | undefined): RegExp => {
   return allMatch;
 };
 
+const replaceXCode = (value: string): string => {
```
lucasponce, line 32: Please, just add a minor comment to help with the context something like

"4XX and 5XX expressions are allowed in the code field"

src/types/ErrorRate/__tests__/GraphEdgeStatus.test.ts
```diff
@@ -0,0 +1,141 @@
...
+
+  it('getTotalRequest', () => {
+    // Check with annotation to set 400
+    const resp: Responses = {
+      '200': { hosts: {}, flags: { '-': '3', XX: '4', YYY: '2' } },
+      '400': { hosts: {}, flags: { '-': '1', XX: '1', YYY: '6' } }
+    };
+    expect(getTotalRequest(resp)).toBe(17);
+  });
+  describe('calculateStatusGraph', () => {
```
lucasponce, line 38: These tests are good.
Perhaps we'll add more in the future as I can find some cases that worth to add.
But good to have them, otherwise it's hard to review the GraphEdgeStatus and validate if there is anything missing.

lucasponce (approved):
I'm ok to consolidate, with latest changes I didn't find any regression and all cases I've tested seems to work.

Health config is a big area and there are cornercases and scenarios that we need to discuss, but that's out of the scope of this work.

Thanks @aljesusg, I know this feature is not an easy one.

hhovsepy:
Post merge verification.
Screenshot from 2021-02-15 10-28-57
Screenshot from 2021-02-15 10-28-47

PR #2059, opened 2021-01-21
Show Mesh nav entry only when needed

israel-hdez:
The back-end config endpoint should be reporting cluster information only when multi-cluster is enabled in Istio. Since the "Mesh" nav entry is a page related with multi-cluster, this should show it only when needed.

Related to kiali/kiali#3592
Back-end PR: kiali/kiali#3607


jshaughn (approved):
LOK

PR #2063, opened 2021-01-26
Envoy dump tables - linking services, destination rules, and virtual services to its detail page

xeviknal:
This issue belongs to the epic https://github.com/kiali/kiali/issues/3346
Needs https://github.com/kiali/kiali/pull/3626

The idea is to easy the task to inspect entities that kiali has detailed information: services, dr, vs.

proxy-linking


hhovsepy (approved):
Verified:
Screenshot from 2021-01-26 17-49-16
Screenshot from 2021-01-26 17-51-58

xeviknal:
@hhovsepy I see that the FQDN entries in the clusters table don't have links. They should be there. See that in the video in the description. The links should be present also in the Routes table. Do you mind re-checking this please?

hhovsepy:
It shows links partly:
Screenshot from 2021-01-26 18-35-06
Screenshot from 2021-01-26 18-37-06

xeviknal:
oh good! But actually, you make me realize that the Destination Rules and Virtual services may link to an unexistent pages, right?
I understand that you have let access to bookinfo2/3 but not to bookinfo into this kiali instance. hence, the services living in bookinfo are not linked. However, DR and VR living in the bookinfo, shouldn't be linked right? Or are they? If you click in the DR `details.bookinfo` does the link work?

hhovsepy:
Actually all bookinfo namespaces are accessible to kiali, and all services there are linkable.

lucasponce:
A first feature check shows ok:
image

image

I will add more comments at code level.

src/components/Envoy/EnvoyModal.tsx
```diff
@@ -42,7 +46,11 @@ const toolbarSpace = style({
   marginLeft: '1em'
 });
 
-type EnvoyDetailProps = {
+type ReduxProps = {
+  namespaces: Namespace[] | undefined;
```
lucasponce, line 50: I don't think "undefined" is necessary.
When you connect the component with Redux to sync with namespaces you'll get always a value (or an empty array).

At least other components that are connected to the "namespaces" prop doesn't add the undefined.

src/components/Envoy/tables/ClusterTable.ts
```diff
@@ -1,24 +1,24 @@
...
 
 export class ClusterTable implements SummaryTable {
   summaries: ClusterSummary[];
   sortingIndex: number;
   sortingDirection: 'asc' | 'desc';
+  namespaces: Namespace[] | undefined;
+  namespace: string;
 
-  constructor(summaries: ClusterSummary[], sortBy: ISortBy) {
+  constructor(summaries: ClusterSummary[], sortBy: ISortBy, namespaces: Namespace[] | undefined, namespace: string) {
```
lucasponce, line 16: Not sure if that "undefined" is necessary or if there is a case where namespace array is undefined instead of an empty one.

Anyway, it's minor.

I saw that in other places as well.

hhovsepy (approved):
Verified now:
Screenshot from 2021-01-27 11-12-48
Screenshot from 2021-01-27 11-14-11

lucasponce:
I left some comments in the backend/frontend code.
No blockers.

src/components/Envoy/tables/ClusterTable.ts
xeviknal: Approached

lucasponce (approved):

PR #2064, opened 2021-01-26
[WIP] Server-side UI Defaults

cfcosta:
Part of https://github.com/kiali/kiali/issues/3371.

This PR implements some UI defaults coming from the server CR configuration. Currently, this supports:

- [x] Namespace selector on the graph
- [ ] Time range for traffic metrics per refresh
- [ ] Graph automatic refresh duration

### DIFF
--- a/src/actions/NamespaceThunkActions.ts
+++ b/src/actions/NamespaceThunkActions.ts
@@ -3,6 +3,8 @@ import { KialiAppState } from '../store/Store';
 import * as Api from '../services/Api';
 import { KialiAppAction } from './KialiAppAction';
 import { NamespaceActions } from './NamespaceAction';
+import { serverConfig } from 'config';
+import Namespace from 'types/Namespace';
 
 const shouldFetchNamespaces = (state: KialiAppState) => {
   if (!state) {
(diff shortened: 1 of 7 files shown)


jshaughn (commented):
@cfcosta I'm not sure but I think maybe the approach here on the ui-side is overcomplicated.  I was thinking the UIDefaults would immediately be set in redux when the user logs in, and then you could basically forget about them.  So maybe we don't need to make changes to components like NamespaceDropdown, as they should, I think, just drive off of redux.  I haven't run this so maybe I'm missing something.  But I originally thought we'd be able to set the defaults in redux when we pull server config in AuthenticationController.doPostLoginActions().  Take a look there and see if it makes sense.

cfcosta:
@jshaughn I actually agree, I tried doing exactly that in the beginning but it was really unstable, sometimes it would get just the default server config instead of the one fetched from the CR. I'll try it again with a clean mind, maybe that will work better.

jshaughn:
> sometimes it would get just the  default server config instead of the one fetched from the CR

Hmm, that sounds like a server-side issue.  Perhaps the server side PR was not deployed at the time?  I think the UI defaults should be applied at login time as suggested above, and we need to assume the server will return the proper settings.

jshaughn:
@cfcosta Closing this WIP in favor of https://github.com/kiali/kiali-ui/pull/2081.  This PR takes the approach I suggested above, and it seems to work fine.

PR #2066, opened 2021-01-27
Redesign tracing page toolbar

jotak:
- Toolbar entirely grouped into a single dropdown, like in graph
- Some removed fields: trace intervals, status code
- Some new fields: percentiles, show spans average
- New component TracesDisplayOptions
- Some styles from GraphSettings are moved to a common place and made
  accessible to TracesDisplayOptions
- removed now useless function, like buildTags, which btw had a lib
  dependency that could be removed as well

Fixes https://github.com/kiali/kiali/issues/3604

Capture d’écran de 2021-01-27 15-39-22


src/components/JaegerIntegration/TracesComponent.tsx
```diff
@@ -153,130 +99,61 @@ class TracesComponent extends React.Component<TracesProps, TracesState> {
...
+    const tags = this.getTags();
     if (tags) {
       url += `&tags=${tags}`;
     }
     return url;
   };
 
-  private handleStatusCode = (value: string) => {
+  private onQuerySettingsChanged = (settings: QuerySettings) => {
+    console.log(settings);
```
jotak, line 145: oops; // FIXME / REMOVE

src/components/JaegerIntegration/TracesComponent.tsx
```diff
@@ -153,130 +99,61 @@ class TracesComponent extends React.Component<TracesProps, TracesState> {
...
-
-  private filterTraces = (): JaegerTrace[] => {
-    if (!this.state.selectedIntervalDuration) {
-      return this.state.traces;
-    }
-    const min = this.state.selectedIntervalDuration.min * this.state.selectedIntervalDuration.scale;
-    const max = this.state.selectedIntervalDuration.max * this.state.selectedIntervalDuration.scale;
-    return this.state.traces.filter(trace => trace.duration >= min && trace.duration <= max);
+  private onDisplaySettingsChanged = (settings: DisplaySettings) => {
+    console.log(settings);
```
jotak, line 151: oops; // FIXME / REMOVE

jotak:
Update: new menu:
Capture d’écran de 2021-01-28 16-01-27

- Percentile values are displayed for reference
- Removed Time axis options as they're aren't very relevant anymore

jotak:
PS: don't hesitate to comment on texts / labels also

hhovsepy (commented):
Verified the functionality.
Just line up of checkbox and values will look better.

jotak:
Thanks @hhovsepy 

> Just line up of checkbox and values will look better.

Not sure if you mean to add a separator line between option groups, in case yes perhaps that would be better, however I've copied the design from the graph display dropdown, where there isn't anything like that

hhovsepy:
I mean the label text of checkbox or radiobutton are a bit higher:
Screenshot from 2021-01-29 10-37-00

jotak:
@jshaughn can you have a quick look at my last commit https://github.com/kiali/kiali-ui/pull/2066/commits/7a1b49ff46d549bbf83be5f859cb71aca0bbddd7 ?
Since I'm reusing here styles from the graph, this is affecting graph code (but should have no visible effect). I'm moving a couple of CSS style to typestyle.

jotak:
> I mean the label text of checkbox or radiobutton are a bit higher:

I see, it should be fixed now

hhovsepy (approved):
Verified now:
Screenshot from 2021-01-29 12-31-00

jshaughn (changes requested):
> @jshaughn can you have a quick look at my last commit 7a1b49f ?

@jotak,  I don't think the CSS change you pointed out in that commit is an issue, the alignment of the dropdown margin for the checkbox or radio elements looks unchanged.  But there does seem to be a regression in the graph Display dropdown, I think due to a margin change around the info icons:

image

Before:
image


After:

image

jotak:
@jshaughn good catch! fixed

lucasponce:
A first comment, I see this new toaster when clicking in a trace from the graph:
image

It's odd as I guess I didn't click anything special so I don't see need for this warning.

lucasponce:
Minor, I think the help icons could follow same pattern used in Graph Display:
image

And to be placed on value and at right:
image

lucasponce:
I see a problem in the graph, the versioned app graph doesn't work well on my local env.
I'm going to test master and check if the problem is still there.

package.json
```diff
@@ -75,7 +75,6 @@
     "js-yaml": "3.13.1",
     "json-beautify": "1.0.1",
     "lodash": "4.17.20",
-    "logfmt": "1.2.1",
```
lucasponce, line 78: unused ?

jotak:
> Minor, I think the help icons could follow same pattern used in Graph Display:
> And to be placed on value and at right:

It can definitely be placed to the right, but it's meant to give information on the group of options, not each single value.

jotak:
> A first comment, I see this new toaster when clicking in a trace from the graph:

Good catch, I didn't notice, it's because it's fetching inbound+outbound percentiles ... but only inbound queries is allowed on services (so you wouldn't see this warning on workloads or apps traces).
Anyway, it's now fixed for services.

src/components/JaegerIntegration/TracesComponent.tsx
```diff
@@ -379,8 +262,9 @@ class TracesComponent extends React.Component<TracesProps, TracesState> {
                     </Toolbar>
                   </RenderHeader>
                   <JaegerScatter
-                    fixedTime={!this.state.adjustTime}
-                    traces={this.filterTraces()}
+                    fixedTime={false}
```
lucasponce, line 265: I guess only fixedTime was used for Jaeger I think.
Perhaps then that parameter could be removed, but it's fine.

lucasponce (approved):
Tested locally and code looks ok to me.
I didn't see any blockers.
I'll wait that @jshaughn is ok with it.

aljesusg (approved):
LGFM, sorry I had it in pending :( tested it !!!

jshaughn (approved):
LGTM!

src/components/JaegerIntegration/TracesComponent.tsx
jotak: +1, I've removed it

jotak:
Thanks all

PR #2067, opened 2021-01-27
Fix BoxLayout issue

jshaughn:
Fix BoxLayout issue with synthetic edge generation. It can no longer assume the box type of a parent node.

Also, some minor cleanup and ensure that all layout type props are applied at all times.

fixes https://github.com/kiali/kiali/issues/3623


lucasponce (approved):

mattmahoneyrh:
Verified fixed.

PR #2071, opened 2021-01-28
Prevent duplicate graph fetch on double-click

jshaughn:
We don't need to both update the URL and update redux on a double-click. Pushing a new node graph URL will already cause a redux update when the page initializes. So, remove the redundancy and prevent two calls to the backend.

Furthermore, fix an issue on operation node double-click by properly setting the target node params.

Fixes https://github.com/kiali/kiali/issues/3649

To test:
Double-click of a node should still properly generate a node detail graph.  The user should still be able to change the target node within the detail graph by double clicking a different node.

The double-click should generate only a single backend call for graph generation, not two.

### DIFF
--- a/src/pages/Graph/GraphPage.tsx
+++ b/src/pages/Graph/GraphPage.tsx
@@ -67,6 +67,8 @@ import { JaegerThunkActions } from 'actions/JaegerThunkActions';
 
 // GraphURLPathProps holds path variable values.  Currenly all path variables are relevant only to a node graph
 type GraphURLPathProps = {
+  aggregate: string;
+  aggregateValue: string;
   app: string;
   namespace: string;
   service: string;
@@ -199,6 +201,10 @@ export class GraphPage extends React.Component<GraphPageProps, GraphPageState> {
   private graphDataSource: GraphDataSource;
 
   static getNodeParamsFromProps(props: RouteComponentProps<Partial<GraphURLPathProps>>): NodeParamsType | undefined {
+    const aggregate = props.match.params.aggregate;
+    const aggregateOk = aggregate && aggregate !== UNKNOWN && aggregate !== 'undefined';
+    const aggregateValue = props.match.params.aggregateValue;
+    const aggregateValueOk = aggregateValue && aggregateValue !== UNKNOWN && aggregateValue !== 'undefined';
     const app = props.match.params.app;
     const appOk = app && app !== UNKNOWN && app !== 'undefined';
     const namespace = props.match.params.namespace;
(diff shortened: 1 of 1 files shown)


src/pages/Graph/GraphPage.tsx
```diff
@@ -199,6 +201,10 @@ export class GraphPage extends React.Component<GraphPageProps, GraphPageState> {
   private graphDataSource: GraphDataSource;
 
   static getNodeParamsFromProps(props: RouteComponentProps<Partial<GraphURLPathProps>>): NodeParamsType | undefined {
+    const aggregate = props.match.params.aggregate;
+    const aggregateOk = aggregate && aggregate !== UNKNOWN && aggregate !== 'undefined';
```
israel-hdez, line 205: I see `aggregate !== 'undefined'`. Is it true that `aggregate` can be the `"undefined"` string? or are you testing if it's an undefined variable?

If it's the latter, I think the very first test is already checking for `undefined` variable (not string).

src/pages/Graph/GraphPage.tsx
```diff
@@ -199,6 +201,10 @@ export class GraphPage extends React.Component<GraphPageProps, GraphPageState> {
...
+    const aggregateValue = props.match.params.aggregateValue;
+    const aggregateValueOk = aggregateValue && aggregateValue !== UNKNOWN && aggregateValue !== 'undefined';
```
israel-hdez, line 207: Same `undefined` comment.

israel-hdez (approved):
I no longer see the duplicated graph fetch.
Very hidden issue.

I couldn't test the operations node issue, but the changes to `getNodeParamsFromProps` make sense.

I left two notes. Not blockers.

src/pages/Graph/GraphPage.tsx
jshaughn: good point, should not be testing a string.

src/pages/Graph/GraphPage.tsx
jshaughn: yes, I'll fix. Thanks!

src/pages/Graph/GraphPage.tsx
jshaughn: @israel-hdez I removed the invalid tests here and in a few other places where the same problem had been originally.

jshaughn:
@mattmahoneyrh Not sure if you want to test this or not.  It's a code-only fix.  If you want to test just ensure you see no regressions in node double-click to drill into the node detail graph.  Otherwise, just apply the rubber stamp. Thanks.

mattmahoneyrh (approved):

PR #2075, opened 2021-02-02
If the login expired, we must remove the auth cookie since it is no longer valid

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/3643

### DIFF
--- a/src/pages/Login/LoginPage.tsx
+++ b/src/pages/Login/LoginPage.tsx
@@ -127,7 +127,7 @@ export class LoginPage extends React.Component<LoginProps, LoginState> {
     }
     if (this.props.status === LoginStatus.expired) {
       messages.push(
-        this.renderMessage('Your session has expired or was terminated in another window.', 'warning', 'sessionExpired')
+        this.renderMessage('Your session has expired or was terminated in another window!', 'warning', 'sessionExpired')
       );
     }
     if (this.props.status === LoginStatus.error) {
@@ -162,6 +162,12 @@ export class LoginPage extends React.Component<LoginProps, LoginState> {
         </ListItem>
       </>
     );
+
+    if (this.props.status === LoginStatus.expired) {
+      // we must clean out the token cookie because it is no longer valid
+      document.cookie =
+        'kiali-token=; Expires=Thu, 01 Jan 1970 00:00:00 GMT; HttpOnly; MaxAge=-1; SameSite=Strict; Path=' + '/';
+    }
 
     let loginPane: React.ReactFragment;
     if (authenticationConfig.strategy === AuthStrategy.token) {


src/pages/Login/LoginPage.tsx
```diff
@@ -163,6 +163,11 @@ export class LoginPage extends React.Component<LoginProps, LoginState> {
       </>
     );
 
+    if (this.props.status === LoginStatus.expired) {
+      // we must clean out the token cookie because it is no longer valid
+      document.cookie = 'kiali-token=; Max-Age=0; path=/';
```
jmazzitelli, line 168: Note - this doesn't work - maybe someone with more expertise in this area can tell me why (or better yet, how to get it to delete this cookie).

jmazzitelli: I would have liked to do something like this here, but apparently, this is a read-only property and you can't change it:

```
this.props.authenticated = false;
```

jshaughn: mazz, try Max=Age -1

jmazzitelli: putting this here

```
this.setState({ isValidToken: false });
```
 blows up the UI - after it detects the session is invalid, the browser screen turns white and I get some error in the console that eventually involves this error "Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops."

jmazzitelli: Just to continue logging what I've tried - I tried setting the isValidToken to false in the handleSubmit function - this didn't work either:
```
  handleSubmit = (e: any) => {
    e.preventDefault();

    if (isAuthStrategyOAuth()) {
      if (this.props.status === LoginStatus.expired) {  // <---- added this if-stmt
        this.setState({ isValidToken: false });
      }
      // If we are using OpenShift or OpenId strategy, take the user back to the authorization endpoint
      window.location.href = authenticationConfig.authorizationEndpoint!;
```

jmazzitelli: > try Max=Age -1

tried..did not work.

jshaughn: Here are the fill set of attrs we use in the server code when deleting the kiali cookie:
```
Name:     cookieName,
Value:    "",
Expires:  time.Unix(0, 0),
HttpOnly: true,
MaxAge:   -1,
Path:     conf.Server.WebRoot,
SameSite: http.SameSiteStrictMode,
```
I would suggest just setting all of these, even if they are all not needed.  'Expires', in particular, may be needed.

jmazzitelli: That's the Set-Cookie header settings - I'm not sure some of these are used for actual Cookie values (like HttpOnly and SameSite). I'll have to look at what `document.cookie` values can be (I thought they were actual Cookie headers).

jmazzitelli: The latest code now in the PR still doesn't work:

      document.cookie =  'kiali-token=; Expires=Thu, 01 Jan 1970 00:00:00 GMT; HttpOnly; MaxAge=-1; SameSite=Strict; Path=' + '/';

jmazzitelli: Need @israel-hdez to look at this - we need to somehow force the Kiali UI to make that call to `api/authenticate` but it is not. Somehow we need state.authenticated to be set to false, or state.isValidToken set to false and have it handle that (i.e. it needs to get the new token via a new cookie that api/authenticate returns).

jmazzitelli: I may be thinking of this problem all wrong - maybe we should be clearing the token server-side when the server returns that 401. That might be the better way to do this. I'll experiment with that.

jmazzitelli:
I'm closing this. I got some success in my server-side experiment. I'm going to create a PR to server-side to try to fix this.

PR #2078, opened 2021-02-09
Fix graph replay regression introduced when removing code that was ca…

jshaughn:
Fix graph replay regression introduced when removing code that was causing duplicate layouts (in what was thought to be all cases).  But we still need to ensure an initial layout is performed for any new cy graph, even if the elements are exactly the same as before, which can happen for graph replay (although, the fix is not specific to graph replay just in case there is another scenario not considered).  The fix still maintains the initial goal of avoiding duplicate/unnecessary layouts.

Fixes https://github.com/kiali/kiali/issues/3684


lucasponce:
It seems fixed
image

lucasponce (approved):

pbajjuri20:
this issue is fixed perfectly

before PR:

before_replay_graph


after PR:

after_replay_graph

pbajjuri20 (approved):

PR #2080, opened 2021-02-09
Update envoy response flags that we know about

jshaughn:
Update envoy response flags that we know about for the Flags tab in graph side-panel.

Closes https://github.com/kiali/kiali/issues/3685

### DIFF
--- a/src/utils/ResponseFlags.ts
+++ b/src/utils/ResponseFlags.ts
@@ -1,4 +1,4 @@
-// The Envoy response flags can be found here:
+// The Envoy response flags can be found here (search for %RESPONSE_FLAGS% on the page):
 // https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log/usage
 const responseFlags = {
   DC: { code: '500', help: 'Downstream connection termination', short: 'Downstream cx term' },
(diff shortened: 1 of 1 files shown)


pbajjuri20:
Tested and passed on Kiali UI 1.30.0 (b5fef87824ec670d43871d9a9874489f2cd7df94)

pbajjuri20 (approved):

src/utils/ResponseFlags.ts
```diff
@@ -1,15 +1,20 @@
-// The Envoy flags can be found here:
-// https://github.com/envoyproxy/envoy/blob/master/source/common/stream_info/utility.cc
+// The Envoy response flags can be found here:
+// https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log/usage
```
jmazzitelli, line 2: I don't see how this link is relevant, compared to the old one. The old utility.cc I can see the 2-char codes.. but I don't see those in the new link.

jmazzitelli: Oh, I see.. %RESPONSE_FLAGS%... should probably mention that in the comments since it is half-way down on the page.

jmazzitelli (approved):
I think the comment should be alittle more descriptive, but that's trivial ... approving..

jmazzitelli (approved):

PR #2081, opened 2021-02-09
UI support for UI defaults

jshaughn:
Apply UI Defaults as soon as the user successfully authenticates. We immediately update the Redux store with the user-configured defaults.  Note that they can be overwritten almost immediately by anything set/bookmarked in the URL.
These defaults apply only for "vanilla" URLs.

Notes:
- We needed to now fetch the namespaces at auth time, before applying the
  defaults, to ensure that any default namespaces were valid.
- Small unrelated change to NamespaceDropdown for clarity, just renaming a method.

Fixes https://github.com/kiali/kiali/issues/3371

Requires Server PR: https://github.com/kiali/kiali/pull/3688

TESTING NOTE
When testing this remember that the UI Defaults are set in the Kiali CR.  And changes won't take effect until the Kiali pod restarts.  Furthermore, the configured UI defaults have less priority than URL settings, so you need to test with vanilla URLs, I suggest a vanilla bookmark that you then open an incognito window.

### DIFF
--- a/src/app/AuthenticationController.tsx
+++ b/src/app/AuthenticationController.tsx
@@ -14,7 +14,7 @@ import { ServerStatus } from '../types/ServerStatus';
 import InitializingScreen from './InitializingScreen';
 import { isKioskMode } from '../utils/SearchParamUtils';
 import * as AlertUtils from '../utils/AlertUtils';
-import { setServerConfig } from '../config/ServerConfig';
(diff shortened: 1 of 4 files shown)


jmazzitelli (approved):
I ran tests - this works.

jmazzitelli (approved):
i tried with the latest refactor and it still works. LGTM.

pbajjuri20 (approved):

PR #2082, opened 2021-02-11
Use the typescript way to check undefined

xeviknal:
Replacing vanilla js undefined checks to use typescript native way.
Although the vanilla js check works fine, it is worth tho move it to typescript.

cc/ @jshaughn @israel-hdez @mtho11


jshaughn (approved):
@xeviknal +1 to the change, good for consistency.

israel-hdez:
A little late, but here is my thumb up :+1:

PR #2084, opened 2021-02-15
Improve match builder

Alan-Cha:
** Describe the change **

Exposing a function and constants as well as changing the `MatchBuilder` component to be more reusable.

** Issue reference **

I am working on this [issue](https://github.com/iter8-tools/kiali-ui/issues/34).

Lucas mentioned in this [comment](https://github.com/kiali/kiali-ui/pull/2015#issuecomment-736032444) that we should consider components. 

Unfortunately, we could not use all of the components out of the box. Specifically, we needed more control over the `MatchBuilder` component. In our use case, we only want to a subset of the match options (i.e. only `URI` and `HEADERS`) and we do not want the operator options to change in the case of `HEADERS`. In this PR, we made the match and operator options properties so that we can provide our own.

In addition, we also export some constants and pulled out another function that we would like to reuse.

** Backwards compatible? **

[Yes] Is your pull-request introducing changes in behavior?

** Screenshot **

There are no UI changes.

** Documentation **

There is no need for documentation changes.

### DIFF
--- a/src/components/IstioWizards/RequestRouting.tsx
+++ b/src/components/IstioWizards/RequestRouting.tsx
@@ -28,9 +28,17 @@ type State = {
   validationMsg: string;
 };
 
-const MSG_SAME_MATCHING = 'A Rule with same matching criteria is already added.';
-const MSG_HEADER_NAME_NON_EMPTY = 'Header name must be non empty';
-const MSG_HEADER_VALUE_NON_EMPTY = 'Header value must be non empty';
+export const MSG_SAME_MATCHING = 'A Rule with same matching criteria is already added.';
+export const MSG_HEADER_NAME_NON_EMPTY = 'Header name must be non empty';
+export const MSG_HEADER_VALUE_NON_EMPTY = 'Header value must be non empty';
+
+export function isMatchesIncluded(rules: Rule[], newRule: Rule) {
+  return rules.some(rule => {
+    return (
+      rule.matches.length === newRule.matches.length && rule.matches.every(match => newRule.matches.includes(match))
+    );
+  });
+}
 
 class RequestRouting extends React.Component<Props, State> {
   constructor(props: Props) {
@@ -78,21 +86,6 @@ class RequestRouting extends React.Component<Props, State> {
       validationMsg: ''
     };
   }
-
-  isMatchesIncluded = (rules: Rule[], rule: Rule) => {
-    let found = false;
-    for (let i = 0; i < rules.length; i++) {
-      const item = rules[i];
-      if (item.matches.length !== rule.matches.length) {
-        continue;
-      }
-      found = item.matches.every(value => rule.matches.includes(value));
-      if (found) {
-        break;
-      }
-    }
-    return found;
-  };
 
   isValid = (rules: Rule[]): boolean => {
     // Corner case, an empty rules shouldn't be a valid scenario to create a VS/DR
@@ -163,7 +156,7 @@ class RequestRouting extends React.Component<Props, State> {
         if (prevState.timeoutRetryRoute.isRetry && prevState.timeoutRetryRoute.isValidRetry) {
           newRule.retries = prevState.timeoutRetryRoute.retries;
         }
-        if (!this.isMatchesIncluded(prevState.rules, newRule)) {
+        if (!isMatchesIncluded(prevState.rules, newRule)) {
           prevState.rules.push(newRule);
           return {
             matches: prevState.matches,
--- a/src/components/IstioWizards/RequestRouting/MatchBuilder.tsx
+++ b/src/components/IstioWizards/RequestRouting/MatchBuilder.tsx
@@ -56,7 +56,7 @@ class MatchBuilder extends React.Component<Props, State> {
     };
   }
 
-  onMathOptionsToggle = () => {
+  onMatchOptionsToggle = () => {
     this.setState({
       isMatchDropdown: !this.state.isMatchDropdown
     });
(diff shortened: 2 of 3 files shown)


src/components/IstioWizards/RequestRouting.tsx
```diff
@@ -79,21 +87,6 @@ class RequestRouting extends React.Component<Props, State> {
     };
   }
 
-  isMatchesIncluded = (rules: Rule[], rule: Rule) => {
```
Alan-Cha, line 82: This function has been rewritten in a functional manner is now exported.

src/components/IstioWizards/RequestRouting/MatchBuilder.tsx
```diff
@@ -67,20 +69,19 @@ class MatchBuilder extends React.Component<Props, State> {
...
   render() {
-    const renderOpOptions: string[] = this.props.category === HEADERS ? [PRESENCE, ...opOptions] : opOptions;
     return (
       <InputGroup>
         <Dropdown
-          toggle={<DropdownToggle onToggle={this.onMathOptionsToggle}>{this.props.category}</DropdownToggle>}
+          toggle={<DropdownToggle onToggle={this.onMatchOptionsToggle}>{this.props.category}</DropdownToggle>}
           isOpen={this.state.isMatchDropdown}
-          dropdownItems={matchOptions.map((mode, index) => (
+          dropdownItems={this.props.matchOptions.map((mode, index) => (
```
Alan-Cha, line 77: `matchOptions` is now a property.

src/components/IstioWizards/RequestRouting/MatchBuilder.tsx
```diff
@@ -98,7 +99,7 @@ class MatchBuilder extends React.Component<Props, State> {
         <Dropdown
           toggle={<DropdownToggle onToggle={this.onOperatorToggle}>{this.props.operator}</DropdownToggle>}
           isOpen={this.state.isOperatorDropdown}
-          dropdownItems={renderOpOptions.map((op, index) => (
+          dropdownItems={this.props.opOptions.map((op, index) => (
```
Alan-Cha, line 102: `opOptions` is now a property.

src/components/IstioWizards/RequestRouting/RuleBuilder.tsx
```diff
@@ -84,7 +84,11 @@ class RuleBuilder extends React.Component<Props, State> {
         <Tabs isFilled={true} activeKey={this.state.ruleTabKey} onSelect={this.ruleHandleTabClick}>
           <Tab eventKey={0} title={'Request Matching'}>
             <div style={{ marginTop: '20px' }}>
-              <MatchBuilder {...this.props} />
+              <MatchBuilder
+                {...this.props}
+                matchOptions={[HEADERS, URI, SCHEME, METHOD, AUTHORITY]}
```
lucasponce, line 89: I understand that the motivation of this change is to reuse the MatchBuilder in a Iter8 form.
I'd suggest that if you need to use these optional matchOptions/opOptions you could make these parameters optional, but leave the default behaviour inside the MatchBuilder logic.
Perhaps for "matchOptions" is easy, but the conditional array filled in the "opOptions" is something that belongs to the MatchBuilder logic, I think.

So, I'd suggest that when matchOptions/opOptions are present MatchBuilder can use them, but if not, it can get the default used for Kiali Wizards.

I think with that minimal change perhaps can be cleaner.

Also, typically this is an auxiliar component for the RequestRouting scenario, that's reason that this class is located in a deep path like "src/components/IstioWizards/RequestRouting/RuleBuilder.tsx", if you consider better, you can also clone the RuleBuilder/MatchBuilder and put a version of that clase under Iter8 logic and add that logic there.

In some cases some duplication sounds cleaner if justified, instead to make an aux component to work for both scenarios.

I'm fine as well with that strategy too if that simplifies your work.

Also, thinking in the future, cloning this aux components may give you more freedon to change them if the current logic can be maintained for both Iter8 / Istio needs.

Let me know what do you think.

@Alan-Cha Thanks for your PR and sorry for my delay in the review, past week was the IstioCon and most of the team was a little bit busy than usual.

Alan-Cha: @lucasponce Both options, creating a default behavior for `MatchBuilder` and cloning the component, sound good to me. I actually originally started out cloning the component but I figured I should try to reduce duplication as the only thing that I had to change was the `matchOptions` and the `opOptions`.

I am on the fence, perhaps leaning slightly towards reducing duplication. I feel that I do not have enough experience with Kiali or with Iter8 to make an informed decision.

Do you feel strongly about either option? If it is easier, I can create two PRs, my original version with duplication and another version that builds off of this PR, and perhaps by doing so, you can have a more complete view on what I am trying to do. Let me know what you think as well!

lucasponce: I'd prefer a clone as a first attempt, as Iter8 and Istio have different domains and that would give you flexibility in case that you need to change something.

The idea is that the users share a similar experience but from implementation side, it's reasonable to allow a little bit duplication on this case.

Also, I think it's easier to refactor in the future joining a cloned class that the other direction, as when you progress you'll need to check potential regressions in the Istio form.

Alan-Cha: Okay! We can do that instead. I need to first clarify some edge case behavior with Yew Huey first, however. She said that she will get back to me next week.

lucasponce:
Hi Alan, I was reviewing pending PRs.
So, then your strategy will be to clone instead to adapt the existing components, isn't it ?
Then you'll send a different PR for that work or you want to reuse this one ?

Let me know, or feel free to update/close it if necessary.

Thanks.

Alan-Cha:
@lucasponce Yes, sorry. I have been caught up in a lot of other work. I will be sending a new PR later today. We can close this for now.

PR #2087, opened 2021-02-19
Fix broken query for trace percentiles

jotak:
Since the grpc jaeger client PR, minDuration now has to be passed as an
integer (in microseconds) instead of a string


jotak:
Hey @hhovsepy I think this will fix the bug you're seeing.
The bug happens when we select a percentile in tracing display options. I you don't do that manually and still see the bug, I guess it's because it's stored in your browser storage.

The regression must have appeared while rebasing my GRPC PR, as it conflicted with the PR about tracing toolbar redesign.

hhovsepy (approved):
Verified

aljesusg (approved):
LGFM

jotak:
thx!

PR #2090, opened 2021-02-25
Add links from Graph side panel validation status to istio config list

xeviknal:
fixes https://github.com/kiali/kiali/issues/2218

### DIFF
--- a/src/components/Link/IstioConfigListLink.tsx
+++ b/src/components/Link/IstioConfigListLink.tsx
@@ -20,33 +20,34 @@ class IstioConfigListLink extends React.Component<Props> {
   validationToParams = () => {
     let params: string = '';
 
-    if(this.props.warnings) {
+    if (this.props.warnings) {
       params += 'configvalidation=Warning';
     }
 
-    if(params !== '') {
-      params += '&'
+    let errorParams: string = '';
+    if (this.props.errors) {
+      errorParams += 'configvalidation=Not+Valid';
     }
 
-    if(this.props.errors) {
-      params += 'configvalidation=Not+Valid';
+    if (params !== '' && errorParams !== '') {
+      params += '&';
     }
+
+    params += errorParams;
 
     return params;
   };
 
   render() {
     let params: string = this.namespacesToParams();
-    if(params !== '') { params += '&'; }
-    params += this.validationToParams();
+    const validationParams: string = this.validationToParams();
+    if (params !== '' && validationParams !== '') {
+      params += '&';
+    }
+    params += validationParams;
 
-    return (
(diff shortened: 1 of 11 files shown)


hhovsepy (commented):
Tested the cases:
1. One WARN only.
2. One Error Only.
3. Many WARN and Errors.
4. One Error and many Warnings.
In all cases are OK.

Only difference with the initial issue is that Filters are not applied:
Screenshot from 2021-02-25 17-44-09

xeviknal:
@hhovsepy I've have added the filters into the links pointing to the istio config list.
1) When there are only errors, it filters by errors.
2) when there  are only warning, it filters by warnings.
3) when there are both, it filters by both warnings and errors.

WRT to linking to one specific istio config, I understood to not implementing it by comment in the issue. This would add more work in the endpoints side. So far, we don't know how many objects have errors/warnings. We know how many validations (one object may have multiple ones). I am okay to modify the backend though topping this extra info.

On the other hand, I've standarized the links. This behavior should not only be in the graph validation link but in overview too.

Finally, there was a bug in the validations icon in the istio config list and service details page:

Screenshot of Kiali (5)

Screenshot of Kiali (6)

Thank you!

jshaughn (approved):
@xeviknal You should yell at me if I don't respond to a review request after 24 hours!  I am sorry, I missed this.  Graph-related changes look fine to me, tested and working.

hhovsepy:
It seems the link from graph is not constantly putting the correct filters.

hhovsepy:
When clicking to namespace icon in graph and then going back to graph page several time with selecting different namespaces, the link from icon does not set filters anymore.

jshaughn:
> When clicking to namespace icon in graph and then going back to graph page several time with selecting different namespaces, the link from icon does not set filters anymore.

I tested this and for me it seemed to work, but I defer to @hhovsepy.  Do you have a reproducible pattern?

xeviknal:
I am working with this. I found a couple of inconsistencies. I'll update this PR soon :) Thanks jay!

xeviknal:
I can see what happens now. The filters in the istio config list don't prioritize the ones in the URL. It always takes the ones set in the state.

To reproduce:
1. Click a link the graph
2. Remove the filter (or add one more)
3. Go back to the graph and click the same link
4. The filters don't appear anymore (or both appears even if only one is intended)

It looks like this is a bug in the StatefulFilters. I'll give it a quick look now.

xeviknal:
@jshaughn @hhovsepy I've found what was the problem. I am resetting filters before sending people to the istio config list. This way, the param filters will have priority.

Let me know!

hhovsepy (approved):
Verified now.

PR #2091, opened 2021-03-02
Fold status+managedFields fields in yaml editor

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3600

### DIFF
--- a/src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
+++ b/src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
@@ -206,8 +206,8 @@ class IstioConfigDetailsPage extends React.Component<RouteComponentProps<IstioCo
       // tslint:disable-next-line
       editor.onChangeAnnotation();
 
-      // Fold Status field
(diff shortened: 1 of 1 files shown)


hhovsepy (approved):
Verified:
Screenshot from 2021-03-02 16-45-28

aljesusg (approved):
LGFM

xeviknal:
@lucasponce I see this is working although it is assuming that the `managedFields` and `status` will be rendered contiguously. It seems fair, however, I can't see where the order of this fields is placed. I am not able to see how the `metadata` fields are sorted neither.
My only concern is that if there is one more field between `managedFields` and `status` this will be hidden.

lucasponce:
> I am not able to see how the metadata fields are sorted neither.

image

'managedFields' is the last entry for the ObjectMeta structure, so, unless kubernetes updates this type, I think that the current design is safe.

We could create a loop per individual field, but I guess on this case it was easy to share the same logic for both status/managedFields.

xeviknal (approved):
LGTM!

PR #2092, opened 2021-03-03
Fix overview filter for Healthy case

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3719

@hhovsepy let me know if this PR looks good for that use case.

Yes, in highly populated namespaces, probably the delay of the async fetch is visible, but I checked that the "loading" icon is present on those cases.


aljesusg (approved):
LGFM

hhovsepy (approved):
Verified:
Screenshot from 2021-03-03 17-33-47

PR #2096, opened 2021-03-08
Metrics page: more details & links from spans

aljesusg:
# New proprosal

Add to the tooltip the method of the span

Screenshot from 2021-03-24 11-21-05

When we click the span this redirect us to the traces tab with the span selected. This span will be expanded with a blue border on the right
Screenshot from 2021-03-24 11-20-31

### DIFF
--- a/src/app/History.ts
+++ b/src/app/History.ts
@@ -32,6 +32,7 @@ export enum URLParam {
   JAEGER_PERCENTILE = 'percentile',
   JAEGER_SHOW_SPANS_AVG = 'showSpansAvg',
   JAEGER_TRACE_ID = 'traceId',
+  JAEGER_SPAN_ID = 'spanId',
   NAMESPACES = 'namespaces',
   OPERATION_NODES = 'operationNodes',
   OVERVIEW_TYPE = 'otype',
--- a/src/components/JaegerIntegration/JaegerResults/SpanTableItem.tsx
+++ b/src/components/JaegerIntegration/JaegerResults/SpanTableItem.tsx
@@ -12,9 +12,19 @@ import { MetricsStats } from 'types/Metrics';
 import { CellProps, createListeners, Expandable, renderExpandArrow } from 'components/Expandable';
 import { formatDuration, isErrorTag } from 'utils/tracing/TracingHelper';
 import responseFlags from 'utils/ResponseFlags';
+import { getSpanId } from '../../../utils/SearchParamUtils';
 
 const dangerErrorStyle = style({
   borderLeft: '3px solid var(--pf-global--danger-color--100)'
+});
+
+const selectedErrorStyle = style({
+  borderRight: '3px solid var(--pf-global--info-color--100)',
+  borderLeft: '3px solid var(--pf-global--danger-color--100)'
+});
+
+const selectedStyle = style({
+  borderRight: '3px solid var(--pf-global--info-color--100)'
 });
 
 const kebabDropwdownStyle = style({
@@ -36,28 +46,70 @@ type RowProps = RichSpanData & {
   isExpanded: boolean;
 };
 
+const getClassName = (isError: boolean, isSpan: boolean): string | undefined => {
+  return isSpan ? (isError ? selectedErrorStyle : selectedStyle) : isError ? dangerErrorStyle : undefined;
+};
+
 export const buildRow = (props: RowProps) => {
   const expandListeners = createListeners();
+  const isSpan = props.spanID === getSpanId();
   expandListeners.push(props.onExpand);
+  const isExpandable = getSpanId() ? isSpan : props.isExpanded;
   return {
-    className: props.tags.some(isErrorTag) ? dangerErrorStyle : undefined,
+    className: getClassName(props.tags.some(isErrorTag), isSpan),
     isOpen: false,
     cells: [
       {
         title: (
           <>
-            {renderExpandArrow(expandListeners, props.isExpanded)} {formatDuration(props.relativeStartTime)}
+            {renderExpandArrow(expandListeners, isExpandable)} {formatDuration(props.relativeStartTime)}
           </>
         )
       },
       {
-        title: <Expandable {...props} clickToExpand={false} listeners={expandListeners} innerComponent={OriginCell} />
+        title: (
+          <Expandable
+            {...props}
+            clickToExpand={false}
+            isExpanded={isExpandable}
+            listeners={expandListeners}
+            innerComponent={OriginCell}
+          />
+        )
       },
       {
-        title: <Expandable {...props} clickToExpand={true} listeners={expandListeners} innerComponent={SummaryCell} />
+        title: (
+          <Expandable
+            {...props}
+            clickToExpand={true}
+            isExpanded={isExpandable}
+            listeners={expandListeners}
+            innerComponent={SummaryCell}
+          />
+        )
       },
-      { title: <Expandable {...props} clickToExpand={true} listeners={expandListeners} innerComponent={StatsCell} /> },
-      { title: <Expandable {...props} clickToExpand={false} listeners={expandListeners} innerComponent={LinksCell} /> }
+      {
+        title: (
+          <Expandable
+            {...props}
+            clickToExpand={true}
+            isExpanded={isExpandable}
+            listeners={expandListeners}
+            innerComponent={StatsCell}
+          />
+        )
+      },
+      {
+        title: (
+          <Expandable
+            {...props}
+            clickToExpand={false}
+            isExpanded={isExpandable}
+            listeners={expandListeners}
+            innerComponent={LinksCell}
+          />
+        )
+      }
     ],
     spanID: props.spanID
   };
--- a/src/components/JaegerIntegration/TracesComponent.tsx
+++ b/src/components/JaegerIntegration/TracesComponent.tsx
@@ -29,6 +29,7 @@ import { getTimeRangeMicros } from 'utils/tracing/TracingHelper';
 import { TracesDisplayOptions, QuerySettings, DisplaySettings, percentilesOptions } from './TracesDisplayOptions';
 import { Direction, genStatsKey, MetricsStatsQuery } from 'types/MetricsOptions';
 import { MetricsStatsResult } from 'types/Metrics';
+import { getSpanId } from 'utils/SearchParamUtils';
 
 interface TracesProps {
   namespace: string;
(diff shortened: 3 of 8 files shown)


aljesusg:
Screenshot from 2021-03-08 17-53-57


Added Jaeger Links for spans I need to check the last Span details changes to focus in a specific span

lucasponce:
I think the PR is still in a draft, I didn't comment on that.

These links should be consistent with same wording and expression used both in Graph and in Traces details.
(Jaeger is not mentioned, but tracing, and the icon to express it's outside of Kiali should be there).

I guess if all are "show" there is no need to create another section, or it will deviate from the context menu in the graph, too.

aljesusg:
# Update

Menu
Screenshot from 2021-03-09 10-54-16
Span selected when we clicked the menu
Screenshot from 2021-03-09 10-59-11
Span selected when we clicked the menu and is error
Screenshot from 2021-03-09 11-02-18

aljesusg:
I am stucked with this ....for some reason sometimes the tippy moves to the top-left....researching about it

aljesusg:
# menu 
Screenshot from 2021-03-18 10-03-02

# How looks like a span selected from metrics (right blue border)
Screenshot from 2021-03-18 10-02-38

hhovsepy:
Menu not constantly jumps to top left corner, happens on both FF and Chrome:
Screenshot from 2021-03-18 12-36-31

aljesusg:
> Menu not constantly jumps to top left corner, happens on both FF and Chrome:
> Screenshot from 2021-03-18 12-36-31

Yes... :( I am researching about it . It's annoying this behavior.

lucasponce (changes requested):
I think @hhovsepy's comment would need to be addressed.
I'll review it as well, but I wanted to mark it to avoid merge it with that issue.

aljesusg:
Ok it seems that now is more stable

hhovsepy:
Better now, but still in some cases menu is shown a bit lower.
Screenshot from 2021-03-22 12-14-23

But the link to Trace or Spans contains wrong workload name, it takes app name not the workload with version.
Screenshot from 2021-03-22 12-14-37

lucasponce:
Hi, some feedback and suggestions about this work.

The menu links to Kiali are nice, but I find a little bit redundant to add 4 external links to Tracing (Jaeger):
image

From business perspective, when user jumps from this chart to the Traces tab is to combine the service mesh telemetry information (+ service mesh logs in the future) with tracing info. And he could navigate into the exterrnal tool for more detail.

I'd recommend to keep only the external link to the span and remove others, as the parent/trace links will be implicit added in the same page (the main difference will be if that info is showed collapsed or opened in jaeger).

This is what user will see and from that page user can navigate easily into the parent/child or context, I don't see the need to add more external links from the chart.
image

I also hit the problems reported from @hhovsepy, it's annoying, I may have a couple of suggestions that perhaps may help to attack that problem from a different angle:

- First, one thing that I miss is that when I select a span in the chart, I'd like to see that span selected. Note that this happens in the traces tab in the main component, so it would keep consistency in that behaviour (perhaps to review if worth to sync style/colors, but that's more an implementation detail).
- Second, once that span is selected, instead to use a contextual menu, from the location of the span, perhaps you can implement a fixed contextual menu always in the same location, because the contextual menu is keep opened until the user closes it. I haven't reviewed the code, but perhaps that would help to mitigate the problem to not control where the context menu is opened, or to keep the open status on the refresh on the same position (I can confirm the problem and also there are other corner cases like opened very far from the mouse as well). I haven't tried to implement it, but it sounds that it may simplify this.

image

aljesusg:
> Hi, some feedback and suggestions about this work.
> 
> The menu links to Kiali are nice, but I find a little bit redundant to add 4 external links to Tracing (Jaeger):
> image
> 
> From business perspective, when user jumps from this chart to the Traces tab is to combine the service mesh telemetry information (+ service mesh logs in the future) with tracing info. And he could navigate into the exterrnal tool for more detail.

Ok in this term I'll keep only the span view tracing link (if user want to see parent/child span then he can go to that view in kiali and then go to tracing view)
> 
> I'd recommend to keep only the external link to the span and remove others, as the parent/trace links will be implicit added in the same page (the main difference will be if that info is showed collapsed or opened in jaeger).
> 
> This is what user will see and from that page user can navigate easily into the parent/child or context, I don't see the need to add more external links from the chart.
> image
> 
> I also hit the problems reported from @hhovsepy, it's annoying, I may have a couple of suggestions that perhaps may help to attack that problem from a different angle:

Yes this is so annoying and is taking so time
> 
> * First, one thing that I miss is that when I select a span in the chart, I'd like to see that span selected. Note that this happens in the traces tab in the main component, so it would keep consistency in that behaviour (perhaps to review if worth to sync style/colors, but that's more an implementation detail).

we can do that.

> * Second, once that span is selected, instead to use a contextual menu, from the location of the span, perhaps you can implement a fixed contextual menu always in the same location, because the contextual menu is keep opened until the user closes it. I haven't reviewed the code, but perhaps that would help to mitigate the problem to not control where the context menu is opened, or to keep the open status on the refresh on the same position (I can confirm the problem and also there are other corner cases like opened very far from the mouse as well). I haven't tried to implement it, but it sounds that it may simplify this.

I'm agree about to keep a menu in the same place....I am thinking about something similar to the "side panel" in the graph page. Keep there the links and maybe in a future improve the information. What do you think?

> 
> image

lucasponce:
> I'm agree about to keep a menu in the same place....I am thinking about something similar to the "side panel" in the graph page. Keep there the links and maybe in a future improve the information. What do you think?

To me a side panel at this chart 

image

sounds redudant with the flow of the traces tab:

image

So, to me as the spans filter is a visualization option, the business story should be:
- "I see a suspicius spike on a chart"
- "I enable spans to see what can be the source"
- "I can jump to traces tab to see the span + service mesh telemetry + service mesh logs (in the future)"
- "Optionally, if I'm versed in tracing, I can jump to the external tool because I'm going to compare or perform other details operations that are outside the Kiali scope"

My suggestion about the fix option is to address the problem we are having with that component, but I don't think we'll increase the info in this context but to drive the user to the traces tab where we may want to consolidate.

In a first thought I was thinking that a center position for the contextual would help here.

One thing that I forgot in my previous comment if that perhaps it can be useful to add an internal link to the Graph when the span is highlighted, that also would make sense.

P.S.: Also, we need to put in context future efforts on this area, for example, in a completely separated task, I'd like to study if we could try to get a better dynamic draw of all charts to avoid scrolling and let the user to see all info available, and play more with the maximize. That's completely separated but it also give visibility that probably we don't want to add too much feature in a chart, because if it's resized perhaps that may impact as well.

aljesusg:
> > I'm agree about to keep a menu in the same place....I am thinking about something similar to the "side panel" in the graph page. Keep there the links and maybe in a future improve the information. What do you think?
> 
> To me a side panel at this chart
> 
> image
> 
> sounds redudant with the flow of the traces tab:
> 
> image
> 
> So, to me as the spans filter is a visualization option, the business story should be:
> 
> * "I see a suspicius spike on a chart"
> * "I enable spans to see what can be the source"
> * "I can jump to traces tab to see the span + service mesh telemetry + service mesh logs (in the future)"
> * "Optionally, if I'm versed in tracing, I can jump to the external tool because I'm going to compare or perform other details operations that are outside the Kiali scope"
> 
> My suggestion about the fix option is to address the problem we are having with that component, but I don't think we'll increase the info in this context but to drive the user to the traces tab where we may want to consolidate.
> 
> In a first thought I was thinking that a center position for the contextual would help here.
> 
> One thing that I forgot in my previous comment if that perhaps it can be useful to add an internal link to the Graph when the span is highlighted, that also would make sense.
> 
> P.S.: Also, we need to put in context future efforts on this area, for example, in a completely separated task, I'd like to study if we could try to get a better dynamic draw of all charts to avoid scrolling and let the user to see all info available, and play more with the maximize. That's completely separated but it also give visibility that probably we don't want to add too much feature in a chart, because if it's resized perhaps that may impact as well.

So, we can keep onClick to redirect to the traces tab to the selected span information and show  in the traces tab > span tab all the information and the links.

aljesusg:
@lucasponce I did the changes. Things to improve this:

In the Graph Side panel we have a link to the trace view , I think taht we should add a link to span view if the user select a span too with the same behavior.

In the span details we can see other spans of the same trace, maybe we should add some icon (arrow down, arrow upper) to show parent/child by span selected perspective. (It's an idea)

In the right menu we have the enough links I think.

# New proposal

Add to the tooltip the method of the span

Screenshot from 2021-03-24 11-21-05

When we click the span this redirect us to the traces tab with the span selected. This span will be expanded with a blue border on the right
Screenshot from 2021-03-24 11-20-31

hhovsepy:
+1 on New proposal.
The issue https://github.com/kiali/kiali/issues/3221 makes the click not happening sometimes.

aljesusg:
Added **show span** link to the summary panel in the graph
Screenshot from 2021-03-25 10-12-04

lucasponce (approved):
+1 this proposal works well.

As @hhovsepy commented this issue kiali/kiali#3221 is also impacting, I wonder @aljesusg if now you have all this area fresh, if you'd like to take a look.

aljesusg:
> +1 this proposal works well.
> 
> As @hhovsepy commented this issue [kiali/kiali#3221](https://github.com/kiali/kiali/issues/3221) is also impacting, I wonder @aljesusg if now you have all this area fresh, if you'd like to take a look.

Ok I'll take this one, anyway we can merge this after @hhovsepy review and I'll work in the another one

hhovsepy:
When clicking to trace in graph side panel for first time, the 'Show Span' is not loading:
Screenshot from 2021-03-25 14-14-35

aljesusg:
> When clicking to trace in graph side panel for first time, the 'Show Span' is not loading:
> Screenshot from 2021-03-25 14-14-35

You are in the traces information, you need to click one trace of the list with N spans and then you'll see the information related with the spans and the link

hhovsepy:
Sometimes not constantly, when clicking on span in metrics page, the redirection to Traces page is done after 3-4 sec delay which confuses.

aljesusg:
> Sometimes not constantly, when clicking on span in metrics page, the redirection to Traces page is done after 3-4 sec delay which confuses, and without spans.
> Screenshot from 2021-03-25 14-10-22

Yes, I experienced that behaviour too in master, I don't know if is related with victory or what I can address it too in the issue that @lucasponce commented

hhovsepy (approved):
Verifying and leaving the non constant issues fix in another PR.

PR #2099, opened 2021-03-10
Fix Health broken

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3757
Fixes https://github.com/kiali/kiali/issues/3756

In short, as filters are stateful, they shouldn't be reset on render but on an action, otherwise it may have side effects.

This scenario was similar as this old one:

https://github.com/kiali/kiali-ui/blob/master/src/components/BreadcrumbView/BreadcrumbView.tsx#L120

I fixed and also added more tests on the filter case as I was not sure at first time if the problem was with some changes I did there.

### DIFF
--- a/src/components/FilterList/__tests__/OverviewPageHelper.test.ts
+++ b/src/components/FilterList/__tests__/OverviewPageHelper.test.ts
@@ -1,0 +1,105 @@
+import * as FilterHelper from '../FilterHelper';
+import NamespaceInfo from '../../../pages/Overview/NamespaceInfo';
+import * as Filters from '../../../pages/Overview/Filters';
+import { FilterSelected } from '../../Filters/StatefulFilters';
+
+const allNamespaces: NamespaceInfo[] = [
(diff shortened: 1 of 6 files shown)


lucasponce:
@hhovsepy I think I'll fix https://github.com/kiali/kiali/issues/3756 on this PR as well as it's touching same files

xeviknal (approved):
Dummy me! LGTM!

lucasponce:
@xeviknal @hhovsepy updated with both fixes

xeviknal (approved):
LGTM! Good tests in there. I guess the fixes came after approaching the addition of links from sidepanel to istioconfig list. Thanks for approaching this.

hhovsepy (approved):
Verified both issues:
Screenshot from 2021-03-11 15-40-30
Screenshot from 2021-03-11 16-05-42
Screenshot from 2021-03-11 16-06-05

PR #2100, opened 2021-03-11
Adding new istio component status: NotReady - replicas set to 0

xeviknal:
needs backend end: https://github.com/kiali/kiali/pull/3766

### DIFF
--- a/src/components/IstioStatus/IstioComponentStatus.tsx
+++ b/src/components/IstioStatus/IstioComponentStatus.tsx
@@ -2,7 +2,12 @@ import * as React from 'react';
 import { ComponentStatus, Status } from '../../types/IstioStatus';
 import { IconType } from '@patternfly/react-icons/dist/js/createIcon';
 import { PFAlertColor } from '../Pf/PfColors';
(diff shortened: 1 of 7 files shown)


lucasponce:
I've just wonder if for consistent you may like to change the icon when not-ready to the blue one, as it's used in similar places for same state.

xeviknal:
@lucasponce I've did both changes: texts and colors. However, the blue to the not ready states is so dark that is not working with the black background. I've added a lighter blue for them. Check it out:

Screenshot of Kiali (8)

hhovsepy:
Screenshot from 2021-03-12 16-49-20

lucasponce (approved):
Ok.
Perhaps some minor tweak to align colors and wording, but nothing to not merge this work.
Thanks @xeviknal

xeviknal:
@lucasponce do you have an idea of which tweaks could be done?
Do you think we should change the tooltip background to be consistent with the `Not Ready` color across the console?
And about the wording, what did I miss? Is there a typo or a word not consistent?

PR #2101, opened 2021-03-11
Fix wrong collapsed state on charts

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3693

How to reproduce:
- Identify a workload with traffic in only one side (Inbound or Outbound):

image

- Click in the tab with no traffic and check that charts are collapsed:

image

- Click in the tab with traffic and verify that charts are expanded (instead to see them collapsed by default):

image

### DIFF
--- a/src/components/Charts/KChart.tsx
+++ b/src/components/Charts/KChart.tsx
@@ -77,6 +77,15 @@ class KChart<T extends LineInfo> extends React.Component<KChartProps<T>, State> 
     };
   }
 
+  componentDidUpdate(prevProps: KChartProps<T>) {
+    // If it starts collapsed because it's empty, then checks if there is new data to expand
+    if (this.state.collapsed && this.props.data.length !== prevProps.data.length) {
+      this.setState({
+        collapsed: false
+      });
+    }
+  }
+
   render() {
     return (
       <Expandable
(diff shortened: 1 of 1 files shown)


aljesusg (approved):
LGFM, thinking about show some kind of message like "No traffic"

lucasponce:
> thinking about show some kind of message like "No traffic"

Well, when no traffic, the charts show correctly that message, adding a toast perhaps may populate a lot the messages.
(It should send a message per chart and there are 8 metrics - i.e. when http is present tcp are not and viceversa, so it could be a mix -).

At least this fix would help to avoid that confusion with the collapse message.

hhovsepy:
1. First open Outbound Tab (without metrics). Metrics are collapsed.
2. Then switch to Inbound tab (with metrics) then switch to Outbound Tab (without metrics), the empty metrics are expanded. Is this expanded behavior expected?
Screenshot from 2021-03-15 09-48-20

lucasponce:
> Then switch to Inbound tab (with metrics) then switch to Outbound Tab (without metrics), the empty metrics are expanded. Is this expanded behavior expected?

Mmm I think it should be consistent, so when empty it should start collapse, I'll refine the fix.

Thanks @hhovsepy for testing it

lucasponce:
@hhovsepy I've updated this PR with a better logic, my initial approach was buggy (still catching-up a bit with that class).
Now it tests when there is change in the props, and then the initial render is collapsed if empty.
User can also open/close the tab after that but it will solve this issue better.

Let me know if this works ok.

hhovsepy (approved):
Verified.

PR #2104, opened 2021-03-17
Enhanced logs view with access log detail

jshaughn:
Part of Epic https://github.com/kiali/kiali/issues/3499

Backend PR: https://github.com/kiali/kiali/pull/3785

This is an attempt to re-vamp the logs view with an eye on usability and future composability.  Goals:

- better use of space/more logs on screen
- easier correlation between container logs for the pod
- enhanced handling of access logs
- better full screen handling

Features of the PR:
- a more compact toolbar area
- replace the split view with a time-sorted, color-coded, integrated view of selected containers
- provide the standard toolbar features in Full-Screen
- add a parsed access log viewer (via info tooltip)
- improve timestamp sorting/viewing

Before:
Screenshot from 2021-03-17 16-45-00

After:
Screenshot from 2021-03-17 17-01-04

Screenshot from 2021-03-17 17-02-15


jshaughn:
@jmazzitelli @mattmahoneyrh , OK, the PR has passed UX review and has informally been viewed by @lucasponce .  I think it's in pretty good shape for dev/qe review with the hope of making it into v1.33.  This revised view, and enhanced handling of access logs, is a step towards a more advanced handling of logs, and access logs in particular.

pbajjuri20:
Deployed and tested the latest Features of logs



log_1


log_2

jshaughn:
> Deployed and tested the latest Features of logs

@pbajjuri20 Any issues?

pbajjuri20:
> > Deployed and tested the latest Features of logs
> 
> @pbajjuri20 Any issues?

@jshaughn No issues found everything works perfectly, I'm waiting for dev approval.

jmazzitelli (changes requested):
The "C" icon for "Containers" needs to change - it is identical to the "C" icon for "Cluster" in the graph view - the only way you differentiate is if you hover and look at the tooltip for "Container" (there is no tooltip for "Cluster")

Compare:
image

to:

image

lucasponce:
Some weird feedback:
image

Note the lack of Envoy log on the first half of the tab and there were logs there.
I haven't debugged, but worth to test if something was missing.

lucasponce:
I'd test well the custom timeselector, I did several attempts and I'm not sure which logic applies:
image

(Does it get the N lines from the beginning / the most recent, I guess it will depend on which logs are available from k8s api but worth to take a close look on this test - not a blocker -)

jshaughn:
> The "C" icon for "Containers" needs to change

@jmazzitelli , actually, I think it's OK to re-use the same badge letters in different contexts, but I also don't have a problem making a change.  Instead of changing the 'C' for containers I'd prefer to change to 'CL' for clusters as it makes it more distinct in the graph boxing.  I pushed the change.

As an aside, I think we should probably consolidate our PfBadge definitions in the same sort of way we have done for PfColors and KialiIcons, so that we ensure consistent lettering and styling.  As for the styling, I think we should make a change to smaller PF badges, more like console.  Not necessarily just because it's more like console but because I prefer that they take up less space, as in console.

jmazzitelli (approved):

pbajjuri20 (approved):

jshaughn:
> Note the lack of Envoy log on the first half of the tab and there were logs there.

@lucasponce There is some subtle manipulation here that you may be noticing.   Envoy timestamps its log messages, that first field in the entry is the timestamp as generated by Envoy at log-time.  But the logs are then submitted to k8s for storage and k8s assigns a timestamp to each entry.  The log entries are typically submitted as a batch, and so the k8s timestamps are both later than the log entry timestamps, and also very close together.   When Kiali fetches a time range you are querying against the k8s API, and the k8s timestamps on the entries.   So, you should get back access logs with k8s timestamps >= to the query time.   You may not get back all access log entries with envoy timestamps >= to the query time.   Furthermore, because the integrated log view works better this way, we prefer the access log timestamps and use it to replace the k8s timestamp on these entries.

I'm not 100% sure it explains what you are seeing but maybe it does, as I'm fairly sure the time range handling is correct.  But I'll take another look.  You can also turnon the "Show Timestamps" option in the kebab, to perhaps help.  Finally, I do see in the screenshot that you could scroll up and maybe see more access logs.

jshaughn:
> (Does it get the N lines from the beginning / the most recent, I guess it will depend on which logs are available from k8s api but worth to take a close look on this test - not a blocker -)

@lucasponce First, see my other comment about the timestamps and the fetching.   As for the "N lines", those are "Tail" lines, as in general users want to see the most recent log lines when limiting by lines.  So, looking at the screenshot, it looks correct as the end of the time range is the :09 minute, which is what the logs are showing (even though the TZ is not local, so the hour doesn't match).

pbajjuri20:
> > The "C" icon for "Containers" needs to change
> 
> @jmazzitelli , actually, I think it's OK to re-use the same badge letters in different contexts, but I also don't have a problem making a change. Instead of changing the 'C' for containers I'd prefer to change to 'CL' for clusters as it makes it more distinct in the graph boxing. I pushed the change.
> 
> As an aside, I think we should probably consolidate our PfBadge definitions in the same sort of way we have done for PfColors and KialiIcons, so that we ensure consistent lettering and styling. As for the styling, I think we should make a change to smaller PF badges, more like console. Not necessarily just because it's more like console but because I prefer that they take up less space, as in console.

Tested the "C" icon changed to "CL" for the cluster, No issues found.


Cluster_CL

PR #2105, opened 2021-03-18
Relax matching on breadcrumb paths

jshaughn:
Don't hold all parts of the breadcrumb path to valid k8s namespace value restrictions.  Let other parts conform to valid label value restrictions (see https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set)

Fixes https://github.com/kiali/kiali/issues/3776


xeviknal (approved):
I've been playing around it looks good.

jshaughn:
@mattmahoneyrh I think a simple regression test ensuring breadcrumbs are working well is all that is needed.

mattmahoneyrh (approved):

PR #2108, opened 2021-03-22
Fix the traffic animation issue

jshaughn:
Fix the traffic animation issue, which turned out to just be a symptom of a deeper problem.  It happened only when animation was enabled and the user enabled boxing, changed to an "app" graph type, or navigated to a graph with boxing (and yes, this took a long time to realize :)). The root cause was the fact that boxLayouts finish in an async fashion, and animation was started prior to the layout completing. Theoretically it could also happen without any boxing, but just using cola or cose-bilkent
layouts, which can also run async.

The solution is to introduce some promise logic to ensure we don't complete a graph update before a layout is complete.

* unrelated but included, a small change to legend labels based on some confusion reported by a community member.

Fixes https://github.com/kiali/kiali/issues/3755

### DIFF
--- a/src/components/CytoscapeGraph/Layout/BoxLayout.ts
+++ b/src/components/CytoscapeGraph/Layout/BoxLayout.ts
@@ -130,6 +130,10 @@ export default class BoxLayout {
     this.runAsync();
   }
 
+  emit(events) {
+    this.emit(events);
+  }
+
   // Discrete layouts (dagre) always stop before layout.run() returns. Continuous layouts (cose,cola)
   // are started by layout.run() but may stop after run() returns. Because outer boxes require the inner
   // box layouts to complete, we need to force discrete behavior regardless of layout, and that is why
@@ -208,6 +212,8 @@ export default class BoxLayout {
         boxNode.removeScratch(STYLES_KEY);
         boxNode.removeScratch(PARENT_POSITION_KEY);
       });
+
+      this.emit('boxlayoutstop');
     });
 
     layout.run();
--- a/src/components/CytoscapeGraph/Layout/BoxLayout.ts
+++ b/src/components/CytoscapeGraph/Layout/BoxLayout.ts
@@ -130,8 +130,12 @@ export default class BoxLayout {
     this.runAsync();
   }
 
-  emit(events) {
-    this.emit(events);
+  /**
+   * This is a stub required by cytoscape to allow the layout impl to emit events
+   * @param _events space separated string of event names
+   */
+  emit(_events) {
+    // intentionally empty
   }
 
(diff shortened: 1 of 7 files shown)


jshaughn:
@mattmahoneyrh Please do a pretty good regression test of graph features for this one.  Including find/hide, trace overlay, different layouts, navigating away and back, focus animation (when navigating from a detail page graph), and of course play with animation.

mattmahoneyrh (approved):

src/components/CytoscapeGraph/TrafficAnimation/TrafficRenderer.ts
```diff
@@ -298,10 +298,10 @@ export default class TrafficRenderer {
    */
   processStep = () => {
     try {
-      if (this.previousTimestamp === undefined) {
-        this.previousTimestamp = Date.now();
-      }
       const nextTimestamp = Date.now();
+      if (!this.previousTimestamp) {
+        this.previousTimestamp = nextTimestamp;
+      }
```
israel-hdez, line 304: Good changes, to make all variables match.

src/components/CytoscapeGraph/Layout/BoxLayout.ts
```diff
@@ -130,6 +130,10 @@ export default class BoxLayout {
     this.runAsync();
   }
 
+  emit(events) {
+    this.emit(events);
```
israel-hdez, line 134: Do you need this?
I have the idea that this is calling itself. But I'm  not sure how cytoscape extensions work. May be it's being overridden somewhere?

israel-hdez (commented):
I only have a question about a method. Once resolved I can approve.

src/components/CytoscapeGraph/Layout/BoxLayout.ts
jshaughn: You're right, that's confusing and I need to clean it up.  I do need the `emit()` because I need to emit the `layoutstop` event when BoxLayout finally finishes, and that will allow the rest of the graph processing to then continue.  But this is just a contractual stub, the body isn't called, and that is why we don't get into a loop.

BoxLayout is a Cy layout that we have created (we register it for use in CytoscapeReactWrapper).   It means that Cytoscape will invoke it without question as a Layout (sort of like an Interface) and it is expected to support all of the necessary entry points.  Because we are the only consumer, our layout doesn't actually support everything a public layout would support, in fact up until now we only implemented the `run` entry point.  Now we have added the `emit`. 
The promise in processGraphUpdate waits on that event to be emitted.

I am going to make a commit to clean this up...  Thanks for pointing it out.

israel-hdez: All good.

israel-hdez (approved):
LGTM!

PR #2111, opened 2021-03-24
Update ExperimentTrafficForm

Alan-Cha:
** Describe the change **

Improve the match rule section of the traffic control section in the experiment page by reusing components from Istio (request routing).

** Issue reference **

Related to #2084

** Backwards compatible? **

[No] Is your pull-request introducing changes in behavior?

This PR is not introducing changes in behavior. It is only changing the UI.

** Screenshot **

Before:
Screen Shot 2021-02-09 at 3 53 45 PM

After:
Screen Shot 2021-03-24 at 2 11 27 PM

### DIFF
--- a/src/components/IstioWizards/WizardActions.ts
+++ b/src/components/IstioWizards/WizardActions.ts
@@ -709,21 +709,21 @@ export const getInitWeights = (
           });
         }
       });
-    }
 
-    // Convention: we place the mirror routes as last position
-    if ((route as HTTPRoute).mirror) {
-      const httpRoute = route as HTTPRoute;
-      // Check mirror on HTTP Route
-      if (httpRoute.mirror && httpRoute.mirror.subset && wkdVersionName[httpRoute.mirror.subset]) {
-        const mirrorPercentage = httpRoute.mirrorPercentage ? httpRoute.mirrorPercentage.value : 100;
-        wkdWeights.push({
-          name: wkdVersionName[httpRoute.mirror.subset],
-          weight: mirrorPercentage,
-          locked: false,
-          maxWeight: 100,
-          mirrored: true
-        });
+      // Convention: we place the mirror routes as last position
+      if ((route as HTTPRoute).mirror) {
+        const httpRoute = route as HTTPRoute;
+        // Check mirror on HTTP Route
+        if (httpRoute.mirror && httpRoute.mirror.subset && wkdVersionName[httpRoute.mirror.subset]) {
+          const mirrorPercentage = httpRoute.mirrorPercentage ? httpRoute.mirrorPercentage.value : 100;
+          wkdWeights.push({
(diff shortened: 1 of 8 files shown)


src/components/IstioWizards/WizardActions.ts
```diff
@@ -709,21 +709,21 @@ export const getInitWeights = (
           });
         }
       });
-    }
 
-    // Convention: we place the mirror routes as last position
-    if ((route as HTTPRoute).mirror) {
```
Alan-Cha, line 715: Extra attention here: this is not an Iter8 extension change. I believe this was meant to go inside the if statement. I got an error because it was not.

lucasponce: Right. I guess from the Service Wizards there is no case to get that situation, with a undefined route, which is what it should generated the error, right ?

The change seems ok to me.

lucasponce:
Hi @Alan-Cha,
Thanks a lot for the work.

In the Istio Config Forms I introduced some refactoring that may apply here as well, but I comment, as extensions have other rythm I'm +1 that you decide if you may want to work on them in iterative work and consolidate this PR
(I prefer to put something stable in master and send minor changes as a general strategy).

So, I will approve this work, but here are my comments for your consideration:

- Namespace selection: In the Istio Config forms now there is no duplication to select the namespace, the main toolbar is used for namespace selection.

Iter8:
image

Istio Config:
image

So, this change will save space in the form and reduce the duplication. The page may need to "connect" to Redux to get that state and that may trigger some change internally at implementation label.

- Left label description: we've refactored the labeling to avoid two lines in the left side of the horizontal form, saving space

Iter8:
image

image
(Here perhaps Show can be omitted)

Istio Config:
image

- Buttons for adding rows and new object: in Istio Config we had also different patters for adding items to a list/table and adding the result of the list to a different section as well. We unified that pattern.

Iter8:
image

Also, note that we moved the subtable/sublist into the right part of the horizontal form, again, just to unify and to not have multiple patterns for same use case.

Istio Config:
image

- Reduce the informative/helper texts or move them into an info (i) icon:

Iter8:
image

In forms like this, where there is a lot of space, you can save the helper text just adding an info icon at the right.
Then probably forms will show a less populated

In general, take a look to the recent changes to the Istio Config form and let me know.

From my side I'm +1 to apply these changes as they are scoped to the Iter8 extension only and you decide the rythm.

Let me know if some comment is not clear enough, also count on me if you need some support at code level as well.

lucasponce (approved):
I'll approve in case that @Alan-Cha would like to consolidate this first and send followup changes with some of the comments suggested.

lucasponce:
>  Could I make the changes at a later date?

Of course, as these changes are located into the Iter8 page. 
I'm going to consolidate them in master, otherwise they may have conflicts, so feel free to work on them in other PRs.

I will be happy to review them.

PR #2112, opened 2021-03-26
Span detail can't be open when navigating from Graph

aljesusg:
Fix https://github.com/kiali/kiali/issues/3814

### DIFF
--- a/src/components/JaegerIntegration/JaegerResults/SpanTable.tsx
+++ b/src/components/JaegerIntegration/JaegerResults/SpanTable.tsx
@@ -24,6 +24,7 @@ import MetricsStatsThunkActions from 'actions/MetricsStatsThunkActions';
 import { RichSpanData } from 'types/JaegerInfo';
 import { sameSpans } from 'utils/tracing/TracingHelper';
 import { buildQueriesFromSpans } from 'utils/tracing/TraceStats';
+import { getSpanId } from '../../../utils/SearchParamUtils';
 
 type SortableCell<T> = ICell & {
   compare?: (a: T, b: T) => number;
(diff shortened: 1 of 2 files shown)


hhovsepy:
Now when clicking on span from graph, the opened Span Details spans are collapsed, but are expandable.
Is this the expected behavior? Because original issue says that spans are not expandable at all.

aljesusg:
> Now when clicking on span from graph, the opened Span Details spans are collapsed, but are expandable.
> Is this the expected behavior? Because original issue says that spans are not expandable at all.

The behaviour is that the span selected in the graph should be opened in the trace details tab and others must be collapsed.
I think that there are another issue not related with this, sometimes the spanId is wrong in the link of the graph I think that is related with the fetch of the span but I need to check it. Can you verify this ?

lucasponce (approved):

lucasponce:
While testing other PR, I've found a problem related.

On multiple spans with error, it seems that spans can't expand, like this

image

Please, check that this PR also solves that situation as well cc @hhovsepy.

Thanks.

hhovsepy:
> > Now when clicking on span from graph, the opened Span Details spans are collapsed, but are expandable.
> > Is this the expected behavior? Because original issue says that spans are not expandable at all.
> 
> The behaviour is that the span selected in the graph should be opened in the trace details tab and others must be collapsed.
> I think that there are another issue not related with this, sometimes the spanId is wrong in the link of the graph I think that is related with the fetch of the span but I need to check it. Can you verify this ?
I confirm that spanId sometimes gets wrong number in link from graph. But the linking functionality is working in general.
Will this spanId issue fixed in another PR?

hhovsepy (approved):

aljesusg:
> I confirm that spanId sometimes gets wrong number in link from graph. But the linking functionality is working in general.
Will this spanId issue fixed in another PR?


Yes because is not related, we should open an issue with that because is hard to reproduce it, it seems is related with the fetch information of spans from graph that the id is not loading it correctly.

PR #2117, opened 2021-03-30
Apply UX feedback on graph boxing, and update PF color application

jshaughn:
This PR addresses UX feedback regarding the recent MC-related additions of namespace and cluster boxing. Specifically:
- restrict box colors to the [light] gray PF palette for uniformity and in order to accentuate status-oriented color use in the graph
- darken the namespace and cluster box borders to make them more distinct
- add PF badging, and increase font size, on box labels to make them easier to locate and read
- reduce highlighting "flashing" by not highlighting/dimming the contents of namespace and cluster boxes
  - in general, users may just be navigating over these boxes to reach a more specific node

While working on applying the PF palette I resolved issues in our "PfColors.tsx":
- Colors were hardcoded
- Many colors were out of date as the values were set using PF3
- Many defined colors were unused
- The color definitions were somewhat haphazard

I've overhauled PfColors.tsx to resolve the issues.  With the changes:
- We now define only the colors we use
- All colors are now resolved via PF variables such that any changes in PF will be picked up moving forward
  - note: we do still have a couple of custom colors defined by UX
- There are now two specific ways to use the defined colors, the common CSS mechanism, and another for when the specific value must be resolved (for cytoscape use).

So, with the conversion to the PF variables some colors have been updated.  In general the changes will likely not be noticed, but now we comply with PF4 (at least the version we are using).

The bulk of the PR "meat" is in GraphStyles.ts, PfColors.tsx, and GraphHighlighter.tsx.  Most everything else is just a touch due to the color refactor.

Closes https://github.com/kiali/kiali/issues/3675


mattmahoneyrh (approved):

lucasponce:
I see odd to use a different font size on the box labels on default status (no hover).
image

Visaully it produces a little bit of overhead to me, as now there are different font-sizes in default status and when hovering.
Also, the box labels have a different background + label badges, so it seems to me enough, you get the desired behaviour with no need to increase the size.

I'd recommend to keep the font size similar of the other elements, when hovering, all font sizes are increased and visually are ok.

lucasponce (approved):
One comment for your consideration but this PR is good.

jshaughn:
@lucasponce Thanks for the review.  I will take a final look at the box badge size and perhaps revert to a consistent default size while maintaining the larger hover size as a way to more easily identify the larger blocks.  Going forward I think we may need to look at condensed nodes (a node that represents an entire namespace, or cluster, or maybe app) that can be expanded/collapsed.

PR #2118, opened 2021-04-07
Navigation to remote Kialis from graph context menu and side panel

israel-hdez:
If there is an accessible remote Kiali:

mc-links

If node is remote, but there is NO accessible remote Kiali:

image

Requires back-end PR: https://github.com/kiali/kiali/pull/3853

Fixes kiali/kiali#3717


israel-hdez:
Apparently, this is ready.
This is dealing only with the context menu of the graph.

There are other links in the side panel, so I'm doing one (or more) additional PRs before marking kiali/kiali#3717 as done.

lucasponce:
Have you covered the flow: zoom into a node > double click > navigate into the details page ?
(That also will impact to this work as well)

I wouldn't create a confirmation message, that's annoying from user experience, to me, just simply opening in a different tab is enough. In a mid future, idea is that all clusters view could be seen in a single Kiali.

(I would reserve the confirmation menu on situations like the Istio Config edit, when a user has something not saved and he/she may lost some work, but otherwise, just allowing to jump in different tab would be enough).

This also has impact into the details pages that I'm actively working, as from the details view we can jump to other details, so when a node is remote, I'll follow same pattern (to jump into the remote cluster, if possible, do nothing if not visible).

I'll cc you in the changes we are doing in the overview tabs because we need to sync on them as well.

israel-hdez:
> Have you covered the flow: zoom into a node > double click > navigate into the details page ?
> (That also will impact to this work as well)
> 

It's not covered. Even double-click in the main graph is not covered.
This is touching only the context menu.

I would like to cover those flows in another PR.

> I wouldn't create a confirmation message, that's annoying from user experience, to me, just simply opening in a different tab is enough. In a mid future, idea is that all clusters view could be seen in a single Kiali.
> 

TBH, after playing with this a little, I agree that the dialog is annoying.
But I think a warning will be appreciated. So, I'll replace that annoying dialog with a simple message inside the context menu.

> (I would reserve the confirmation menu on situations like the Istio Config edit, when a user has something not saved and he/she may lost some work, but otherwise, just allowing to jump in different tab would be enough).
> 
> This also has impact into the details pages that I'm actively working, as from the details view we can jump to other details, so when a node is remote, I'll follow same pattern (to jump into the remote cluster, if possible, do nothing if not visible).
> 
> I'll cc you in the changes we are doing in the overview tabs because we need to sync on them as well.

Yes! I appreciate your help :-)

lucasponce:
> But I think a warning will be appreciated. So, I'll replace that annoying dialog with a simple message inside the context menu.

I'd recommend to follow same pattern we do with external links (Jaeger and Grafana) just add the "External" icon as the way to indicate user they exit from the app to another.

In the traces details or other you have examples, I don't think we need a new pattern here.

But sure, happy to help, let me know if you want me to review something.

jshaughn:
> I'd recommend to follow same pattern we do with external links (Jaeger and Grafana) just add the "External" icon as the way to indicate user they exit from the app to another.

I totally agree, I'd like to see the external link used here.  It, sstandard Patternfly and will go a long way to indicating the situation.

image

As an aside, I think the graph labeling may need to be enhanced, the cluster should be part of the label when the mode is in a remote cluster AND there is no cluster boxing.

israel-hdez:
@lucasponce @jshaughn Changes done!

israel-hdez:
Animation at main comment is updated.

lucasponce:
> Animation at main comment is updated.

In the animation I don't see also main links when you select a node (app/workload/service) and you can browse to those elements from the sidepanel too (from direct links, and from the kebap menu located there).

I think those also should be addressed as well.

jshaughn (changes requested):
I like the external link icons.

I would prefer a more concise message when the remote kiali is not available:
Current: `No options. This node represents a resource in <strong>{linkParams.cluster}</strong> cluster where an accessible Kiali instance couldn't be found.`

Maybe something like: `No remote links, Kiali is not available on the <strong>{linkParams.cluster}</strong> cluster. `

When my remote cluster had an older kiali (1.32.0) server installed (via Helm) a click on a remote graph node resulted in a UI crash:

```
TypeError: Cannot read property 'url' of undefined
getOptionsFromLinkParams
src/components/CytoscapeGraph/ContextMenu/NodeContextMenu.tsx:232
  229 | if (cluster.length !== 0 && cluster !== serverConfig?.clusterInfo?.name) {
  230 |   const externalClusterInfo = serverConfig.clusters[cluster];
  231 |   const kialiInfo = externalClusterInfo.kialiInstances.find(instance => instance.url.length !== 0);
> 232 |   const externalKialiUrl = kialiInfo!.url.replace(/\/$/g, '') + '/console';
      | ^  233 | 
  234 |   for (let idx = 0; idx < options.length; idx++) {
  235 |     if (options[idx].target !== '_blank') {
```

israel-hdez:
@jshaughn The UI crash is fixed.

@lucasponce I added the changes needed to also address the side panel main links and kebab menu. I pasted a screenshot of the sidepanel, but I didn't update the animation.

jshaughn (changes requested):
@israel-hdez The new message looks good.  The reported crash-on-click was resolved, and the side-panel links are gone, so that is all good.   New issues with my East-West setup:
- When on West my East-productpage is giving me a local context menu. I expected the "no kiali on 'East' cluster" message.
- When on East if I select a restricted node (unknown cluster), like PassthroughCluster, I get a crash:

image

israel-hdez:
@jshaughn Last commit should protect against the undefined value.
I hope it's fine now.

israel-hdez:
I'm checking the issue with productpage.

jshaughn (approved):
Nice!  The crash when selecting a restricted node is fixed.  The context menu and side panel links all look good for various node types and also edges.  Without kiali.io/external-url annotations (on the kiali service) I get the proper 'No Options" messaging and with annotations I get the proper links with the external link icon.  The mesh yable also looks good, no links for kiali instances w/o annotation and active links with the annotation.  Not a blocker but we could also add the external link icon on those table links.

code looks good, approving...

israel-hdez:
@mattmahoneyrh A sanity check of the graph, just to make sure this is not breaking the single cluster case.

lucasponce:
> @lucasponce I added the changes needed to also address the side panel main links and kebab menu. I pasted a screenshot of the sidepanel, but I didn't update the animation.

I missed the screenshot, perhaps it was not uploaded ?

israel-hdez:
@lucasponce 

> I missed the screenshot, perhaps it was not uploaded ?

I could swear I added it. I probably forgot to push the Save button.
Anyway, I updated the animation. I hope it looks good.

jshaughn:
> Anyway, I updated the animation. I hope it looks good.

@israel-hdez @lucasponce @mattmahoneyrh , I would like for this work to be in the 1.33 release so let's not block unless there is a major issue. Thanks.

lucasponce:
Ok from my side.

mattmahoneyrh (commented):
PR looks good.

mattmahoneyrh (approved):

israel-hdez:
Thanks! I'm merging.

PR #2119, opened 2021-04-08
Allow nav from detail view mini-graph to graph page node graph.

jshaughn:
A couple of other minor fixes:
- include aggregate nodes when checking for node change
- disable namespace selector in node graph view
- avoid race condition when unsetting node on return to main graph

Closes https://github.com/kiali/kiali/issues/3696

### DIFF
--- a/src/components/CytoscapeGraph/MiniGraphCard.tsx
+++ b/src/components/CytoscapeGraph/MiniGraphCard.tsx
@@ -17,6 +17,8 @@ import { DecoratedGraphElements, EdgeLabelMode, GraphType, NodeType } from '../.
 import CytoscapeGraph, { GraphNodeDoubleTapEvent } from './CytoscapeGraph';
 import { CytoscapeGraphSelectorBuilder } from './CytoscapeGraphSelector';
 import { DagreGraph } from './graphs/DagreGraph';
+import { GraphUrlParams, makeNodeGraphUrlFromParams } from 'components/Nav/NavUtils';
+import { store } from 'store/ConfigStore';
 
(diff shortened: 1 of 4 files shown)


jshaughn:
@mattmahoneyrh (@lucasponce ) With this change there is a new kebab option as shown in the screenshot below.  Now you can navigate directly to the main graph's node detail page (think of it as an "expand" of the mini-graph).

For QE testing try all the ways you can think of navigating around between the detail page, node graph page, full graph page, and also starting from various bookmarked URLs.  Ensure the graph types are as expected.  Also, this tangentially touched the trace overlay so just sanity check trace overlay and that a bookmarked URL with a trace overlay restores the overlay.  And of course any other evil QE usage path you may come up with.

image

lucasponce:
Let me hold this PR until I'm presenting my proposal for Overview refactoring.
It will affect this, as probably we may want to move the Detail -> Graph navigation to the page instead of the tab -> card.

jshaughn:
@lucasponce It's fine if you want to hold it, although I'm not sure the PR relies much on where the option is located.

lucasponce:
I'm reviewing this one, finally I didn't include this change in #2112 but probably in the next iteration.

lucasponce (approved):
As commented, as part of the refactoring work we're doing on this location, I think that graph links belongs to the page instead to the tab, but that can be discussed in a different issue.

It works ok and code LGTM.

pbajjuri20 (approved):
Tested and Passed on Kiali UI 1.33.0 (7bddff4b41f869a62c6caaf7ef8de6e0a53463b7)

PR #2120, opened 2021-04-08
Remove prod dep on ssri, it seems unnecessary and there is a CVE

jshaughn:
@israel-hdez Snyk complained about https://snyk.io/vuln/SNYK-JS-SSRI-1085630.  It wanted to update the ssri version but I removed it from the prod deps and it seems maybe we don't even need it.  It is still pulled in by react-scripts, but that is a dev dep.  What do you think, can we just get rid of it?

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,7 +93,6 @@
     "reselect": "4.0.0",
     "screenfull": "5.0.2",
     "seedrandom": "3.0.1",
-    "ssri": "6.0.1",
     "tippy.js": "3.4.1",
     "typesafe-actions": "3.2.1",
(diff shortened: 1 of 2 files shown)


israel-hdez (approved):
I see SSRI was added long, long ago in #187 to fix yet another CVE issue with that library.
Apparently, it was never used by Kiali.

As you say, react-scripts is a dev dep. So, I hope this change will remove the CVE warning.

FWIW, I see in the past, react-scripts were listed as a non-dev dependency, probably that was what raised the CVE warning in first place. But we are not using it in the final package. We should be safe.

:+1:

PR #2124, opened 2021-04-14
Consolidate terminology on inbound/outbound

jshaughn:
Consolidate terminology on inbound/outbound by converting instances of incoming/outgoing (both in the UI and the code).

Fixes https://github.com/kiali/kiali/issues/3873

@lucasponce This PR currently covers only the terminology change.  If we feel a need to change the Inbound traffic reporting we can expand the PR (see related comments in the issue). Setting to DNM until we have agreed on the scope.

### DIFF
--- a/src/components/CytoscapeGraph/CytoscapeTrace.ts
+++ b/src/components/CytoscapeGraph/CytoscapeTrace.ts
@@ -68,8 +68,8 @@ const showSpanSubtrace = (cy: Cy.Core, graphType: GraphType, span: Span) => {
   }
 
   // Inbound service entry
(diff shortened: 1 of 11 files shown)


jshaughn:
@lucasponce The PR has been updated to limit inbound traffic for cluster and namespace boxes to only the traffic entering the box.  It no longer counts traffic generated within the box.  As you suggested, I think this is more intuitive.  Note that the full, unbounded graph, will still count the generated traffic as inbound to the graph as a whole.  No more changes planned.

lucasponce (approved):
Tested locally.
Code changes make sense to me.
Thanks !

mattmahoneyrh (approved):

PR #2129, opened 2021-04-20
Ensure serverConfig has proper defaults

jshaughn:
Ensure serverConfig has proper defaults in case the server does not provide certain information.

Fixes https://github.com/kiali/kiali/issues/3913

### DIFF
--- a/src/components/CytoscapeGraph/ContextMenu/NodeContextMenu.tsx
+++ b/src/components/CytoscapeGraph/ContextMenu/NodeContextMenu.tsx
@@ -125,7 +125,6 @@ export class NodeContextMenu extends React.PureComponent<Props> {
     let menuOptions: React.ReactNode = null;
     // use local links if this is the home cluster, or if there is no configured home cluster
     if (!serverConfig.clusterInfo?.name || linkParams.cluster === serverConfig.clusterInfo.name) {
-      linkParams.cluster = '';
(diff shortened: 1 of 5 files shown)


lucasponce:
In a first try, it fixes the first crash but it considers everything in a multi-cluster with wrong or disabled links:
image

Which is not 100% right
image

jshaughn:
@israel-hdez Agreed, I was surprised to see that and I'm not yet sure why it is happening.  Had to break for a meeting.  Will get back to it, let me know if you find something.

israel-hdez (approved):
Code looks good.
It's working on my side.

jshaughn:
@lucasponce @mattmahoneyrh , everything looks good from Edgar's and my side.  We both tested on a non-multicluster-configured env,  and then I sanity-checked the MC env again.  Context menus, side panel links, and side-panel kebabs are all behaving as expected from what we've seen.

Assuming QE and lponce approvals we'll merge, then backport to 1.33 and release a 1.33.1.

lucasponce (approved):
Tested locally against a minikube from a development configuration (kiali running outside the cluster connecting remotelly).

Thanks for the fix.

pbajjuri20 (approved):
Tested and passed on Kiali UI 1.34.0 (63206f8a611312cd630566ebe71dd50a07a27533)

PR #2132, opened 2021-04-28
PoC Research Metrics tab main layout

aljesusg:
Relates to https://github.com/kiali/kiali/issues/3948
Backend https://github.com/kiali/kiali/pull/3956

charts
maximized

1º iteration @lucasponce Thinking about group in cards  Responses, Requests and TCP


aljesusg:
I am thinking about to have one legend for "resources" and spanduration to apply to all charts and remove the legends to have more space

aljesusg:
I worked in this new proposal with cards moving the TCP one to the down-right zone.
It makes no sense have tcp in the first line of vision. Users usually read from left to right and from top to bottom, so It makes more sense in this position

Screenshot from 2021-04-29 09-50-54

aljesusg:
new feature

Link from minigraph edge to metrics


edge_link

If edge click has same source and destination (x-server => x-server) we don't do nothing, otherwise this redirect to inbound/oubound metric tab page with only the source/dest clicked enabled.

lucasponce:
It's in the good direction, I could fetch it locally to give a try.

A first goal, the idea is to adjust dynamically the metrics, I see that the height is changed but it's again fixed, so we go back to current situation of scrolling:

image

Or:

image

To me this should be dynamic or built on resize.

That page has the advantage that height is passed from the tab on resize, so when you render you can tell each component how to present that (the overview page uses that strategy).

A couple of minor early comments as well, more in the implementation side:

- You could play with the single card instead of a card for tooling and other for cards type:
image
- Then you can also align in how the logs and traces are setting the visualization options as well.
- I think it can be interesting to add the "collapse" option in the kebap menu, and also use the same icon for expand
image

Also, use more examples, the electronic-shop combines situations where http and tcp traffic are present and it will give more good scenarios to test:

image

In summary, the work in the good direction, I'll play to use the full tab space without scrolling and this stage, also as we know the height in advanced, on smaller heights it can render one type or another of chart to show a better summary.

But nice progress, just keep experimenting with those examples and that direction, thanks.

lucasponce:
Another thought, I guess with the new relayout the expanded title to show/hide may have less sense:
image

I think it would be interesting to show the user that a metric is empty, without allow them to "expand", and this can be used in next iterations to add a descriptive message about "why" (instead to show 'empty', but that's out of the scope of this work).

aljesusg:
> It's in the good direction, I could fetch it locally to give a try.
> 
> A first goal, the idea is to adjust dynamically the metrics, I see that the height is changed but it's again fixed, so we go back to current situation of scrolling:
> 
> image
> 
So we want to remove that scroll.
> Or:
> 
> image
> 
> To me this should be dynamic or built on resize.
> 
> That page has the advantage that height is passed from the tab on resize, so when you render you can tell each component how to present that (the overview page uses that strategy).
> 
> A couple of minor early comments as well, more in the implementation side:
> 
> * You could play with the single card instead of a card for tooling and other for cards type:
>   image
> * Then you can also align in how the logs and traces are setting the visualization options as well.
> * I think it can be interesting to add the "collapse" option in the kebap menu, and also use the same icon for expand
>   image
> 
We have to set this menu for each chart to let the user maximize one of them.
> Also, use more examples, the electronic-shop combines situations where http and tcp traffic are present and it will give more good scenarios to test:
> 
> image
> 
I'll take a look to this example
> In summary, the work in the good direction, I'll play to use the full tab space without scrolling and this stage, also as we know the height in advanced, on smaller heights it can render one type or another of chart to show a better summary.
Let me play in render with the actual height to fit the space.
> 
> But nice progress, just keep experimenting with those examples and that direction, thanks.
Thanks

lucasponce:
> So we want to remove that scroll.

Mostly yes, as we have a fixed number of charts (for these tabs).
The user should know how many metrics are available and in a first look which can be interesting to zoom-in.

Note, that this can be discussed and in narrow heights (as we have that info) we can use a different chart summary if needed.

> We have to set this menu for each chart to let the user maximize one of them.

The menu can be in the options toolbar only, but yes, the icon/button should be the same per every chart, as it works today but same align.

aljesusg:
- Changed Icon aligned workload logs
- Removed Card style (One for all)
- Auto resize height calculation

Peek 2021-05-04 11-36

hhovsepy:
There is a small overlap between one graph and the right  side graph vertical axis labels.
Screenshot from 2021-05-05 10-07-05

aljesusg:
> There is a small overlap between one graph and the right side graph vertical axis labels.
> Screenshot from 2021-05-05 10-07-05

yep, working on it thanks

aljesusg:
Added gutter
Screenshot from 2021-05-05 10-24-21

src/components/Charts/Dashboard.tsx
```diff
@@ -40,27 +39,44 @@ export class Dashboard<T extends LineInfo> extends React.Component<Props<T>, Sta
     };
   }
 
+  renderCharts(charts: ChartModel[]) {
+    const heightChart = (window.innerHeight - TOP_PADDING - SPACE_CARD) / 2;
```
lucasponce, line 43: Instead to use this strategy, I suggest to look how the RenderComponentScroll is doing it in the overview tab and just use same way here.

aljesusg: It's the same strategy @lucasponce 

https://github.com/kiali/kiali-ui/blob/master/src/components/Nav/Page/RenderComponentScroll.tsx#L31-L34

lucasponce:
As commented, it's still missing to adjust the height to avoid the scroll:

image

aljesusg:
> As commented, it's still missing to adjust the height to avoid the scroll:
> 
> image

You shouldn't see the charts 2-2-4....you must to see them like 4-4, this is weird

Screenshot from 2021-05-05 10-24-21

lucasponce:
From metrics layout, I guess we could try to merge request/response throughput and request/response size on the same chart.
As are same metrics with same magnitudes.
Then perhaps it can help the visualization, and as a user I can "compare" in the same chart request/response throughput and sizes.

aljesusg:
> From metrics layout, I guess we could try to merge request/response throughput and request/response size on the same chart.
> As are same metrics with same magnitudes.
> Then perhaps it can help the visualization, and as a user I can "compare" in the same chart request/response throughput and sizes.

I agree with this we can try it maybe in a next step

lucasponce:
> I agree with this we can try it maybe in a next step

But that may condition this work as we will reduce two charts.

hhovsepy:
Still some overlap, but not critical:
Screenshot from 2021-05-05 11-00-16

aljesusg:
> > I agree with this we can try it maybe in a next step
> 
> But that may condition this work as we will reduce two charts.

ok let me do it, I need to make some changes in the backend

aljesusg:
> Still some overlap, but not critical:
> Screenshot from 2021-05-05 11-00-16

mmm I think that first I'll address the reduction of charts proposed by lucas before, this could help to increase the space and avoid overlap

aljesusg:
Hey @lucasponce here is the demo

Peek 2021-05-06 12-42

What do you think?

By default we'll have the metrics in the same chart with a distribution of 2-4 (this is because we'll have more metrics to show when we merge throughtput and size)

If user decide to keep like before the distribution will be 4-4

lucasponce:
> What do you think?

I'm going to fetch locally, but please, rebase with master, as there changes there that may affect to the work in progress.

I'll comment later.

lucasponce:
The height is not well implemented yet, the "IstioMetrics" component should delegate into the "RenderComponentScroll" for capturing the height of the tab. That info should be passed to the "Dashboard" and there you can calculate the specific height for each component on a change of properties.

But I can test some scenarios, so here are some comments.

To me, this view looks good:
image

I guess that having two rows, with 4 charts is clean (yes, joining request/response in the same chart, now I don't see it as clear as in my previous comment, thanks for the demo to test it).

Even in a narrow screen, user can see a summary of the information and decide which chart to expand for futher details:
image

I recommend you use also the whole demos, specially the travels and the electronic-shop will help you to visualize more rich scenarios:

image

From "feature/business" perspective, I think that's goal that the user can visualize all metrics to fix the problem that some metrics were hidden and a user may not know that it exists. Also user can expand a particular chart for details and correlation.

I guess it's in good direction, but now there are more UI implementation aspects to address:
- Fix the height in a proper way.
- Add a little separation between toolbar and charts (and take that into consideration when calculating height) [1]
- Discuss if the "Span duration" can be moved close to the "Expand" button, that may save some space [2]
- Check if some technique/ellipsis can be used on long legends [3] where text overlaps and it shows out of the chart.
- Consider to put the "unit" in the chart one time on top/bottom and leave units without repetitive text [4]
- On expanded charts, no problem to use the whole tab space [5]

I haven't performed more tests, I expect we will iterate on these topics to be use that it's stable.

Also, please, apply this approach to the custom metrics, in special the "Envoy Metrics" that are shipped by default on deployments (other runtime scenarios needs more steps from the user, but Envoy charts will be always there by default, so they shouldn't be missed from this works as well - at least in the distribution and aspect -).

Thanks

[1]
image

[2] 
image

[3] 
image

[4]
image

[5] 
image

aljesusg:
I am thinking about have one spans button on top instead for each chart, the spans are the same and we fetch them one time for all charts.
spans
What do you think ? I think it makes more sense 1 button for all charts

aljesusg:
What do you think about have only one button for spans? We show the same results and we only fetch them one time. 

# By chart
spans

# One button
showSpans_option_top

aljesusg:
- [X] Height Fixed
- [X] Units in axis for spans and others
- [X] Maximized chart now with full height
- [X] Increased legend for 2 items by row in case so items
- [X] Fixed toolbar
- [X] Spans options like workload logs
- [X] Added Grid for horizontal axis (I think that we don't need it for vetical what do you think @lucasponce?)


charts
maximized

lucasponce:
This PR has still problems when deploying in a cluster:
image

The style in the chart title is wrong.
The main important point of this work, which is the adjust of the height is not yet working:
image

I thought we commented that spans checkbox can go to the toolbar to enable all/none.

About the span style:
image

I refered most to sync the checkbox (in the logs is blue when checked, in the first approache of this PR was black when checked) but the black background in the "spans" title is not necesarry to bring it here, it's used in logs because the logs window is black, that's all.

Also important, Envoy Metrics are shown empty:
image

I'd like to see that the runtime metrics have no regressions as well.

lucasponce:
I haven't reviewed the impl details but the expanded view should be also calculated accordingly to show scenarios like this:
image

In this state, the user will like to enable more details, like spans, quantiles, potential groups, so height should try to show these scenarios without scroll (perhaps it's just a matter to adjust the offset and check that in most of the demo scenarios work well).

aljesusg:
> This PR has still problems when deploying in a cluster:
> image
> 
I can't understand this. I see that correctly and I tried it with different screen, let me check
> The style in the chart title is wrong.
> The main important point of this work, which is the adjust of the height is not yet working:
> image

The same
> 
> I thought we commented that spans checkbox can go to the toolbar to enable all/none.
> 
I remember that we decided to have it by chart , I redo this
> About the span style:
> image
> 
> I refered most to sync the checkbox (in the logs is blue when checked, in the first approache of this PR was black when checked) but the black background in the "spans" title is not necesarry to bring it here, it's used in logs because the logs window is black, that's all.
> 
> Also important, Envoy Metrics are shown empty:
> image

I am working on that now and in the runtimes. My first step was fix the new style
> 
> I'd like to see that the runtime metrics have no regressions as well.

aljesusg:
> I haven't reviewed the impl details but the expanded view should be also calculated accordingly to show scenarios like this:
> image
> 
> In this state, the user will like to enable more details, like spans, quantiles, potential groups, so height should try to show these scenarios without scroll (perhaps it's just a matter to adjust the offset and check that in most of the demo scenarios work well).

Let me check that example to accurate the calculation, the problem here is the resize when legend changes, that no depend of the render scroll component. I am going to play with more items by row or cut the name with a tooltip if I can't find a solution for that.

I'll ping you. Thanks @lucasponce

hhovsepy:
When resizing, the span checkbox selection is reset.

aljesusg:
Well I found how to solve the problem with height but now the deal is how to do it when legend have so items.

Screenshot from 2021-05-12 15-41-43

Should we keep the same height for all charts and reduce them? or it seems ok if we only reduce charts with more items ? (Dinamically)

Don't focus in the space of the legend I am working to fit all I can but in some cases we'll have more items

aljesusg:
Well enabled envoy tab

Screenshot from 2021-05-13 18-14-40

For this tab, my concern is when user add a new chart, I don't know if we want to fit this too. For example, up to 8 charts is fine because it would be like the inbound/outbound metrics. But what happen if the user include more metrics?

Now user can configure the spans for this in the dashboard https://github.com/kiali/helm-charts/blob/master/kiali-server/templates/dashboards/envoy.yaml 

So, we cant fit the height for the window like overview and others, and autoresize charts but I am concerned that the user set so charts...maybe we can "fit it" if the user set 8 or less and let the scroll in other cases.

lucasponce:
I'm testing the latest commit.
When spans enabled, the spans units are in the left instead to be on the right:
image

lucasponce:
The helpers lines are nice
image

But I wonder if you could test a dashed approach like the material theme proposed in Victory:

image

(I'm only interested in the horizontal, but perhaps both horizontal/vertical may help)

[1] https://formidable.com/open-source/victory/guides/themes

aljesusg:
> The helpers lines are nice
> image
> 
> But I wonder if you could test a dashed approach like the material theme proposed in Victory:
> 
> image
> 
> (I'm only interested in the horizontal, but perhaps both horizontal/vertical may help)
> 
> In some populated examples the helpers lines are strong and are on top of the graph charts, not sure if that's also happening here, so those lines should be also sent fo the background (perhaps it's inside the Victory wrapper).
> 
> [1] https://formidable.com/open-source/victory/guides/themes

Yes I can enable vertical too, no problem

aljesusg:
> I'm testing the latest commit.
> When spans enabled, the spans units are in the left instead to be on the right:
> image

i don't understand this

lucasponce:
The height is not calculated correctly:

image

It's fine that when legend is populated we need to have a decision, if we have fixed grid boxes, it's find to adjust the chart to fit the box, but in that situation legend should match in the bottom.

In the example, "cars and flights" legend line should match "avg,hotels and avg,insurances" line.

I haven't reviewed the code yet, I'll add more comments around this if I spot something.

lucasponce:
Also there is some wrong offset calculation at the bottom of the page:
image

It can be adjusted.

aljesusg:
> The height is not calculated correctly:
> 
> image
> 
> It's fine that when legend is populated we need to have a decision, if we have fixed grid boxes, it's find to adjust the chart to fit the box, but in that situation legend should match in the bottom.
> 
> In the example, "cars and flights" legend line should match "avg,hotels and avg,insurances" line.
> 
> I haven't reviewed the code yet, I'll add more comments around this if I spot something.

I thought about that...maybe we can maximized the chart when user enable percentiles to avoid this.

lucasponce:
> I thought about that...maybe we can maximized the chart when user enable percentiles to avoid this.

-1

It's going to happen when user combines other metrics, like local/remote/respose, the calculation of the chart should be correct.

aljesusg:
> > I thought about that...maybe we can maximized the chart when user enable percentiles to avoid this.
> 
> -1
> 
> It's going to happen when user combines other metrics, like local/remote/response code, the calculation of the chart should be correct.

Ok let me play with heights to fit them and we can review after the behaviour. Thanks @lucasponce

lucasponce:
Another example of this broken:

image

On corner cases, the charts + legends should fit the space of the tab.
It's understandable that if I have a fix height + weidth for element; the legend may change the chart height, but both should fit.

It's not a problem if some of this charts looks narrow, but user should have perception that the space of the tab is optimized.

lucasponce:
image

aljesusg:
Please @lucasponce let me make changes and test, I'll ping you after them. I don't want waste your time.

aljesusg:
Ok I did some improvements in the legend calculation:

- truncate items if we have more than 7 in the legend
- Fix the space with so items
- increase the items by row and boxSizes

Peek 2021-05-24 11-09


What do you think @lucasponce ?

lucasponce:
I will test later but just looking the screenshot 
image
Shows that the scrolling is not addressed.

aljesusg:
> I will test later but just looking the screenshot
> image
> Shows that the scrolling is not addressed.

yep I enabled it again to play with the legend. I'll redo that, I want to be sure that legend scale correctly

aljesusg:
Added Vertical Grid.

Screenshot from 2021-05-25 09-40-52

ChartAxis use PF theme when grid is true, so to use VictoryTheme Material we need to migrate this to VictoryAxis of victory library

lucasponce:
> Please @lucasponce let me make changes and test, I'll ping you after them. I don't want waste your time.

@aljesusg let me know when the PR is ready to continue testing this PR.

aljesusg:
> > Please @lucasponce let me make changes and test, I'll ping you after them. I don't want waste your time.
> 
> @aljesusg let me know when the PR is ready to continue testing this work.

We can review it. I tested it with some examples. Other task is the envoy tab and how to deal with it. We can address it in another PR.

lucasponce:
> Other task is the envoy tab and how to deal with it. We can address it in another PR.

I'm not sure about this.

Please rebase the backend + frontend PRs for this work, I'll start to review the comments I put in the previous revision.

lucasponce:
This is not ready to review:
image
I don't know which tests are you doing, but not the basic on the demo ones.

When the basic scope of the feature is tested we can review from Dev and QE.

aljesusg:
> This is not ready to review:
> image
> I don't know which tests are you doing, but not the basic on the demo ones.
> 
> When the basic scope of the feature is tested we can review from Dev and QE.

let me check because i can't reproduce this

lucasponce:
image

image

aljesusg:
Screenshot from 2021-05-27 22-24-41
Screenshot from 2021-05-27 22-24-23
Screenshot from 2021-05-27 22-23-58

aljesusg:
All the data is mocked to test that Dashboards are rendering correctly.

# Storybook
## Play with ChartHeight

in show addons you can set the chartheight to check how the dashboard change. (See demo)
Peek 2021-06-02 11-27

## Types

### Inbound/outbound metrics
Screenshot from 2021-06-02 11-24-46
### With All percentiles
Screenshot from 2021-06-02 11-24-57
### With spans
Screenshot from 2021-06-02 11-25-08
### Envoy Metrics
Screenshot from 2021-06-02 11-26-12

lucasponce:
The CI is broken with a problem that also affects locally when trying to start a debug session:

```
There might be a problem with the project dependency tree.
It is likely not a bug in Create React App, but something you need to fix locally.

The react-scripts package provided by Create React App requires a dependency:

  "babel-loader": "8.1.0"

Don't try to install it manually: your package manager does it automatically.
However, a different version of babel-loader was detected higher up in the tree:

  /home/lponce/Projects/repos/kiali-ui/node_modules/babel-loader (version: 8.2.2) 
```

aljesusg:
> The CI is broken with a problem that also affects locally when trying to start a debug session:
> 
> ```
> There might be a problem with the project dependency tree.
> It is likely not a bug in Create React App, but something you need to fix locally.
> 
> The react-scripts package provided by Create React App requires a dependency:
> 
>   "babel-loader": "8.1.0"
> 
> Don't try to install it manually: your package manager does it automatically.
> However, a different version of babel-loader was detected higher up in the tree:
> 
>   /home/lponce/Projects/repos/kiali-ui/node_modules/babel-loader (version: 8.2.2) 
> ```

I was fixing it :) the problem was the storybook and react-scripts. Added it in resolutions. Rebased with master too.

lucasponce:
I could build and test the PR.

This is one stopper:

image

Legend should be built within the space of the chart, otherwise in some narrow situation will overlap or overflow without control as it's visible here:

image

I think this can be addressed and tested without running a complete flow but using the storybooks tests and just working with similar scenarios.

aljesusg:
> I could build and test the PR.
> 
> This is one stopper:
> 
> image
> 
> Legend should be built within the space of the chart, otherwise in some narrow situation will overlap or overflow without control as it's visible here:
> 
> image
> 
> I think this can be addressed and tested without running a complete flow but using the storybooks tests and just working with similar scenarios.

I was thinking about this ....

- Are users  going to reduce to that width.?
- The solution to that is calculate the width vs the space of the legend. (I can do it). In this case, items by row in legend will be 2 or 1, and in this case the chart sill be so small. I am going to address this solution to check the behavior.

lucasponce:
The solution should be that the chart should be rendered in the height and width given.
In corner cases is ok that graph can be smaller, that's something for a follow-up task, to change the representation in a good way.
But the way that legend is rendered is not correct.

I don't think that the implementation is addressing this.

aljesusg:
Ok Fixed, now the legend fit the width
Screenshot from 2021-06-03 08-58-41

The problem here is that now when the legend is so big so the chart moves
Screenshot from 2021-06-03 08-58-58

I am thinking about to reduce the px of fonts or maybe add some scroll to fix this

aljesusg:
Ok, What do you think in this 

Screenshot from 2021-06-03 09-20-57

lucasponce:
I'm still seeing some issues that would require more followup work, but I think the PR could be consolidate if we solve two issues:

In a 1280x1024 window, the collapse menu actions trigger a bad resize of the components:
image
image

Onclick event in the legend doesn't refresh the on/off color status:
image

There are more things related the style, how legend is being built, but I'll create a separated issue for that.

The main business goal of this work is addressed and also the adding of the "storybook" tests will help to modify and these components better by other devs.

lucasponce:
@hhovsepy @prachiyadav as discussed, I think this work could be consolidated if no major regression tests.

There will be a followup work, so, if this may break the integration tests, we can run a basic testing here and wait to the next work, but I think that it's mostly ok, followup changes can be related to the style/CSS/adjustment but no in terms of basic features.

So let me know if you prefer to test this or wait for next one.
I've tested both in local + deployed version, but another sanity check to validate that nothing is missing would be great.

aljesusg:
I am going to try to fix theses things, i am researching about a method of victoreCore to  getDimensions of legends based in some props, I'm triying to calculate the height and width that legend will have to fit the space with the correct fontSize, let me some time

hhovsepy (approved):
Verified.

aljesusg:
Ok, As far as I know ...it seems that the render with PF4 is not correctly resizing the Victory components, I checked the documentation and these components are 100% resizable so...I'm going to test to migrate all to victory

aljesusg:
> I'm still seeing some issues that would require more followup work, but I think the PR could be consolidate if we solve two issues:
> 
> In a 1280x1024 window, the collapse menu actions trigger a bad resize of the components:
> image
> image
> 
> Onclick event in the legend doesn't refresh the on/off color status:
> image
> 
> There are more things related the style, how legend is being built, but I'll create a separated issue for that.
> 
> The main business goal of this work is addressed and also the adding of the "storybook" tests will help to modify and test these components better by other devs.

Hey @lucasponce I fixed the render when the menu is collapsed, it seems that the svg rendered by victory not applied the style correctly.

About the legendItem I can disable/enable items and Is ee the gray colors in disable. I'm going to deploy this to check in the build.

aljesusg:
Fixed the legendItem color and the render in collapse menu

lucasponce:
Runtime metrics are not adjusted, at least, it requires a minimun layout.

Kiali, Go Metrics and Internal Metrics:
image

JVM Metrics:
image

Node.js:
image

I have more comments, but that will be addressed in a different task.

I'd like to have these minimul layouts in a good state and I'll consolidate this work.

aljesusg:
> Runtime metrics are not adjusted, at least, it requires a minimun layout.
> 
> Kiali, Go Metrics and Internal Metrics:
> image
> 
> JVM Metrics:
> image
> 
> Node.js:
> image
> 
> I have more comments, but that will be addressed in a different task.
> 
> I'd like to have these minimul layouts in a good state and I'll consolidate this work.

yeah we talked about this :) I did the changes only for envoy tab not for runtime because we didn't know how many charts we are going to receive. The changes are small but I don't know how to deal for example with 9-10 charts.

## Options
A) Let the scroll for that cases
B) reduce the size or charts to fit all of them in 2 rows
C) Fit charts if the number is <= 8 and keep like now for more charts.

In all cases we'll remove the "spans" options in the CR Dashboards.

lucasponce:
@aljesusg it's ok.

Please, then don't add more commits to this PR.
That will be added in a followup work.

I'll wait that @prachiyadav can take a look to this work.

prachiyadav (approved):

lucasponce (approved):
As commented, we are going to consolidate this work.
There are some tasks to address around style + runtime metrics that will be addressed in a followup PR.

PR #2134, opened 2021-04-28
Improve handling when adding cluster discriminators to graph node labels

jshaughn:
Some Istio versions set cluster='Kubernetes' as a default in the telemetry, but do not set it as a default on the control plane itself.

Fixes https://github.com/kiali/kiali/issues/3955


jshaughn:
@israel-hdez For your review please confirm that things still work as expected when configured for MC.

israel-hdez (approved):
It works on the MC setup.
Let's hope nobody will use  `Kubernetes` as cluster id on a real MC setup :-)

lucasponce (approved):
image

PR #2135, opened 2021-04-29
Bump node-sass to 5.0

nrfox:
Bumps node-sass version to `5.0`. Prior versions of node sass do not compile on systems without python2 due to [this](https://github.com/sass/node-sass/issues/2877) issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -116,7 +116,7 @@
     "husky": "1.3.1",
     "jest-canvas-mock": "2.2.0",
     "jest-localstorage-mock": "2.4.2",
-    "node-sass": "4.13.1",
+    "node-sass": "5.0.0",
     "prettier": "2.0.5",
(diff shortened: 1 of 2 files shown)


jshaughn (approved):
Seems fine, just a dev dependency.   I guess what you are saying is that without this we required Python 2 to build?

nrfox:
> Seems fine, just a dev dependency. I guess what you are saying is that without this we required Python 2 to build?

Yes exactly. If you alias `python=python3` you should see this same issue if you've got Python 2 on your system.

israel-hdez (approved):
I invoked

`yarn && yarn build && yarn start`

It works. As long as it works, I'm happy :-)

jshaughn:
@nrfox You can merge this, as just a change to dev deps it does not require QE.

PR #2136, opened 2021-05-03
Fix wrong render of workload list

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3973

The list of workloads is matched with the list of workloads provided by health data, so there was a bug in the match code.


hhovsepy (approved):
solution lgtm
Screenshot from 2021-05-03 15-38-09

prachiyadav (approved):
Screenshot from 2021-05-03 20-57-14

PR #2137, opened 2021-05-03
Refine message for remote Kiali

israel-hdez:
This standardizes the behavior of the graph context menu when there are no available options to show. Concretely, no options because of a node living in a remote cluster where an accessible Kiali instance couldn't be found.

When Kiali UI boots, it will check which clusters does NOT have accessible Kialis. A message will be logged in the message center telling about this fact and that no links to these clusters will be available. In the graph, context menus for relevant nodes won't be present.

image

Fixes kiali/kiali#3907

### DIFF
--- a/src/app/AuthenticationController.tsx
+++ b/src/app/AuthenticationController.tsx
@@ -19,6 +19,7 @@ import { TLSStatus } from '../types/TLSStatus';
 import { MeshTlsActions } from '../actions/MeshTlsActions';
 import { AuthStrategy } from '../types/Auth';
 import { JaegerInfo } from '../types/JaegerInfo';
+import { ServerConfig } from '../types/ServerConfig';
 import { LoginActions } from '../actions/LoginActions';
 import history from './History';
 import { NamespaceActions } from 'actions/NamespaceAction';
@@ -183,6 +184,7 @@ class AuthenticationController extends React.Component<AuthenticationControllerP
       this.props.setNamespaces(configs[0].data, new Date());
       setServerConfig(configs[1].data);
       this.applyUIDefaults();
+      this.checkConfiguredRemoteKialis(configs[1].data);
 
       if (this.props.landingRoute) {
         history.replace(this.props.landingRoute);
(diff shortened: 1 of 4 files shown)


israel-hdez:
@abonas I'm hoping this is better.

jshaughn (changes requested):
I'm sorry but I absolutely do not want to see a toast message every time I navigate to the graph.  In general I don't like proactive warnings, they are too distracting.

To be honest. I was totally fine with the current behavior/message.  Or, the message could be changed based on the original issue complaint, perhaps, `Cluster {linkParams.cluster} not configured for remote Kiali links.`.  

At most the current message would be OK if shown one time.  Personally, I'd still rather not see a Toast and rather a silent addition to the Message Center.  Remember that it may be totally intended to not have remote linking.

As for the current message, if it stays it would be nice for it to perhaps mention the Mesh page to see which Kiali instances are configured for remote links.

israel-hdez:
FWIW, I also prefer contextual help because feedback is more natural. And IMHO toast messages cannot be easily associated with the context menu and people may think that its absence could be a bug.

I'd be happy to revert all this and do a simple message change to `No options. No Kiali configured for <name> cluster.`

But I also like consistency and I admit that this is the pattern that has been used in the past. I'm not sure we will be able to easily reach a consensus about what pattern to use. 

So, TBH, I think that if nobody complains, I'll end up reverting this and doing a simple message change, to go inline with what is present in v1.33.

jshaughn:
@israel-hdez Let's get opinions for the other interested people.  Fwiw, I don't think `No options. No Kiali configured for <name> cluster` solves the original issue.  Alissa's problem was that the current message makes it sound like there is no Kiali on <name>. 
But, there may be a Kiali that is not configured with an external url.

lucasponce:
Note that an external Kiali is not different that an external Jaeger where these use cases are deployed for some time and experience should be similar.

So, my suggestion (just how External links integration works):
- If a external link is not available, send a one time toast informative message.
- Do not render the link (on this case, in case of the link you can use a short - not available  or similar wording -)..
- If the external link is broken (and you can detect that), send a warning toast if it makes sense.

In case of doubt, check how external links with Grafana and Jaeger works and try to align then but this is not a new use case that needs to be re-implemented, Kiali connects with an external app (another Kiali) and a link will open a new tab/window, that's all.

abonas:
My original issue was about the text itself and not about the behavior.
But if we're at it :),
Then I +1 for a one time message displayed (and not repeating), and indeed follow the rest of external links for consistency.

jshaughn:
If we must have a toast then one time only, not one time per graph mount, preferably at the same time as we would display the jaeger.  Also, one message for all discovered Kiali instances not configured for external linking, not one message per instance.

israel-hdez:
I moved the toast to be shown when the UI loads. The toast will be shown once and it won't popup again unless a browser refresh is done.

I updated the image in the main comment to show the toast.
The Graph page won't show context menus if the remote Kiali is not accessible.

I hope this is fine.

lucasponce (approved):
LGTM
I haven't tested the multi-cluster config, only code review and a smoke test on a local single cluster.

src/app/AuthenticationController.tsx
```diff
@@ -257,6 +259,30 @@ class AuthenticationController extends React.Component<AuthenticationControllerP
...
+            clustersWithoutKialis.push(cluster);
+          }
+        }
+      }
+
+      if (clustersWithoutKialis.length > 0) {
+        AlertUtils.addWarning(
+          "Not all remote clusters have reachable Kiali instances.",
+          undefined, undefined,
+          "Context menus are diasabled for remote cluster nodes without a discovered Kiali or if it is not configured for remote links."
```
jshaughn, line 280: diasabled -> disabled

Actually, a slight recommended change:

Context menus are disabled for remote cluster nodes if a Kiali instance is not discovered, or if the remote Kiali is not configured with an external URL.

jshaughn (changes requested):
minimally, a typo to be fixed.  Also, a recommended change to the detail text.

src/app/AuthenticationController.tsx
israel-hdez: It's now fixed.

PR #2139, opened 2021-05-05
 Application Traffic - some outbound traffic is not shown

jshaughn:
Fix regression in Application detail's Traffic tab.  The fix maintains the change to a versionedapp graph for application mini-graphs (the cause of the problem) but restores the traffic tab to its original behavior. 

While investigating I ended up fixing a few things that, in the end, did not contribute to the reported problem:
- fix target node selection for application nodes for node detail and application mini-graphs
- fix two double-loading issues for the graph, one related to node graphs and the other related to bookmarked graphs

Fixes https://github.com/kiali/kiali/issues/3974

### DIFF
--- a/src/components/CytoscapeGraph/CytoscapeGraph.tsx
+++ b/src/components/CytoscapeGraph/CytoscapeGraph.tsx
@@ -180,7 +180,7 @@ export default class CytoscapeGraph extends React.Component<CytoscapeGraphProps>
       // pre-select node if provided
       const node = this.props.graphData.fetchParams.node;
       if (node && cy && cy.$(':selected').length === 0) {
-        let selector = "[nodeType = '" + node.nodeType + "']";
+        let selector = `[namespace = "${node.namespace.name}"][nodeType = "${node.nodeType}"]`;
         switch (node.nodeType) {
           case NodeType.AGGREGATE:
             selector =
--- a/src/components/CytoscapeGraph/CytoscapeGraph.tsx
+++ b/src/components/CytoscapeGraph/CytoscapeGraph.tsx
@@ -7,6 +7,7 @@ import ReactResizeDetector from 'react-resize-detector';
 import { GraphData } from 'pages/Graph/GraphPage';
 import { IntervalInMilliseconds, TimeInMilliseconds } from '../../types/Common';
 import {
+  BoxByType,
   CLUSTER_DEFAULT,
   CytoscapeBaseEvent,
   CytoscapeClickEvent,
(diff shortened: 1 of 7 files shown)


hhovsepy (approved):
Verified the issue.

lucasponce:
I tested this PR and it looks ok from my side.
I have a corner case that not sure if this in PR or also in master around the navigation from the minigraph.

In the hotels-v1 or other workloads in the travel-agency (the "inner" workloads between the "travels" and "mysql") had some issue while navigating, I suspect perhaps same regression I introduced during the refactoring:

image

In the rest of the cases it works just fine, so not a blocker, I'd like to consolidate this and resolve the issue later if it grows.

jshaughn:
> I have a corner case that not sure if this in PR or also in master around the navigation from the minigraph.

@lucasponce, it looks like that mini-graph nav issue is in master as well, I'll see if I can fix it as part of the pr.

jshaughn:
@lucasponce , I fixed the click-nav issue.

@pbajjuri20 , no more commits planned so please perform final QE.   Note that in addition to the original issue there were a couple of other fixes:
- When selecting app detail the app-box of the target app should now be the initially selected/highlighted node in the minigraph.
- In a workload detail minigraph, when clicking on a different node, particularly a different workload node, you will nav to the new workload detail, and now the minigraph should update promptly.  Previosuly it was leaving the old minigraph until there was a refresh.

pbajjuri20:
@jshaughn: Issues are perfectly fixed.

before:  Application Mini-Graph

before_PR_2139

after:  Application Mini-Graph

PR_2139_1

Rating Outbound Traffic in Application

PR_2139

Selecting the different Workload in Mini-Graph

Peek 2021-05-05 12-59

pbajjuri20 (approved):

pbajjuri20:
@jshaughn: Thanks for identifying that new issue,(the newly-tapped node is not selected in the new graph) It is resolved with a new commit.


Peek 2021-05-05 14-37

pbajjuri20 (approved):

lucasponce:
I couldn't test this PR.
I think something has changed, I see a minor regression (probably it was before).
When an action is finished (i.e. in the service details) the overview tab should trigger a refresh.

This will update the service details and also the Istio Config, BUT I think it's not happening in the minigraph, that has a side effect that if user has created some new config, the minigraph doesn't refresh the badges.

jshaughn:
@lucasponce created https://github.com/kiali/kiali/issues/3983 for your issue above.

PR #2141, opened 2021-05-06
Protect against code-drop-through

jshaughn:
Fixes https://github.com/kiali/kiali/issues/3982

@mattmahoneyrh IMO, no QE required for this non-reproducible fix. Just assigning as an FYI.


jmazzitelli (approved):
lgtm

PR #2142, opened 2021-05-06
Fix mini-graph update after detail page Actions

jshaughn:
- also maintains previous fix for update after mini-graph navigation

Fixes https://github.com/kiali/kiali/issues/3983

### DIFF
--- a/src/pages/AppDetails/AppInfo.tsx
+++ b/src/pages/AppDetails/AppInfo.tsx
@@ -46,7 +46,7 @@ class AppInfo extends React.Component<AppInfoProps, AppInfoState> {
     if (
       this.props.duration !== prev.duration ||
       this.props.lastRefreshAt !== prev.lastRefreshAt ||
-      this.props.app?.name !== prev.app?.name
+      this.props.app !== prev.app
     ) {
       this.fetchBackend();
     }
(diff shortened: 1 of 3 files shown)


jshaughn:
@lucasponce I've updated the PR.  See the commit comment for details about the root cause of the double-fetch.

lucasponce:
Thanks @jshaughn, I've tested and I didn't find any major problem but I'm going to cc @hhovsepy @skondkar as there are several scenarios to test, this refresh can introduce regressions quite easily and I remember I introduced a couple in the past.

@hhovsepy @skondkar would you mind to check if I didn't miss anything ? In special I'd like to check following scenarios:
- User navigates into the details pages App/Workload/Service and a change in the duration and refresh works correctly in all tabs (overview, traffic, logs, metrics, traces).
- When user generates some configuration using Wizards, page is refreshed, then i.e. when a new virtualservice is created minigraph is updated with the proper badge.
- No suspicious queries from the browser (we detected a duplication of the graph query due the bug that Jay has fixed).

In theory all tabs are now connected in a same way, but in the past there was some hard bugs to detect due the page was refreshed but not the content of the tab.

Thanks !

skondkar (approved):
Verified as per the testing steps. LGTM
Verified on Kiali UI :   1.35.0 (d8f352da00e7c5c03973f897ecf6978b2e7d6b19)

hhovsepy (approved):
Verified.

PR #2146, opened 2021-05-07
Fix ServiceEntry graph links

nrfox:
ServiceEntries should link to the namespace where they are defined which isn't always the node namespace.

Relates to: https://github.com/kiali/kiali/issues/3798
Requires backend changes: https://github.com/kiali/kiali/pull/3980

These ServiceEntry links were broken before when the ServiceEntry definition was outside of the graph node's namespace but should now properly direct to the Istio Config of the Service Entry.

LinkDetailsScreenshot

NodeSummaryDetailsLink

### DIFF
--- a/src/components/CytoscapeGraph/ContextMenu/NodeContextMenu.tsx
+++ b/src/components/CytoscapeGraph/ContextMenu/NodeContextMenu.tsx
@@ -56,7 +56,7 @@ type LinkParams = { cluster: string; namespace: string; name: string; type: stri
 export class NodeContextMenu extends React.PureComponent<Props> {
   static derivedValuesFromProps(node: DecoratedGraphNodeData): LinkParams | undefined {
     const cluster: string = node.cluster;
-    const namespace: string = node.namespace;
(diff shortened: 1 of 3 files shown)


nrfox:
Note that the tooltip still has the namespace as `bookinfo`:

Screenshot from 2021-05-07 16-07-11

Even though that's not where the ServiceEntry object exists at, it is still the namespace of the node in the graph at present. I went back and forth on updating the tooltip but I'm happy to change if others think that makes more sense.

jshaughn:
I would leave the Tooltip alone.  We're presenting the ServiceEntry as the "instance" that is in the bookinfo namespace.  Only the definition (base service entry) is in the default namespace.  Until users complain I think it's fine, and consistent, as-is.

jshaughn (approved):
LGTM

mattmahoneyrh (commented):
After applying external service: https://github.com/kiali/kiali/blob/master/tests/e2e/assets/bookinfo-ext-service-entry.yaml

The following errors occurs when clicking on External Service 'Details' link:
    "Could not fetch Istio object type [serviceentries] name [external-service-foo-se]"

mattmahoneyrh (approved):

PR #2150, opened 2021-05-10
Add namespace to istio error toaster message

nrfox:
Adds the namespace to the toaster error message that appears when an Istio config object does not exist.

Before:
Before

After:
After

Relates to kiali/kiali#3798


pbajjuri20:
Tested and Passed on Kiali UI 1.35.0 (aa626b6923f8b5bce23a5ca63130f159d6f87f2b)

SE

pbajjuri20 (approved):

PR #2151, opened 2021-05-10
Don't include the home cluster among remote clusters when checking for external URLs

jshaughn:
We should not generate a warning toast unless we discover at least one remote cluster without a Kiali configured with an external URL.

Fixes https://github.com/kiali/kiali/issues/3995


jmazzitelli (approved):
I no longer see the warning message popup

pbajjuri20 (approved):

PR #2153, opened 2021-05-11
Bump deps

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3998

Fixes https://github.com/kiali/kiali-ui/pull/2152
Fixes https://github.com/kiali/kiali-ui/pull/2149
Fixes https://github.com/kiali/kiali-ui/pull/2148
Fixes https://github.com/kiali/kiali-ui/pull/2144
Fixes https://github.com/kiali/kiali-ui/pull/2143
Fixes https://github.com/kiali/kiali-ui/pull/2097

This PR bumps several dependencies to address the warnings generated from the bots.
A local smoke test seems ok in major features but it would require a regression test to ensure nothing is broken.

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "eventemitter3": "4.0.0",
     "js-yaml": "3.13.1",
     "json-beautify": "1.0.1",
-    "lodash": "4.17.20",
+    "lodash": "4.17.21",
     "m-react-splitters": "1.2.0",
     "moment": "2.24.0",
     "react": "16.13.1",
@@ -128,9 +128,13 @@
   "resolutions": {
     "ajv": "6.12.3",
     "dot-prop": "4.2.1",
+    "elliptic": "6.5.4",
+    "handlebars": "4.7.7",
     "hoist-non-react-statics": "3.3.0",
+    "hosted-git-info": "2.8.9",
     "ini": "1.3.7",
     "kind-of": "6.0.3",
+    "url-parse": "1.5.1",
     "victory": "34.3.12",
     "y18n": "5.0.5"
   },
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
     "@types/cytoscape": "3.14.0",
     "@types/enzyme": "3.10.5",
     "@types/jest": "23.3.10",
-    "@types/lodash": "4.14.150",
+    "@types/lodash": "4.14.169",
     "@types/node": "12.12.14",
     "@types/react": "16.9.34",
     "@types/react-dom": "16.9.7",
@@ -134,6 +134,7 @@
     "hosted-git-info": "2.8.9",
     "ini": "1.3.7",
     "kind-of": "6.0.3",
+    "lodash": "4.17.21",
     "url-parse": "1.5.1",
     "victory": "34.3.12",
     "y18n": "5.0.5"
--- a/yarn.lock
+++ b/yarn.lock
@@ -3512,10 +3512,15 @@ bluebird@^3.5.5:
   resolved "https://registry.yarnpkg.com/bluebird/-/bluebird-3.7.1.tgz#df70e302b471d7473489acf26a93d63b53f874de"
   integrity sha512-DdmyoGCleJnkbp3nkbxTLJ18rjDsE4yCggEwKNXkeV123sPNfOCYeDoeuOY+F2FrSjO1YXcTU+dsy96KMy+gcg==
 
-bn.js@^4.0.0, bn.js@^4.1.0, bn.js@^4.1.1, bn.js@^4.4.0:
+bn.js@^4.0.0, bn.js@^4.1.0, bn.js@^4.1.1:
   version "4.11.9"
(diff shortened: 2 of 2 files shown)


aljesusg (approved):
build correctly 

```bash
File sizes after gzip:

  1004.07 KB (+103.03 KB)  build/static/js/2.144df867.chunk.js
  244.82 KB (+6.55 KB)     build/static/js/main.37951fa8.chunk.js
  74.38 KB (-515 B)        build/static/css/2.0c18ac1a.chunk.css
  24.44 KB (-151 B)        build/static/css/main.77964c1a.chunk.css
  735 B                    build/static/js/runtime-main.df90d992.js

```
I check views and it seems correctly.

I didn't see anything url-parse that affect us.

In handlebars there are so changes https://github.com/handlebars-lang/handlebars.js/blob/master/release-notes.md . cytoscape use handlebars 4.7.6 

LGFM

israel-hdez (changes requested):
`yarn why lodash` tells that lodash v4.17.20 is still being used.
So,https://github.com/kiali/kiali-ui/pull/2152 and https://github.com/kiali/kiali-ui/pull/2148 are not fully covered.

I think you would need to do the same process of adding lodash to `resolutions` to properly update the `yarn.lock` file and then remove it.

jshaughn:
> yarn why lodash tells that lodash v4.17.20 is still being used.

That's weird, I would not have guessed that a version other than the once declared would get used.  I guess that's for the direct dependency to kiali but that sub-deps still honor the older deps if not forced by resolutions?

> I think you would need to do the same process of adding lodash to resolutions to properly update the yarn.lock file and then remove it.

By adding and removing resolution entries doesn't it make it sort of hard to know what we've actually pinned in yarn.lock?   I guess comments aren't really supported so I guess there isn't much we can do outside of a commit comment.

israel-hdez:
> > yarn why lodash tells that lodash v4.17.20 is still being used.
> 
> That's weird, I would not have guessed that a version other than the once declared would get used. I guess that's for the direct dependency to kiali but that sub-deps still honor the older deps if not forced by resolutions?
> 

I think that yarn tries to keep as much as possible whatever versions are in the lockfile. If you start clean (without a `yarn.lock` file) it can use the latest versions of all deps. But then, when you want/need to update a sub-dep, it's not very trivial, because yarn does not have a command that lets you update individual sub-deps :-/

Forcing a version with `resolutions` is one way to force the upgrade, but you may end with an incompatibility because it forces the same version of a package for all deps. For example, one dep may want package `foo-v1.y.z` and another dep may want package `foo-v2.y.z` (same package, different major). If you use resolutions, you will force either `v1` or `v2` of `foo`. Then, because of semver semantics, the API of `foo` can change and one of your deps may break because of the unexpected API.

The method that I like to use for the upgrade is the one mentioned in [this post](https://medium.com/@ayushya/upgrading-javascript-packages-deep-dependencies-using-yarn-8b5983d5fb6b); in short: in the `yarn.lock` file you manually delete the lines that are "locking" a version of a package (I like to think that I'm "unlocking it"). Then, run `yarn install` to let it (re-)resolve the versions that are compatible with the specifiers of each dep.

> > I think you would need to do the same process of adding lodash to resolutions to properly update the yarn.lock file and then remove it.
> 
> By adding and removing resolution entries doesn't it make it sort of hard to know what we've actually pinned in yarn.lock? I guess comments aren't really supported so I guess there isn't much we can do outside of a commit comment.

Isn't that the goal of the lockfile? to pin/lock versions for replicatable builds? :-/ (I may be misunderstanding)

Anyway... for this specific PR, I think we just want to move to patched packages rather than pinning. The "resolutions" have specific use cases that I don't think are applicable here. See https://classic.yarnpkg.com/en/docs/selective-version-resolutions/.  This is why I want to encourage keeping the `resolutions` list clean. I reckon that it's easier to use "resolutions" to force the upgrade rather than finding what lines to delete in the `yarn.lock` file. So, I think it's OK to use "resolutions" to upgrade and revert/cleanup if there are no incompatibilities -- it's harder to do the cleanup later!.

Indeed, we do have some old "resolutions" incompatibilities that have been there for a long time. See these lines in the CI run: https://github.com/kiali/kiali-ui/pull/2153/checks?check_run_id=2554736607#step:5:7, and we were OK when we introduced those "incompatibilities" because we tested and nothing was broken. I see no new incompatibilities in CI, so we should be good if these new `resolutions` are reverted.

lucasponce:
> I think you would need to do the same process of adding lodash to resolutions to properly update the yarn.lock file and then remove it.

Thanks @israel-hdez, good catch, yes, there are some nested deps that were pulling the old version.
I'll update the PR.

I personally don't want to put the 3rd party dependencies into the first line / direct dependencies, otherwise is pretty noisy.
The resolutions was a tradeoff to make more clear to the user what deps have the app, and how to deal with the potential edge cases.

But we have +1300 deps so, this will be an ongoing discussion, let's patch the resolutions and if we find that it's not enough we'll figure out a better way.

israel-hdez (approved):
LG!

prachiyadav (approved):

hhovsepy (approved):
Full regression passed.

PR #2155, opened 2021-05-12
Update to using new isProxy flag to identify proxy containers for logs view

jshaughn:
Client-side changes utilizing the new flag to identify when the log view should treat logs as proxy-logs.

PartOf https://github.com/kiali/kiali/issues/3977

Requires Server PR: https://github.com/kiali/kiali/pull/3981

### DIFF
--- a/src/pages/WorkloadDetails/WorkloadPodLogs.tsx
+++ b/src/pages/WorkloadDetails/WorkloadPodLogs.tsx
@@ -469,36 +469,6 @@ class WorkloadPodLogs extends React.Component<WorkloadPodLogsProps, WorkloadPodL
         >
           {this.hasEntries(this.state.filteredLogs)
             ? this.state.filteredLogs.map((le, i) => {
-                /*
-              if (!le.accessLog && le.message.startsWith('[2021')) {
-                le.accessLog = {
-                  authority: 'foo',
-                  bytes_received: 'foo',
-                  bytes_sent: 'foo',
-                  duration: 'foo',
-                  forwarded_for: 'foo',
-                  method: 'foo',
-                  protocol: 'foo',
-                  request_id: 'foo',
-                  response_flags: 'foo',
-                  status_code: 'foo',
-                  tcp_service_time: 'foo',
-                  timestamp: 'foo',
-                  upstream_service: 'foo',
-                  upstream_service_time: 'foo',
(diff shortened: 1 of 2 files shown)


src/pages/WorkloadDetails/WorkloadPodLogs.tsx
```diff
@@ -731,27 +701,28 @@ class WorkloadPodLogs extends React.Component<WorkloadPodLogsProps, WorkloadPodL
...
+    });
+    let appContainers = 0;
     let containers = podContainers.map(c => {
       const name = c.name;
-      const displayName = c.name;
+      if (c.isProxy) {
+        return { color: proxyContainerColor, displayName: name, isProxy: true, isSelected: true, name: name };
+      }
+      const version = pod.appLabel && pod.labels ? pod.labels[serverConfig.istioLabels.versionLabelName] : undefined;
+      const displayName = !version ? name : `${name}-${version}`;
```
lucasponce, line 720: I see this confusing:
image

You're appending a version label to the container name and that's not really true.
I'd keep with the container name because it's what I expect to see as a user:

```
      containers:
        - name: cars
          image: quay.io/kiali/demo_travels_cars:v1
```

Appending the version at the container level is not helping here.
(Note that pod is always visible)

lucasponce:
Feature looks ok:
image

image

lucasponce (changes requested):
A requested change to avoid confusion between Pod vs Container names, but feature looks good.

src/pages/WorkloadDetails/WorkloadPodLogs.tsx
jshaughn: I'm not sure, I think that is a holdover from long ago and not new to this PR.  But I agree with you, I'm not sure why it's like that and I'll change it.

jshaughn:
@mattmahoneyrh The primary before/after takeaway here is that logs for istio gateways (istio-ingressgateway, istio-egressgateway) should be treated like envoy logs, not application container logs.  That means in the "before", the workload logs tab would show standard log entries (in white, no clickable info icon), and "after" they should be treated as envoy logs (gold, clickable info icon).

lucasponce (approved):

pbajjuri20 (approved):

PR #2156, opened 2021-05-12
Add sourceprincipal and destprincipal operands to Graph Find/Hide.

jshaughn:
Note that use of these operands will auto-enable the Security Display option as it is required for populating the principal fields.

Closes https://github.com/kiali/kiali/issues/4000


jmazzitelli (approved):
lgtm

mattmahoneyrh (approved):

PR #2157, opened 2021-05-13
[v1.34] Don't include the home cluster among remote clusters when checking for external URLs

israel-hdez:
Back-port of https://github.com/kiali/kiali-ui/pull/2151


jshaughn (approved):
LG

PR #2158, opened 2021-05-17
Better CSS fix for the ACE editor

lucasponce:
Fixes https://github.com/kiali/kiali-ui/pull/2154

Moved the fontSize from the ace builtint properties to the scss file.

### DIFF
--- a/src/pages/IstioConfigDetails/IstioConfigDetailsPage.scss
+++ b/src/pages/IstioConfigDetails/IstioConfigDetailsPage.scss
@@ -8,6 +8,7 @@
   position: relative;
   min-height: 200px;
   border: 1px solid #8b8d8f;
+  font-size: 16px !important;
 }
 
 .istio-validation-error {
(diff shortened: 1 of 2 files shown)


hhovsepy:
still facing the same issue, it disappears only when removing '!' before the important

lucasponce:
I've seen that I have in my env some *.css that are ignored but may affect when building the global CSS.
I guess that can explain the mistery to not be able to reproduce from my side.

I'll investigate.

lucasponce:
image

@hhovsepy would you be able to debug the styles that are being loaded in your test ?
Just to check which one is being applied.

hhovsepy:
By first load I see 16px in style which is applied and is big:
Screenshot from 2021-05-17 19-29-34
But when size is not !important the 'element''s 12px is applied:
Screenshot from 2021-05-17 19-30-43

lucasponce:
But the "host" links in the side panel should have also this style:
image

Are you having a different one in your test ?

hhovsepy:
> Are you having a different one in your test ?
For the 'host' link there is not style applied:
Screenshot from 2021-05-18 09-14-09

hhovsepy:
Screenshot from 2021-05-18 09-25-03

hhovsepy (approved):
Verified now.

prachiyadav (approved):

PR #2159, opened 2021-05-17
Introduce Find/Hide presets

jshaughn:
Graph Find/Hide now offers preset expressions via new dropdowns on each input.  The preset expressions are supplied via the UIDefaults supplied by the server, and so can be configured by users via the CR.

Fixes https://github.com/kiali/kiali/issues/3708

SERVER PR:  https://github.com/kiali/kiali/pull/4020
Operator PR: https://github.com/kiali/kiali-operator/pull/315

### DIFF
--- a/src/pages/Graph/GraphHelpFind.tsx
+++ b/src/pages/Graph/GraphHelpFind.tsx
@@ -37,15 +37,16 @@ export default class GraphHelpFind extends React.Component<GraphHelpFindProps> {
       color: '#fff',
       backgroundColor: '#003145',
       width: contentWidth,
-      height: '71px',
+      height: '80px',
       padding: '5px',
       resize: 'none',
       overflowY: 'hidden'
     });
     const preface =
       'You can use the Find and Hide fields to highlight or hide graph edges and nodes. Each field accepts ' +
-      'expressions using the language described below. Hide takes precedence when using Find and Hide ' +
-      'together. Uncheck the "Compress Hidden" display option for hidden elements to retain their space.';
+      'expressions using the language described below. Preset expressions are available via the dropdown. ' +
+      'Hide takes precedence when using Find and Hide together. Uncheck the "Compressed Hide" Display ' +
+      'option for hidden elements to retain their space.';
 
     return (
       <>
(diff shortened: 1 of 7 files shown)


jshaughn:
@lucasponce @lwrigh Sorry for the launch failure in today's meeting.  Below is a video of the feature enhancement.  In short, the Find/Hide inputs now each have a dropdown with Preset expressions, defined in English.  The expressions are configurable in the Kiali CR, what is shown are just some possible defaults.  When selected the actual expression will populate the input box, and can then be edited if desired.  Selection will also replace anything already in the input box.  Any thoughts?

kiali#3708

lwrigh:
Thank you for sharing the video @jshaughn ! One thing that catches my attention is that the find/hide dropdown menu goes off to the side instead of expanding right below the find/hide fields. I would expect those menus to open up right below the find/hide fields just like the dropdowns behave.

jshaughn:
@lwrigh The reason for that is because the dropdown is actually just the down arrow, not the entire input box.  But I do know what you are saying, it does look offset.  I could move the dropdown to the left side, or maybe I can move the selections under the input box to sort of "deceive" the user.

lwrigh:
> @lwrigh The reason for that is because the dropdown is actually just the down arrow, not the entire input box. But I do know what you are saying, it does look offset. I could move the dropdown to the left side, or maybe I can move the selections under the input box to sort of "deceive" the user.

That sounds good to me so it feels more standard. Thanks!

jshaughn:
@lwrigh This is what it looks like with the leading dropdown.  I had to add a gap between Find and Hide to make it distinct:

image

As it turned out, it's not super-simple to shift the dropdown contents to the left without some fancier CSS work.  I'd have to dive into that, but a more simple alternative is to anchor the options on the right.  This approach may end up looking odd, though, as a longer option it will run farther to the left, under say the Description dropdown.  

image

abonas:
This is a great enhancement. 
What happens if the added presets have the wrong syntax?

jshaughn:
> This is a great enhancement.

You can thank @lucasponce for the idea.

> What happens if the added presets have the wrong syntax?

It would generate an error on-screen, as if you had typed an incorrect expression.  I think this is fine, if a person is adding new expressions they can easily test and fix.

jshaughn:
@lucasponce @lwrigh  OK, I solved the CSS issues and I prefer the menu-option on the right side, so, proposing we stick with:

image

lucasponce:
Thanks @jshaughn, from feature perspective and/or UI I don't have too many comments, this implementation is very good.

I've noticed a regression, when you use the find helper, in the "Show Edge Labels" there is a problem (it can be selected again):

image

lucasponce:
Ah, no, it's also in master, so, when a find expression is present, it seems that the "show edge labels" are locked to avoid user may break the expression ?

If that's correct then please ignore my previous comment.

lucasponce (approved):

jshaughn:
@lucasponce This is probably a little-known behavior but it is intentional.   If you enter a find or hide expression that requires a Display option be enabled, for example `rt > 1`, the option will be automatically enabled (and you will get a Toast message telling you the same).   Now, if you try to unset that option what happens is that it immediately gets re-enabled on the next refresh, which is typically immediate.  So, the options are not "locked", but we do ensure the necessary options stay enabled.

mattmahoneyrh (approved):

PR #2160, opened 2021-05-18
Support WorkloadGroup Istio resources

lucasponce:
Backend needed https://github.com/kiali/kiali/pull/4024


hhovsepy (approved):
Verified, also the case when non cluster user has view only access to bookinfo and cache is disabled:
Screenshot from 2021-05-19 17-08-02
Screenshot from 2021-05-19 17-07-48

PR #2161, opened 2021-05-18
Support new Iter8Info endpoint

jadeyliu939:
Output detail error message if Iter8 v2.x was installed and using the new Iter8Info endpoint in pull request [#4028](https://github.com/kiali/kiali/pull/4028)

** Screenshot **

Screen Shot 2021-05-18 at 7 18 21 PM


abonas:
I'm commenting only the phrasing and not the different iter8 versions availability and their compatibility with Kiali.
I think that adding hard coded future release dates to code is less optimal. I would go with "version 2.X is not supported ..."

jadeyliu939:
> I'm commenting only the phrasing and not the different iter8 versions availability and their compatibility with Kiali.
> I think that adding hard coded future release dates to code is less optimal. I would go with "version 2.X is not supported ..."

Agree, I will change the phrases. thank you

lucasponce (approved):

PR #2162, opened 2021-05-19
Hide Logout option for auth strategies that don't support it

israel-hdez:
Strategies that don't support logout are anonymous and header.

Fixes kiali/kiali#3934


jshaughn (approved):
Works for me using anonymous.

pbajjuri20:
I have tested and verified with strategy: anonymous, Kiali Logout button is disabled successfully.

Before PR:

before_logout

After PR: 

after_logout


I have tested the OpenShift and Token strategies, Logout option works perfectly.

pbajjuri20 (approved):

PR #2163, opened 2021-05-20
Upgrade tippy dependency to latest version

nrfox:
Upgrades the `tippy.js` dependency to the latest version. There were some breaking API changes between major versions and [this migration guide](https://github.com/atomiks/tippyjs/blob/master/MIGRATION_GUIDE.md) was followed to fix those. The two props the UI was previously using that I am unsure of the exact effects are [multiple](https://github.com/atomiks/tippyjs/blob/master/MIGRATION_GUIDE.md#if-you-were-using-multiple-or-relying-on-its-behavior) and [size](https://github.com/atomiks/tippyjs/blob/master/MIGRATION_GUIDE.md#if-you-were-using-size). I haven't noticed any regressions or changes in behavior but regression tests should be run on this PR to ensure nothing has changed with the version bump.

Fixes https://github.com/kiali/kiali/issues/4035

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,7 +95,7 @@
     "reselect": "4.0.0",
     "screenfull": "5.0.2",
     "seedrandom": "3.0.1",
-    "tippy.js": "3.4.1",
+    "tippy.js": "6.3.1",
     "typesafe-actions": "3.2.1",
     "typestyle": "2.0.4",
     "visibilityjs": "2.0.2"
--- a/package.json
+++ b/package.json
@@ -95,7 +95,7 @@
     "reselect": "4.0.0",
     "screenfull": "5.0.2",
     "seedrandom": "3.0.1",
-    "tippy.js": "5.2.1",
+    "tippy.js": "6.3.1",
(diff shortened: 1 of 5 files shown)


nrfox:
Actually just discovered a bug and will convert this to a draft until fixed.

jshaughn (approved):
For me I didn't see any regressions when trying a variety of things, health hovers, custom date range, sparkline/chart hovers, graph tour, context menus.  So For me it is OK.  It did not seem to solve any of the potentially related issues listed in the main issue description, but with the update I think investigating those is now on possible.  They all seem to have to do with losing tooltips after some sort of update or refresh.

mattmahoneyrh (approved):

nrfox:
This is subtle but I noticed the diamond on the tooltip is sticking out too far now with the tippy bump. Marking this as "do not merge" until it's fixed.

nrfox:
Re-adding folks after rebase/updates

jshaughn (approved):
I don't see any ill-effects from the latest change, Approving.

PR #2164, opened 2021-05-21
Indicate in a different way the reserved word mesh

leandroberetta:
**Description**

This PR adds the same styles to the special "mesh" gateway as other gateways have in the VirtualService Overview and it also adds a tooltip with more details about its meaning.

**Issue**

[#1391](https://github.com/kiali/kiali/issues/1391)

**Documentation**

The following is a demo of how the change looks:

mesh_tooltip

### DIFF
--- a/src/pages/IstioConfigDetails/IstioObjectDetails/VirtualServiceOverview.tsx
+++ b/src/pages/IstioConfigDetails/IstioObjectDetails/VirtualServiceOverview.tsx
@@ -2,10 +2,21 @@ import * as React from 'react';
 import { checkForPath, highestSeverity } from '../../../types/ServiceInfo';
 import { Host, HTTPRoute, ObjectValidation, TCPRoute, TLSRoute, VirtualService } from '../../../types/IstioObjects';
 import VirtualServiceRoute from './VirtualServiceRoute';
-import { Stack, StackItem, Text, TextVariants, Title, TitleLevel, TitleSize } from '@patternfly/react-core';
+import {
+  Stack,
+  StackItem,
+  Text,
+  TextVariants,
+  Title,
+  TitleLevel,
+  TitleSize,
+  Tooltip,
+  TooltipPosition
+} from '@patternfly/react-core';
 import GlobalValidation from '../../../components/Validations/GlobalValidation';
 import IstioObjectLink from '../../../components/Link/IstioObjectLink';
 import ServiceLink from './ServiceLink';
+import { PFColors } from '../../../components/Pf/PfColors';
 
 interface VirtualServiceProps {
   namespace: string;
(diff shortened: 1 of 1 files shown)


lucasponce (changes requested):
"Mesh" should not share the same style of a link as it can't "jump" to the "mesh" definition.
Instead, it could be good to have a regular "mesh" word with an "info" icon with the tooltip to explain that.

We are using that method (the info icon + tooltip) as a standard way to introduce more information for an element across all pages.

But PR is good :) thanks for sending it !

leandroberetta:
> "Mesh" should not share the same style of a link as it can't "jump" to the "mesh" definition.
> Instead, it could be good to have a regular "mesh" word with an "info" icon with the tooltip to explain that.
> 
> We are using that method (the info icon + tooltip) as a standard way to introduce more information for an element across all pages.
> 
> But PR is good :) thanks for sending it !

@lucasponce thanks for the comment and help, yes I agree with you so I've just committed some code with your idea!.

mesh2

src/pages/IstioConfigDetails/IstioObjectDetails/VirtualServiceOverview.tsx
```diff
@@ -66,7 +83,15 @@ class VirtualServiceOverview extends React.Component<VirtualServiceProps> {
       childrenList.push(
         <li key={'gateway_' + host.service + '_' + j}>
           {host.service === 'mesh' || !isValid ? (
```
hhovsepy, line 85: When it is mesh or not valid can be separated, because this causes existing or wrong gateways to be shown as invalid and mesh in the same time when one of gateways is invalid:
Screenshot from 2021-05-25 07-55-35

leandroberetta: @hhovsepy Thanks for testing this issue, I made son changes to the code to fix this situation:

gateway_ok
gateway_ok_mesh
gateway_error
gateway_error_mesh

hhovsepy (approved):
Verified:
Screenshot from 2021-05-27 15-39-43

leandroberetta:
New tooltip description according istio.io docs:

mesh4

jshaughn:
If this is a verbatim match to Istio docs then I guess it's OK.   Otherwise I would update to:

`The reserved word, "mesh", implies all of the sidecars in the mesh.`

lucasponce:
+1 with suggestion from @jshaughn

One style minor change and I think it's ready: would you mind to align the test of the toolip ? I'm trying to consolidate the style as there are some tooltips looking very different.

image

So, I propose if you can just change it to something like this:

image

Thanks !

leandroberetta:
> If this is a verbatim match to Istio docs then I guess it's OK. Otherwise I would update to:
> 
> `The reserved word, "mesh", implies all of the sidecars in the mesh.`

+1 too.

@lucasponce @jshaughn 

mesh_left

lucasponce (approved):
Thx @leandroberetta

lucasponce:
@prachiyadav I'm going to merge to consolidate this with other PRs, it can be tested as well from master branch.

PR #2165, opened 2021-05-24
Add kiali traffic scenario graph badges to Services

nrfox:
Adds graph badges for the traffic scenarios supported by Kiali wizards: request routing, fault injection etc.

Fixes: kiali/kiali#1477
Backend changes: https://github.com/kiali/kiali/pull/4043


nrfox:
Here's are some examples of what multiple traffic badges look like on the graph:
Screenshot from 2021-05-24 11-58-02

Screenshot from 2021-05-21 12-56-19

Screenshot from 2021-05-24 12-19-03

lucasponce:
One early comment: when a Kiali scenario is present, it can be assumed that there will be a VS, so I wouldn't show that badge in those cases, only when you haven't identified a Kiali scenario, then yes, a VS badge would be good.

Otherwise is going to look redundant (and perhaps confusing to some users).

nrfox:
> One early comment: when a Kiali scenario is present, it can be assumed that there will be a VS, so I wouldn't show that badge in those cases, only when you haven't identified a Kiali scenario, then yes, a VS badge would be good.
> 
> Otherwise is going to look redundant (and perhaps confusing to some users).

:+1: that sounds good

nrfox:
Here's the updated legend with the latest icons:

Screenshot from 2021-05-25 15-49-35

lucasponce:
It needs a rebase but LGTM.
Really happy to see this feature.

I've tested on Graph, MiniGraph and Find/Hide.

image

nrfox:
Rebased and added a check to ensure that only one traffic shifting icon is displayed at a time

src/components/CytoscapeGraph/graphs/GraphStyles.ts
```diff
@@ -200,17 +205,47 @@ export class GraphStyles {
     const isOutside = node.isOutside;
 
     let icons = '';
+    // Put a little more space between icons when a badge has multiple icons
+    const iconMargin = () => (icons === '' ? style(defaultIconMargin) : style({ marginRight: '2px' }));
     if (node.isRoot) {
-      icons = `<span class="${NodeIconRoot} ${iconMargin}"></span> ${icons}`;
+      icons = `<span class="${NodeIconRoot} ${iconMargin()}"></span> ${icons}`;
```
jshaughn, line 211: I realize it was like this prior to the PR, but I think for trafficSource nodes (aka "root" nodes), the root icon should always be the leftmost icon.  Looking here, I think it ends up being the rightmost icon.  Let's make that change.

src/components/CytoscapeGraph/graphs/GraphStyles.ts
```diff
@@ -52,9 +52,14 @@ let NodeColorFillHoverDegraded: PFColorVal;
 let NodeColorFillHoverFailure: PFColorVal;
 const NodeHeight = '25px';
 const NodeIconCB = icons.istio.circuitBreaker.className; // bolt
+const NodeIconFaultInjection = icons.istio.faultInjection.className; // ban
 const NodeIconMS = icons.istio.missingSidecar.className; // exclamation
 const NodeIconRoot = icons.istio.root.className; // alt-arrow-circle-right
 const NodeIconVS = icons.istio.virtualService.className; // code-branch
+const NodeIconRequestRouting = icons.istio.requestRouting.className; // code-branch
+const NodeIconRequestTimeout = icons.istio.requestTimeout.className; // clock
+const NodeIconTCPTrafficShifting = icons.istio.tcpTrafficShifting.className; // share-alt
```
jshaughn, line 61: Until we are going to use distinctive icons I don't think we need this duplicate definition.  Let's just have NodeIconTrafficShifting and we can remove icons.istio.tcpTrafficShifting.

src/components/CytoscapeGraph/graphs/GraphStyles.ts
```diff
@@ -200,17 +205,47 @@ export class GraphStyles {
...
+        if (node.hasFaultInjection) {
+          icons = `<span class="${NodeIconFaultInjection} ${iconMargin()}"></span> ${icons}`;
+        }
+        if (node.hasTrafficShifting) {
+          icons = `<span class="${NodeIconTrafficShifting} ${iconMargin()}"></span> ${icons}`;
+        }
+        // Don't show icon twice since the same for both http and tcp traffic shifting
+        if (node.hasTCPTrafficShifting && !node.hasTrafficShifting) {
+          icons = `<span class="${NodeIconTCPTrafficShifting} ${iconMargin()}"></span> ${icons}`;
+        }
```
jshaughn, line 241: If you take the recommendation to remove `NodeIconTCPTrafficShifting` then this can just be:

```
if (node.hasTrafficShifting || node.hasTCPTrafficShifting) {
...
```

src/config/Icons.ts
```diff
@@ -26,8 +31,11 @@ const mutIcons = {
       color: 'red'
     },
     mtls: { className: 'pf-icon pf-icon-locked', type: 'pf', name: 'locked', ascii: '\uE02a ' },
-    disabledMtls: { className: 'pf-icon pf-icon-unlocked', type: 'fa', name: 'unlock', ascii: '\uE065 ' },
+    requestRouting: { className: 'fa fa-code-branch', type: 'fa', name: 'code-fork', ascii: '\uf126 ' },
+    requestTimeout: { className: 'fa fa-clock', type: 'fa', name: 'clock', ascii: '\uf017 ' },
     root: { className: 'fa fa-arrow-alt-circle-right', type: 'fa', name: 'arrow-alt-circle-right', ascii: '\uf35a ' },
+    tcpTrafficShifting: { className: 'fa fa-share-alt', type: 'fa', name: 'share-alt', ascii: '\uf1e0 ' },
```
jshaughn, line 37: can remove if you like

src/config/KialiIcon.tsx
```diff
@@ -94,9 +97,13 @@ export const KialiIcon: { [name: string]: React.FunctionComponent<IconProps> } =
   PlayCircle: (props: IconProps) => <PlayCircleIcon className={props.className} />,
   Regex: (props: IconProps) => <AsteriskIcon className={props.className} />,
   Repository: (props: IconProps) => <RepositoryIcon className={props.className} />,
+  RequestRouting: (props: IconProps) => <CodeBranchIcon className={props.className} />,
+  RequestTimeout: (props: IconProps) => <OutlinedClockIcon className={props.className} />,
   Services: (props: IconProps) => <ServiceIcon className={props.className} />,
   Stop: (props: IconProps) => <StopIcon className={props.className} />,
   Topology: (props: IconProps) => <TopologyIcon className={props.className} />,
+  TCPTrafficShifting: (props: IconProps) => <ShareAltIcon className={props.className} />,
```
jshaughn, line 105: can remove if you like

src/pages/Graph/GraphToolbar/GraphSettings.tsx
```diff
@@ -314,12 +311,6 @@ class GraphSettings extends React.PureComponent<GraphSettingsProps, GraphSetting
     ];
 
     const badgeOptions: DisplayOptionType[] = [
-      {
-        id: 'filterCB',
-        labelText: 'Circuit Breakers',
-        isChecked: showCircuitBreakers,
-        onChange: toggleGraphCircuitBreakers
-      },
```
jshaughn, line 322: This removal is fine, but now that the `Virtual Service` option is more complex/powerful, in that it is used to render not only VS but also the related scenario badges, it deserves an info tooltip.  See above for an example of adding the `tooltip` attribute to the option.

src/pages/Graph/SummaryPanelNode.tsx
```diff
@@ -137,7 +147,19 @@ export class SummaryPanelNode extends React.Component<SummaryPanelNodeProps, Sum
...
+    hasCB?: boolean,
+    hasVS?: boolean,
+    hasMissingSC?: boolean,
+    isDead?: boolean,
+    hasRequestRouting?: boolean,
+    hasFaultInjection?: boolean,
+    hasTrafficShifting?: boolean,
+    hasTCPTrafficShifting?: boolean,
+    hasRequestTimeout?: boolean
+  ) => {
```
jshaughn, line 160: With this many params would be nice if they were alphabetized :-)  No change required, though.

jshaughn (changes requested):
Looking good, a few comments in the code.  Also, you can remove:
 
- `ActionKeys.GRAPH_TOOLBAR_TOGGLE_GRAPH_CIRCUIT_BREAKERS`
- instances of `showCircuitBreakers` in CytoscapeGraph.tsx, MiniGraphcard.tsx, GraphPage.tsx

src/components/CytoscapeGraph/graphs/GraphStyles.ts
```diff
@@ -200,17 +205,47 @@ export class GraphStyles {
     const isOutside = node.isOutside;
 
     let icons = '';
+    // Put a little more space between icons when a badge has multiple icons
+    const iconMargin = () => (icons === '' ? style(defaultIconMargin) : style({ marginRight: '2px' }));
```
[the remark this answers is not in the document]
nrfox: This is a closure that uses the `icons` var from line 207. I can move this up but would just need to pass `icons` to the function instead of using it from the outside scope.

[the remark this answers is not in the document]
jshaughn: Yes, I see that,  totally up to you.  For me it just looked a little out of place but you can leave it as-is if you like.

src/pages/Graph/SummaryPanelNode.tsx
nrfox: I turned the param into an object so it doesn't matter in what order they are passed but that did make the definition kind of gnarly. Let me know what you think.

jshaughn: Yeah, I don't love that, sorry for opening a can of worms :-)  How about something similar:

```
  private renderBadgeSummary = (nodeData: DecoratedGraphNodeData) => {
    const {
      hasCB,
      hasFaultInjection,
      hasMissingSC,
      hasRequestRouting,
      hasRequestTimeout,
      hasTCPTrafficShifting,
      hasTrafficShifting,
      hasVS,
      isDead
    } = nodeData;

```

nrfox: That's much better. I'll update.

pbajjuri20 (approved):
Tested and passed on Kiali-UI 1.36.0 (1325f3d75900f572299231f22872d9f028eaa61e)

PR #2170, opened 2021-06-01
Move Envoy details to Workload details

leandroberetta:
**Description**

To facilitate the accessibility of Envoy details this PR moves Envoy details from the modal to different tabs at the Workload level.

**Issue**

[#4008](https://github.com/kiali/kiali/issues/4008)

**Documentation**

The following represents how Envoy details are shown in the different tabs:

Listeners, clusters and routes are shown in a table view:

env_details_listeners

Bootstrap and full config dump are shown in the editor:

env_details_bootstrap

**Notes**

The Envoy Metrics tabs is added at runtime via a Monitoring Dashboard, so even moving the Traces tab there is no guarantee that Envoy related tabs will be grouped because it may exists another runtimes dashboards in between. One solution is to order the tabs given the numerical key that already exists in the code.


lucasponce:
Thanks @leandroberetta, it looks very good.

One early comment:

image

I'd keep the following order of the tabs in the workload:

- "Overview", "Traffic", "Logs", "Inbound" (yes, moved from "Inbound Metrics" -> "Inbound"), "Outbound", "Traces", and then the Envoy ones: "Envoy Clusters", "Envoy Listeners", "Envoy Routes", "Envoy Config", "Envoy Status" ("Envoy Metrics" -> "Envoy Status").

The order of the Envoy tabs is also on purpose as to try to follow the logic you'd examine them (I think).

Also the "Inbound/Outbound Metrics" -> "Inbound/Outbound" should be propagated into the App and Service detail pages.

Now them it would be better sync with the Graph sidepanel:

image

Also, after this refactor, worth to check if the context menu are ok in the graph and to/from the span details (just wording).

aljesusg:
IMO we'll have so tabs...what do you think about to have a envoy tab with a subtabs?

https://patternfly-react.surge.sh/components/tabs#tabs-with-sub-tabs

lucasponce:
> IMO we'll have so tabs...what do you think about to have a envoy tab with a subtabs?

We are doing the opposite in the overall design for the details pages, to get rid of the inner subtabs.
This was previously discussed with UX in the context of the whole refactoring of the details pages and the pattern to follow is to put the tab at the entity level.

aljesusg:
> We are doing the opposite in the overall design for the details pages, to get rid of the inner subtabs.
> This was previously discussed with UX in the context of the whole refactoring of the details pages and the pattern to follow is to put the tab at the entity level.

yes I understand that but my concern is that will have 11 tabs and maybe the screen is not enough big to see all of them...so Will it appear an scroll ? Will the user know that there is more information?

Anyway if UX has taken this so for my great :)

leandroberetta:
The amount of tabs is a concern to take into account for sure, but also the consistency between them, because we have two separate tabs for Inbound and Outbound Metrics but just one for Inbound and Outbound Traffic. 

Another concern is that if we rename the metrics tabs removing "Metrics", would it be clear that we are talking about metrics?.

lucasponce:
> Another concern is that if we rename the metrics tabs removing "Metrics", would it be clear that we are talking about metrics?.

Yes, I agree that the "naming" of the tabs is open as there may be multiple approaches.

But the design of the tabs is not new. For example, in the Kiali workload, when more dashboards are present:

image

image

So, now what we are doing with this work is to treat the Envoy information as similar pattern.

The discussion is then, more how to name/order them.

Note that tabs component support cornercases scenarios like narrowed widths.

About naming: I'm still undecided if using the form "Envoy Clusters", "Envoy Listeners", "Envoy Routes", or skip the "Envoy".

Also another comment that I forgot to add in my first feedback: this info makes sense when the sidecar is present, when not they shouldn't show up (instead to show them with empty content). The overview tab will add details about the missing sidecar.

leandroberetta:
> The discussion is then, more how to name/order them.

Perfect.

> About naming: I'm still undecided if using the form "Envoy Clusters", "Envoy Listeners", "Envoy Routes", or skip the "Envoy".

I know that is information is intended for advanced users trying to debug problems or just wanting to have more information about what they're trying to do but as a user that is not fully aware of Envoy internals, it's nice to see that these names are Envoy related, maybe if we remove the Envoy part, inside of every tab we can put a little bit of what are these configurations. 

> Also another comment that I forgot to add in my first feedback: this info makes sense when the sidecar is present, when not they shouldn't show up (instead to show them with empty content). The overview tab will add details about the missing sidecar.

Yes, you're right!. Coding this validation, I assumed that if at least one pod (related to a Workload) does not contain an Istio proxy injected, Envoy tabs won't be shown. In general, these pods will be created by a higher order entity (a ReplicaSet for example) and they will have the same annotations, so they will have sidecars injected. Of course there is a case of self-managed pods (very strange IMO) with different set of containers (one with sidecar and the other without it), if we will take this case into account, we need to show the Envoy tabs and restrict the pod lists from the combo to show only the pods that has sidecars.

lucasponce:
Hi @lwrigh this is one of the topics that we can review with you in the UX meeting.

lucasponce:
The discussion about the naming and grouping of the tabs I guess it could be open, so I propose to discuss this with @lwrigh in the UX call, to see if we find a design that doesn't force to change the current one.

One thing that I guess it can be addressed is the Envoy Config tab.
In the current approach is fixed:

image

But I think it should also follow the one used in the Istio Config where editor is adjusted to the tab content size:

image

(Buttons don't apply yet as there are no editable actions yet).

lucasponce:
Also, another proposal could be to combine all feedback discussed as well.

@aljesusg proposed to use inner tabs, a possibility could be to have something in between, with a common pod selector like the pods:

image

And combining with the sub tabs used like in the trace details:

image

Personally, I'd prefer to try to have some option with main tab, but I understand that this perhaps may solve and meet all feedbacks.

@leandroberetta @aljesusg any comments ?

In any case @lwrigh would also take a look.

abonas:
I'm wondering about the checkbox "show only bootstrap configuration":
- whether the checkbox is the right control here for limiting what info is displayed.
- should it say "show" or "display" (for consistency with other places in UI, like topology, etc)
UX will recommend what's best :)

leandroberetta:
> Also, another proposal could be to combine all feedback discussed as well.
> 
> @aljesusg proposed to use inner tabs, a possibility could be to have something in between, with a common pod selector like the pods:
> 
> image
> 
> And combining with the sub tabs used like in the trace details:
> 
> image
> 
> Personally, I'd prefer to try to have some option with main tab, but I understand that this perhaps may solve and meet all feedbacks.
> 
> @leandroberetta @aljesusg any comments ?
> 
> In any case @lwrigh would also take a look.

I think I prefer having only one Envoy tab, with the pod selector and sub tabs. The pod selector can be the same to the tabs, or it can be placed inside of every tab as it is today.

leandroberetta:
> * should it say "show" or "display" (for consistency with other places in UI, like topology, etc)

Good observation, I will check other places in the UI, thanks!.

lucasponce:
> > * should it say "show" or "display" (for consistency with other places in UI, like topology, etc)
> 
> Good observation, I will check other places in the UI, thanks!.

I saw that comment and perhaps not sure if it can create some confusion.
The pattern used is similar as the checkbox you can see in the "logs" or the "traces" details.

In both descriptions are short and also you can use "bootstrap".

leandroberetta:
Some progress:

envoy_listeners
envoy_config
envoy_metrics

Some comments:

- I had to play around a little bit with the CustomMetrics component, because it was designed to be in tabs at workload level (it added a RenderScrollComponent among others, and in the sub tabs design, I'm adding these components in an outer place). So I added a new option to CustomMetrics, allowing it to be embedded in other component without the addition of some components that are not needed).
- Not sure if the checkbox with the "Bootstrap" legend is fine, I was thinking in bringing back the Bootstrap tab, any comments?.
- Above the sub tabs, it would be a nice place to put some information about things like: outlier detection, retries, timeouts, I will create a issue to continue in the future if makes sense.

lucasponce:
> * I had to play around a little bit with the CustomMetrics component, because it was designed to be in tabs at workload level (it added a RenderScrollComponent among others, and in the sub tabs design, I'm adding these components in an outer place). So I added a new option to CustomMetrics, allowing it to be embedded in other component without the addition of some components that are not needed).

Thanks, I'll review in the PR, also thinking that perhaps for "Runtimes" we could apply same pattern here of a main "Runtime" tab and also subtabs for any specific runtime dashboard. Main Inbound/Outbound metrics should remain as it is.

> * Not sure if the checkbox with the "Bootstrap" legend is fine, I was thinking in bringing back the Bootstrap tab, any comments?.

Sounds ok, I'm with you 50/50, but fine to bring back the Bootstrap tab, now would be more clear.

> * Above the sub tabs, it would be a nice place to put some information about things like: outlier detection, retries, timeouts, I will create a issue to continue in the future if makes sense.

Yes, this is linked with one the previous work I did for the CircuitBreaker pattern in general, but it's an Epic task as the Envoy version used by Istio (the "Istio Proxy") doesn't have this info enabled by default, to improve performance and reduce metrics cardinality.

But totally +1 it's part of the high level discussion https://github.com/kiali/kiali/discussions/4082 and now that that "Envoy" tab is getting shape, having a summary of the relevant information to user makes perfectly sense.

Thanks for bringing it  :)

leandroberetta:
@lucasponce I like the idea of putting all the runtimes in a single tab, and I even would like to merge the Inbound and Outbound metrics tabs (like the traffic tab). I will create an issue at least for grouping the runtimes.

I will remove the "do not merge" label as I think it's stable now:

- The bootstrap tab is back
- I tested this with the recent changes of the custom metrics (removal of the CRD, etc)

envoy_bootstrap

lucasponce:
Adding @prachiyadav as this is ready for review.

How to test:
- Verify that "Envoy" tab is present under Workload details.
- Check the selector of different Pods.

leandroberetta:
> The work looks good.
> I have one comment from feature perspective (let me know if you'd like to address it on this PR or in a following on).
> 
> Now that we have the Envoy tab under the workload (where it makes sense), then probably the "Envoy Metrics" coming from the Runtime Metrics could make less sense:
> 
> image
> 
> So, I suggest to filter it from the Application details then.
> 
> Don't worry if the metrics tab may have some adjustment (it's something I'm addressing on #2176).
> 
> Also, we could group the "Runtimes" metrics (that are less used in general) in a similar pattern, but that could be for other followup work.
> 
> So, let me know what do you think.

@lucasponce thanks for the review. Regarding the Application details, as it was an small change, I've just pushed the modification.

For the runtimes grouping, I will create a new issue.

lucasponce (approved):
This is ready to consolidate.
Note that the metrics tab may need the work of https://github.com/kiali/kiali-ui/pull/2176 but I think this one could get merged when @prachiyadav is ok.

lucasponce:
@prachiyadav this PR is required for two more work https://github.com/kiali/kiali-ui/pull/2178 and https://github.com/kiali/kiali-ui/pull/2176 so I'm going to consolidate it.

Current validation can just go from the master branch, any issue will be addressed in a follow-up PR.

prachiyadav:
Screenshot from 2021-06-16 20-07-58

lucasponce:
Discussed with @prachiyadav that it seems that the Kiali QE job used for master was not picking the latest commits to including this work.

prachiyadav:
Screenshot from 2021-06-16 23-38-25

prachiyadav:
Verified, its working with Master

PR #2171, opened 2021-06-02
Added new Warning message on workload pod in case of failing containers.

hhovsepy:
Potential fix of https://github.com/kiali/kiali/issues/3529
UI changes required https://github.com/kiali/kiali/pull/4075

In the case when Workload's Pod has failing non istio containers, but the Kubernetes returns that pod's state as Running, Kiali should show warning on that pod.

### DIFF
--- a/src/pages/WorkloadDetails/WorkloadInfo.tsx
+++ b/src/pages/WorkloadDetails/WorkloadInfo.tsx
@@ -127,6 +127,21 @@ class WorkloadInfo extends React.Component<WorkloadInfoProps, WorkloadInfoState>
     const pendingPod: ObjectCheck = { message: 'Pod is in Pending Phase', severity: ValidationTypes.Warning, path: '' };
     const unknownPod: ObjectCheck = { message: 'Pod is in Unknown Phase', severity: ValidationTypes.Warning, path: '' };
     const failedPod: ObjectCheck = { message: 'Pod is in Failed Phase', severity: ValidationTypes.Error, path: '' };
+    const failingPodContainer: ObjectCheck = {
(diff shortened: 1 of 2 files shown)


lucasponce:
Perhaps this solution is not completed.
I haven't tested but the idea would be to check if it covers a situation when a pod is in a READY N/M where N < M, not necessarily 0, typically 1/2 (on Istio deployments, when sidecar or app container doesn't start for some reason).

On this case, the Pod Phase may be ok, but I think Kiali has not enough info on those cases.

Yes, adding the warning is good, perhaps just to check if those cases are covered.

hhovsepy:
It should display different warning message. 
> The workload status is correct:

hhovsepy:
Ready now:
Screenshot from 2021-06-08 15-32-50

hhovsepy:
How to test: Create a Deployment in 'bookinfo' namespace using OpenShift UI for instance and add 'hhovsepy/exit1' docker image into that.
In Kiali UI, for newly created Workload 'example' it will show the pod status alerts.
Regression testing required for missing 'app' and 'version' labels, for instance these labels can be edited in OpenShift UI going to Deployment details page.

lucasponce (approved):
Tested with the instructions commented in https://github.com/kiali/kiali-ui/pull/2171#issuecomment-856740418

image

hhovsepy:
Tooltip right after new deployment, before all contains are started:
Screenshot from 2021-06-09 11-46-03

skondkar (approved):
Verified LGTM

Screenshot from 2021-06-10 12-57-59

Screenshot from 2021-06-10 12-57-51

PR #2173, opened 2021-06-04
Wrapper for icon tooltips

nrfox:
Creates a small wrapper for icon tooltips to ensure they are created with a span to avoid patternfly svg issues.

Fixes kiali/kiali#3583


jshaughn (commented):
Another possibility, instead of adding a new component perhaps export a new function from KialiIcon.

nrfox:
> Another possibility, instead of adding a new component perhaps export a new function from KialiIcon.

Something like?
```
KialiIcon.tsx
...

export function createTooltipIcon(icon: any) {
  return (
     <span>icon</span>
  )
}

HealthIndicator.tsx
...

import { createTooltipIcon } from KialiIcon

<Tooltip>
...
  {createTooltipIcon(icon)}
</Tooltip>
```

jshaughn:
Sure, or even the same function you had originally.  Whatever you like best.  Go ahead and create what you think is best and promote the PR, let's get a fix in for this.

src/config/KialiIcon.tsx
```diff
@@ -120,3 +120,10 @@ Object.keys(KialiIcon).forEach(key => {
     className: iconStyle
   };
 });
+
+// createTooltipIcon wraps the icon in a span element. Tooltip child elements that are
+// SVGs (icons) need to be wrapped in something to avoid the tooltip from disappearing on refresh.
+// See: https://github.com/kiali/kiali/issues/3583 for more details.
+export function createTooltipIcon(icon: any) {
+  return <span>{icon}</span>;
+}
```
jshaughn, line 129: Thinking on this further, is there any reason we should depend on the caller to have to do this?  Maybe KialiIcon should just wrap every icon it returns and then we don't have to export anything, and users of KialiIcon are protected for free.  And then, if the bug is fixed we can remove the wrapping at some point.

nrfox: That sounds good to me. `<span>` elements shouldn't affect layout?

In this case though the caller is not using `KialiIcon` to create the health icon. Instead it's calling a [helper method](https://github.com/kiali/kiali-ui/blob/master/src/components/Health/Helper.ts#L6-L8) which is just to say that not everything is creating icons through `KialiIcon` but maybe it should be?

jshaughn: Yeah, ignore my comment.  Adding unnecessary <span> does not seem like a good idea, it's hard to know if it will affect layout but it might.

jshaughn (approved):
Looks fine to me as long as the HealthIndicator is the only svg with a ToolTip.

nrfox:
> Looks fine to me as long as the HealthIndicator is the only svg with a ToolTip.

I found another instance where `createIcon` was being used without a wrapper and updated it to use the new helper function.

PR #2174, opened 2021-06-08
Show CB badge even when no VS present

nrfox:
Displays the Circuit Breaker badge on the graph even if a Virtual Service is not present.

Fixes kiali/kiali#4076


jshaughn (approved):
I forgot that CBs need only a Destination Rule and don't require a VS.  As far as I know this is the only Traffic Management feature that doesn't need a VS, so the code looks good to me.

pbajjuri20:
cb-pr

Tested and Passed on Kiali-UI (1.36.0 (803675b6347ecf6472047a1144f506689a86ad8e)

pbajjuri20 (approved):

PR #2176, opened 2021-06-10
Metrics tab layout - followup

lucasponce:
Follow up work for https://github.com/kiali/kiali/issues/3948

Backend PR https://github.com/kiali/kiali/pull/4091

This PR (backend + frontend) introduces several changes in the Metrics tab layout.

Introduces "ellipsis" pattern for legend, to have a consistent layout of charts even in narrow screens:
image

Large legend selections will we covered with an ellipsis on the legend, and an extra layer for full visibility (also with scrolling for overflow):
image

Again, this is a corner case and in those cases user should use the maximized chart:
image

Note that user will have all series tooltip, but there are not support for "hovering" from the "ellipsis legend yet", it's the only limitation (that will be addressed in a next work).

Smaller screens keep consistency in metrics layouts:
image

Minor changes but included:

Align traces tab with istio + runtime metrics, there were some different styles in controls and layout:
image

Adjusted the way to draw the layout, removing some intermediate grids that were creating differences in style on charts and empty status:
image

Align Envoy metrics:
image

Align Runtime metrics:
image
image
image
image

Fix resize when sidebar menu is toggle on/off:
image

Minor fixes related to styles.
Moved to use the dashboard spans for grid layout as a previous step to deprecate them and control the layout internally.

How to test (cc @skondkar):
- Verify the new layout on Istio metrics (Inbound/Outbound), Envoy Metrics and Runtime Metrics (using the runtimes example is enough, other personalizations are out of the scope of this work).

I'd like to merge first Leandro's work @leandroberetta as runtimes may need adjustments on the new subtabs model, but I think this work is ready.

[1] https://github.com/kiali/demos/tree/master/runtimes-demo


aljesusg:
Good idea to keep the lengend in a tooltip but it seems that when you collapse the menu the dots are not aligned

not_aligned


I only have this minor comment, I am going to review the code

src/components/Charts/Charts.scss
```diff
@@ -1,8 +1,3 @@
 .pf-c-chart svg {
   overflow: visible !important;
-  height: 100% !important;
-}
-
-.pf-c-chart div {
-  height: 100% !important;
 }
```
aljesusg, line 3: We can remove this file if you removed the import

lucasponce:
> Good idea to keep the legend in a tooltip but it seems that when you collapse the menu the dots are not aligned

Does this happen in your example when you collapse the menu by resizing the window or click on the toggle menu button ?

aljesusg (commented):
Code looks good for me. Only remove the charts.scss file from my side

aljesusg:
> > Good idea to keep the legend in a tooltip but it seems that when you collapse the menu the dots are not aligned
> 
> Does this happen in your example when you collapse the menu by resizing the window or click on the toggle menu button ?

mm It's no happening every time with the full window, but if you resize the window you can see that. From my side is not much important

src/components/Charts/Charts.scss
lucasponce: The Charts.scss is still used to overwrite the svg overflow (used for the axis).
I thought the same but I realized that it can't be removed yet.

Probably for a next round of refactoring.

lucasponce:
> mm It's no happening every time with the full window, but if you resize the window you can see that. From my side is not much important

Ok, I guess I know what is happening, when the metrics toolbar stacks is when I see that problem.
I'm going to try one workaround.

Thanks.

leandroberetta (commented):
The code looks good to me. When testing on my local environment I see that some texts are overlapping though.

metrics_jvm_1
metrics_nodejs_1

The graph settings are shown in the pictures.

lucasponce:
> The code looks good to me. When testing on my local environment I see that some texts are overlapping though.

Thanks Leandro, good catch, I'll work on it.

lucasponce:
> Is there a fixed legend size before which legend is shown partly under the graph, and after exceeding this size whole legend is in tooltip only?

Yes,

https://github.com/kiali/kiali-ui/pull/2176/files?file-filters%5B%5D=.tsx#diff-baadec6fa1fe868977a415223d543a6e36dcd83aeba79a158a3968cdddf2a76fR467

This logic "calculates" (aprox) the size of the legend and it stops when exceeds of the current width (also counting the space of padding or ellipsis).

So, it also work in maximized too, if the legend is too long, it enables the ellipsis and put the full legend info into the new layer.

lucasponce:
Fixed the problem with the legend overlapping:
image

Also, note that there is an expected issue with the legend:
image

As VictoryLegend uses absolute positions to place the text, and it has to "calculate" the size before to show it, for that it uses some "aprox" functions:
image

With some "dictionaries" for this
image

In short, that behaviour depends of the victory library version, so, we can workaround a little bit with the current code but probably it would need some revision in the future, when we upgrade Patternfly + Victory versions (not for this release).

lucasponce:
@aljesusg @leandroberetta @hhovsepy @skondkar 

This PR is ready from my side, please let me know if it looks ok to consolidate.

Thanks for the reviews.

aljesusg (approved):
LGFM

hhovsepy (approved):
Functionality LGTM

skondkar (approved):
Verified with runtimes-demo. LGTM.

PR #2177, opened 2021-06-10
Display throughput on the graph edges

jshaughn:
This PR adds UI support for optionally displaying HTTP throughput (bytes/sec) on edges.  I took the liberty to make some additional changes under this umbrella.  Because throughput has two options, request and response, it made sense to update the graph Display menu to better support options for a particular selection.  That seemed to look good so I expanded it to response time edge labels, to allow the user more control over the response time percentile.  Then I got annoyed that I could only see one edge label at a time, so I got rid of the radio approach and changed to a checkbox approach that allows users to get multiple edge labels at a time (the values are stacked and ordered on the same fashion as the display menu options).

So, here is a video of the new throughput option and also an example of having both request throughput and median response time...

kiali#2897

Closes https://github.com/kiali/kiali/issues/2897

**SERVER PR:** https://github.com/kiali/kiali/pull/4093


jshaughn:
@lucasponce Added you as a reviewer because you were a proponent of the feature.  But if you are too busy let me know and I will re-assign.

jmazzitelli:
Without examining the code, this build and ran successfully for me:

(side note: the popup help text for response time and throughput was good - explained that missing those 2 on some edges is expected)

image

lucasponce:
Starting to review, a minor comment but important:
image

I think tooltips from "Display" options should open to the right, opening on the "top" creates a little bit of noise while testing this.
In the right would be clean IMHO.

lucasponce:
I'd add units for consistency:
image

You have a "5ms" and next an absolute number that I need to check what it means in the tooltip info.

lucasponce:
Now it's possible to show Request Distribution, Response Time and Throughput in the graph, I'll enable some defaults on three categories in the minigraph as well:
image

jshaughn:
@lucasponce PR Updated for the PR feedback, units on the tcp and throughput rates, and right-positioned tooltips.  Also added a tooltip for the entire edge labeling group to explain multiple selection.

lucasponce:
If units are only showed when hovering, that should be added in minigraph, too
image

Then user won't know that value represents:
image

As commented, it could be a minor change, but request distribution, response time and throughtput should be enabled into the details view.

jshaughn:
> but request distribution, response time and throughtput should be enabled into the details view.

Sorry, lost in translation, it sounded to me like you were going to do this:

> in the graph, I'll enable some defaults on three

jmazzitelli:
I would be careful enabling all three in the mini-graph on the details view - or, at least, make it somehow easy for the user to know what all three are.

This is my original question before when viewing the multiple values in the main graph page: "how do I know what each number represents?" And the answer was, "They are stacked in the order they appear in the dropdown menu on the graph page".

But of course, there is no dropdown menu in the mini-graph. So how is a user going to know what each number represents while in the mini-graph? People are going to easily forget "which one is response time? which one is throughput?"

lucasponce:
> I would be careful enabling all three in the mini-graph on the details view - or, at least, make it somehow easy for the user to know what all three are.

Enable them, yes, that's purpose of the details page.

Explain them, yes, the user needs to know what those values mean (no matter how - i.e. on the edge, on an additional legend, etc. - )

jmazzitelli (commented):
I don't know if I like this or not, but ... it seems like the protocol should not prefix the first number but instead should maybe be at the top (or the bottom?) of the stack.

When I hover, I see "http" show up in the edge, but it prefixes the first number in the stack:
Peek 2021-06-14 11-49

lucasponce:
> I don't know if I like this or not, but ... it seems like the protocol should not prefix the first number but instead should maybe be at the top (or the bottom?) of the stack.

I personally +1 to skip the protocol, Kiali+Istio only manages two types: http and tcp, and those are clear defined on edges I think.

Perhaps I don't need the "html" but a "rs" indicating that is "request per second"

jmazzitelli:
> http and tcp, and those are clear defined on edges I think.

If those protocol labels will only ever be either "http" or "tcp" then yes, I agree. The  TCP edges are clear (those are blue lines with the arrow head having the line). HTTP edges are green with only the triangle arrowhead. So, I don't see a need to even show "http" here, unless that could be something else other than http or tcp?? I don't know what other values it could be though.

lucasponce:
> HTTP edges are green with only the triangle arrowhead. So, I don't see a need to even show "http" here, unless that could be something else other than http or tcp?

jmazzitelli:
> Well I missed that grpc is labeled

Ah... that is an important one. So, yeah, I think we should keep this label.

So do we want "http" (or "grpc") to PREFIX the number? Or appear on its own line in the stack of values?

For example, should the edge label be this?

```
http 0.33
34.5%
```

or this?

```
http
0.33
34.5%
```

jshaughn:
@lucasponce @jmazzitelli I pushed another round of revisions.  Feel free to look+comment, or we could meet tomorrow briefly and talk through any final revisions.  In this push:
- show units on:
  - multiple labels
  - hover
  - select
- show error rates only on hover and selection, to reduce verbosity
- consistently stack values including protocol and security badge.  More consistent and less overlap
- move request distribution to bottom of stack
- enable all labels for mini-graph
  - note that mini-graph has no legend or info tooltip, do we need one?

jmazzitelli (commented):
I think the detail mini-graph is way too noisy. Even for the basic bookinfo, look at how crowded the edge labels are (look at "reviews"):
image

Also, in the main graph I even see some overlap. I'm not sure what could be done about this - and this is minor because you can always zoom in and out to see what the labels are if they do overlap (but note you don't have that option in the mini-graph - look at the image above and you can see overlap of the "productpage" label that covers most of the "100%" value). But it is "ugly":
Peek 2021-06-14 15-41

jshaughn:
> I think the detail mini-graph is way too noisy

I agree, I don't think we need everything enabled in the mini-graph.    I think we should just get on a call and talk this through interactively.  I don't want this feature to make anything worse for the user.  I just wanted them to have the option to see more.

Note that I don't really expect people to turn everything on, I can imagine maybe combining rate and response time may be popular.

> Also, in the main graph I even see some overlap. I'm not sure what could be done about this...

I actually spent some time trying to change this, such that when the edge is being highlighted the edge labeling shows on top of the node.  There is supposed to be an option to allow this but for some reason it's not available in our build.  After this PR I was going to propose we update cytoscape (probably after 1.36), and revisit this issue.

Anyway. let's see if we can get the three of us on a call and find a happy medium.

lucasponce:
I'm personally ok with this view
image

From user's perspective, when I browse to the details I like to see these data available. I don't find noisy because those data gives information.

Perhaps I may hay one optimization, the "100%" labels are redundant, I see interesting to show them when there are on below values, otherwise it can be seen clear that if there is one single edge there is a "100%" but when there are more the split % is a data I'd like to see.

jmazzitelli (approved):
Dev folks had discussions on this earlier. This is approved.

lucasponce (approved):
Discussed and tested locally.

mattmahoneyrh (approved):

PR #2183, opened 2021-06-21
Sort sublists details pages

xeviknal:
the details description component didn't apply any sorting for both apps, workloads and services (nor in backend/frontend). Therefore, some refreshing of the page made some of the list change the order all the time despite of there wasn't any information change. It was just a sorting change. It bugged me a little bit and distracted me from putting attention on other sections of the detail pages.

Same thing happens with the pod list and the istio config section.

list-glitch

### DIFF
--- a/src/components/IstioConfigCard/IstioConfigCard.tsx
+++ b/src/components/IstioConfigCard/IstioConfigCard.tsx
@@ -64,30 +64,32 @@ class IstioConfigCard extends React.Component<Props> {
       return this.noIstioConfig();
     }
     let rows: IRow[] = [];
-    this.props.items.map((item, itemIdx) => {
-      rows.push({
-        cells: [
-          {
-            title: (
-              <span>
-                <PFBadge badge={IstioTypes[item.type].badge} position={TooltipPosition.top} />
-                {this.overviewLink(item)}
-              </span>
-            )
-          },
-          {
-            title: (
-              <ValidationObjectSummary
-                id={itemIdx + '-config-validation'}
-                validations={item.validation ? [item.validation] : []}
-                style={{ verticalAlign: '-0.5em' }}
-              />
-            )
-          }
-        ]
+    this.props.items
+      .sort((a: IstioConfigItem, b: IstioConfigItem) => (a.type < b.type ? -1 : a.name < b.name ? -1 : 1))
+      .map((item, itemIdx) => {
+        rows.push({
+          cells: [
+            {
+              title: (
+                <span>
+                  <PFBadge badge={IstioTypes[item.type].badge} position={TooltipPosition.top} />
+                  {this.overviewLink(item)}
+                </span>
+              )
+            },
+            {
+              title: (
+                <ValidationObjectSummary
+                  id={itemIdx + '-config-validation'}
+                  validations={item.validation ? [item.validation] : []}
+                  style={{ verticalAlign: '-0.5em' }}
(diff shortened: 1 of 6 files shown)


src/pages/ServiceDetails/ServiceDescription.tsx
```diff
@@ -69,18 +69,20 @@ class ServiceDescription extends React.Component<ServiceInfoDescriptionProps, St
     const workloads: AppWorkload[] = [];
     if (this.props.serviceDetails) {
       if (this.props.serviceDetails.workloads) {
-        this.props.serviceDetails.workloads.forEach(wk => {
-          if (wk.labels) {
-            const appName = wk.labels[serverConfig.istioLabels.appLabelName];
-            if (!apps.includes(appName)) {
-              apps.push(appName);
+        this.props.serviceDetails.workloads
+          .sort((w1: WorkloadOverview, w2: WorkloadOverview) => (w1.name > w2.name ? -1 : 1))
```
lucasponce, line 73: This logic is flipped, look into the other side:
image
(sorry mazz)

src/components/IstioConfigCard/IstioConfigCard.tsx
```diff
@@ -64,30 +64,32 @@ class IstioConfigCard extends React.Component<Props> {
...
-              <ValidationObjectSummary
-                id={itemIdx + '-config-validation'}
-                validations={item.validation ? [item.validation] : []}
-                style={{ verticalAlign: '-0.5em' }}
-              />
-            )
-          }
-        ]
+    this.props.items
+      .sort((a: IstioConfigItem, b: IstioConfigItem) => (a.type < b.type ? -1 : a.name < b.name ? -1 : 1))
```
lucasponce, line 68: Not sure if this is correct, so name should fall only when == 0, otherwise you may have some random sort.

lucasponce (changes requested):
A couple of minor issues for your review.
I think this one could also be backported to the 1.36 as it's a pure bug fix that won't affect into the feature CI testing.

src/pages/ServiceDetails/ServiceDescription.tsx
hhovsepy: But it sorts correctly.
Screenshot from 2021-06-22 17-53-07

src/components/IstioConfigCard/IstioConfigCard.tsx
aljesusg: In this case if type of a is lower show a first. Otherwise compare name ....this is wrong

I think that this should be:
If the type is the same then compare name, if the name is the same too then not care about order, a and then b or b and then a

xeviknal:
@lucasponce comments approached!

lucasponce (approved):

lucasponce:
Note that this one once is approved by @skondkar can be backported to 1.36.
It's a bug fix.

skondkar (approved):
Verified on Kiali UI  1.37.0 (5ecee976070f9fff18ec4ba4c20b3aaed41a848e)
Verified sorting works correctly and the original issue after refresh does not exist.
Verified on Details page of Applications, Workloads and Services. LGTM.

hhovsepy (approved):

lucasponce:
@xeviknal Please, could you prepare the backport fo the v1.36 branch for this one ? Thanks

PR #2184, opened 2021-06-24
Upgrade namespace in overview page to newer installed Istio.

hhovsepy:
Fix for issue https://github.com/kiali/kiali/issues/3988

Requires backend: https://github.com/kiali/kiali/pull/4122

This PR contains a feature for executing Upgrade/Downgrade To Istio on Namespace action in Overview page.

When Kiali CR is configured "IstioUpgradeAction" and appropriate current and newer Istio revisions: "IstioCurrentRevision" and "IstioUpgradeRevision". Then Kiali allows user to upgrade to the newest Istio by setting the 'istio.io/rev' label to namespace.
Then all workload's pods are restared in that namespace to inject the newest Istio Envoy.

### DIFF
--- a/src/components/IstioWizards/WizardActions.ts
+++ b/src/components/IstioWizards/WizardActions.ts
@@ -1387,9 +1387,15 @@ export const buildSidecar = (name: string, namespace: string, state: SidecarStat
   return sc;
 };
 
-export const buildNamespaceInjectionPatch = (enable: boolean, remove: boolean): string => {
+export const buildNamespaceInjectionPatch = (enable: boolean, remove: boolean, revision: string | null): string => {
   const labels = {};
-  labels[serverConfig.istioLabels.injectionLabelName] = remove ? null : enable ? 'enabled' : 'disabled';
+  if (revision) {
+    labels[serverConfig.istioLabels.injectionLabelName] = null;
+    labels[serverConfig.istioLabels.injectionLabelRev] = revision;
+  } else {
+    labels[serverConfig.istioLabels.injectionLabelName] = remove ? null : enable ? 'enabled' : 'disabled';
+    labels[serverConfig.istioLabels.injectionLabelRev] = null;
+  }
   const patch = {
     metadata: {
       labels: labels
--- a/src/config/ServerConfig.ts
+++ b/src/config/ServerConfig.ts
@@ -58,15 +58,21 @@ const defaultServerConfig: ComputedServerConfig = {
   istioAnnotations: {
     istioInjectionAnnotation: 'sidecar.istio.io/inject'
   },
+  istioCanaryRevision: {
+    current: '',
+    upgrade: ''
+  },
   istioIdentityDomain: 'svc.cluster.local',
   istioNamespace: 'istio-system',
   istioLabels: {
     appLabelName: 'app',
     injectionLabelName: 'istio-injection',
+    injectionLabelRev: 'istio.io/rev',
     versionLabelName: 'version'
   },
   kialiFeatureFlags: {
-    istioInjectionAction: true
+    istioInjectionAction: true,
+    istioUpgradeAction: false
   },
   prometheus: {
     globalScrapeInterval: 15,
--- a/src/pages/Overview/OverviewPage.tsx
+++ b/src/pages/Overview/OverviewPage.tsx
@@ -575,7 +575,7 @@ export class OverviewPage extends React.Component<OverviewProps, State> {
     }
 
     if (serverConfig.istioNamespace !== nsInfo.name) {
-      if (serverConfig.kialiFeatureFlags.istioInjectionAction) {
+      if (serverConfig.kialiFeatureFlags.istioInjectionAction && !serverConfig.kialiFeatureFlags.istioUpgradeAction) {
         namespaceActions.push({
           isGroup: false,
           isSeparator: true,
@@ -604,8 +604,9 @@ export class OverviewPage extends React.Component<OverviewProps, State> {
         };
         if (
           nsInfo.labels &&
-          nsInfo.labels[serverConfig.istioLabels.injectionLabelName] &&
-          nsInfo.labels[serverConfig.istioLabels.injectionLabelName] === 'enabled'
+          ((nsInfo.labels[serverConfig.istioLabels.injectionLabelName] &&
+            nsInfo.labels[serverConfig.istioLabels.injectionLabelName] === 'enabled') ||
+            nsInfo.labels[serverConfig.istioLabels.injectionLabelRev])
         ) {
           namespaceActions.push(disableAction);
           namespaceActions.push(removeAction);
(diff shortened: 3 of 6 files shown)


src/pages/Overview/OverviewPage.tsx
```diff
@@ -620,6 +620,37 @@ export class OverviewPage extends React.Component<OverviewProps, State> {
           namespaceActions.push(enableAction);
         }
       }
+      if (serverConfig.kialiFeatureFlags.istioUpgradeAction) {
+        const upgradeAction = {
+          isGroup: false,
+          isSeparator: false,
+          isDisabled: !canWrite,
+          title: 'Upgrade to Newer Istio',
```
lucasponce, line 628: Perhaps here I'd use the values from the "istio.io/rev" label.
It's a minor thing but I'd separate the labeling of the namespace with the "tags" [1] as there can be another level of indirect pointers here.

[1] https://istio.io/latest/docs/setup/upgrade/canary/

src/pages/Overview/OverviewPage.tsx
```diff
@@ -620,6 +620,37 @@ export class OverviewPage extends React.Component<OverviewProps, State> {
...
+          isGroup: false,
+          isSeparator: false,
+          isDisabled: !canWrite,
+          title: 'Downgrade to Current Installed Istio',
+          action: (ns: string) => this.onUpgradeDowngradeIstio(ns, serverConfig.kialiFeatureFlags.istioCurrentRevision)
+        };
+        if (
+          nsInfo.labels &&
+          ((nsInfo.labels[serverConfig.istioLabels.injectionLabelRev] &&
+            nsInfo.labels[serverConfig.istioLabels.injectionLabelRev] === 'prod') ||
```
lucasponce, line 641: The 'prod' constant can be configured by the user, so I'd read this from the config coming from the backend.

src/pages/Overview/OverviewPage.tsx
```diff
@@ -620,6 +620,37 @@ export class OverviewPage extends React.Component<OverviewProps, State> {
...
+            (nsInfo.labels[serverConfig.istioLabels.injectionLabelName] &&
+              nsInfo.labels[serverConfig.istioLabels.injectionLabelName] === 'enabled'))
+        ) {
+          namespaceActions.push(upgradeAction);
+        } else if (
+          nsInfo.labels &&
+          nsInfo.labels[serverConfig.istioLabels.injectionLabelRev] &&
+          nsInfo.labels[serverConfig.istioLabels.injectionLabelRev] === 'canary'
```
lucasponce, line 649: Same here 'prod' and 'canary' are used in the example but it can be defined by the user [1]

[1] https://istio.io/latest/docs/reference/commands/istioctl/#istioctl-experimental-revision-tag-generate

lucasponce:
@hhovsepy a couple of comments but it's the same direction I'd implement it.

I understand there is a pending debate if we may want to ask the user to rollout the deployments or not, in general, Kiali doesn't enter in this operation, as it's more a kubernetes management operation that can be trigger by the user when he/she thinks is appropiate.

So +1 to keep consistency with adding labels and annotations from Istio semantic.

Perhaps, when ready, in the info, there could be an additional message indicating that deployments would need to be re-deployed.

hhovsepy:
The functionality of Upgrades require newer version of Istio installed in the Control Plane.
Also both installed versions of Istio needs to be tagged.
More information about Istio installation for upgrade purposes can be found in https://istio.io/latest/docs/setup/upgrade/canary/

To test this PR:
In kiali ConfigMap make sure that these 3 configurations are set:
`istio:
   istio_current_revision: prod
   istio_upgrade_revision: canary
kiali_feature_flags:
   istio_upgrade_action: true`
the Istio revision values are the tags required while installing both versions of Istio (as required above).

Once they are set, in Overview page, new option is added in namespace actions to upgrade the namespace into newer installed Istio or to downgrade into current installed Istio.

Screenshot from 2021-07-01 16-52-41
Screenshot from 2021-07-01 16-52-52

hhovsepy:
Ready for reviewing.

src/components/IstioWizards/WizardActions.ts
```diff
@@ -1387,9 +1387,14 @@ export const buildSidecar = (name: string, namespace: string, state: SidecarStat
   return sc;
 };
 
-export const buildNamespaceInjectionPatch = (enable: boolean, remove: boolean): string => {
+export const buildNamespaceInjectionPatch = (enable: boolean, remove: boolean, revision: string | null): string => {
```
lucasponce, line 1390:
```
revision: string | null
```
Do you mean 'null' or 'undefined' ?

Typically you can use the expression:

```
revision?: string
```

For undefined/undeclared fields, which I think is what you may want to express here.

src/config/ServerConfig.ts
```diff
@@ -58,15 +58,21 @@ const defaultServerConfig: ComputedServerConfig = {
...
   istioLabels: {
     appLabelName: 'app',
     injectionLabelName: 'istio-injection',
+    injectionLabelRev: 'istio.io/rev',
     versionLabelName: 'version'
   },
   kialiFeatureFlags: {
-    istioInjectionAction: true
+    istioInjectionAction: true,
+    istioUpgradeAction: false
```
lucasponce, line 75: Ah, I put the same comment in the backend :) I see here the default is explicit.
Thanks !

src/components/IstioWizards/WizardActions.ts
hhovsepy: this function can be called from two places, one with revision param value, another without that param at all

src/pages/Overview/OverviewPage.tsx
```diff
@@ -620,6 +620,41 @@ export class OverviewPage extends React.Component<OverviewProps, State> {
...
+      if (
+        serverConfig.kialiFeatureFlags.istioUpgradeAction &&
+        serverConfig.istioCanaryRevision.upgrade &&
+        serverConfig.istioCanaryRevision.current
+      ) {
+        const upgradeAction = {
+          isGroup: false,
+          isSeparator: false,
+          isDisabled: !canWrite,
+          title: 'Upgrade to ' + serverConfig.istioCanaryRevision.upgrade + ' Istio',
```
lucasponce, line 632: As this action should be linked with [1] semantic, what do you think to use something like:

"Upgrade to XYZ revision"

That would help user to understand better the mapping between Kiali and this feature, as also users would need to deal with "tags" and also with other potential naming.

[1] https://istio.io/latest/docs/setup/upgrade/canary/

hhovsepy: agree

src/components/IstioWizards/WizardActions.ts
lucasponce: Ah, and I see that the "null" value is used in the patch to remove the parameter.
Ok.

hhovsepy:
There is a suggestion on original issue about the behavior:
Question to community:
wdyt, should "Enable/Disable/Remove Auto Injection" functionality still be available on Overview Page when Canary upgrade functionality is enabled?
Because if can cause mess in labeling on namespace. I think it is better to hide "Enable/Disable/Remove" actions when Canary upgrade is enabled.
https://github.com/kiali/kiali/issues/3988#issuecomment-875601547

hhovsepy:
Now when upgrade functionality is activated, the enable/disable/remove is not shown:
Screenshot from 2021-07-08 10-45-46

hhovsepy:
And when "istio_upgrade_action: false" then only the Injection actions are shown:
Screenshot from 2021-07-08 12-06-09

hhovsepy:
And when both Upgrade or Injection are disabled:
Screenshot from 2021-07-08 12-11-09

hhovsepy:
Question to community:
Namespace labeling will require workloads to be restarted to have the label applied and the canary Istio envoy to be injected on workloads.
And workloads needs to be restarted in a loop one by one.
So where is the correct place to redeploy workloads? In a UI call to backend per workloads or in a single backend call which will loop over all workloads?
On leave the redeployment to be done manually by Kiali user via kubectl?

lucasponce (approved):

PR #2186, opened 2021-06-25
Fix sync state on URL on Metrics tab

lucasponce:
Fixes https://github.com/kiali/kiali/issues/4106

It's a workaround, Kiali should review how state is shared between components and URL and unify it, right now there is some technical debt from the early designs that have a mix of strategies on this.

### DIFF
--- a/src/components/Metrics/CustomMetrics.tsx
+++ b/src/components/Metrics/CustomMetrics.tsx
@@ -252,6 +252,7 @@ class CustomMetrics extends React.Component<Props, MetricsState> {
                 <MetricsSettingsDropdown
                   onChanged={this.onMetricsSettingsChanged}
                   onLabelsFiltersChanged={this.onLabelsFiltersChanged}
+                  direction={this.state.dashboard?.title || 'dashboard'}
                   labelsSettings={this.state.labelsSettings}
                   hasHistograms={hasHistograms}
                 />
--- a/src/components/Metrics/IstioMetrics.tsx
+++ b/src/components/Metrics/IstioMetrics.tsx
@@ -117,6 +117,7 @@ class IstioMetrics extends React.Component<Props, MetricsState> {
       if (this.props.direction !== prevProps.direction) {
         const settings = MetricsHelper.retrieveMetricsSettings();
         this.options = this.initOptions(settings);
+        this.setState({ labelsSettings: settings.labelsSettings });
       }
       this.spanOverlay.reset();
       this.refresh();
@@ -288,6 +289,7 @@ class IstioMetrics extends React.Component<Props, MetricsState> {
               <MetricsSettingsDropdown
                 onChanged={this.onMetricsSettingsChanged}
                 onLabelsFiltersChanged={this.onLabelsFiltersChanged}
+                direction={this.props.direction}
                 labelsSettings={this.state.labelsSettings}
                 hasHistograms={true}
               />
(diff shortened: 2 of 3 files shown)


hhovsepy (changes requested):
I can still see the original problem.

lucasponce:
> I can still see the original problem.

Oh, don't you see a change ? I mean, the state is reset when you change from the tab, but at least the state is the same that the query performed in the backend with the reset state.

hhovsepy:
What I can see:
The 'quantiles' param is reset when changing the tab and back. As well as in UI. So this is consistent.
The 'byLabels' param is reset only in backend 'dashboards' call, but remain selected in UI filters, after tab change and back. So here is the inconsistency between selected UI filters and REST call params.

lucasponce:
> The 'byLabels' param is reset only in backend 'dashboards' call, but remain selected in UI filters, after tab change and back. So here is the inconsistency between selected UI filters and REST call params.

A good catch, it seems the fix is not complete, I'll work on it.

lucasponce:
@hhovsepy ok, I've spotted the root cause of this problem, you're right.
Basically the IstioMetrics and the MetricsSettingsDropdown have internal state that is shared and synced through the url.

That designs create edges that are out of sync easily, so, I'm going to refactor this as a previous step to keep a single state in the future.

lucasponce:
@hhovsepy I've proposed a different fix for this problem.

It's a workaround, not really happy with the fix itself, but to address proper this, I think that URL + state of the components should be unified and today some components are still using a combination or URL + inner state.

So, to not create an unnecessary refactoring, when tab changes, the state is reset to the defaults.

Let me know how it looks like.

leandroberetta:
> So, to not create an unnecessary refactoring, when tab changes, the state is reset to the defaults.
> 

@lucasponce I was testing this change but I don't see the state being reset, for example:

- Go to a Workload then Inbound Metrics, select a non default option like "Local version"
- Change the tab to Outbound Metrics, "Local version" is not selected
- Come back to Inbound Metrics and "Local version" remains selected

I was expecting to see only default values.

lucasponce:
> Come back to Inbound Metrics and "Local version" remains selected

Ah, if you click the "Inbounb tab" you should see the data updated, but if you move back using the browser, probably you'd keep the old url, and then the tab should use the values defined there.

I'm testing that, it seems in that case there is something that is missing in the fix.

hhovsepy:
> > Come back to Inbound Metrics and "Local version" remains selected
> 
> Ah, if you click the "Inbounb tab" you should see the data updated, but if you move back using the browser, probably you'd keep the old url, and then the tab should use the values defined there.
> 
> I'm testing that, it seems in that case there is something that is missing in the fix.

I can see that selected labels are reset after switching between tabs.

lucasponce:
> I can see that selected labels are reset after switching between tabs.

Yes, there is a corner cose missing that I think I'm finishing, when there is a tab change but URL is populated, some of the parameters where wrongly reset. 

I'll update the fix.

lucasponce:
@hhovsepy @leandroberetta ok, PR updated, it's a temporal fix but to better address this the design of the shared state should change and I prefer to address that refactoring in a separated effort.

hhovsepy (approved):
lgtm

lucasponce:
Hi @skondkar this is ready for QE.

skondkar (approved):
Verified on Kiali UI 1.37.0 (00870e8c8bc3a50f49d6a4daa4c6f644c178df0a)

LGTM.

leandroberetta (approved):
Verified.

PR #2187, opened 2021-06-29
Fix duplicate container options in workload pod logs

jshaughn:
Fix reference-assignment issue.  Also, improve clarity with some variable renaming and remove a redundant check in componentDidUpdate.

Fixes https://github.com/kiali/kiali/issues/4130

To Recreate issue for Testing:
1) Go to a workload detail that has both a proxy and app container.
2) Go to Logs tab
    - container options should be correct at the top of the logs window
3) Change (or reselect) the pod
    - the app container name should be duplicated

After the fix the duplication should go away and everything else should work the same as always for the logs tab.

### DIFF
--- a/src/pages/WorkloadDetails/WorkloadPodLogs.tsx
+++ b/src/pages/WorkloadDetails/WorkloadPodLogs.tsx
@@ -45,7 +45,7 @@ export interface WorkloadPodLogsProps {
   lastRefreshAt: TimeInMilliseconds;
 }
 
-interface Container {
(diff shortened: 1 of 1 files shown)


pbajjuri20:
Tested and Passed on Kiali UI 1.37.0 (473362272f66efc6e0a5ea94fc9a9adf10915e53)

pbajjuri20 (approved):

lucasponce (approved):
Tested.
image

LGTM

PR #2189, opened 2021-06-29
Display hostnames in network section of service details

leandroberetta:
**Description**

This PR enhance the Service Overview page adding a lists of hostnames that can reach the service according to the VirtualServices/Gateways configurations.

**Issue**

[#4067](https://github.com/kiali/kiali/issues/4067)

**Documentation**

If there are VirtualServices related to the Service, the hostnames will be showed as follows:

ex1p

The configuration is the following:

```yaml
kind: Gateway
metadata:
  name: bookinfo-gateway-1
spec:
  selector:
    istio: ingressgateway
  servers:
  - port:
      number: 80
      name: http
      protocol: HTTP
    hosts:
    - bookinfo.com
---
kind: Gateway
metadata:
  name: bookinfo-gateway-2
spec:
  selector:
    istio: ingressgateway
  servers:
  - port:
      number: 80
      name: http
      protocol: HTTP
    hosts:
    - bookinfo.net
---
kind: VirtualService
metadata:
  name: bookinfo
spec:
  hosts:
  - my.bookinfo.com
  - my.bookinfo.net
  gateways:
  - bookinfo-gateway-1
  - bookinfo-gateway-2
  http:
  - match:
    - uri:
        exact: /productpage
    - uri:
        prefix: /static
    - uri:
        exact: /login
    - uri:
        exact: /logout
    - uri:
        prefix: /api/v1/products
    route:
    - destination:
        host: productpage
        port:
          number: 9080
```

If the VirtualService has a wildcard host (*) then it will try to get hosts from the related gateways:

ex2p

The configuration is the following:

```yaml
kind: Gateway
metadata:
  name: bookinfo-gateway-1
spec:
  selector:
    istio: ingressgateway
  servers:
  - port:
      number: 80
      name: http
      protocol: HTTP
    hosts:
    - bookinfo.com
---
kind: Gateway
metadata:
  name: bookinfo-gateway-2
spec:
  selector:
    istio: ingressgateway
  servers:
  - port:
      number: 80
      name: http
      protocol: HTTP
    hosts:
    - bookinfo.net
---
kind: VirtualService
metadata:
  name: bookinfo
spec:
  hosts:
  - "*"
  gateways:
  - bookinfo-gateway-1
  - bookinfo-gateway-2
  http:
  - match:
    - uri:
        exact: /productpage
    - uri:
        prefix: /static
    - uri:
        exact: /login
    - uri:
        exact: /logout
    - uri:
        prefix: /api/v1/products
    route:
    - destination:
        host: productpage
        port:
          number: 9080
```

And finally if at least a Gateway has a wildcard host, it will show the wildcard:

ex3p

The configuration is the following:

```yaml
kind: Gateway
metadata:
  name: bookinfo-gateway-1
spec:
  selector:
    istio: ingressgateway
  servers:
  - port:
      number: 80
      name: http
      protocol: HTTP
    hosts:
    - '*'
---
kind: Gateway
metadata:
  name: bookinfo-gateway-2
spec:
  selector:
    istio: ingressgateway
  servers:
  - port:
      number: 80
      name: http
      protocol: HTTP
    hosts:
    - bookinfo.net
---
kind: VirtualService
metadata:
  name: bookinfo
spec:
  hosts:  
  - '*'
  gateways:
  - bookinfo-gateway-1
  - bookinfo-gateway-2
  http:
  - match:
    - uri:
        exact: /productpage
    - uri:
        prefix: /static
    - uri:
        exact: /login
    - uri:
        exact: /logout
    - uri:
        prefix: /api/v1/products
    route:
    - destination:
        host: productpage
        port:
          number: 9080
```

### DIFF
--- a/src/pages/ServiceDetails/ServiceDetailsPage.tsx
+++ b/src/pages/ServiceDetails/ServiceDetailsPage.tsx
@@ -19,13 +19,13 @@ import Namespace from '../../types/Namespace';
 import * as AlertUtils from '../../utils/AlertUtils';
 import { PromisesRegistry } from '../../utils/CancelablePromises';
 import { ServiceDetailsInfo } from '../../types/ServiceInfo';
-import { PeerAuthentication, Validations } from '../../types/IstioObjects';
+import { Gateway, PeerAuthentication, Validations } from '../../types/IstioObjects';
 import ServiceWizardDropdown from '../../components/IstioWizards/ServiceWizardDropdown';
 import TimeControl from '../../components/Time/TimeControl';
 
 type ServiceDetailsState = {
   currentTab: string;
-  gateways: string[];
+  gateways: Gateway[];
   serviceDetails?: ServiceDetailsInfo;
   peerAuthentications: PeerAuthentication[];
   validations: Validations;
@@ -93,14 +93,13 @@ class ServiceDetails extends React.Component<ServiceDetailsProps, ServiceDetails
             namespaces.map(ns => API.getIstioConfig(ns.name, ['gateways'], false, '', ''))
           )
           .then(responses => {
-            let gatewayList: string[] = [];
+            let gateways: Gateway[] = [];
             responses.forEach(response => {
-              const ns = response.data.namespace;
               response.data.gateways.forEach(gw => {
-                gatewayList = gatewayList.concat(ns.name + '/' + gw.metadata.name);
+                gateways.push(gw);
               });
             });
-            this.setState({ gateways: gatewayList });
+            this.setState({ gateways: gateways });
           })
           .catch(gwError => {
             AlertUtils.addError('Could not fetch Gateways list.', gwError);
@@ -131,6 +130,16 @@ class ServiceDetails extends React.Component<ServiceDetailsProps, ServiceDetails
         AlertUtils.addError('Could not fetch PeerAuthentications.', error);
       });
   };
+
+  private getGatewaysAsList(): string[] {
+    let gatewayList: string[] = [];
+
+    this.state.gateways.forEach(gateway => {
+      gatewayList.push(gateway.metadata.namespace + '/' + gateway.metadata.name);
+    });
+
+    return gatewayList;
+  }
 
   private renderTabs() {
     const overTab = (
@@ -203,7 +212,7 @@ class ServiceDetails extends React.Component<ServiceDetailsProps, ServiceDetails
         workloads={this.state.serviceDetails.workloads || []}
         virtualServices={this.state.serviceDetails.virtualServices}
         destinationRules={this.state.serviceDetails.destinationRules}
-        gateways={this.state.gateways}
+        gateways={this.getGatewaysAsList()}
         peerAuthentications={this.state.peerAuthentications}
         tlsStatus={this.state.serviceDetails.namespaceMTLS}
         onChange={this.fetchService}
--- a/src/pages/ServiceDetails/ServiceDetailsPage.tsx
+++ b/src/pages/ServiceDetails/ServiceDetailsPage.tsx
@@ -93,13 +93,7 @@ class ServiceDetails extends React.Component<ServiceDetailsProps, ServiceDetails
             namespaces.map(ns => API.getIstioConfig(ns.name, ['gateways'], false, '', ''))
           )
           .then(responses => {
-            let gateways: Gateway[] = [];
-            responses.forEach(response => {
-              response.data.gateways.forEach(gw => {
-                gateways.push(gw);
-              });
-            });
-            this.setState({ gateways: gateways });
+            this.setState({ gateways: responses.map(response => response.data.gateways).flat() });
           })
           .catch(gwError => {
             AlertUtils.addError('Could not fetch Gateways list.', gwError);
@@ -132,13 +126,7 @@ class ServiceDetails extends React.Component<ServiceDetailsProps, ServiceDetails
   };
 
   private getGatewaysAsList(): string[] {
-    let gatewayList: string[] = [];
-
-    this.state.gateways.forEach(gateway => {
-      gatewayList.push(gateway.metadata.namespace + '/' + gateway.metadata.name);
-    });
-
-    return gatewayList;
+    return this.state.gateways.map(gateway => gateway.metadata.namespace + '/' + gateway.metadata.name);
   }
 
   private renderTabs() {
--- a/src/pages/ServiceDetails/ServiceInfo.tsx
+++ b/src/pages/ServiceDetails/ServiceInfo.tsx
@@ -4,7 +4,7 @@ import { Grid, GridItem, Stack, StackItem } from '@patternfly/react-core';
 import ServiceId from '../../types/ServiceId';
 import ServiceDescription from './ServiceDescription';
 import { ServiceDetailsInfo } from '../../types/ServiceInfo';
-import { ObjectValidation, PeerAuthentication, Validations } from '../../types/IstioObjects';
+import { Gateway, ObjectValidation, PeerAuthentication, Validations } from '../../types/IstioObjects';
 import { RenderComponentScroll } from '../../components/Nav/Page';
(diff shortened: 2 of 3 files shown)


lucasponce:
Really nice @leandroberetta, it's implemented as it was thought.

An early comment:
image

In the new "hostnames" I'd add a tooltip (similar as it's shown in the "endpoints") with more details about where these hostnames come from. There user can get details if the resolution happens in Gateway, VirtualService or both.

lucasponce:
I have an scenario that may be also related / synced with https://github.com/kiali/kiali/issues/4068 that @israel-hdez is working on.

I may have a Gateway defining:

```
    - hosts:
        - '*.istio-cluster.org'
```

With an open wildcard that can be restricted in the VirtualService:

```
  hosts:
    - '*.travel-control.istio-cluster.org'
``` 

So in that case, the Service details is showing the interested info at this level:

image

But the nice thing in https://github.com/kiali/kiali/issues/4068 would be that the '*.istio-cluster.org' is identified visually at Gateway level (linked on this case at Ingress).

jshaughn:
Nice addition and as mentioned this can be leveraged in kiali/kiali#4068.   One minor comment/suggestion.  Perhaps when there are multiple values for a Network attribute (Endpoints, Ports, Hostnames), the values should be dropped below the attribute name and indented.   For Hostnames especially,  this may prevent some horizontal scrolling.  It's fine as is if we prefer to minimize vertical space used.

lucasponce:
Not sure what you suggest here @jshaughn have you looked into other services ?

image

image

I'd say it's pretty much the spected style for that card

src/pages/ServiceDetails/ServiceDetailsPage.tsx
```diff
@@ -93,14 +93,13 @@ class ServiceDetails extends React.Component<ServiceDetailsProps, ServiceDetails
...
+            let gateways: Gateway[] = [];
             responses.forEach(response => {
-              const ns = response.data.namespace;
               response.data.gateways.forEach(gw => {
-                gatewayList = gatewayList.concat(ns.name + '/' + gw.metadata.name);
+                gateways.push(gw);
               });
             });
-            this.setState({ gateways: gatewayList });
+            this.setState({ gateways: gateways });
```
israel-hdez, line 102: You probably can simplify lines 96-102 to a single statement:

```js
this.setState({ gateways: responses.map(response => response.data.gateways).flat() });
```

src/pages/ServiceDetails/ServiceDetailsPage.tsx
```diff
@@ -132,6 +131,16 @@ class ServiceDetails extends React.Component<ServiceDetailsProps, ServiceDetails
...
   };
 
+  private getGatewaysAsList(): string[] {
+    let gatewayList: string[] = [];
+
+    this.state.gateways.forEach(gateway => {
+      gatewayList.push(gateway.metadata.namespace + '/' + gateway.metadata.name);
+    });
+
+    return gatewayList;
```
israel-hdez, line 141: Also, you probably can simplify with `map`:

```suggestion
    return this.state.gateways.map(gateway => gateway.metadata.namespace + '/' + gateway.metadata.name);
```

That `map` function is magical :-)

israel-hdez:
I was hoping I can borrow some Leandro's code, but it is only UI changes, while I need Go code.

BTW, I cannot clearly see any GW filtering/matching. I have the idea that this will show hostnames of all available gateways in the namespace, regardless if they do not match the service being viewed.

src/pages/ServiceDetails/ServiceDetailsPage.tsx
leandroberetta: Thans for your review @israel-hdez! sure, good simplifications, totally agree!

src/pages/ServiceDetails/ServiceDetailsPage.tsx
leandroberetta: Yes, sure, thanks!.

leandroberetta:
@lucasponce I added tooltips to describe where every hostname is coming from:

ex1-tooltip
ex2-tooltip
ex3-tooltip

@israel-hdez I'm displaying every possible source of traffic for this specific service, not for all the namespace, maybe at graph level that is ok, but in the minigraph I imagine showing same values (or nothing as they are already shown in the network card).

leandroberetta:
> Nice addition and as mentioned this can be leveraged in [kiali/kiali#4068](https://github.com/kiali/kiali/issues/4068). One minor comment/suggestion. Perhaps when there are multiple values for a Network attribute (Endpoints, Ports, Hostnames), the values should be dropped below the attribute name and indented. For Hostnames especially, this may prevent some horizontal scrolling. It's fine as is if we prefer to minimize vertical space used.

Horizontal scrolling is a concern I have for long hostnames that indeed are very common to have, I checked what happens with a large one and the results are:

hostname_large

So it's dropping below but with no identation, looks odd I think. I will try some things to improve this situation. Thanks @jshaughn for noticing this.

lucasponce:
> So it's dropping below but with no identation, looks odd I think. I will try some things to improve this situation. 

Ah, but note that the card it's designed to be flexible on narrow screens, that's why I commented.

On large names you can use the ellipsis pattern used in the titles of the metrics, I think that code is easily reusable, it's just style.

But note that corner cases will apply to all places, so, any field (name, namespace) can be populated in a very long name and we will let on those cases that style cover that (pods names are an example, but the card is different).

leandroberetta:
> On large names you can use the ellipsis pattern used in the titles of the metrics, I think that code is easily reusable, it's just style.

I like this idea, I will look that example and add it to this case.

leandroberetta:
I added the styles for overflowing hostnames:

hostname_1
hostname2

The only thing I don't like is that the icon goes away with the ellipsis, I tried to keep it but in the case that the ellipsis is not required (enough space), the icon lays aligned to the right, making it different compared to the icon in endpoints for example.

I think that this design for now it's ok but a card redesign will be required for this new situations where long texts (with no spaces) are needed.

lucasponce:
> The only thing I don't like is that the icon goes away with the ellipsis, I tried to keep it but in the case that the ellipsis is not required (enough space), the icon lays aligned to the right, making it different compared to the icon in endpoints for example.

Personally I'd like to keep the "info" tooltip icon where the long text can be there as well without issues.
Perhaps then, "the ellipsis" here can be implemented "manually" just setting a max length in the text to show and using tooltip for  details.

leandroberetta:
@lucasponce I guess I was able to fix the situation thanks to css flex.

hostname_21
hostname_23_tooltip
hostname_22

lucasponce:
Thanks @leandroberetta, CSS issues always need iterations. I understand how time consuming they can be.

Please, one last request, can we align the tooltip ?

image

We're doing an effort that tooltip style looks consistent across elements, some if missing, but I'd like if you can align to left as shown in the screenshot.

leandroberetta:
Sure @lucasponce, I also modified the endpoint's tooltip with your suggestion.

The result is:

tooltip_left

As you can see, I left aligned the text inside the tooltips. Not 100% convinced about the tooltip content, what do you think?.

lucasponce:
> Not 100% convinced about the tooltip content, what do you think?.

Gateway/Virtualservice <name>:
<hostname>

```
Gateway bookinfo-gateway:
longnamehost....com
```

I think rest of cards in the details follow a simple pattern for tooltips as the context is clear.

leandroberetta:
@lucasponce thanks for your comments.

The results are:

tooltip_new

lucasponce (approved):

prachiyadav (approved):
Screenshot from 2021-07-08 17-11-53

prachiyadav:
Screenshot from 2021-07-08 17-15-00

lucasponce:
Hi @israel-hdez, is this PR ok from your side ?

israel-hdez (approved):
I'm good!
Merge it to make it available for 1.37! :-D

prachiyadav:
Test Case > SWSQE-1850

PR #2196, opened 2021-07-05
Add hostnames list for gateway and virtual service nodes

israel-hdez:
In the main graph, when clicking on nodes representing ingress gateways, the side panel will provide the list of configured hostnames for the selected gateway.

The main graph will be showing a small text "x hosts" to denote that certain node has a list of configured hostnames to serve.

hosts-hover2

Fixes kiali/kiali#4068

**SERVER PR:** https://github.com/kiali/kiali/pull/4138


israel-hdez:
OK, @jshaughn @lucasponce, I have this in the UI part:

image

Notice that new icon/badge representing gateways. It only affects ingresses. Honestly, I'm not sure I like it. If you don't like it, I'll happily revert and leave only the hostnames list in the side panel.

Hostnames are for gateways. This won't show whatever is configured in a VS, because this is for the GW node. If it is of any value, I can try to complement this work with hostnames for Service nodes, in case they have a VS attached to them.

lucasponce:
The info in the side panel is good, but I think that as user to have some option to enable the hostnames in the graph would give me a lot of most useful information:

image

On this screenshot the "control" workload would work as "ingress" with a generic "*.my-portal.com" that are personalized with VirtualServices on the next level of nodes.

When user is working with inner traffic, probably services and workloads names are more interesting, but when it's an scenario based on multiple gateways I'm sure users would appreciate this option.

jshaughn:
> Notice that new icon/badge representing gateways. It only affects ingresses. Honestly, I'm not sure I like it. If you don't like it, I'll happily revert and leave only the hostnames list in the side panel.

I think I like the Gateway badge, but I'm not sure it should replace the TrafficSource badge.  I'd maybe like to see how things look with the TrafficSource badge as well, and the TS badge would always be the left-most badge.   

> Hostnames are for gateways. This won't show whatever is configured in a VS, because this is for the GW node. If it is of any value, I can try to complement this work with hostnames for Service nodes, in case they have a VS attached to them.

This seems like a good idea.  Whether a GW or VS the defined hosts are valuable.

As for the side panel, I'm not sure if I love the badged list in the header, it's a little verbose, but I also don't know what else to do with it.  It does need to be prominent.  Maybe something to show to UX when everything is nearly done.

> The info in the side panel is good, but I think that as user to have some option to enable the hostnames in the graph would give me a lot of most useful information:

I think initially I'd like to see just the side-panel.  A Display option is possible but the information could be very verbose and the graph is already dense.  If we did do it then probably the existing HTML label could have a detail section below the current information, maybe something like a better version of:

image

israel-hdez:
@jshaughn @lucasponce OK, with the last commits, this is what I have:

image

I definitely wouldn't like to make the hostnames too distracting. So, I'm using subtle styles.
So far, no display option to show/hide hostnames from graph.

israel-hdez:
Added to tomorrow's UX agenda.

lucasponce:
cc-ing @andrew-ronaldson, even if we add a topic in the UX agenda, if we can add Andrew sooner, the better, then he can catch up before the meeting about the topic.

andrew-ronaldson:
I think we discussed something like this in last weeks meeting. Adding it to the label area and let the user turn on this view in the display options. This is a rough screenshot that I cut up from your example. The text should match more closely but I think it could still be a bit smaller as it is the secondary information.
hostname option

israel-hdez:
@andrew-ronaldson It looks better than my PoC.
The text of the hostnames, if I remember correctly, we want a bigger font size. Notice that in my screenshot, the graph is truncated because it is zoomed-in. On the initial non-zoomed size, using this font-size the hostnames are hard to read. So, we want a font-size that is closer in size to the main label, but still somehow differentiate from the main label.

andrew-ronaldson:
Could be a thin horizontal line to separate the two sections and it wouldn't add too much vertical space. Would this be visible by default or enabled in the display options?

israel-hdez:
@andrew-ronaldson I have this:

image

When the node is selected, or the mouse cursor is on the node, the hosts part does not change style:

image



> Would this be visible by default or enabled in the display options?

Honestly, I would like to avoid new display options. I understand this is important information to have, so IMHO having this information hidden by default doesn't sound right. But making it visible by default can result in a busy graph. So, if a clean graph cannot be achieved without a new display option, then I would say it's OK to add it. But I would like to avoid it if possible.


Playing a little bit with the graph control, I see it is possible to show the hosts when the nodes are with the "highlight" style:

hosts-hover

Also, it is possible to have a small text telling that there is some extra information, and then expand the extra information when the mouse is over the small block of text:

hosts-hover2


Just dropping some ideas, although I'm not sure if this is good on an UX perspective.

jshaughn:
@israel-hdez Great stuff, looks really nice.   I like all of these although I think maybe your last option is best, with the expandable counts.  That gives a consistent label size across nodes, provides relevant information while hiding details until the user wants to see it.

I think perhaps we should still handle large lists by perhaps showing a max (5, 10?) and if longer then force a selection of the node and defer to the side panel.  In the side panel I think I'd like to see something scrollable as well.

I think what you have incorporates the current feedback from @andrew-ronaldson , and seeing that he is currently out, I think we can continue with this.  If you like, update the PR and assign me for review.   Even if we merge we can still iterate as needed with Andrew.

israel-hdez:
OK, this is finished.

The final behavior is the following:

hosts-hover2

> @jshaughn   In the side panel I think I'd like to see something scrollable as well.

I didn't do this. I wonder if you meant that we want nested scrollbar?
We currently have a global scrollbar for the sidepanel:

image

Nesting scrollbars is probably not very natural, but it's doable. Let me know if you want a nested scrollbar to add it.

andrew-ronaldson:
Sorry I wasn't around to review this earlier but this is looking good @israel-hdez

src/components/CytoscapeGraph/graphs/GraphStyles.ts
```diff
@@ -361,8 +385,24 @@ export class GraphStyles {
       return `<div class="${labelDefault} ${labelBox}" style="${labelStyle}">${icons}${contentSpan}</div>`;
     }
 
-    const contentSpan = `<div class="${contentClasses}" style="${contentStyle}">${contentText}</div>`;
-    return `<div class="${labelDefault}" style="${labelStyle}">${icons}${contentSpan}</div>`;
+    let hosts: string[] = [];
+    node.hasVS?.hostnames?.forEach(h => hosts.push(h === '*' ? '(All hosts)' : h));
```
jshaughn, line 389: I'm not sure, maybe just showing `*` would be better than `(All Hosts)`.  I actually found `(All Hosts)` a little confusing.  Or, perhaps both `* (all hosts)`

src/components/CytoscapeGraph/graphs/GraphStyles.ts
```diff
@@ -361,8 +385,24 @@ export class GraphStyles {
...
+    node.isGateway?.ingressInfo?.hostnames?.forEach(h => hosts.push(h === '*' ? '(All hosts)' : h));
+
+    let htmlHosts = '';
+    if (hosts.length !== 0) {
+      let hostsToShow = hosts;
+      if (hostsToShow.length > config.graph.maxHosts) {
+        hostsToShow = hosts.slice(0, config.graph.maxHosts);
+        hostsToShow.push((hosts.length - config.graph.maxHosts) === 1 ? "(and one more host)" : `(and ${hosts.length - config.graph.maxHosts} more hosts)`);
```
jshaughn, line 397: maybe ellipses and no parentheses?  Also, maybe only numbers (I'm also OK with just keeping 'one')?

1 more host...
5 more hosts...

src/pages/Graph/SummaryLink.tsx
```diff
@@ -127,10 +127,10 @@ const getLink = (nodeData: GraphNodeData, nodeType?: NodeType) => {
 
 export const renderBadgedHost = (host: string) => {
   return (
-    <span>
+    <div>
       <PFBadge badge={PFBadges.Host} />
-      {host}
-    </span>
+      {host === "*" ? "(All hosts)" : host}
```
jshaughn, line 132: Same here, I like keeping the `*` as that reflects the actual config entry (right?).

src/pages/Graph/SummaryPanelNode.tsx
```diff
@@ -100,13 +102,31 @@ export class SummaryPanelNode extends React.Component<SummaryPanelNodeProps, Sum
...
+  };
+
+  private renderVsHostnames = (nodeData:  DecoratedGraphNodeData) => {
+    return (
+      <div>
+        {nodeData.hasVS?.hostnames?.map(hostname => renderBadgedHost(hostname))}
+      </div>
+    );
+  };
+
```
jshaughn, line 129: I think this where we need to be more creative to avoid a lot of vertical space and an annoying number of badges.   Maybe just one badge and if there is just one host, show it.  Otherwise show a fixed scrolling box with the entries?

jshaughn (changes requested):

src/pages/Graph/SummaryLink.tsx
israel-hdez: Right, the star reflects the config entry.

I'm taking your suggestion of keeping both: `* (all hosts)`

src/components/CytoscapeGraph/graphs/GraphStyles.ts
israel-hdez: OK, looks cleaner, I'll use your provided example.

src/pages/Graph/SummaryPanelNode.tsx
israel-hdez: you mean using a comma separated list, rather than a vertical list with "bullets" (where the bullet is the badge)?

israel-hdez: We can do scrolling, but this is how it will look:

image

I don't like that much the double scrollbars in the side panel. But if it is OK for you, I can add them.

andrew-ronaldson: What about an [expandable section](https://www.patternfly.org/v4/components/expandable-section)?  It could reduce the scrolling without using a secondary scroll.

jshaughn: I think an expandable section could work, and would not need to scroll because, as you said, we already have vertical scrolling. The only suggestion I have is that we move it to the "badge summary area":

image

The above is just to show, as a reminder, that we have this section.  I think we should add an "Is Gateway" to this, and for gateways and relevant VS we could add an expandable section underneath.  > X hosts... or something like that.

What do you think?  Hopefully not a CSS nightmare.

israel-hdez: It shouldn't be hard. I'll try it.

israel-hdez: This is how it looks:

image

It's in the last commit.

jshaughn: Can we have it directly below the "Is gateway" and not detached with the Workload link in between?

israel-hdez: This is the updated look in commit https://github.com/kiali/kiali-ui/pull/2196/commits/01225bc02e5b8448123543fa9b66d4e85101f404

image

I also moved around the hosts list for the VirtualService/RequestRouting case.


> Any idea what happens if the value is long? Do we need ellipses and a tooltip?

The text should wrap to the next line... Probably it won't look that good.

jshaughn: This looks good to me, let me when you are complete and I will pull and test.

israel-hdez: Done:

image

jshaughn (approved):
LGreat2M

pbajjuri20:
LGTM! 

graph_1



graph

pbajjuri20 (approved):

PR #2205, opened 2021-07-15
Make request routing icon the leftmost for consistency

nrfox:
Aside from root nodes, this moves the request routing icon to be the leftmost icon for consistency's sake so that the current VS icon always appears as the leftmost one.

Screenshot from 2021-07-15 14-49-48

Relates to kiali/kiali#4090 but is not required and can be merged independently of those changes.


jshaughn (approved):
Good idea, change looks fine.

pbajjuri20 (approved):
Tested and verified on Kiali UI 1.38.0 (d04454693a3f0df8f609ca7a71f65a657b9831b2)


kiali_wizard-1

PR #2208, opened 2021-07-22
Prefix validation message with the now explicit code

jmazzitelli:
Part of: https://github.com/kiali/kiali/issues/4197

This goes with PR https://github.com/kiali/kiali/pull/4205 - see that for more details.

I don't like this PR because there should be a common function you can call that takes a "check" object as an input parameter and returns the full message string (which is prefixed with `code` if it is not empty - e.g. `return (check.code ? check.code + ' ' : '') + check.message).` Right now that expression is everywhere that `check.message` was before. Someone enhance this PR so it does that :)

### DIFF
--- a/src/components/Validations/GlobalValidation.tsx
+++ b/src/components/Validations/GlobalValidation.tsx
@@ -63,7 +63,7 @@ class GlobalValidation extends React.Component<Props> {
 
   message(): string {
     const checks = this.globalChecks();
-    let message = checks.map(check => check.message).join(',');
+    let message = checks.map(check => (check.code ? check.code + ' ' : '') + check.message).join(',');
 
     if (!message.length && !this.isValid()) {
       message = 'Not all checks passed!';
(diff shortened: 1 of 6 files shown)


jmazzitelli:
I'm going to take this out of draft mode - it works, and all tests pass. So we can open it up for review now

xeviknal (approved):
LGTM!

xeviknal:
Regarding to that: https://github.com/kiali/kiali/pull/4205#issuecomment-885728836
I think we can consolidate this work and approach it in a subsequent issue. I think we should see how community is using the feature and see where we can put the warning afterwards. We shouldn't be pretty intrusive showing it all the time, the masthead starts to be a bit crowded. wdyt? thanks anyway!

jmazzitelli:
> I think we should see how community is using the feature and see where we can put the warning afterwards.

Agree.

PR #2210, opened 2021-07-26
Move React-scripts and node-sass to dependencies

aljesusg:
Fix https://github.com/kiali/kiali/issues/4201

Temporary solution until new feature


aljesusg:
mmm I am going to work with my workaround...I am not happy with snyk test you were right @israel-hdez

PR #2212, opened 2021-07-29
Add a new section to get internal certificates information

leandroberetta:
UI changes for [#4227](https://github.com/kiali/kiali/pull/4227)


src/components/IstioCertsInfo/IstioCertsInfo.tsx
```diff
@@ -0,0 +1,181 @@
...
+    );
+  };
+
+  render() {
+    return (
+      <Modal
+        isSmall={true}
+        isOpen={this.state.showModal}
+        onClose={this.close}
+        title="Certicates information"
```
abonas, line 110: Typo - Certicates should be certificates :)

leandroberetta: @abonas thanks! it's fixed now.

src/components/IstioCertsInfo/IstioCertsInfo.tsx
```diff
@@ -0,0 +1,181 @@
...
+import { PFColors } from 'components/Pf/PfColors';
+import { KialiIcon } from 'config/KialiIcon';
+import { infoStyle } from 'styles/DropdownStyles';
+
+type IstioCertsInfoState = {
+  showModal: boolean;
+  certsError: Boolean;
+};
+
+type IstioCertsInfoProps = {
```
xeviknal, line 33: We usually follow one convention for Redux properties. We split the strict native properties from the component from the Redux one. We do this for better readability. I find it pretty useful for troubleshooting and understand what's part of this component and what's from redux (and therefore used beyond the component).

See the example here:
https://github.com/kiali/kiali-ui/blob/ff87e9a66894ef0f13c01a7b8f086c7a282a9aa4/src/components/IstioActions/IstioActionsButtons.tsx#L10-L24

xeviknal (commented):
Just one little comment about conventions and redux. Pretty nitpicking though.

src/components/IstioCertsInfo/IstioCertsInfo.tsx
lucasponce: Note that this convention is not followed 100% as we rely as well in comments (not new).
image

xeviknal: Actually, this type only contains Redux props. The component props are split and declared in the previous type `IstioMetricsProps`. So the main difference between my example and yours is the way you intersect both types. However, this last example makes it a bit harder to understand because devs need the comment to make that explicit. If devs see the whole type definition, then they understand that those props are only from Redux.

leandroberetta: I'm OK to introduce this change, it's added now. Thanks!.

xeviknal (approved):
LGTM!

lucasponce (approved):

PR #2219, opened 2021-08-12
Fix a bug where when view only mode is set, workload actions dropdown is not disabled

leandroberetta:
**Description**

This PR fixes a bug that occurs when viewOnlyMode is set to true, the Workload's actions dropdown is not disabled.

**Issue**

Fixes [#4202](https://github.com/kiali/kiali/issues/4202)

**Related PRs**

Requires [#4273](https://github.com/kiali/kiali/pull/4273)


lucasponce:
Some suggestions about the behaviour, for consistency in all the flows:

- In the workloads, instead to disable all actions, I'll disable the action item, then user should see what it's available:
image

- I need to check if it's my environment, but I think when "read_only_mode: true", all actions should be disabled, on the Istio Config details, if should the disabled the "Delete" action item and "Save" buttons (perhaps is my environment):
image

- Same comments apply for the Services actions, don't disable the actions drowndown, but the items, then users should know which actions are available:
image

leandroberetta:
> Some suggestions about the behaviour, for consistency in all the flows:
> 
> * In the workloads, instead to disable all actions, I'll disable the action item, then user should see what it's available:
>   image
> * I need to check if it's my environment, but I think when "read_only_mode: true", all actions should be disabled, on the Istio Config details, if should the disabled the "Delete" action item and "Save" buttons (perhaps is my environment):
>   image
> * Same comments apply for the Services actions, don't disable the actions drowndown, but the items, then users should know which actions are available:
>   image

@lucasponce ok, makes sense to me, this PR only tried to mimic how the Services' Actions dropdown behaves but I guess we can disabled the items instead of all the dropdown, let me check it.

Regarding the Istio Config view, i'm not seeing that in my environment, delete and save buttons are disabled.

leandroberetta:
I updated a little bit how the items are displayed when disabled due to view only mode in true.

Workloads:

workload_view

Services:

services_view

Istio Config:

istio_config_4


Also, I noticed that Istio Config actions needed the validation too:

istio_config_view_3

leandroberetta:
This can be tested setting view_only_mode property to true in Kiali CR as follows:

```yaml
apiVersion: kiali.io/v1alpha1
kind: Kiali
metadata:    
  name: kiali
  namespace: kiali-operator
spec:
  auth:
    strategy: anonymous
  deployment:
    ...
    view_only_mode: true
    ...
  istio_namespace: istio-system
  version: default
```

Then checking the Actions dropdown from Workloads, Services, IstioConfig (both the list and the page when one config is selected).

 Also, a regression would be needed to check when view_only_mode is set to false to check that everything stills working (related to Actions dropdown).

@lucasponce what do you think of the latest changes?.

src/components/IstioActions/IstioActionsDropdown.tsx
```diff
@@ -56,23 +59,34 @@ class IstioActionDropdown extends React.Component<Props, State> {
     this.props.onDelete();
   };
 
+  renderTooltip = (key, position, msg, child): any => {
```
lucasponce, line 62: Better to move the "any" types per proper ones, on this case "React.ReactFragment" could be the appropiate I think.
There could be some "any" use in the code but when we can, let's try to move it to strong types, that would help future migrations.

src/components/IstioActions/IstioActionsNamespaceDropdown.tsx
```diff
@@ -33,23 +47,45 @@ class IstioActionsNamespaceDropdown extends React.Component<Props, State> {
     history.push('/istio/new/' + type);
   };
 
+  renderTooltip = (key, position, msg, child): any => {
```
lucasponce, line 50: Same comment, try to use the type that better matches this one (JSX.Element or React.ReactFragment).

lucasponce (approved):
A couple of minor changes to replace "any" types but this would be ready for QE to test.
Code reviewed and I've tested locally

@leandroberetta would be good if you can add a comment defining the QE flow that they should test. Thanks !

jmazzitelli:
I think this should be backported to 1.36

PR #2220, opened 2021-08-17
Add WorkloadEntry appender

nrfox:
Adds WorkloadEntry appender to default appender list.

Backend-changes: kiali/kiali#4284

Relates to kiali/kiali#4223

### DIFF
--- a/src/services/GraphDataSource.ts
+++ b/src/services/GraphDataSource.ts
@@ -180,7 +180,7 @@ export default class GraphDataSource {
     }
 
     // Some appenders are expensive so only specify an appender if needed.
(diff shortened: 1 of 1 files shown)


jshaughn (approved):
This is fine, I'll approve, but what's the point of calling the appender if we're not doing anything with the new information?  Why not add this with the code that deals with the new flag?

nrfox:
> This is fine, I'll approve, but what's the point of calling the appender if we're not doing anything with the new information? Why not add this with the code that deals with the new flag?

That's a good point. I'll just close this and merge these changes into #2221.

PR #2221, opened 2021-08-20
Add VM badge to workloads with WorkloadEntries

nrfox:
Adds a VM badge to workload nodes that have `WorkloadEntries`.

Fixes kiali/kiali#4220

Screenshot from 2021-08-18 16-18-58

Backend changes: kiali/kiali#4284


lucasponce:
@nrfox feel free to add @andrew-ronaldson (UX)
Typically the flow we have is:
- Start adding UX folks into issues/PRs
- Then decide, if the issue is relevant to bring it to the UX agenda or if it can be resolved inside the PR/issue (because it's minor, or clear).

Having said that, the icon you choose is already "taken" for the "mirror" feature in the Wizards, so probably we'd like to use a different one (the previous icon was also discussed with @lwrigh - former UX - in the past).

nrfox:
> @nrfox feel free to add @andrew-ronaldson (UX)
> Typically the flow we have is:
> 
>     * Start adding UX folks into issues/PRs
> 
>     * Then decide, if the issue is relevant to bring it to the UX agenda or if it can be resolved inside the PR/issue (because it's minor, or clear).
> 
> 
> Having said that, the icon you choose is already "taken" for the "mirror" feature in the Wizards, so probably we'd like to use a different one (the previous icon was also discussed with @lwrigh - former UX - in the past).

Thanks for the context. I'll add @andrew-ronaldson as a reviewer. I'm happy to change the icon to something else that works as a stand-in for a VM. As a starting point, some potential options that remind me of VMs from the [pf icons list](https://www.patternfly.org/v4/guidelines/icons/):

- pf-icon-cluster
- fa-desktop
- fa-hdd

Sorry I can't drop svgs into github comments directly apparently.

jshaughn:
> ...the icon you choose is already "taken" for the "mirror" feature in the Wizards, so probably we'd like to use a different one

I think we should keep Nick's choice of the VM icon to represent actual VMs.  We could possibly use the same icon for mirroring or perhaps change the mirror icon to the font-awesom [clone icon](https://fontawesome.com/v5.15/icons/clone?style=regular)?

andrew-ronaldson:
I noticed that the badge in the screenshot above is the correct VM icon in PF so I was thinking the same thing as @jshaughn 
Would changing the mirror feature icon cause any confusion for existing users?

jshaughn:
By the way, when we add node flags we also add the capbility to perform graph Find/Hide on the flag.  So, as part of this PR please make the relevant updates to graphFind.tsx, graphHelpFind.tsx and graphFind.test.tsx

lucasponce:
Or perhaps leave both icons as they are.
I'm revisiting the discussion, and the original icon was different, and the motivation was to bring the Workload one:
https://github.com/kiali/kiali-ui/pull/2017#issuecomment-736985274

So, as the contexts are different I don't think they would bring confusion.

lucasponce:
(Sorry, closed/reopened by mistake) :)

nrfox:
> Or perhaps leave both icons as they are.
> I'm revisiting the discussion, and the original icon was different, and the motivation was to bring the Workload one:
> [#2017 (comment)](https://github.com/kiali/kiali-ui/pull/2017#issuecomment-736985274)
> 
> So, as the contexts are different I don't think they would bring confusion.

It sounds like the VM icon could coexist for both use cases for now?

mattmahoneyrh (approved):
Verified

image

PR #2223, opened 2021-08-24
Convention for redux props usage into props components types

xeviknal:
I had the conception that we were using the following convention for specifying redux props:

https://github.com/kiali/kiali-ui/blob/ff87e9a66894ef0f13c01a7b8f086c7a282a9aa4/src/components/IstioActions/IstioActionsButtons.tsx#L10-L24

@lucasponce proof me wrong here, which I thank for it.
https://github.com/kiali/kiali-ui/pull/2212/files#r694932209

So I thought that we could kind of agree on a way to go altogether.
I propose to use the way as linked right above.

What do you think?

### DIFF
--- a/STYLE_GUIDE.adoc
+++ b/STYLE_GUIDE.adoc
@@ -91,29 +91,26 @@ of the rules needed:
 The declaration of the properties of a component connected to Redux should clearly indicate whether each property comes from Redux or are strictly from the component.
 This convention should help developers easily see which properties comes from the Redux state management without the need to be either versed with the whole Redux catalog or the component itself.
 
-The convention is to
-* Declare a type A which only contains the Redux properties.
-* Declare a type B which only contains the properties from the component.
-* Intersect type A into type B.
-* Use type B for the component declaration.
+The convention is consist in:
+
+* Declaring a type `ReduxProps` which only contains the Redux properties, sorted alphabetically.
+* Declaring a type `<ClassName>Props` which only contains the properties from the component, sorted alphabetically.
+* Intersecting type `ReduxProps` into type `<ClassName>Props`.
+* Using `<ClassName>Props` type for the component declaration.
 
 See an example:
 
 [source, typescript]
 ----
 type ReduxProps = {
- setLastRefreshAt: (lastRefreshAt: TimeInMilliseconds) => void;
+ // just the Redux props, alphabetical
 };
 
-type Props = ReduxProps & {
- objectName: string;
- readOnly: boolean;
- canUpdate: boolean;
- showOverview: boolean;
- overview: boolean;
-}
+type <ClassName>Props = ReduxProps & {
+ // non-Redux props, alphabetical
+};
 
-class StyleComponent extends React.Component<Props> {
+class <ClassName>Component extends React.Component<Props> {
 ...
 }
 ----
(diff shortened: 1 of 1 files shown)


jshaughn:
@xeviknal @lucasponce ,

It is not adopted in all code but I,  and I think others, have been using the convention:

```
 type ReduxProps = { 
   // just the Redux props, alphabetical
 }; 
  
 type <ClassName>Props = ReduxProps & { 
   // non-Redux props, alphabetical
 }; 
```

I personally update to this convention when I'm touching a class, if it's not in place already (and I happen to notice).

nrfox:
I'm of two minds because part of me likes making it explicit where the props are coming from but part of me thinks that it should not matter where the props are coming from, they're just props. Explicit is still better and the advice on how to approach [this stuff](https://redux.js.org/style-guide/style-guide#use-the-react-redux-hooks-api) changes each time I look at it so +1 to this change.

israel-hdez:
Like Nick, I'm with one foot in and the other out.

React won't make any distinction. It's more a conceptual or convention matter.

Also, it's not that strict. When using a component a caller may choose to connect some props to the redux store, while another caller may choose to pass it's own processed value (i.e. without connecting the redux store). But I think we don't have such cases in Kiali.

I'm :+1:, but I cannot promise that my brain won't forget to use this convention.

leandroberetta:
I think it's good for documenting and explicitly indicate the Redux props, +1.

aljesusg:
+1 to this, clean and readable code

hhovsepy:
+ 1 to have this convention documented, also would be good to have a separate issue/task to go through the code and fix the places (if any) when the convention is not met.

jshaughn:
>  go through the code and fix the places (if any) when the convention is not met.

In general I avoid refactoring for the sake of matching convention, or some other stylistic reason.  Every time you touch working code you can introduce a regression.  Personally, I'd recommend making this sort of update when you're already in the code for some other reason.

lucasponce:
I understand the suggestion, but I'm -1 to force it.
There should be a balance between code style and to allow developers to discuss if this convention makes sense to apply them or not.

For the Redux case I'm in favor to make it optional, to me, the important is not the properties declaration but the mapping between redux into the component which is where you typically see the connection.

I've leave developers + reviewers to discuss about it but case by case.

Note that the code is old (+3 years) and there are a lot of conventions used (classes, hooks, redux, non-redux), so, trying to update the guide won't help as this discussion will happen again.

So, I vote to discuss these style convention in the PR when it happens and see if within the context makes sense or not (i.e. don't break the consistency of a particular code just to force a convention), but not as a general rule.

lucasponce (approved):
I think this PR can be merged.
Perhaps we need to trigger another disccussion as the STYLE_GUIDE.adoc is very outdated and IMHO is not helping in the main objectives like:
- Defining more clarity about when and how to add tests in the UI.
- Indicate when to use Redux and when not.
- Extend in the same way to other components (there is a mix of styles).

As discussed, we are not going to refactor old code to adjust new style.

Also, discussions about conventions and code style can be open and subjective, so I think that author + reviewers should comment and agree on those in each individual PR.

But this convention could be useful, thanks @xeviknal.

PR #2225, opened 2021-08-26
Full Bookmarking for Graph Page

jshaughn:
Fixes kiali/kiali#3840

The graph page should now support full bookmarking via the URL.  The Toolbar options including Find/Hide, the Display options, and Replay.   Things intentionally not bookmarked (should they be?):
* graph tour
* graph-find help
* pan/zoom
* replay frame (it will reset to the start)


pbajjuri20 (approved):
LGTM!

PR #2229, opened 2021-09-01
Display Envoy tab for workloads running Istio Proxy without Sidecar

aljesusg:
Fix https://github.com/kiali/kiali/issues/4165

I was thinking about to add some check in this case to avoid error in envoy tab (check that there is a configDump)

But this is going to increase the consuming and maybe it's not necessary I don't think there are many cases

Screenshot from 2021-09-01 11-02-44

### DIFF
--- a/src/pages/WorkloadDetails/WorkloadDetailsPage.tsx
+++ b/src/pages/WorkloadDetails/WorkloadDetailsPage.tsx
@@ -163,7 +163,6 @@ class WorkloadDetails extends React.Component<WorkloadDetailsPageProps, Workload
         </Tab>
       );
     }
-
     if (this.state.workload && this.hasIstioSidecars(this.state.workload)) {
       const envoyTab = (
         <Tab title="Envoy" eventKey={10} key={'Envoy'}>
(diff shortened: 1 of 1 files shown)


nrfox (approved):
LGTM

nrfox:
> I was thinking about to add some check in this case to avoid error in envoy tab (check that there is a configDump)
>
> But this is going to increase the consuming and maybe it's not necessary I don't think there are many cases

The gateway pods seem to have a config dump at least and that was why this issue was opened to begin with. I'm not sure if there are other workloads that would be running the istio proxy standalone and not as a sidecar.

aljesusg:
> > I was thinking about to add some check in this case to avoid error in envoy tab (check that there is a configDump)
> > But this is going to increase the consuming and maybe it's not necessary I don't think there are many cases
> 
> The gateway pods seem to have a config dump at least and that was why this issue was opened to begin with. I'm not sure if there are other workloads that would be running the istio proxy standalone and not as a sidecar.

Mmm ok we can go ahead with this use case and we’ll see if there are other cases

jshaughn:
@mattmahoneyrh I just noticed you were set as a Reviewer, not as an Assignee.  @aljesusg , when ready for QE, add Matt as an assignee, and then he will set a QE person as the reviewer.  Also, if something sits for more than a couple of days, it has likely been missed, make sure you bring it up in scrum or dm QE to ask about it. I will assign Matt now...

pbajjuri20:
LGTM!

pbajjuri20 (approved):

aljesusg:
Thabks @pbajjuri20

PR #2230, opened 2021-09-02
if non-Maistra environment, use the pod label for the injection patch

jmazzitelli:
part of https://github.com/kiali/kiali/issues/4268

server PR: https://github.com/kiali/kiali/pull/4309

This introduces a new section to the StatusState object. Along with all the version information (already found in the ExternalServicesInfo[] section), this now has "IstioEnvironment" that will provide specific details about the known Istio environment. The only flag in there now is "isMaistra" but I envision other things will go in here in the future (other flags to indicate other Istio features that may or may not be available - federation and multi-cluster features are one thing I can see).


jmazzitelli:
selecting add/remove/disable auto-injection on a workload when in a non-Maistra environment will now update the label. In a Maistra environment, it will continue updating the annotation.

lucasponce:
Thanks @jmazzitelli I'll review it today, but after our talk I was thinking if it's worth to discuss the releases cadence.
I mean, Kiali 1.40 is target to Istio 1.11.
So, this version won't reach Maistra releases, as the latest is based on 1.9, probably when Maistra plan to move to 2.2, probably they will use a Istio 1.11 or newer, so they would need to use the label logic.

In short, perhaps, the UI logic won't be necessary after all, as it can rely always in the new label patch.

The PR looks mostly ok, I'll fetch locally to check in detail if there is something, but I mean that as the current version of Maistra won't ship this Kiali version, this condition would never happen (unless Maistra decides to not use the label mechanism in the future of 2.2).

jmazzitelli:
> probably when Maistra plan to move to 2.2, probably they will use a Istio 1.11 or newer, so they would need to use the label logic.

That is not necessarily true. Maistra has their own implementation for how they determine what to auto-inject (remember - Maistra doesn't even support the namespace label and that has been in Istio since the beginning - Maistra purposefully doesn't support it). So it is possible even when Maistra upgrades to the 1.11 they also won't support the pod label just like they chose not to look at the namespace label (i.e. Maistra team does not necessarily follow the lead of upstream Istio when it comes to when to inject sidecars)

jmazzitelli:
The above discussion made me think of this - so I created a new enhancement request: https://github.com/kiali/kiali/issues/4327

src/types/StatusState.ts
```diff
@@ -8,14 +8,19 @@ export enum StatusKey {
...
 
-export interface Component {
+export interface ExternalServiceInfo {
   name: string;
   version?: string;
   url?: string;
 }
 
+export interface IstioEnvironment {
+  isMaistra: boolean;
```
jshaughn, line 18: It's fine but I sort of wish this were an Enum like "distribution" or something like that.

jshaughn (approved):
This looks good to me, I like the included refactor to remove the confusing ServerStatus and consilidate on StatusState (Although, in the future I wouldn't mind seeing StatusState be renamed to something less generic, maybe even ServerStatus :)).

One thing I wonder about is whether this will be reflected well in the About box.  Ity would seem that for Maistra (or OSSM) it would be nice to show the Maistra version and maybe also the underlying Istio version.

src/types/StatusState.ts
jmazzitelli: we could add something like that. the enum values would need to be propagated to the UI as well. I can look into seeing how to do this.

jshaughn: Not needed for this PR, just something to think about going forward.

src/components/Nav/Masthead/HelpDropdown.tsx
```diff
@@ -6,12 +6,12 @@ import { Dropdown, DropdownToggle, DropdownItem } from '@patternfly/react-core';
...
 import { connect } from 'react-redux';
 import { isUpstream } from '../../UpstreamDetector/UpstreamDetector';
-import { Status, Component, StatusKey } from '../../../types/StatusState';
+import { Status, ExternalServiceInfo, StatusKey } from '../../../types/StatusState';
 import { config } from '../../../config';
 
 type HelpDropdownProps = {
   status: Status;
-  components: Component[];
+  components: ExternalServiceInfo[];
```
lucasponce, line 14: Minor, but just for consistency:
image

if the renaming Component -> ExternalServiceInfo is to add clarity (+1 on that), I'd also propagate to the properties that is used if it's a direct reference as it shows here.

(Non a blocker from my side, it can optional to do it in the future, code is clear to me).

src/reducers/HelpDropdownState.ts
```diff
@@ -5,18 +5,20 @@ import { HelpDropdownActions } from '../actions/HelpDropdownActions';
...
 
-// This Reducer allows changes to the 'graphDataState' portion of Redux Store
 const HelpDropdownState = (state: StatusState = INITIAL_STATUS_STATE, action: KialiAppAction): StatusState => {
   switch (action.type) {
     case getType(HelpDropdownActions.statusRefresh):
       return {
         ...INITIAL_STATUS_STATE,
         status: action.payload.status,
-        components: action.payload.components,
+        externalServices: action.payload.components,
```
lucasponce, line 21: Minor, perhaps worth to adjust the "payload" variable as well (just for clarity).
Action/Reducers are using same variable in the payload, but renaming it.

Again, non blocker, but as actions/reducers in Redux are not trival to read, I think those would help others as well.

image

lucasponce (approved):
Reviewed and tested locally.
I put a comment about a pending rename in a variable, but it's non blocker and optional.

src/reducers/HelpDropdownState.ts
jmazzitelli: I thought about changing the property names/variable names also to match, but I thought the amount of code changes/churn would be too much. But I understand this would make it more clear if the names all matched. Let me try to do this and see if I can get the names all consistent.

jmazzitelli:
@lucasponce variable rename in commit https://github.com/kiali/kiali-ui/pull/2230/commits/4ff6041ef3dcd908597fe44ad479b9a168a68518

lucasponce:
Thanks @jmazzitelli, after consolidate previous work, it seems there is now a conflict
image

We can merge it as soon as you rebase it.

jmazzitelli:
@lucasponce conflicts resolved

PR #2236, opened 2021-09-08
Add ability to set proxy log level for workload pod

nrfox:
Adds a way to set a workload pod's proxy log level from the kebab dropdown.

Any error that occurs when setting the log level will result in an error notification/alert.

Screenshot from 2021-09-08 14-15-33

On error:
Screenshot from 2021-09-07 14-13-30

Relates to: kiali/kiali#1525
Server side PR: kiali/kiali#4312

### DIFF
--- a/src/config/Config.ts
+++ b/src/config/Config.ts
@@ -155,6 +155,7 @@ const conf = {
       pod: (namespace: string, pod: string) => `api/namespaces/${namespace}/pods/${pod}`,
       podLogs: (namespace: string, pod: string) => `api/namespaces/${namespace}/pods/${pod}/logs`,
       podEnvoyProxy: (namespace: string, pod: string) => `api/namespaces/${namespace}/pods/${pod}/config_dump`,
+      podEnvoyProxyLogging: (namespace: string, pod: string) => `api/namespaces/${namespace}/pods/${pod}/logging`,
       podEnvoyProxyResourceEntries: (namespace: string, pod: string, resource: string) =>
         `api/namespaces/${namespace}/pods/${pod}/config_dump/${resource}`,
       serverConfig: `api/config`,
--- a/src/pages/WorkloadDetails/WorkloadPodLogs.tsx
+++ b/src/pages/WorkloadDetails/WorkloadPodLogs.tsx
@@ -6,6 +6,7 @@ import {
   CardBody,
   Grid,
   GridItem,
+  Popover,
   TextInput,
   Toolbar,
   ToolbarGroup,
@@ -21,7 +22,7 @@ import {
   DropdownSeparator
 } from '@patternfly/react-core';
 import { style } from 'typestyle';
-import { addError } from 'utils/AlertUtils';
+import { addError, addSuccess } from 'utils/AlertUtils';
 import { Pod, LogEntry, AccessLog, PodLogs } from '../../types/IstioObjects';
 import { getPodLogs, getWorkloadSpans, setPodEnvoyProxyLogLevel } from '../../services/Api';
 import { PromisesRegistry } from '../../utils/CancelablePromises';
@@ -493,6 +494,26 @@ export class WorkloadPodLogs extends React.Component<WorkloadPodLogsProps, Workl
         </DropdownItem>
       );
     });
+    const dropdownGroupLabel = (
+      // nowrap is needed for the info icon to appear on same line as the label text
+      <Popover
+        headerContent="Setting Proxy Logger Level"
+        bodyContent={
+          <span>
+            This action configures the <b>application</b> log level for the proxy but will not affect the <b>access</b>{' '}
+            logs. Setting the log level to 'off' will <b>not</b> disable access logging. To hide all proxy logging from
+            the logs view, un-check the 'istio-proxy' container.
+          </span>
+        }
+      >
+        <div style={{ whiteSpace: 'nowrap' }}>
+          Set Proxy Log Level
+          <Button variant={ButtonVariant.link} style={{ paddingLeft: '6px' }}>
+            <KialiIcon.Info className={defaultIconStyle} />
+          </Button>
+        </div>
+      </Popover>
+    );
 
     const kebabActions = [
       <DropdownItem key="toggleToolbar" onClick={this.toggleToolbar}>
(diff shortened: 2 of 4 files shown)


jshaughn (changes requested):
@nrfox This looks good and it's pretty cool.  A couple of comments:
- When the pod does not have (or is not itself) a proxy should we still show these options in the kebab?  Maybe not, or maybe they should be disabled (disabling is more our convention for non-applicable actions).
- I'm not sure we have a convention but it may look better to indent the options under the section title (ignore if you like, not a blocker)
- I successfully set Trace/Debug/Info but when setting Off/Warning/Error/Critical I still saw access logs, is that expected?  

One more note, and some for a follow-up issue, it looks like we are identifying all proxy logEntries as access logs.  Only access logs should have the info icon to see the parsed view.

nrfox:
Thanks for taking a look!

> When the pod does not have (or is not itself) a proxy should we still show these options in the kebab? Maybe not, or maybe they should be disabled (disabling is more our convention for non-applicable actions).

Yes! I should've planned for that. Good catch. I'm happy with either hiding or disabling. I like the idea of not taking up screen real estate with options the user can't interact with but maybe that would hurt discoverability of the feature if users are looking for it and can't find it. How about disabling with a tooltip explaining why?

>  I'm not sure we have a convention but it may look better to indent the options under the section title (ignore if you like, not a blocker)

:+1: 

>  I successfully set Trace/Debug/Info but when setting Off/Warning/Error/Critical I still saw access logs, is that expected?

I saw that same thing and assumed it was expected but I'm not 100% sure. When you get the list of loggers through the admin interface `access` is not included which might indicate that you can't configure the access logs through the admin interface. I could see how that would be confusing for users who set the log level to "off" but still see istio proxy logs. Perhaps we can add a tooltip on hover of the "off" option that explains that and to check the "istio-proxy" box if they don't want to see any istio-proxy logs?

```
istio-proxy@ratings-v2-59658c8774-b2db5:/$ curl -XPOST localhost:15000/logging                                                                                                                
active loggers:                                                                                                                                                                               
  admin: warning
  aws: warning
  assert: warning
  backtrace: warning
  cache_filter: warning
  ...
```

If you attempt to set the logging to an invalid level you'll get an error back so the admin interface recognizes those options as valid.

jshaughn:
> ... but maybe that would hurt discoverability of the feature if users are looking for it and can't find it. How about disabling with a tooltip explaining why?

Instead of a tooltip I think I'd go for an info icon, like what you see in the graph display menu.  I don't want it hover-activated but instead a click.  And the text could be longer than a typical tooltip.  For this scenario we could leave the header, now iwth the always-on info icon, and not show the options (to simplify and save the real-estate)

> Perhaps we can add a tooltip on hover of the "off" option that explains that and to check the "istio-proxy" box if they don't want to see any istio-proxy logs?

Again yes, but as part of the info tooltip).  Checking the istio-proxy box will disable the fetch completely so that is a good way to do it.  Thinking about this, I do believe the access logging may be handled separately from these loggers.  Whether it can be controlled per-pod is something maybe to investigate.  I think at the moment it is controlled mesh-wide as an istio-setting:  https://istio.io/latest/docs/tasks/observability/logs/access-log/#enable-envoy-s-access-logging

nrfox:
> Whether it can be controlled per-pod is something maybe to investigate. I think at the moment it is controlled mesh-wide as an istio-setting: https://istio.io/latest/docs/tasks/observability/logs/access-log/#enable-envoy-s-access-logging

Envoy seems to distinguish between "[application logging](https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/application_logging)" and "[access logs](https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log/access_log)". I haven't seen anything in the docs that says the admin interface logging endpoint is only for "application logging" and excludes access logs but it seems that way.

jmazzitelli:
I'm OK disabling it (or even removing it) for kubernetes , but we should leave it as-is for OpenShift because it is used when deploying on OpenShift - it creates a Route and that normally always works with our default template.

See: https://github.com/kiali/kiali-operator/blob/master/roles/default/kiali-deploy/templates/openshift/route.yaml and https://github.com/kiali/kiali-operator/blob/master/roles/default/kiali-deploy/tasks/openshift/os-main.yml#L40-L51

lucasponce:
> I'm not sure we have a convention but it may look better to indent the options under the section title (ignore if you like, not a blocker)

There is. Discussed with @lwrigh when talking about menus, so no idendation when using "title" in dropdown, please:
image
image

Otherwise we'll start again to create micro-styles on each page/tab.

lucasponce:
Another suggestion @nrfox, if the call is positive, I'd recommend to send an "ok" message to the alert center, to be in sync with the pattern used when backend is modified:
image

nrfox:
@jshaughn how's this look?

Screenshot from 2021-09-13 14-15-11

jshaughn:
@nrfox   OK, as per @lucasponce / @lwrigh , no indentation.  Also, I'm +1 for an OK message to the message center when setting.  I like the info icon but I think I made a mistake earlier when I said no tooltip.  I meant no tooltip on the "Set proxy log level" title itself.  You shouldn't have to click on the info icon, it should be like the graph's Display menu, and just show a tooltip when you hover over the info icon.  Sorry for any confusion.  As for the tooltip, use the same styling you see in the Display menu tooltips. (I think what you have above looks more like the graph tour).  So, you can drop the title.  A couple of text nitpicks:
- maybe say "proxy logger level" as opposed to "application log level", I don't want the user to think this affects the app logging (non-proxy containers) in some way.
- maybe say "proxy container" instead of "istio-proxy" container.  Almost all, but not all proxies are named 'istio-proxy".

nrfox:
@jshaughn take-two. Used a `<Tooltip>` instead of a `<Popover>` and I think it's closer to what you described.

Screenshot from 2021-09-13 16-21-37

jshaughn:
@nrfox Yes, that's what I had in mind, sorry for messing up the initial description.  Two nits and a question:
- style like the Display options and left-align the tooltip text
- "logs view including" --> logs view, including"
- would it make sense to add a line like, "This option is disabled for pods with no proxy container."?

nrfox:
> * would it make sense to add a line like, "This option is disabled for pods with no proxy container."?

Yep that would be good!

nrfox:
One thing I noticed that I can't explain is that the tooltip moves when a refresh happens:

tooltip_moving

It's subtle but it doesn't appear to happen on the other tooltips that I've tested this on. Is that a blocker?

jshaughn:
>  Is that a blocker?

No, just an annoyance. let's get this nice feature in!

nrfox:
Alright pushed the changes:

Dropdown with no proxy container 
Screenshot from 2021-09-14 12-01-16

Updated tooltip:
Screenshot from 2021-09-14 12-01-03

jshaughn (approved):
LG!

pbajjuri20 (approved):
LGTM!

PR #2241, opened 2021-09-15
Generate link to istio config for workload entries

nrfox:
In the graph node summary panel for workloads with workload entries, the link now links to the istio config page for the workload entry instead of the workloads page.

Fixes kiali/kiali#4219
Back-end changes: kiali/kiali#4372

### DIFF
--- a/src/pages/Graph/SummaryPanelNode.tsx
+++ b/src/pages/Graph/SummaryPanelNode.tsx
@@ -35,7 +35,7 @@ type ReduxProps = {
   jaegerState: JaegerState;
 };
 
-type SummaryPanelNodeProps = ReduxProps & SummaryPanelPropType;
+export type SummaryPanelNodeProps = ReduxProps & SummaryPanelPropType;
 
 const expandableSectionStyle = style({
   fontSize: 'var(--graph-side-panel--font-size)',
@@ -101,7 +101,6 @@ export class SummaryPanelNode extends React.Component<SummaryPanelNodeProps, Sum
     const actions =
       options.length > 0 ? [<DropdownGroup label="Show" className="kiali-group-menu" children={options} />] : undefined;
 
-    // TODO: Summary Panel node tests
     let workloadLinks;
     if (nodeData.hasWorkloadEntry) {
       nodeData.hasWorkloadEntry.forEach(we => {
@@ -149,7 +148,6 @@ export class SummaryPanelNode extends React.Component<SummaryPanelNodeProps, Sum
             {shouldRenderService && <div>{renderBadgedLink(nodeData, NodeType.SERVICE)}</div>}
             {shouldRenderApp && <div>{renderBadgedLink(nodeData, NodeType.APP)}</div>}
             {shouldRenderWorkload && workloadLinks}
-            {/* Add case for WE entries and rendering each workload entry */}
           </div>
         </div>
         {shouldRenderTraces ? this.renderWithTabs(nodeData) : this.renderTrafficOnly()}
--- a/src/pages/Graph/SummaryPanelNode.tsx
+++ b/src/pages/Graph/SummaryPanelNode.tsx
@@ -7,7 +7,15 @@ import { summaryHeader, summaryPanel, summaryBodyTabs, summaryFont } from './Sum
 import { decoratedNodeData } from '../../components/CytoscapeGraph/CytoscapeGraphUtils';
 import { KialiIcon } from 'config/KialiIcon';
 import { getOptions, clickHandler } from 'components/CytoscapeGraph/ContextMenu/NodeContextMenu';
-import { Dropdown, DropdownGroup, DropdownItem, DropdownPosition, Expandable, KebabToggle, Tab } from '@patternfly/react-core';
+import {
+  Dropdown,
+  DropdownGroup,
+  DropdownItem,
+  DropdownPosition,
+  Expandable,
+  KebabToggle,
+  Tab
+} from '@patternfly/react-core';
 import { KialiAppState } from 'store/Store';
 import { SummaryPanelNodeTraffic } from './SummaryPanelNodeTraffic';
 import SummaryPanelNodeTraces from './SummaryPanelNodeTraces';
@@ -93,6 +101,29 @@ export class SummaryPanelNode extends React.Component<SummaryPanelNodeProps, Sum
     const actions =
       options.length > 0 ? [<DropdownGroup label="Show" className="kiali-group-menu" children={options} />] : undefined;
 
+    // TODO: Summary Panel node tests
+    let workloadLinks;
+    if (nodeData.hasWorkloadEntry) {
+      nodeData.hasWorkloadEntry.forEach(we => {
+        if (!workloadLinks) {
+          workloadLinks = [];
+        }
+        const generateWorkloadEntryLink = () => ({
+          link: `/namespaces/${encodeURIComponent(nodeData.namespace)}/istio/workloadentries/${encodeURIComponent(
+            we.name
+          )}`,
+          displayName: we.name,
+          key: `${nodeData.namespace}.wle.${we.name}`
+        });
+
+        workloadLinks.push(
+          <div>{renderBadgedLink(nodeData, NodeType.WORKLOAD, undefined, generateWorkloadEntryLink)}</div>
+        );
+      });
+    } else {
+      workloadLinks = <div>{renderBadgedLink(nodeData, NodeType.WORKLOAD)}</div>;
+    }
+
     return (
       <div ref={this.mainDivRef} className={`panel panel-default ${summaryPanel}`}>
         <div className="panel-heading" style={summaryHeader}>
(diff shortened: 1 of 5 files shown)


nrfox:
Just realized I made some bad assumptions about how workload entries are named... converting to draft until that gets fixed.

nrfox:
> Just realized I made some bad assumptions about how workload entries are named... converting to draft until that gets fixed.

Fixed. Now the workload entries are added to the node metadata by the backend.

nrfox:
Screenshot from 2021-09-20 13-07-39

Should there be a ceiling on how many of the workload entry links get added to the side panel? If there are 100 workload entries there will be 100 links and that probably won't look great. Perhaps if there are more than 5 or 10 then a "see more" link can link to the istio config summary page filtering on workload entries?

jshaughn:
> Should there be a ceiling on how many of the workload entry links

Absolutely.  Look at what @israel-hdez did with the Hostnames stuff in summarynodepanel, I think it can look the same.

nrfox:
With badge:
Screenshot from 2021-09-21 14-48-21

Without badge:
Screenshot from 2021-09-21 14-53-49

I'm not sure I like the badge in the expandable section but maybe it's worth keeping it so that the links are consistent with the other links in the node panel?

jshaughn:
What about if the badge were at the top level and not repeated in the dropdown.

nrfox:
> What about if the badge were at the top level and not repeated in the dropdown.

I think better?
Screenshot from 2021-09-21 19-40-01

jshaughn:
@nrfox That looks better, at least to me.  It's less busy.  When the dust settles we need to ensure that the Hosts dropdown has the same styling, to be consistent.

cc @israel-hdez in case he has any comments, Edgar was the one that recently added the Hosts stuff.

src/pages/Graph/SummaryPanelNode.tsx
```diff
@@ -117,19 +132,53 @@ export class SummaryPanelNode extends React.Component<SummaryPanelNodeProps, Sum
...
+    return (
+      <>
+        {getBadge(nodeData, NodeType.WORKLOAD)}
+        <Expandable
+          toggleText={
+            nodeData.hasWorkloadEntry.length === 1
+              ? '1 workload entry'
+              : `${nodeData.hasWorkloadEntry.length} workload entries`
+          }
+          className={classes(expandableSectionStyle, style({ display: 'inline' }))}
```
israel-hdez, line 169: (Nitpick: performance?)

I wonder if it is better to declare that `style({ display: 'inline' })` in a `const`?

I'm not sure if `typestyle` is smart to reuse styles. If not, repeated invocations to `style` may be injecting several classNames to the document with the same css definition. 

I didn't check, so I cannot tell.

israel-hdez (commented):
I don't have important comments, so I'm approving.

> we need to ensure that the Hosts dropdown has the same styling, to be consistent.

Probably, we need to pass this to UX. I see we are using two different kind of badges: related to Istio Config (Has VS, Is Gateway, etc) and related to the kind of node (**WE**, **A**pp, **S**ervice, **W**orkload, etc). So, I'm not sure if we can unify so that both the Hosts and the WE use the same style.

src/pages/Graph/SummaryPanelNode.tsx
nrfox: > I wonder if it is better to declare that style({ display: 'inline' }) in a const?

:+1: 

> I'm not sure if typestyle is smart to reuse styles. If not, repeated invocations to style may be injecting several classNames to the document with the same css definition.

It looks like the [classes method](https://typestyle.github.io/#/core/classes) just adds the two styles as separate classes relying on css composition: e.g. `<p class="class1 class2">`. In this case the only thing defined in `class2` would be `{display: 'inline'}`. I'm happy to update if that's not desirable.

nrfox:
> Probably, we need to pass this to UX. I see we are using two different kind of badges: related to Istio Config (Has VS, Is Gateway, etc) and related to the kind of node (**WE**, **A**pp, **S**ervice, **W**orkload, etc). So, I'm not sure if we can unify so that both the Hosts and the WE use the same style.

The difference between when to use badge vs. icon seems to be whether it is associated with a link or not?

src/pages/Graph/SummaryPanelNode.tsx
israel-hdez: > In this case the only thing defined in `class2` would be `{display: 'inline'}`. I'm happy to update if that's not desirable.

That's perfectly fine.

I understand you would see `<p class="class1 class2">` in the first call. However, since you are calling repeatedly the `style` function, I was wondering if you would see `<p class="class1 class3">`, then `<p class="class1 class4">`, ..., `<p class="class1 classN">` in subsequent calls/renders. This is what I think wouldn't be desirable. But if you always see `<p class="class1 class2">`, that's fine.

israel-hdez:
> The difference between when to use badge vs. icon seems to be whether it is associated with a link or not?

Apparently, yes... Or the way I see it: whether there is a Kiali page that the user can jump to see more details, or not.

src/pages/Graph/SummaryPanelNode.tsx
nrfox: Ah I gotcha, I'll pull that out so it's not inlined.

nrfox:
Adjusted margin a bit so that the dropdown the links are aligned with one another:

Screenshot from 2021-09-28 15-41-58

israel-hdez (approved):

pbajjuri20:
I have applied this Workload_entry.yaml

```
apiVersion: networking.istio.io/v1beta1
kind: WorkloadEntry
metadata:
  name: reviews-v1
  namespace: bookinfo
spec:
  serviceAccount: bookinfo-ratings
  address: 76.187.71.26
  labels:
    app: reviews
    version: v1
  ports:
    http: 9080
```
~               

workload badge is there perfectly but, I see some lock symbol in the Graph, Is it expected? 



we_pr

pbajjuri20:
WE_PR

nrfox:
> I have applied this Workload_entry.yaml
> 
> ```
> apiVersion: networking.istio.io/v1beta1
> kind: WorkloadEntry
> metadata:
>   name: reviews-v1
>   namespace: bookinfo
> spec:
>   serviceAccount: bookinfo-ratings
>   address: 76.187.71.26
>   labels:
>     app: reviews
>     version: v1
>   ports:
>     http: 9080
> ```
> 
> ~
> 
> workload badge is there perfectly but, I see some lock symbol in the Graph, Is it expected?
> 
> we_pr

Can you try applying the workload entry off the `master` branch? I wonder if it's just related to the workload entry itself and not the PR changes?

pbajjuri20:
> > I have applied this Workload_entry.yaml
> > ```
> > apiVersion: networking.istio.io/v1beta1
> > kind: WorkloadEntry
> > metadata:
> >   name: reviews-v1
> >   namespace: bookinfo
> > spec:
> >   serviceAccount: bookinfo-ratings
> >   address: 76.187.71.26
> >   labels:
> >     app: reviews
> >     version: v1
> >   ports:
> >     http: 9080
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ~
> > workload badge is there perfectly but, I see some lock symbol in the Graph, Is it expected?
> > we_pr
> 
> Can you try applying  the workload entry off the `master` branch? I wonder if it's just related to the workload entry itself and not the PR changes?

I have applied for the Workload Entry against the master, It shows the lock symbol
we_latest

nrfox:
> > > I have applied this Workload_entry.yaml
> > > ```
> > > apiVersion: networking.istio.io/v1beta1
> > > kind: WorkloadEntry
> > > metadata:
> > >   name: reviews-v1
> > >   namespace: bookinfo
> > > spec:
> > >   serviceAccount: bookinfo-ratings
> > >   address: 76.187.71.26
> > >   labels:
> > >     app: reviews
> > >     version: v1
> > >   ports:
> > >     http: 9080
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > ~
> > > workload badge is there perfectly but, I see some lock symbol in the Graph, Is it expected?
> > > we_pr
> > 
> > 
> > Can you try applying  the workload entry off the `master` branch? I wonder if it's just related to the workload entry itself and not the PR changes?
> 
> I have applied for the Workload Entry against the master, It shows the lock symbol we_latest

And if you remove the Workload Entry that lock symbol goes away after the metrics time range set (1m)? I think it's ok here since the workload entry is only there to test that the host links appear in the node summary panel. For testing purposes, it shouldn't matter that the specific workload entry applied does not work, only that the links appear.

pbajjuri20:
> > > > I have applied this Workload_entry.yaml
> > > > ```
> > > > apiVersion: networking.istio.io/v1beta1
> > > > kind: WorkloadEntry
> > > > metadata:
> > > >   name: reviews-v1
> > > >   namespace: bookinfo
> > > > spec:
> > > >   serviceAccount: bookinfo-ratings
> > > >   address: 76.187.71.26
> > > >   labels:
> > > >     app: reviews
> > > >     version: v1
> > > >   ports:
> > > >     http: 9080
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > ~
> > > > workload badge is there perfectly but, I see some lock symbol in the Graph, Is it expected?
> > > > we_pr
> > > 
> > > 
> > > Can you try applying  the workload entry off the `master` branch? I wonder if it's just related to the workload entry itself and not the PR changes?
> > 
> > 
> > I have applied for the Workload Entry against the master, It shows the lock symbol we_latest
> 
> And if you remove the Workload Entry that lock symbol goes away after the metrics time range set (1m)? I think it's ok here since the workload entry is only there to test that the host links appear in the node summary panel. For testing purposes, it shouldn't matter that the specific workload entry applied does not work, only that the links appear.

After deleting the Workload Entry. It is taking some time to disappear.

pbajjuri20:
@nrfox: As, I discussed with Nick, about this PR. Workload Entry in the Graph, It looks good.

pbajjuri20 (approved):

PR #2244, opened 2021-09-23
Graph shorcuts in Graph Tour

leandroberetta:
This PR adds a new stop in the graph tour to show the shortcuts available in the graph.

This is a work in progress draft PR to get feedback:

shortcuts

Resolves [#4133](https://github.com/kiali/kiali/issues/4133)

### DIFF
--- a/src/components/Tour/TourStop.tsx
+++ b/src/components/Tour/TourStop.tsx
@@ -13,7 +13,8 @@ import { PFColors } from 'components/Pf/PfColors';
 
 export interface TourStopInfo {
   name: string; // displayed in the tour stop header.
-  description: string; // displayed as the tour stop body
+  description?: string; // displayed as the tour stop body
+  htmlDescription?: JSX.Element;
   position?: PopoverPosition;
   offset?: string; // tippy prop: 'xOffset, yOffset'
   isValid?: boolean; // internal use, leave unset
@@ -185,7 +186,7 @@ class TourStop extends React.PureComponent<TourStopProps> {
                   <span>{info.name}</span>
                 </div>
               }
-              bodyContent={info.description}
+              bodyContent={info.description ? info.description : info.htmlDescription}
               footerContent={
                 <div>
                   {this.backButton()}
--- a/src/pages/Graph/GraphHelpTour.tsx
+++ b/src/pages/Graph/GraphHelpTour.tsx
@@ -1,21 +1,7 @@
 import * as React from 'react';
-import { Chip, Grid, GridItem, PopoverPosition } from '@patternfly/react-core';
+import { PopoverPosition } from '@patternfly/react-core';
 import { TourStopInfo, TourInfo } from 'components/Tour/TourStop';
-
-const shortcuts = (): JSX.Element => {
-  return (
-    <Grid gutter={'md'}>
-      <GridItem span={5}>
-        <Chip isReadOnly>Mouse Wheel</Chip>
-      </GridItem>
-      <GridItem span={7}>Zoom in and out</GridItem>
-      <GridItem span={5}>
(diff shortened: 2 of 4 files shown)


leandroberetta:
@lucasponce @jshaughn based on the comments on the issue, I tried to mix some ideas and comments resulting in the screenshot above.

There are shortcuts missing but I created this to start discussing about the design.

jshaughn:
@leandroberetta I like the idea of making "Shortcuts" the first tour stop.  It leverages the fact that we already have this feature in place.  And if the user just wants to see Shortcuts he can exit the tour immediately.  We may want to rename "Graph tour" to "Shortcuts and tips", or something like that, to ensure users know they can look there.  I'm not sure "Graph tour" has ever been particularly enticing.  Also, the word "Graph" is pretty useless, it's already implied.

leandroberetta:
> @leandroberetta I like the idea of making "Shortcuts" the first tour stop. It leverages the fact that we already have this feature in place. And if the user just wants to see Shortcuts he can exit the tour immediately. We may want to rename "Graph tour" to "Shortcuts and tips", or something like that, to ensure users know they can look there. I'm not sure "Graph tour" has ever been particularly enticing. Also, the word "Graph" is pretty useless, it's already implied.

@jshaughn thanks for the comments, I agree, I even thought in removing the description, just leaving the help icon... following the same pattern in the factory reset button

lucasponce:
Reusing the tour for the shortcuts seems ok.

Perhaps a little bit out of context as other users may prefer something direct:

https://github.com/kiali/kiali/issues/4133#issuecomment-899398084

But I agree that adding a tour + another shortcuts info entry may result in too overloaded, so +1 to put it in the tour.

leandroberetta:
The following is a test without any description but a different tooltip:

shortcuts2

leandroberetta:
@lucasponce @jshaughn I'm converting this draft into a proper PR for review:

shortcuts4

jshaughn (approved):
Looks great, and now we have html option for tour stop.

lucasponce (changes requested):
The tour is great, I have a couple of minor comments about the style, as it's hard to test it and we need to rely manually on it.

image

I'd not use the Grid for render the items, it's not flexible for minor panels like this and it creates situations where the margin/padding from the items looks greater. I'd take same marging that you can see in the toolbox (i.e. same margin used between Namespace and Display combobox). I'd use a simple render of divs that you can set your own style for margin.

Another style comment, I'd use font 14px for the description instead of 16px, otherwise it looks different with the next tour:

image

But +1, it looks very nice to see this in the first step of the tour, I've also found it very useful.

leandroberetta:
@lucasponce thanks for your recommendations, I tried to used a combination of table + divs resulting in the following:

shortcuts5

As you can see, I'm using the margin recommended but when text wrapped, the margin between the rows, increased, as the outer canvas is fixed always to a certain width, is this ok for you?. 

Also the size of the text is equal to the following tour steps.

leandroberetta:
New version with @lucasponce suggestions:

shortcuts6

lucasponce:
Thanks @leandroberetta I put an ultra minor change in a https://github.com/leandroberetta/kiali-ui/pull/3 just to align a little bit.

Let me know what do you think and we can consolidate it.

leandroberetta:
> Thanks @leandroberetta I put an ultra minor change in a [leandroberetta#3](https://github.com/leandroberetta/kiali-ui/pull/3) just to align a little bit.
> 
> Let me know what do you think and we can consolidate it.

@lucasponce thanks for the fix, it's merged now.

lucasponce (approved):

skondkar:
Polarion Testcase ->> SWSQE-1995

PR #2247, opened 2021-09-29
Add sidecars when creating traffic policies

leandroberetta:
WIP.


src/components/IstioWizards/WizardActions.ts
```diff
@@ -1137,6 +1137,87 @@ export const buildAuthorizationPolicy = (
   return ap;
 };
 
+export const buildGraphSidecars = (namespace: string, graph: GraphDefinition): Sidecar[] => {
+  const denyAll: Sidecar = {
```
lucasponce, line 1141: I don't think the "denyAll" applies to the sidecars, as it's not like the AuthorizationPolicy, so I guess for Sidecars it can rely on the workloads only.

src/components/IstioWizards/WizardActions.ts
```diff
@@ -1137,6 +1137,87 @@ export const buildAuthorizationPolicy = (
...
+        node.data.workload &&
+        node.data.app &&
+        node.data.version
+      ) {
+        const sc: Sidecar = {
+          metadata: {
+            name: node.data.workload,
+            namespace: namespace,
+            labels: {
+              [KIALI_WIZARD_LABEL]: 'AuthorizationPolicy'
```
lucasponce, line 1175: 'AuthorizationPolicy' -> 'Sidecar'

src/components/IstioWizards/WizardActions.ts
```diff
@@ -1137,6 +1137,87 @@ export const buildAuthorizationPolicy = (
...
+            const edge = graph.elements.edges[j];
+
+            if (node.data.id === edge.data.source) {
+              for (let z = 0; z < graph.elements.nodes.length; z++) {
+                const targetNode = graph.elements.nodes[z];
+
+                if (targetNode.data.id === edge.data.target) {
+                  targetNode.data.destServices?.forEach((ds: DestService) => {
+                    if (sc.spec.egress && ds.namespace !== 'unknown') {
+                      sc.spec.egress[0].hosts.push(`${ds.namespace}/${ds.name}.${ds.namespace}.svc.cluster.local`);
```
lucasponce, line 1204: I'd try to fetch the constants like 'svc.cluster.local' from the config (there is a global 'serverConfig'):

image

Then if those values change we are safe.

lucasponce (commented):
Thanks @leandroberetta I'm happy with the feature.
It works as expected, I put a couple of comments to address but this is a nice addition.

src/components/IstioWizards/WizardActions.ts
leandroberetta: I guess the name is wrong, that is true, but what I wanted to do is to create a general rule that will applies for future workloads, to restrict the egress traffic just to the own namespace.

But, thinking a little bit more, maybe it's not correct to take this assumption, when new workloads are created, user should update the traffic policies.

I'm gonna remove it.

jshaughn: > I'm gonna remove it.

+1, we should not make that assumption but rather only reflect the state at the time this action is performed.

leandroberetta:
All suggestions were addressed.

lucasponce (approved):

prachiyadav:
Test case >> SWSQE- 1966

PR #2248, opened 2021-09-30
Generate models and API methods from Kiali openapi definition

nrfox:
This is a POC that uses the [openapi generator](https://openapi-generator.tech/) and the [Kiali openapi spec](https://github.com/kiali/kiali/blob/master/swagger.json) to autogenerate models and methods to interact with the API.

Some potential benefits are:
- Time saved manually defining api models/methods
- Keeping API/front-end in sync
- Autogenerated docs and clients for kiali API

To generate the models from this branch:
```
yarn && npx @openapitools/openapi-generator-cli generate -g typescript-axios -i ../kiali/swagger.json -o src/generated/ --additional-properties=withSeparateModelsAndApi=true --model-package models --api-package api
```
This assumes that your kiali repo exists locally at `../kiali`. Adjust the command accordingly.

It mostly just works but there are a few things that need to be changed in order to get tests passing. In addition the linting that runs as part of pre-commit should be updated to exclude the generated models. [The Namespace model](src/types/Namespace.ts) is an example where the generated model can be used in place of the existing handrolled model without any changes. An interesting thing to note is that many models within the API today have "required" fields that are not actually marked as required in the API definition. That will be the main source of changes required for the models. The generated API methods will probably be more work to switch to.


lucasponce:
It's nice @nrfox.
Once that https://github.com/kiali/kiali/pull/4389 is progressed and there are strong types, this autogenerator would definitely help.

Today there are some dyncamic types in the backend that are not fully synced in the frontend (no reason but a technical debt) and there are added in the frontend when we needed, but once that strong type is done I don't see a reason to not rely in an autogenerated tool for the APIs.

lucasponce:
It would be nice to continue due experiments to see if the backend is fully typed and documented.
I have the fear that it's not yet, probably swagger is not 100% accurate.

But we could use this as a tester, with the goal that anytime that there is a change in the backend, the frontend types could be automatically updated.

I know this is easy to say, but it's not obvious, as there are multiple aspects to address in both sides (I think we keep having some dynamic typing and probably in the frontend there are some bad unions of types that make impact this).

Not in a rush, it can be something we can experiment but just leaving some comment here for the future.

nrfox:
> It would be nice to continue due experiments to see if the backend is fully typed and documented.
I have the fear that it's not yet, probably swagger is not 100% accurate.

Agreed this change would be a good way to uncover those differences. I'll try to update this when I get a chance.

lucasponce:
@nrfox I'm reviewing the pending PRs in the list. I know this was an experimental Draft, I suggest closing it and we can revisit it later when the APIs are better consolidated.

nrfox:
@lucasponce sure we can close this. I believe there's some circular references in the kiali swagger spec that are causing issues when generating models from it. That'll probably need to be addressed upstream in the generator tool before this can move forward.

PR #2251, opened 2021-10-06
WIP Update backend VS/DR types

lucasponce:
This PR will update the backend types refactored from https://github.com/kiali/kiali/pull/4389.

As discussed in the backend PR, the goal is to update the types with minimal changes, there are other PR like https://github.com/kiali/kiali-ui/pull/2248 that is exploring the possibility to update automatically these types in the future.


xeviknal:
Early comment in my review:
In the istio config editor, the `kind` and `apiVersion` fields has disapeared:
Screenshot of Kiali Console (20)
Screenshot of Kiali Console (17)

Whilst in the main branch is like this:
Screenshot of Kiali Console (19)

This feels like a regression.

xeviknal:
Another catch:

Broken istio config details page when the status field is enabled in istio and the object has one message:

Screenshot of Kiali Console (21)

Kiali backend returns the message.level as an string (in this case `ERROR`):

wrong-istio-message-level

While the original and expected by the frontend value is a number:

right-istio-status-level

The frontend expect the level to be a number:

https://github.com/kiali/kiali-ui/blob/04b759826508c5ebfbf80226c6b3f0ef5250faed/src/pages/IstioConfigDetails/IstioObjectDetails/IstioStatusMessageList.tsx#L18-L24

And

https://github.com/kiali/kiali-ui/blob/bdb6ec1d1b04c9b90f8578916ac41b16a005c85a/src/types/IstioObjects.ts#L77-L82

xeviknal:
Istio status messages are back:
Screenshot of Kiali Console (22)

Kind and api, are back as well for me:
Screenshot of Kiali Console (23)

PR #2253, opened 2021-10-06
Add Trendlines to Metrics

israel-hdez:
* Adding regression-js as a helper library for the linear regression algorithm
* Add a "Trendlines" option in the inbound/outbound metric tabs.
* Trendlines are dashed lines.
* Trendlines are shown and hidden together with its associated serie when the corresponding label is clicked on the legend.
* Trendlines are highlighted together with its associated serie when the mouse cursor is over the corresponding label on the legend.

** Animation **

Related #2997.

cc @jotak

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,6 +92,7 @@
     "redux-persist": "5.10.0",
     "redux-persist-transform-filter": "0.0.18",
     "redux-thunk": "2.3.0",
+    "regression": "^2.0.1",
     "reselect": "4.0.0",
     "screenfull": "5.0.2",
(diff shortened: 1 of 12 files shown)


lucasponce:
Hi @israel-hdez,

When I revisited this issue https://github.com/kiali/kiali/issues/2997#issuecomment-894111363 I was thinking in exactly the same design and implementation as you have worked :), and it's great.

image

So in terms of features I have no comments :) as it matches with the feature description.

I'm testing the feature to provide some feedback in terms of side effects/corner cases, but all that I see looks good.

PR #2255, opened 2021-10-07
Fix null errorMetrics in CardSpakline

lucasponce:
Some backend queries can return a valid null in the errorMetrics.

On my example, "travel-portal" namespace had no info in the errorMetrics, but it has info in the non-error:
image

A potential workaround for these cases could be to move the errorCondition inside the main check.


lucasponce:
Pending post-verification from QE, but discussed with @prachiyadav to consolidate minor PRs in master and review them later.

hhovsepy:
Post merge verified:
Screenshot from 2021-10-13 12-42-03

PR #2257, opened 2021-10-19
Fix form validation for wildcard and mesh gateway

lucasponce:
Fixes https://github.com/kiali/kiali/issues/4416

Added a couple of synced form validations on the VirtualHosts field:

image

And Gateway "include mesh" field:

image

### DIFF
--- a/src/components/IstioWizards/GatewaySelector.tsx
+++ b/src/components/IstioWizards/GatewaySelector.tsx
@@ -128,24 +128,26 @@ class GatewaySelector extends React.Component<Props, GatewaySelectorState> {
   };
 
   isMeshGatewayValid = (): boolean => {
+    const hasVsWildcard = this.props.vsHosts.some(h => h === '*');
+    const hasGwWildcard = this.state.gwHosts.split(',').some(h => h === '*');
     // Gateway added
     if (this.state.addGateway) {
       // Mesh can't use wildcard in the hosts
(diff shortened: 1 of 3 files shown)


lucasponce:
Thanks @hhovsepy, yes, it's a little bit tricky, as there are some edges that were not covered.
I've updated the form validation in the gateway to cover your comments, let me know how it looks like.

hhovsepy:
Unable to remove Gateways while updating the traffic.
Create traffic and select "Create Gateway".
Edit traffic and uncheck "Add Gateway" at all, save.
Screenshot from 2021-10-19 15-41-33

lucasponce:
> Create traffic and select "Create Gateway".

When you create the traffic, can you show the Gateway details (and VHosts ?)

> Edit traffic and uncheck "Add Gateway" at all, save.

When you update/edit the traffic, do you create a new Gateway ?

I can't reproduce completely your scenario, I'd need that info.

lucasponce:
Sorry, I've added some unintended noise, basically the form validation is different, the presence of "*" as a hosts can be valid only when there is a gateway different from the "mesh".

I'll update this PR.

lucasponce:
Updated.
Wildcards in Hosts are not allowed to be in the VS without a non mesh Gateway.
I guess I've tested all combinations in the form, let me know if you find a missing edge on the PR.

PR #2259, opened 2021-10-19
Score and rank graph nodes

nrfox:
Adds a way to score/rank graph nodes based on configurable criteria. The ranking criteria is selected by the "Ranking" display toggles the same way that other graph display options are selected. Find/hide can be used to filter nodes based on their rank.

Fixes: kiali/kiali#2888


abonas:
> @jshaughn this could use slightly more polish but wanted to get some input before marking ready for review if you don't mind taking a look and playing around with it a bit. I'm not sure the UX is quite right by putting the ranking options in the display menu but that seemed like the most natural place to put it. imo it's nice being able to mix and match ranking options. Adding new scoring options in the future should not be difficult.

It makes sense to put it in display menu, although perhaps the wording might need some changing. Perhaps add a verb/action to clarify the meaning, like the one above it which has "show badges".

nrfox:
> I don't think that's where I'd put the hook, it's just got a specific job, instead it would probably be somewhere in the CytoscapeGraph.tsx#processGraphUpdate() logic. The logic is fine where it is now, I'm not really proposing a move. If you feel you'd rather make a change then that's fine too.

I ended up moving the logic here since having it where it used to be in graphDataSource required a data refresh in order for the rankings to update. It wasn't working well with find/hide updates. It's working as expected in `processGraphUpdate`.

> I think a normalized rank-range may make it easier for users to then perform a find/hide. The numbers are relative, so it makes sense to me. The rankings would not change for a dense graph and a sparse graph.

The rank range is now normalized between 1..100. I tested this on a few different node counts and it seemed to produce correct results but if you could take a look as well that'd be much appreciated!

src/pages/Graph/GraphToolbar/GraphSettings.tsx
```diff
@@ -513,6 +523,19 @@ class GraphSettings extends React.PureComponent<GraphSettingsProps, GraphSetting
...
+        isChecked: rank,
+        onChange: toggleRank,
+        tooltip: (
+          <div style={{ textAlign: 'left' }}>
+            Rank graph nodes based on configurable criteria such as 'number of inbound edges'. These rankings can be
+            used in the graph find/hide feature to help highlight the most important workloads, services, and
+            applications. Rankings are normalized to fit between 1..100 and nodes may tie with each other in rank.
+          </div>
+        )
       }
```
jshaughn, line 539: This looks good using the new approach.  But this option should be placed alphabetically in the `Show` section.

I needed more from the tooltip text.  I see the ranks are 1..100 but I don't know if 1 is highest or lowest.  Maybe continue the example, "for example, the node with the highest number of inbound edges would be ranked 1". (or is it 100 :-)).

src/pages/Graph/GraphToolbar/GraphFind.tsx
```diff
@@ -698,6 +701,13 @@ export class GraphFind extends React.Component<GraphFindProps, GraphFindState> {
       case 'op':
       case 'operation':
         return { target: 'node', selector: `[${CyNode.aggregateValue} ${op} "${val}"]` };
+      case 'rank': {
+        if (!this.props.rank || this.props.rankBy.length === 0) {
+          return undefined;
+        }
+        const s = this.getNumericSelector(CyNode.rank, op, val, expression, isFind);
+        return s ? { target: 'node', selector: s } : undefined;
+      }
```
jshaughn, line 710: We have a convention (hidden, yes, known probably only be me, but...) to enable options that are required for the use of an operand used in find or hide.  For example, see `mtls`.  If you use this in an expression we automatically ensure that the `Security` option is enabled.  Try it, I think we want to do the same here.

Also, I think we should generate an error if the `val` is out of range.

jshaughn (changes requested):
I think the repositioning of the logic looks good, I'll take another look on the next round (or hit up someone else while I'm away).  I hit one bug and have a few other suggested changes.   Also, I wasn't sure, but it seemed to me that maybe nodes outside of the requested namespace(s) are not ranked, is that by design?

src/pages/Graph/GraphHelpFind.tsx
```diff
@@ -176,7 +176,8 @@ export default class GraphHelpFind extends React.Component<GraphHelpFindProps> {
       ['http > 0.5', `edges with http rate > 0.5 rps`],
       ['rt > 500', `edges with response time > 500ms. (requires response time edge labels)`],
       ['%httptraffic >= 50.0', `edges with >= 50% of the outbound http request traffic of the parent`],
-      ['node = svc and svc startswith det or !traffic', 'service node starting with "det" or edges with no traffic']
+      ['node = svc and svc startswith det or !traffic', 'service node starting with "det" or edges with no traffic'],
+      ['rank >= 2', 'top 2 highest ranking nodes']
```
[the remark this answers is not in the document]
lucasponce: The main theme/topic is around "score" https://github.com/kiali/kiali/issues/2888 so, if we are going to open the semantic/wording about the terms, let's try to keep the discussion there.

This is a new feature, so, more than "churn" a lot the wording, what I would expect is that the description is provided clear (rank/score/importance, etc).

src/pages/Graph/GraphToolbar/GraphFind.tsx
nrfox: This helped in discovering a subtle bug where node scoring was not zeroing out old scores/ranks correctly when no selection criteria was passed in.

I added in the automatically enable feature. It's a nice touch but just for ranking, users will still need to select their scoring criteria e.g. you can have rank checked but not have any criteria such as "inbound edges" checked.

src/pages/Graph/GraphToolbar/GraphSettings.tsx
nrfox: Let me know if it's more clear now

src/pages/Graph/GraphHelpFind.tsx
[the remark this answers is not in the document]
jshaughn: I'm fine with the `Rank` verbiage as long as we have a good tooltip explaining this Display option.

src/pages/Graph/GraphToolbar/GraphFind.tsx
jshaughn: Nice, glad it surfaced an issue.  It's unlikely that users will use Find/Hide withhout having set any rank criteria, so just ensuring it's enabled should be sufficient.

src/pages/Graph/GraphHelpFind.tsx
```diff
@@ -176,7 +176,8 @@ export default class GraphHelpFind extends React.Component<GraphHelpFindProps> {
       ['http > 0.5', `edges with http rate > 0.5 rps`],
       ['rt > 500', `edges with response time > 500ms. (requires response time edge labels)`],
       ['%httptraffic >= 50.0', `edges with >= 50% of the outbound http request traffic of the parent`],
-      ['node = svc and svc startswith det or !traffic', 'service node starting with "det" or edges with no traffic']
+      ['node = svc and svc startswith det or !traffic', 'service node starting with "det" or edges with no traffic'],
+      ['rank <= 2', 'top 2 highest ranking nodes']
```
jshaughn, line 180: Not sure this is quite right.  It's not guaranteed to highlight 2 nodes.  Shouldn't it be something like, "nodes assigned a top 2 ranking"?

src/pages/Graph/GraphToolbar/GraphSettings.tsx
```diff
@@ -487,6 +497,22 @@ class GraphSettings extends React.PureComponent<GraphSettingsProps, GraphSetting
...
+        isChecked: rank,
+        onChange: toggleRank,
+        tooltip: (
+          <div style={{ textAlign: 'left' }}>
+            Rank graph nodes based on configurable criteria such as 'number of inbound edges'. These rankings can be
+            used in the graph find/hide feature to help highlight the most important workloads, services, and
+            applications. Rankings are normalized to fit between 1..100 and nodes may tie with each other in rank.
+            Ranking starts at 1 for the top ranked nodes so when ranking nodes based on 'number of inbound edges', the
+            node(s) with the most inbound edges would have rank 1. Node(s) with the second most inbound edges would have
+            rank 2.
```
jshaughn, line 512: This is much more clear.  I would maybe change the end to add a little more information:

From: `Node(s) with the second most inbound edges would have rank 2.`

To: `Each selected criteria contributes equally to a node's ranking.  Although 100 rankings are possible, only the required number of rankings are assigned, starting at 1. `

jshaughn (changes requested):
Really close, I think just some wording needs to change...

src/pages/Graph/GraphHelpFind.tsx
nrfox: Yes that's confusing. It's the "top 2" in terms of rank which could be more than 2 nodes and not the number 1 and number 2 nodes.

jshaughn (approved):
Looking good to me.  Tooltip text and verbiage seems clear to me.  I tested single criteria, combined criteria, find/hide, auto-enable for find/hide, error when find/hide has rank-out-of-range, and bookmarking.

I think we should follow up by adding rank display to the node side panel (when active).  I will create an issue.

src/pages/Graph/GraphToolbar/GraphFind.tsx
```diff
@@ -698,6 +702,19 @@ export class GraphFind extends React.Component<GraphFindProps, GraphFindState> {
       case 'op':
       case 'operation':
         return { target: 'node', selector: `[${CyNode.aggregateValue} ${op} "${val}"]` };
+      case 'rank': {
+        if (!this.props.rank) {
+          AlertUtils.addSuccess('Enabling "rank" display option for graph find/hide expression');
+          this.props.toggleRank();
+        }
```
nrfox, line 709: @jshaughn one thing that @mattmahoneyrh brought up during his testing is that when `rank = X` is in the find/hide box, the user is unable to deselect the "rank" display option. This is because of the autotoggle here. Other display options with the autotoggle like response time work similarly. Is this intended or is it a bug?

nrfox: display_options_deselect

jshaughn: It's correct.  It actually gets deselected and re-enabled quickly.  You can't have the find/hide expression and not have it enabled.  The same behavior can be seen with mtls and the security option, for example.

jshaughn: if you look closely you may notice the notification count go up.  anyway, it's correct behavior. ship it.

nrfox: @mattmahoneyrh ^^

mattmahoneyrh (approved):

mattmahoneyrh:
Test case: SWSQE-2161

PR #2260, opened 2021-10-20
Add preview mode in overview page

aljesusg:
Fix https://github.com/kiali/kiali/issues/4433

Still working, I'll move this to next step after input

Screenshot from 2021-10-20 11-08-14


lucasponce:
First comments:
- Modals should have same look and feel, at least in the basic controls and size.
- Copy/Download pattern is already used in the logs view, so, we try to re-use the same pattern, so probably a copy and download button can go there.
- I'd like to see the draft with a combined and highly populated scenario, like the travel-agency demo or similar, I've fetched locally but it seems it has some issues.
- I'm not sure that trying to use generic types is going to help you here as remember that then you'd need to revert them when creating, but perhaps my comment is early on this, I'm not looking at the code yet.

lucasponce:
> Ok know is like the logs but I think that the separation between copy and download is so much 1 em for modal.

Are you sure that the modal is the same that used for the Service Wizards ? which have a style defined that we should use for modal and shared components ?

The multi-tab approach doesn't look bad and it's consistent with the Envoy tabs which have a similar pattern.

For the shared elements, we need to keep consistency but for the aspects that may be related to the new flow we can work with UX, but I'll try to keep the consistency in the flows we already have.

lucasponce:
> OMG, I forgot push the commit. sorry @lucasponce this is how looks like

Thanks for the update, I wonder if we want to apply here the same pattern as used in the Envoy work, having two levels of tabs.

One level for "AuthorizationPolicies" and "Sidecars", and in a second level the enumeration of the resources.
I see more value to have a bigger editor, instead to have two editor at same time.

Also, this pattern would let to grow in case that others resources may be added into this feature (i.e. PeerAuthentication).

Thinking in other use cases, there are more cases where there are three or four types of resources involved.

But my suggestion is also to put this on the agenda of the UX meetings and also involve @andrew-ronaldson on the discussion.

hhovsepy:
Just a reminder to include "apiVersion" and "kind" per config in the export/copy YAML file.

aljesusg:
> Just a reminder to include "apiVersion" and "kind" per config in the export/copy YAML file.

Mmm good comment...we can add the kind easily but We must do some changes in the backend to get apiVersion because for example for sidecars  the apiversion **networking.istio.io/v1alpha3** can change some day.

Any idea @lucasponce ?

lucasponce:
> Mmm good comment...we can add the kind easily but We must do some changes in the backend to get apiVersion because for example for sidecars the apiversion networking.istio.io/v1alpha3 can change some day.

No need to change the backend for this.
It's a fixed API:
image

So, just add it in the buildAuthorizationPolicy used to generate the code from the overview and that's ok:
image

image

If in the future it needs to deal with different APIs then we can think on refactor, but I don't see it necessary yet, in any case, those methods needs to be in sync with the version used.

lucasponce:
Some comments.

The tab style should be consistent in font size and aspect.

image

If you want to use the first tab level with the whole width aspect is ok, as that pattern is used in the Service Wizard:

image

But second level should have a little space and not introduce new font sizing and aspects.

For reference, the second level tab used in the Envoy feature looks like this:

image

I remember that we discussed with @lwrigh if we want to have similar style when using the 2 tabs level pattern but here I defer into @andrew-ronaldson feedback about that.

Also, the current PR has issues in other namespaces:
image

image

hhovsepy:
"apiVersion" and "kind" are exported now, thank you.
It seems "copy to clipboard" does not copy.

hhovsepy:
Suggestion: User can change each config page one by one via YAML editor, and Update Traffic Policies now can reset all of them. But what about having a Preview page on Update Traffic Policies?

aljesusg:
Added Info icon to show that the object changed. I don't know if there is an icon more appropriate for this
Screenshot from 2021-11-18 12-32-34

lucasponce:
> Added Info icon to show that the object changed. I don't know if there is an icon more appropriate for this

Let's use the same pattern used in the Istio Config editor, I think it's just an asterisk on the name when there is a modification.

andrew-ronaldson:
> > Added Info icon to show that the object changed. I don't know if there is an icon more appropriate for this
> 
> Let's use the same pattern used in the Istio Config editor, I think it's just an asterisk on the tab's name when there is a modification.
> 
> image

+1 @lucasponce. I like the consistency with the *

aljesusg:
Fix styles
Screenshot from 2021-11-22 10-49-49

aljesusg:
Working in show the edition tool for update traffic policies. Should we show an option to set by default the traffic policies?

aljesusg:
Ok, new things:

- Added update traffic preview
- Removed code from overviewpage to a new component trafficmanagement (overview page was to big)
- Moved control to this new component

Problems:

- Sometimes the title tab is not aligned with the content of the aceeditor, I am dealing with that now, I can't find the problem :(

lucasponce:
Thanks @aljesusg I'll review your changes when to tell me (to not add you noise while it's a WIP).

One thing that is relevant is to move the permissions query (which is already heavy and it affects to scalability) into the modal, so, that would make to render this page faster in large number of namespaces.

aljesusg:
ok @lucasponce fixed, is ready to review it. Thanks !

lucasponce:
Thanks @aljesusg.

I'm going to start adding comments related to the feature & style and next, I'll work in the code.

A first comment that is relevant:

image

This work should change the logic of the "getPermission", instead to ask for every namespace in a single place, it should let the user enter in the modal, propose the changes and if they have permissions, allow them to write/update them.

I'm focus on this, because in terms of performance, that query is a heavy one and now with this change it can be moved inside the modal in a next step.

lucasponce:
The "Delete All" modal looks broken:
image

Linking this comment with my previous comment, if we move the permissions to the "inside" modal, I think, user should open the "Delete" modal as well, and then, inside, ask specific permission for the namespace.

With that minor change, the penalty performance will be minimized on high number of namespaces.

Again, I know that would require changes in the modal component, but I think it's worth to invest them on this work.

aljesusg:
> Thanks @aljesusg.
> 
> I'm going to start adding comments related to the feature & style and next, I'll work in the code.
> 
> A first comment that is relevant:
> 
> image
> 
> This work should change the logic of the "getPermission", instead to ask for every namespace in a single place, it should let the user enter in the modal, propose the changes and if they have permissions, allow them to write/update them.
> 
> I'm focus on this, because in terms of performance, that query is a heavy one and now with this change it can be moved inside the modal in a next step.

+1
> 
> I know that this change is not just a move of the query, it may trigger other changes and accomodate the logic for it.
> 
> P.S.: I detailed too much, but if the user doesn't have permissions, in the wizards are already some examples about how to show a message indicating that, that style could be reused here as well.

Ok, I didn't move it by the delete option, so are we ok about let user click the delete option and show him the modal telling him if they have not rights? (Delete option just show the modal not the previews)

lucasponce:
> Ok, I didn't move it by the delete option, so are we ok about let user click the delete option and show him the modal telling him if they have not rights? (Delete option just show the modal not the previews)

Yes. Exactly. But the "Delete" button should be disabled if user has not rights. Similar flow as the Preview.

Then, the overview page would not require to query all permissions for all namespaces in advance, only when user wants to perform the action.

aljesusg:
> The "Delete All" modal looks broken: image
> 
> Linking this comment with my previous comment, if we move the permissions to the "inside" modal, I think, user should open the "Delete" modal as well, and then, inside, ask specific permission for the namespace.
> 
> With that minor change, the penalty performance will be minimized on high number of namespaces.
> 
> Again, I know that would require changes in the modal component, but I think it's worth to invest them on this work.

mm weird There is a

> > Ok, I didn't move it by the delete option, so are we ok about let user click the delete option and show him the modal telling him if they have not rights? (Delete option just show the modal not the previews)
> 
> Yes. Exactly. But the "Delete" button should be disabled if user has not rights. Similar flow as the Preview.
> 
> Then, the overview page would not require to query all permissions for all namespaces in advance, only when user wants to perform the action.
> 
> It's one of the few queries that requires a non cached POST and I'd say is the culprit of some problems in this page at large scale.

Ok, Do we want to show preview in delete too ? IMO it makes no sense, maybe we can show directly the modal with the info about we are going to delete or a modal with you have not rights to delete.

lucasponce:
> Ok, Do we want to show preview in delete too ? IMO it makes no sense, maybe we can show directly the modal with the info about we are going to delete or a modal with you have not rights to delete.

I don't think so. I mean to keep the Delete modal, but:
- All users can open the Delete modal.
- The permission query happens inside the modal, indicating the resources that would be deleted.
- If the user has permissions, the user can perform the operation.

What it will change from today's logic is that if the user doesn't have permission, the "Delete" button inside the modal will be disabled instead to get disabled the option in the kebap menu.

aljesusg:
> > Ok, Do we want to show preview in delete too ? IMO it makes no sense, maybe we can show directly the modal with the info about we are going to delete or a modal with you have not rights to delete.
> 
> I don't think so. I mean to keep the Delete modal, but:
> 
> * All users can open the Delete modal.
> * The permission query happens inside the modal, indicating the resources that would be deleted.
> * If the user has permissions, the user can perform the operation.
> 
> What it will change from today's logic is that if the user doesn't have permission, the "Delete" button inside the modal will be disabled instead to get disabled the option in the kebap menu.

Anyway lucas enable/disable autoinjection need permissions too. IMO we should move them to the trafficManagement too

lucasponce:
> Anyway lucas enable/disable autoinjection need permissions too. IMO we should move them to the trafficManagement too

Right. But perhaps better a modal where the query can be performed and ask user for confirmation, then we would follow the same flow.

lucasponce:
I'm testing the modals for Auto-Injection.
Those look good but I think the logic is not well implemented as I have a disabled button:
image

I see that the query happens in the modal's show, but something is not correctly handled as I'd expect to have that button enabled:
image

image

src/pages/Overview/TrafficManagement.tsx
```diff
@@ -0,0 +1,288 @@
...
+
+type State = {
+  confirmationModal: boolean;
+  authorizationPolicies: AuthorizationPolicy[];
+  sidecars: Sidecar[];
+  disableOp: boolean;
+  canaryVersion: string;
+};
+
+export default class TrafficManagement extends React.Component<TrafficManagementProps, State> {
```
lucasponce, line 39: I read "TrafficManagement" a little bit confusing in this context.
I suggest "TrafficPolicies" or "OverviewTrafficPolicies" as in this folder specific components for this page are used in that way.

In future iterations it can be refactored in other way.

src/pages/Overview/OverviewPage.tsx
```diff
@@ -572,43 +548,33 @@ export class OverviewPage extends React.Component<OverviewProps, State> {
...
+          isSeparator: true
         });
         const enableAction = {
           isGroup: false,
           isSeparator: false,
-          isDisabled: !canWrite,
           title: 'Enable Auto Injection',
-          action: (ns: string) => this.onAddRemoveAutoInjection(ns, true, false)
+          action: (ns: string) =>
+            this.setState({ showTrafficManagement: true, nsTarget: ns, opTarget: 'enable', kind: 'injection' })
```
lucasponce, line 563: I understand this design but it's confusing for another developer.
If we want to concentrate some of the actions in the "TrafficManagement" component, I guess that component should get a better name as "OverviewNamespaceModal" or something in that way that would give hint to understand the logic.

From my perspective, now that component performs more tasks: annotate a namespace, create some Istio configuration for an entire namespace, and potentially more in the short term.

src/pages/Overview/TrafficManagement.tsx
```diff
@@ -0,0 +1,288 @@
...
+          break;
+      }
+    }
+  }
+
+  fetchPermission = () => {
+    this.promises.register('namespacepermissions', API.getIstioPermissions([this.props.nsTarget])).then(result => {
+      this.setState({
+        confirmationModal: true,
+        disableOp: !result.data[this.props.nsTarget][AUTHORIZATION_POLICIES][this.props.opTarget]
```
lucasponce, line 90: This sentence won't work here.
The response of the permissions query is similar to:
```
{"authorizationpolicies":{"create":true,"update":true,"delete":true},"gateways":{"create":true,"update":true,"delete":true},"peerauthentications":{"create":true,"update":true,"delete":true},"requestauthentications":{"create":true,"update":true,"delete":true},"serviceentries":{"create":true,"update":true,"delete":true},"sidecars":{"create":true,"update":true,"delete":true}}
```

But you are asking for a "disable/enable" operation.

src/components/IstioConfigPreview/EditResources.tsx
```diff
@@ -0,0 +1,63 @@
...
+  aceEditorRef: React.RefObject<AceEditor>;
+  constructor(props: Props) {
+    super(props);
+    this.state = { resourceTab: 0 };
+    this.aceEditorRef = React.createRef();
+  }
+
+  render() {
+    return (
+      <Grid className={style({ padding: '20px' })}>
```
lucasponce, line 33: I think it could be better to not use a Grid here.
I see a lot of this pattern to use a top layout with a single item, that's a little bit overkill, and we'll pay some offer when moving to another version.

In short, use a simple "div" element if you don't need anything else, but use a grid if you need a grid layout.

src/components/IstioConfigPreview/EditorPreview.tsx
```diff
@@ -0,0 +1,78 @@
...
+  };
+
+  render() {
+    return (
+      <AceEditor
+        ref={this.aceEditorRef}
+        mode="yaml"
+        theme="eclipse"
+        onChange={value => this.onChange(value)}
+        height={'var(--kiali-yaml-editor-height)'}
```
lucasponce, line 67: This style is tricky.
On the IstioConfigPage is set to adjust the height when the whole page scrolls as it needs to calculate some margins to not overflow (and even in that case I see there is still some tech debt there).

But for this use case, this style it can be fixed (i.e. 275px), which would get the benefit to fit in all screens and also to avoid the double scroll bar (editor and modal).

With that change the scroll will happen only in the modal:
image

lucasponce:
It seems that in the enable/disable/remove operations, the main modal is also opened and I don't think it should even if the page has a blur effect:
image

Also, I guess that the "enable" verb should be in blue and perhaps the "disable"/"remove" can be in red.

lucasponce:
I see a problem, I guess, due to the state management of the components.

Steps to reproduce:

1.- Open a "Create Traffic Policies" modal from a populated namespace, i.e., default:

image

2.- Select "Sidecars" tab:
image

3.- Cancel the modal.

4,.- Create a new namespace, it should be empty, i.e. test
image

5.- Cancel de modal.

6.- Open again the "Create Traffic Policies" in the "default" namespace:

image

So, it seems the state is not refreshed correctly, and it causes this problem.

lucasponce:
I'm still seeing the problem.
I've recorded a video, if it can help to reproduce it:
kiali-overview-preview-problem

aljesusg:
> I'm still seeing the problem. I've recorded a video, if it can help to reproduce it: kiali-overview-preview-problem

Thanks !! This helped me . I found the problem in loading trafficPolicies. Fixed

lucasponce:
It seems that there're still issues:
kiali-overview-preview-problem2

lucasponce:
More testing:
kiali-overview-preview-problem3

- The "Update" action should open the wizard again with update config, to be re-write. However, my example seems that the "Update" button doesn't open the modal.
- The confirmation on the "Create" button looks a little bit redundant, but if we want to leave it, I'm fine, perhaps the "Create" button in blue and the "Remove" actions in red.

lucasponce (approved):
Feature perspective, it's more robust and the scenarios discussed in the PR work.
Code perspective, probably we want to align how the state is handled with other wizards, as it's not a trivial aspect and I guess we'd need to unify them, but that can be discussed in the next Sprint.

I'm +1 to consolidate if @hhovsepy and @skondkar can also take a look at this feature.
Nice work.

hhovsepy:
Have tested all the scenarios mentioned above, LGTM. +1 on Update preview page.
Screenshot from 2022-01-11 09-19-28
Screenshot from 2022-01-11 09-19-39
Screenshot from 2022-01-11 09-19-54
Screenshot from 2022-01-11 09-20-14

Only minor comment: Export or Copy of all resources does not put separator between AuthPolicies and Sidecars groups:
Screenshot from 2022-01-11 09-13-50

aljesusg:
> Have tested all the scenarios mentioned above, LGTM. +1 on Update preview page.
> 
> Screenshot from 2022-01-11 09-19-28
> 
> Screenshot from 2022-01-11 09-19-39
> 
> Screenshot from 2022-01-11 09-19-54
> 
> Screenshot from 2022-01-11 09-20-14
> 
> 
> 
> Only minor comment: Export or Copy of all resources does not put separator between AuthPolicies and Sidecars groups:
> 
> Screenshot from 2022-01-11 09-13-50
> 
> 

Mmm Should we add it? I don't know if kubectl throw a problem without the separator. I'm going to add it

lucasponce:
> Mmm Should we add it? I don't know if kubectl throw a problem without the separator. I'm going to add it

+1 adding it, if not I guess a "kubectl apply -f " may not work

aljesusg:
> > Mmm Should we add it? I don't know if kubectl throw a problem without the separator. I'm going to add it
> 
> +1 adding it, if not I guess a "kubectl apply -f " may not work

Done :)

skondkar (approved):
Verified mentioned scenarios. LGTM
Also verified Update traffic policies preview page and copy/download.

Screenshot from 2022-01-12 16-42-05

Screenshot from 2022-01-12 17-00-35

hhovsepy (approved):
Verified.

skondkar:
Polarion testcase ->> SWSQE-2403

PR #2267, opened 2021-11-05
Refactor graphSettings to use redux-toolkit and hooks

nrfox:
This is more along the lines of a POC but if people like the changes it can be merged. The goal was to try out the [Redux Toolkil](https://redux-toolkit.js.org/) library along with using React hooks in an existing component to try and determine what the benefits are and how painful migrating existing components would be. It ended up being a much larger diff than I intended but it's probably a better example that way.

My impressions from the refactor are pretty positive. Here are some highlights:

**Less boilerplate and indirection when using redux**
With redux toolkit, writing the **reducer logic** is **simplified** because you can do mutable updates and they will get converted to immutable updates behind the scenes. **Actions** are **generated automatically** further reducing boilerplate. This is more of a convention but putting the reducer, actions, and selector(s) all together makes it much easier to find and follow the redux logic for me personally rather than having them scattered around.

With hooks, `mapStateToProps` and `mapDispatchToProps` are no longer needed. These get replaced by a [useSelector](https://github.com/kiali/kiali-ui/compare/master...nrfox:redux-rtk?expand=1#diff-40e08717d528d25ecee13a8481d238d1e62e7ab1721d577b3696036c7a4d5516R91) and [useDispatch](https://github.com/kiali/kiali-ui/compare/master...nrfox:redux-rtk?expand=1#diff-40e08717d528d25ecee13a8481d238d1e62e7ab1721d577b3696036c7a4d5516R94) hook inside of the component. I find this now makes it **easier** to **know exactly where state/props are coming from** since props are just props and redux state is brought directly into the component through the selector. It's also more obvious which click handlers dispatch redux actions. Whereas before, working inside a component it was difficult to know which props came from the redux store and which were regular props. The same was true of know which methods fired redux actions.

**Works alongside existing reducers/class components**
The reducers produced by redux toolkit's helper methods are just normal reducers meaning you [can swap out existing reducers easily](https://github.com/kiali/kiali-ui/compare/master...nrfox:redux-rtk?expand=1#diff-59e2efff49298d02d725acb13279742938107a9f2f041c713d777019ed8be05aR22). Existing class components [can import and use the generated actions](https://github.com/kiali/kiali-ui/compare/master...nrfox:redux-rtk?expand=1#diff-97999f8d6e2bddd50afe82ae80a948f7328c91f184b80b8718a6a9c5bc49cef4R952-R960) with `mapXToProps` methods seamlessly.

One downside is that hooks cannot be used inside of class components, so class components would need to be migrated all at once. However the reducers can be migrated to use `createSlice` seamlessly.

Overall its nice to use this new toolkit and feel some improvements in the API that address some potential pain points. I think especially for new components makes a lot of sense to write them in this format. Looking at [GraphReset](https://github.com/kiali/kiali-ui/compare/master...nrfox:redux-rtk?expand=1#diff-b6525968bd562a6528173af3e781b036699119d6a7a4ced9071cb2292819130fR1-R22) you get a good feel for how the reduction in boilerplate makes it easier to implement and follow the data flow.


jshaughn:
Thanks, @nrfox for this investigation.  This PR serves as a nice example.

PR #2271, opened 2021-11-15
Showing appProtocol information in tooltip for Service details Network section.

hhovsepy:
UI part of https://github.com/kiali/kiali/pull/4499

### DIFF
--- a/src/pages/ServiceDetails/ServiceNetwork.tsx
+++ b/src/pages/ServiceDetails/ServiceNetwork.tsx
@@ -145,7 +145,7 @@ class ServiceNetwork extends React.Component<Props> {
                         <span style={{ marginRight: '5px' }}>
                           {port.name} {port.port}/{port.protocol}
                         </span>
-                        {port.appProtocol !== '' ? (
+                        {port.appProtocol && port.appProtocol !== '' ? (
                           <Tooltip
                             position={TooltipPosition.right}
                             content={<div style={{ textAlign: 'left' }}>App Protocol: {port.appProtocol}</div>}
--- a/src/pages/ServiceDetails/ServiceNetwork.tsx
+++ b/src/pages/ServiceDetails/ServiceNetwork.tsx
@@ -142,10 +142,20 @@ class ServiceNetwork extends React.Component<Props> {
                   {(this.props.serviceDetails.service.ports || []).map((port, i) => {
                     return (
                       <div key={'port_' + i}>
-                        <span style={{ marginRight: '10px' }}>
+                        <span style={{ marginRight: '5px' }}>
                           {port.name} {port.port}/{port.protocol}
                         </span>
-                        {this.hasIssue(i) ? this.getPortOver(i) : undefined}
+                        {port.appProtocol !== '' ? (
+                          <Tooltip
+                            position={TooltipPosition.right}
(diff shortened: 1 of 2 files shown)


src/pages/ServiceDetails/ServiceNetwork.tsx
```diff
@@ -142,10 +142,20 @@ class ServiceNetwork extends React.Component<Props> {
...
+                          <Tooltip
+                            position={TooltipPosition.right}
+                            content={<div style={{ textAlign: 'left' }}>App Protocol: {port.appProtocol}</div>}
+                          >
+                            <span style={{ marginRight: '5px' }}>
+                              <KialiIcon.Info className={infoStyle} />
+                            </span>
+                          </Tooltip>
+                        ) : undefined}
+                        <span style={{ marginLeft: '5px' }}>{this.hasIssue(i) ? this.getPortOver(i) : undefined}</span>
```
xeviknal, line 158: Just one idea for this to be a bit more simple.
For each of the two scenarios wrt ports we end up having these three concepts: name, port and protocol. Don't you think that we could keep the UI showing only these three concepts instead of adding the app protocol hidden in a tooltip?

At the end, the appProtocol is the protocol. So if we have appProtocol, we shouldn't have it in the name field (`protocolo-suffix`), right? In case that appProtocol is not present, then we should take the protocol from the name (protocolo-name convention). This way, we shouldn't need to add any tooltip, just rethink the way to compute the `name port/protocol` variables in L46

Does this make sense?

hhovsepy: You mean in the case when name does not contain protocol and AppProtocol is set, then show directly AppProtocol instead of default 'TCP' protocol , in a format of "name port / appProtocol"  instead of "name port / protocol" ?
Could be a better solution than we a agreed, wdyt @lucasponce ?

hhovsepy: but remember that 'protocol' is different than the protocol in 'name' and is always default TCP or UDP.

xeviknal: I understood that you can't define the protocol in both places `name` and `appProtocol` at the same time. Like the example in the istio documentation:
```
  ports:
  - number: 3306
    name: database
    appProtocol: https
  - number: 80
    name: http-web
```

This would be very confusing to me, wouldn't it?

```
  - number: 3306
    name: mysql-database
    appProtocol: https
```

So I understand that you always will want to show: `name port/protocol` independently where the information comes from. I also understood that the `appProtocol` is something new to kubernetes 1.18+ to avoid doing this `protocol-name` unnatural format in the name field.

Sorry, I might be introducing noise here. I don't know if I am getting this right.

hhovsepy: When appProtocol is given, then I consider it as higher priority over name and validate it's format, ignoring name's validation.
In the showing format the '/protocol' is not the protocol in name, but it is TCP or UDP protocol., name always contains what it is set, either protocol-name or just name format.
Thus, the decision was made to not touch any display formatting but only show tooltip in case of given appProtocol.
So existing users will not see format changes.

xeviknal (approved):
LGTM! Comments approached

skondkar (approved):
LGTM.
Verified along with https://github.com/kiali/kiali/pull/4499

PR #2272, opened 2021-11-16
Show rank in node side panel

nrfox:
Shows rank in the node summary side panel.

Fixes kiali/kiali#4494

Screenshot from 2021-11-16 11-22-24
Screenshot from 2021-11-16 11-22-12


src/pages/Graph/SummaryPanelNode.tsx
```diff
@@ -345,6 +353,12 @@ export class SummaryPanelNode extends React.Component<SummaryPanelNodeProps, Sum
             {shouldRenderGatewayHostnames && this.renderGatewayHostnames(nodeData)}
           </>
         )}
+        {shouldRenderRank && (
+          <div>
+            <KialiIcon.Rank />
+            <span style={{ paddingLeft: '4px' }}>Rank: {nodeData.rank !== undefined ? nodeData.rank : 'N/A'}</span>
```
jshaughn, line 359: I don't think there should be an N/A rank.  I think all nodes should have a rank within the [1..100] range.  If ranking by Outbound Edges and a node has 0 outbound edges, I think it should get the lowest rank, not N/A.

src/pages/Graph/SummaryPanelNode.tsx
```diff
@@ -366,7 +380,9 @@ export class SummaryPanelNode extends React.Component<SummaryPanelNodeProps, Sum
 }
 
 const mapStateToProps = (state: KialiAppState) => ({
-  jaegerState: state.jaegerState
+  jaegerState: state.jaegerState,
+  rank: state.graph.toolbarState.rank,
+  rankBy: rankBySelector(state)
```
jshaughn, line 385: I get why you need these both, because we don't clear `rankBy` even if we disable `rank`, right?   But here's my suggestion, add a new redux prop, something like `minRank`.  This would be 0 when `rank` is disabled.  When `rank` is enabled it would store the minRank, something in the [1..100] range.   If we had this it would be the only redux prop needed here, as it would get reset each time ranks were calculated, or when `rank` was disabled.  Then you'd be able to show the current rank range as well as the individual node rank.

jshaughn (changes requested):
I like the icon.  If possible, I think you need to add in the UX feedback about the expand-carat, even though it's only indirectly related to this PR.  Also, see my suggestions and let me know what you think.

src/pages/Graph/SummaryPanelNode.tsx
nrfox: Node rank is `N/A` if there is no scoring criteria selected or if the scoring criteria does not apply to that particular node. Most scoring criteria will probably apply to all nodes and the default rank/score should just be the lowest possible rather than `N/A`. However if the user disables ranking for any reason, I think it makes sense to assign `N/A` rather than the lowest rank since nothing has been scored/ranked.

> If ranking by Outbound Edges and a node has 0 outbound edges, I think it should get the lowest rank, not N/A.

Agreed, if a node has no edges it should get the lowest rank and not `N/A`. That's a bug in the current implementation that I'll fix with this change.

src/pages/Graph/SummaryPanelNode.tsx
nrfox: Adding `minRank` as a redux prop sounds good but we'd still probably need to subscribe to `rankBy` updates since the `minRank` could stay the same for the overall graph even though the scoring criteria changes and the rank for that particular node changed. It'd be nice to subscribe to updates to a particular node and re-render the component when any of the data changes but iirc the graph updates based on the layout changing and since node scoring is a sort of post processing, it doesn't affect the layout.

nrfox:
Here's an updated screenshot of ranking info in the node summary panel with the lowest rank included and the host toggle indented more:
Screenshot from 2021-11-17 10-30-48
Screenshot from 2021-11-17 10-30-38

Instead of `Rank 3/5`, we could also do something like `Rank 3 of 5` or `Rank 3 out of 5`?

nrfox:
Other updates include:
- Rank does not disappear when `Rank` display option is unchecked. When `Rank` is unchecked the node summary panel will show `N/A` for the rank value.
- Rank display option renamed to `Rank Nodes`

src/pages/Graph/SummaryPanelNode.tsx
jshaughn: Totally fine to assign N/A or some indicator that there is no rank, I just don't think the user should ever see that value.  Rank would only appear in the side panel when ranks are active.

src/pages/Graph/SummaryPanelNode.tsx
jshaughn: Yes, you're right if minRank is just an int.  I suppose we could consolidate things into RankInfo, and it could be reassigned with a new object on any change.  Depends on how much you don't want to use rankBy just as a refresher.

jshaughn:
> Here's an updated screenshot...

Looks great!  And for me 3/5 can stay as-is.

src/pages/Graph/SummaryPanelNode.tsx
nrfox: iirc @lucasponce made the case that if `Rank: ...` always shows up in the node summary panel it would encourage more users to take advantage of the feature. Since Rank always shows up in the side panel then we need a value for when there is no rank hence `N/A`. I personally don't have strong feelings one way or another.

jshaughn (changes requested):
> Rank does not disappear when Rank display option is unchecked. When Rank is unchecked the node summary panel will show N/A for the rank value.

I don't really understand this decision.  I would expect no rank to show up when it's not relevant.  This is the way the other side-panel badging works.

> Rank display option renamed to Rank Nodes

I thought it was fine as just `Rank`.  If you are going to change it then I think you need to reverse it so that it flows with the `Show` title.   So, instead of `Show Rank Nodes`, it should be `Show Node Rank`.   Note that this would also change the alphabetized position.

src/pages/Graph/SummaryPanelNode.tsx
nrfox: Consolidating things into a single `RankInfo` object seems like a good approach. It might cause unnecessary renders later for components that just want to subscribe to either `rank` or `rankBy` but imo that's a pretty minor concern. If the node itself was in the store I think the most direct option would be to subscribe to the node by it's node id.

src/pages/Graph/SummaryPanelNode.tsx
nrfox: To keep rank consistent with the other options on the side panel, I'll keep it hidden whenever it is not selected instead of having `N/A` for now. We can revisit that later if need be.

nrfox:
> I thought it was fine as just `Rank`. If you are going to change it then I think you need to reverse it so that it flows with the `Show` title. So, instead of `Show Rank Nodes`, it should be `Show Node Rank`. Note that this would also change the alphabetized position.

I'm also ok with the wording as is. It came up as a potential source of confusion for users but I'm happy to keep it as is.

src/pages/Graph/SummaryPanelNode.tsx
nrfox: The summary target rank/score was not being updated when the graph was re-scored. That was the underlying issue requiring the `rankBy` hack. I updated this PR so that when the nodes are re-scored the summary target data will also get updated triggering a re-render for the node summary panel.

jshaughn (changes requested):
@nrfox I like the feature and the approach, but I feel like the code has gotten a little over-complicated.  I have several comments so to expedite I'm going to submit a PR for your consideration.

https://github.com/nrfox/kiali-ui/pull/2

jshaughn (approved):
All looks good to me!

pbajjuri20:
Tested and Passed on Kiali UI 1.44.0 (3f06954f801f758307b722f7db8a0b6d0adf7695)


rank_pr

rank_pr2

pbajjuri20 (approved):

pbajjuri20:
Polarion Test Case ID: 2167

PR #2274, opened 2021-11-24
Reconciliation status on side panel of Istio configs

leandroberetta:
This PR adds the following information when the control plane is configured to set the status field into the resources:

Resolves [#4457](https://github.com/kiali/kiali/issues/4457)

The status condition of type "Reconciled" will be shown in the side panel of Istio configs.

status-conditions

The status condition of type "Reconciled" will be indicated as follows in the config list per entry:

status-reconciled-2

### DIFF
--- a/src/components/Validations/ValidationObjectSummary.tsx
+++ b/src/components/Validations/ValidationObjectSummary.tsx
@@ -1,11 +1,12 @@
 import * as React from 'react';
-import { ObjectValidation, ValidationTypes } from '../../types/IstioObjects';
+import { ObjectValidation, StatusCondition, ValidationTypes } from '../../types/IstioObjects';
 import ValidationSummary from './ValidationSummary';
 import { CSSProperties } from 'react';
 
 interface Props {
   id: string;
   validations: ObjectValidation[];
+  reconciledCondition?: StatusCondition;
   style?: CSSProperties;
 }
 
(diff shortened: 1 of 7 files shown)


leandroberetta:
The amount of proxies seems wrong. On my first test, it indicated 32 proxies, then I added 5 instances more of details service (5 more proxies) but then the count was 48.

I'm not sure if the number is reliable as this feature is still in alpha.

I will investigate a little bit more.

lucasponce:
Perhaps the absolute number is not really relevant for the user, but having a flag when an object is not reconciled is a good information.

Let me test the feature and I can provide you more feedback.

lucasponce:
Also this draft is related with this epic as well https://github.com/kiali/kiali/issues/4241.

Basically we don't have an uniform side panel for all Istio resources, and it would be good to add that support to have an uniform view there.

Also the "reconciliation status" could be one common information that we could add in all resources.

leandroberetta:
> Also this draft is related with this epic as well [kiali/kiali#4241](https://github.com/kiali/kiali/issues/4241).
> 
> Basically we don't have an uniform side panel for all Istio resources, and it would be good to add that support to have an uniform view there.
> 
> Also the "reconciliation status" could be one common information that we could add in all resources.

@lucasponce sure, let me check the epic and think a little bit possible additions and the design of the side panel for further discussions.

lucasponce:
@leandroberetta do we want to close this one in favor of the work defined in https://github.com/kiali/kiali/issues/4241 ?

leandroberetta:
> @leandroberetta do we want to close this one in favor of the work defined in [kiali/kiali#4241](https://github.com/kiali/kiali/issues/4241) ?

Sure, sorry for the delay, makes sense to do that.

PR #2275, opened 2021-11-24
Graph zoom-aware labeling and scale-usability enhancements

jshaughn:
Closes https://github.com/kiali/kiali/issues/4521

There are server and operator PRs for the new config.

**Server PR:**:  https://github.com/kiali/kiali/pull/4544
**Operator PR:** https://github.com/kiali/kiali-operator/pull/461
 
**UPDATED: Jan 10, '22**

This is a step towards improving usability concerns as graph scale grows.  The changes include:
- add configurable minimum-font thresholds that determine when to show node/edge labels and/or badges.
   - as zoomed-font decreases text is initially removed (as it becomes unreadable), then badges (as they become inscrutable)
   - the thresholds are configurable as part of ui-defaults (i.e. install-wide as part of CR)
- significantly reduce "flashing" when mousing around the graph
  - reasonably fast mouse-over will not trigger highlight/dim
  - instead, encourage hover (300ms),  selection (click) or drag-zoom
    - hover will trigger neighborhood highlighting and contextInfo popup
    - selection will perform full-path highlighting and the side-panel update
      - also, clicking on a selected element will now un-select the element
    - right-click will continue to show contextMenu, now updated with new contextInfo and styling
- cytoscape toolbar now vertically presented bottom-left, to reduce it occluding graph elements
  - legend slightly resized and a few styling changes for consistency

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,4 +1,5 @@
 {
+  "proxy": "http://localhost:20001/kiali",
   "name": "@kiali/kiali-ui",
   "version": "1.45.0",
   "description": "React UI for [Kiali](https://github.com/kiali/kiali).",
@@ -66,7 +67,7 @@
     "csstips": "0.3.0",
     "csx": "9.1.0",
     "cy-node-html-label": "2.0.0",
-    "cytoscape": "3.15.1",
+    "cytoscape": "3.20.0",
     "cytoscape-canvas": "3.0.1",
     "cytoscape-cola": "2.3.1",
     "cytoscape-cose-bilkent": "4.1.0",
--- a/package.json
+++ b/package.json
@@ -1,5 +1,4 @@
 {
-  "proxy": "http://localhost:20001/kiali",
   "name": "@kiali/kiali-ui",
   "version": "1.45.0",
   "description": "React UI for [Kiali](https://github.com/kiali/kiali).",
@@ -67,7 +66,7 @@
     "csstips": "0.3.0",
     "csx": "9.1.0",
     "cy-node-html-label": "2.0.0",
-    "cytoscape": "3.20.0",
+    "cytoscape": "3.15.5",
     "cytoscape-canvas": "3.0.1",
     "cytoscape-cola": "2.3.1",
     "cytoscape-cose-bilkent": "4.1.0",
--- a/src/actions/GraphActions.ts
+++ b/src/actions/GraphActions.ts
@@ -1,5 +1,5 @@
 import { ActionType, createAction, createStandardAction } from 'typesafe-actions';
-import { CytoscapeClickEvent, GraphDefinition, Layout, NodeParamsType, RankResult } from '../types/Graph';
+import { CytoscapeEvent, GraphDefinition, Layout, NodeParamsType, RankResult } from '../types/Graph';
 import { ActionKeys } from './ActionKeys';
 import { TimeInMilliseconds } from 'types/Common';
 
@@ -10,7 +10,7 @@ export const GraphActions = {
   setNode: createStandardAction(ActionKeys.GRAPH_SET_NODE)<NodeParamsType | undefined>(),
   setRankResult: createStandardAction(ActionKeys.GRAPH_SET_RANK_RESULT)<RankResult>(),
   setUpdateTime: createStandardAction(ActionKeys.GRAPH_SET_UPDATE_TIME)<TimeInMilliseconds>(),
-  updateSummary: createStandardAction(ActionKeys.GRAPH_UPDATE_SUMMARY)<CytoscapeClickEvent>()
+  updateSummary: createStandardAction(ActionKeys.GRAPH_UPDATE_SUMMARY)<CytoscapeEvent>()
 };
 
 export type GraphAction = ActionType<typeof GraphActions>;
--- a/src/components/CytoscapeGraph/ContextMenu/EdgeContextMenu.tsx
+++ b/src/components/CytoscapeGraph/ContextMenu/EdgeContextMenu.tsx
@@ -1,0 +1,25 @@
+import * as React from 'react';
+import { style } from 'typestyle';
+import { prettyProtocol } from 'types/Graph';
+import { EdgeContextMenuProps } from '../CytoscapeContextMenu';
+import { getTitle } from 'pages/Graph/SummaryPanelCommon';
+import { renderBadgedName } from 'pages/Graph/SummaryLink';
+import { decoratedNodeData } from '../CytoscapeGraphUtils';
+import { EdgeSingular } from 'cytoscape';
+
+const contextMenu = style({
+  fontSize: 'var(--graph-side-panel--font-size)',
+  textAlign: 'left'
+});
+
+export class EdgeContextMenu extends React.PureComponent<EdgeContextMenuProps> {
+  render() {
+    return (
+      <div className={contextMenu}>
+        {getTitle(`Edge (${prettyProtocol(this.props.protocol)})`)}
+        {renderBadgedName(decoratedNodeData((this.props.element as EdgeSingular).source()), 'From:  ')}
+        {renderBadgedName(decoratedNodeData((this.props.element as EdgeSingular).target()), 'To:        ')}
+      </div>
+    );
+  }
+}
--- a/src/components/CytoscapeGraph/ContextMenu/NodeContextMenu.tsx
+++ b/src/components/CytoscapeGraph/ContextMenu/NodeContextMenu.tsx
@@ -62,7 +62,8 @@ export class NodeContextMenu extends React.PureComponent<Props> {
       case NodeType.APP:
       case NodeType.BOX:
         // only app boxes have full context menus
-        if (node.isBox === BoxByType.APP) {
+        const isBox = node.isBox;
+        if (!isBox || isBox === BoxByType.APP) {
           // Prefer workload links
           if (node.workload && node.parent) {
             name = node.workload;
--- a/src/components/CytoscapeGraph/ContextMenu/NodeContextMenu.tsx
+++ b/src/components/CytoscapeGraph/ContextMenu/NodeContextMenu.tsx
@@ -19,6 +19,7 @@ type ReduxProps = {
   jaegerInfo?: JaegerInfo;
 };
 
+// Note, in the below styles we assign colors to be consistent with PF Dropdown
 const contextMenu = style({
   fontSize: 'var(--graph-side-panel--font-size)',
   textAlign: 'left'
@@ -46,7 +47,7 @@ const contextMenuItem = style({
 });
 
 const contextMenuItemLink = style({
-  color: PFColors.Black600
+  color: PFColors.Black900
 });
 
 type Props = NodeContextMenuProps & ReduxProps;
--- a/src/components/CytoscapeGraph/ContextMenu/NodeContextMenu.tsx
+++ b/src/components/CytoscapeGraph/ContextMenu/NodeContextMenu.tsx
@@ -5,7 +5,7 @@ import { style } from 'typestyle';
 import { ExternalLinkAltIcon } from '@patternfly/react-icons';
 
 import history from 'app/History';
-import { NodeType, DecoratedGraphNodeData } from 'types/Graph';
+import { NodeType, DecoratedGraphNodeData, BoxByType } from 'types/Graph';
 import { JaegerInfo } from 'types/JaegerInfo';
 import { KialiAppState } from 'store/Store';
 import { Paths, serverConfig } from 'config';
@@ -60,14 +60,17 @@ export class NodeContextMenu extends React.PureComponent<Props> {
     let type: string | undefined = undefined;
     switch (node.nodeType) {
       case NodeType.APP:
-      case NodeType.BOX: // we only support app box node graphs, so treat like app
-        // Prefer workload type for nodes backed by a workload
-        if (node.workload && node.parent) {
-          name = node.workload;
-          type = Paths.WORKLOADS;
-        } else {
-          type = Paths.APPLICATIONS;
-          name = node.app;
+      case NodeType.BOX:
+        // only app boxes have full context menus
+        if (node.isBox === BoxByType.APP) {
+          // Prefer workload links
+          if (node.workload && node.parent) {
+            name = node.workload;
+            type = Paths.WORKLOADS;
+          } else {
+            type = Paths.APPLICATIONS;
+            name = node.app;
+          }
         }
         break;
       case NodeType.SERVICE:
@@ -112,6 +115,27 @@ export class NodeContextMenu extends React.PureComponent<Props> {
   }
 
   render() {
+    const isBox = this.props.isBox;
+    const title = isBox ? getTitle(isBox) : getTitle(this.props.nodeType);
+    const header: React.ReactFragment = (
+      <>
+        {title}
+        <div className={contextMenuHeader}>
+          {(!isBox || isBox === BoxByType.APP) && (
+            <>
+              <PFBadge badge={PFBadges.Namespace} style={{ marginBottom: '2px' }} />
+              {this.props.namespace}
+            </>
+          )}
+          {renderBadgedName(this.props)}
+        </div>
+      </>
+    );
+
+    if (this.props.isHover) {
+      return <div className={contextMenu}>{header}</div>;
+    }
+
     const linkParams = NodeContextMenu.derivedValuesFromProps(this.props);
 
     // Disable context menu if we are dealing with an aggregate (currently has no detail) or an inaccessible node
@@ -135,12 +159,7 @@ export class NodeContextMenu extends React.PureComponent<Props> {
 
     return (
       <div className={contextMenu}>
-        {getTitle(this.props.nodeType)}
-        <div className={contextMenuHeader}>
-          <PFBadge badge={PFBadges.Namespace} style={{ marginBottom: '2px' }} />
-          {this.props.namespace}
-          {renderBadgedName(this.props)}
-        </div>
+        {header}
         <hr />
         {menuOptions}
       </div>
--- a/src/components/CytoscapeGraph/ContextMenu/NodeContextMenu.tsx
+++ b/src/components/CytoscapeGraph/ContextMenu/NodeContextMenu.tsx
@@ -13,48 +13,40 @@ import { NodeContextMenuProps } from '../CytoscapeContextMenu';
 import { getTitle } from 'pages/Graph/SummaryPanelCommon';
 import { PFBadge, PFBadges } from 'components/Pf/PfBadges';
 import { renderBadgedName } from 'pages/Graph/SummaryLink';
+import { PFColors } from 'components/Pf/PfColors';
 
 type ReduxProps = {
   jaegerInfo?: JaegerInfo;
 };
 
-const graphContextMenuContainerStyle = style({
-  fontSize: 'var(--graph-side-panel--font-size)'
-  // textAlign: 'left'
+const contextMenu = style({
+  fontSize: 'var(--graph-side-panel--font-size)',
+  textAlign: 'left'
 });
 
-/*
-const graphContextMenuTitleStyle = style({
-  textAlign: 'left',
-  fontSize: '16px',
-  borderBottom: '1px solid black'
+const contextMenuHeader = style({
+  marginBottom: '2px'
 });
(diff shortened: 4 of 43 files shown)


leandroberetta:
@jshaughn I tested this PR and works well, looks promising and the future for the graph.

Some first comments:

When text is missing, far away from any namespace, there is no clue where to zoom in, we might need to display the namespace, it could be inside the namespace box, removing the workloads:

jay_no_ns

When hovering, the side panel looks like the following:

jay_hover

I'm not sure about the delay in the hovering, because to me it seems laggy. My opinion is that the hovering can be replaced with selection (clicking on the interesting workload, edge, etc).

jshaughn:
@leandroberetta Thanks for giving this a try!

> When text is missing, far away from any namespace, there is no clue where to zoom in, we might need to display the namespace, it could be inside the namespace box, removing the workloads:

I agree that there is more we can do around expand/collapse of boxing, and things like that, but I think it probably falls outside of this initial PR scope.  I think with this PR the user experience with a zoomed out graph is to use color to identify an area of the graph of interest, then to do one of: shift-zoom to enlarge the area | hover to see a short summary in the side panel of an element of interest | select an element of interest | graphHide to eliminate part of the graph (effectively zooming in) | manual zoom.

> I'm not sure about the delay in the hovering, because to me it seems laggy. My opinion is that the hovering can be replaced with selection (clicking on the interesting workload, edge, etc).

I agree that the new approach is to favor selection, and that is a main reason for eliminating the prior hover logic and its associated "flashing".   But I think with the zoomed out graph we still need to offer the user an easy way to identify an element of interest, and that is what the new [delayed] hover ability is trying to do.  The 500ms time was my first attempt at a delay that is long enough to allow the user to move the mouse over the graph without interruption, while still not lagging too much on an actual hover.  I can easily reduce the delay, I'll give it a try and make an update.

nrfox:
Nice to see this in action!

Just echoing what has been said above:
I would say +1 what Leandro said about keeping some kind of text to identify groups when zoomed out far enough like large namespace text.

I also agree about the hover delay. It makes the UI feel more laggy. I actually like the current hover behavior as I think it helps you quickly understand that parts of the graph are clickable and entices you a bit to interact more with the graph. Perhaps hovering to highlight could also integrate with grouping/labeling when zoomed out at a large distance. For instance, when zoomed out to where labels have been removed, the namespace boxes keep their text and highlight when hovered over instead of individual nodes allowing the user to click (or doubleclick) to zoom on a specific namespace.

There seems to be a bit of a stutter when zooming to the point where the labels are removed and then again when the badges are removed. The zooming action is smooth until this point is reach.

Will the new zoom settings be configurable from the UI? Different users may want different levels at which their graph gets the labels removed. If these are user specific they'll probably need to be stored client side or even in the react state like the display options.

jmazzitelli:
3 things - some I think @leandroberetta already reported.

1. Delay when hovering in a namespace box - the namespace box doesn't highlight until a second or two after the cursor enters the box
2. The side panel goes gray when hovering in an app box now (no matter the zoom level
zoom-2
)
3. When using the layouts 2 and 3 (cose-bilkent and cola), when you reach the zoom level that removes the labels, the graph radically re-organizes and changes the layout. In this animated video, I am zooming in and out one level (scrolling the mouse wheel one click up, then one click down, then one click up again, then one click down again - just zooming in and out a single zoom level) and you can see the locations of the boxes change. So you can't really follow a particular box or group of boxes because the locations jump on the screen.
zoom-1

jshaughn:
> Just echoing what has been said above: I would say +1 what Leandro said about keeping some kind of text to identify groups when zoomed out far enough like large namespace text.

I can say that I did try something like this, by keeping box labels in place longer, and trying to increase the sizing of the text (you have to sort or fight cytoscape, because cy is trying to scale down the text as you zoom out).  In the end I didn't like what I was seeing. The current idea is that you can just hover over the box and clearly see the side panel updated with the box info, in effect the side-panel becomes a fixed-location, readable label.

So the question is, why is that not working for you?  I think there is an adjustment to the new approach, we are very used to seeing labels, even when we can't read them.  Large labels may actually work against scale, as it will take up space.  Replacing box contents with a readable label may be a solution, but as I said to Leandro, I think some sort of box expand/collapse may be better as a separate effort.

> I also agree about the hover delay. It makes the UI feel more laggy. I actually like the current hover behavior as I think it helps you quickly understand that parts of the graph are clickable and entices you a bit to interact more with the graph. Perhaps hovering to highlight could also integrate with grouping/labeling when zoomed out at a large distance. For instance, when zoomed out to where labels have been removed, the namespace boxes keep their text and highlight when hovered over instead of individual nodes allowing the user to click (or doubleclick) to zoom on a specific namespace.

I'm going to reduce the hover delay and also maybe play with adding back a little bit of immediate hover change, but removing "flashing" was an intial goal.

> There seems to be a bit of a stutter when zooming to the point where the labels are removed and then again when the badges are removed. The zooming action is smooth until this point is reach.

Yes, I am not exactly sure if that can be avoided, zoom is very smooth (cy just manipulating the canvas) until a threshold is crossed.  At that point the labels either come or go, and that causes both a re-render (to add or remove the labels) and a re-layout (to accommodate either the space added or removed for labels).

> Will the new zoom settings be configurable from the UI? Different users may want different levels at which their graph gets the labels removed. If these are user specific they'll probably need to be stored client side or even in the react state like the display options.

They are ui-defaults that can be set in the CR.  As with most everything, we don't currently do user-specific settings.

jshaughn:
@leandroberetta , @nrfox , @jmazzitelli , thanks for runing this and for the comments.  Replies above and below...

> 1. Delay when hovering in a namespace box - the namespace box doesn't highlight until a second or two after the cursor enters the box

That seems longer than I would expect.  I'll test it but it should be more like 1/2 second hover delay.

> 2. The side panel goes gray when hovering in an app box now (no matter the zoom level)

This is actually intentional, although maybe not good.  Not just for app boxes, but whenever hovering the side panel header will display the hover element.  It will not display any other detail unless the element is selected.  To emphasize the hover info I grayed out the non-header portion of the side panel.  In some way we need to ensure that the hover info looks different, and is sort of prominent, and that the user doesn't think the side panel info is pending/loading/not showing up.

> 3. When using the layouts 2 and 3 (cose-bilkent and cola), when you reach the zoom level that removes the labels, the graph radically re-organizes and changes the layout. In this animated video, I am zooming in and out one level (scrolling the mouse wheel one click up, then one click down, then one click up again, then one click down again - just zooming in and out a single zoom level) and you can see the locations of the boxes change. So you can't really follow a particular box or group of boxes because the locations jump on the screen.

Yes, I understand what you are saying but I'm not sure there is anything we can do about it.  When the labels come and go the graph topology changes, due to the space taken by new labels, or the empty space left by the old labels.  When we do the re-layout the layout algorithm will put things where it sees fit.   The only way to avoid it (I think) would be to keep the labels but make them invisible, but that sort of defeats the purpose of scaling, which is (I think) to best use the space available.  That would leave a bunch of blank space that may be hard to explain.

lucasponce:
Some comments from my side.
I think that the threshold should be adjusted, for example, this default can't be read:
image

I'd take some criteria, like to adjust to the min font size used:
image

lucasponce:
The float side panel is nice:
image

But I'd suggest:
- Add a title, there is space and that would help to the user to identify the element (node/edge).
- Not sure if as a user I'd like the automatically hide, it's not obvious that the new float side panel only appear when there is no selection, I don't have a strong suggestion, may be UX may help here, personally I'd put visible the last hover until there is a selection, in that case, user may want to click. But I'm not 100% sure of my comment, either.

lucasponce:
On the implementation side, when the window resizes some events are not correctly handled and it creates some strange behaviours (you expect labels on/off but it happens the opposite):

image

Another case:

image

So, it seems related to some stress with the UI events.

jshaughn:
> I'd take some criteria, like to adjust to the min font size used

That's an interesting idea.  I wonder if instead of specifying zoom levels we should specify the desired minimum font.  I will look into that, it may be an easier task for users to configure a desired font than a zoom level.

jshaughn:
@lucasponce @leandroberetta @nrfox @jmazzitelli , this is ready for another round of review.  I think it incorporates the previous feedback.  Leandro and Nick both commented on maintaining labeling on boxes, but I think that fights against the original idea.  Instead, the hover target popop tries to be the way to identify unlabeled graph elements.  As I mentioned above, I think the use pattern would be to identify an area of interest, whether by color, a "find" highlighting, or whatever, and then either hover over it (and see the popup info), drag-zoom to quickly enlarge, select to see the side panel, or possibly double-click or use the context menu to further investigate.

I tried to address the "lag" complaint with a shorter hover delay, for me it still allows moving the mouse overthe graph without any flashing, but a quicker hover action.  And the hover now re-introduces more of the highlighting/dimming we had before.  The hover popup has, I think, been improved, better sizing, positioning, a title, and it interacts better with the side-panel Show/Hide.  The window resizing issue was a giant pita, but I think it is fixed.

I'm planning in showing this to Andrew (UX) at the next meeting. Any feedback in advance is appreciated, although if you are busy, not required.

lucasponce:
I've performed a quick smoke test.

I see a perf penalty when using zoom (from controls or selection).

image

Even when using the selection tool, the graph hangs for ~1s in that scenario:

image

The "Hover Target" wording is really confusing to me:

image

I'd suggest to just use "Namespace/Workload/Application/Service" etc, I think it's pretty implicit that float is coming from an hover event.

jshaughn:
> The "Hover Target" wording is really confusing to me

+1, I don't really like it, it was just a first attempt to add a title based on the initial comment.

As for the perf hit, I need to maybe see what you're seeing.  It would be expected to see a slight pause when crossing a threshold boundary,  as the graph needs to update and re-layout, but not on each zoom.

jshaughn:
@lucasponce I've pushed a new approach, and although still a little bit of a work in progress, I think this may get rid of the lag when crossing a zoom threshold.

jshaughn:
@lucasponce or any reviewer.  I am hopeful that this is now behaving pretty well.   Please give it a try and let me know your feedback.  Remember to pull the server pr as well.

jmazzitelli (changes requested):
For some reason, after I move the mouse around after deselecting a node, after a while that node gets re-selected (even though I didn't click the mouse at all):
Peek 2021-12-21 16-05

jshaughn:
> For some reason, after I move the mouse around after deselecting a node, after a while that node gets re-selected (even though I didn't click the mouse at all):

This is an unrelated issue, it's in master as well.   Clicking the graph (background) in the node graph should be disabled.

lucasponce:
I still see the performance penalty when using the selection tool.

- Select a group of namespaces a draw a more realistic graph.
- The graph holds 1-2s without progress indication:
image
image

This behavior is not on master where the selection tool for the same graph takes 0 seconds to adjust.

lucasponce:
Comments about the style:
image

"Namespace" and "Current Graph:" should have the same font size/style; otherwise, it looks inconsistent.

jshaughn:
> I still see the performance penalty when using the selection tool.

I didn't notice anything using a smaller drag.  I'll investigate.

> So, I'd reconsider just not using titles in both cases, I think the context is clear IMO and it would look consistent.

+1, titles should have the same styling.

jshaughn:
@lucasponce , I think I have addressed your latest issues.  The lag on drag-zoom should be gone.  I tried a few options for no titles vs. titles, for the hover-tip and the full side-panel.  I think I like consistent titles, and centered.  Also, edge titles now include the protocol.  Let me know what you think.  Finally, as an experiment I have changed the cytoscape-toolbar to get it out of the way and stop occluding the lower center of the graph.  I like it much better, but let me know what you think.

As for the Edge label text in the hover-tip, that's doable, but not simple, as that text is generated in a totally different place.  I'd suggest user's just select the edge if they need more detail.  Otherwise, maybe a separate jira for that feature (we'd have to take make the edge label generator a callable utility fn, or something like that).

Anyone else that has the inclination, feedback welcome.

lucasponce:
I'm testing the latest changes (not yet entered in the code) and the performance issue seems gone.
I see some style issues in the sidebar, as we may have 3 or 4 different styles and fonts for titles:
image

I think those should look similar (align and font size), otherwise, when you use it you perceive that those elements are not synced.

Another style issue is the badge in the labels:
image
It looks different from the badge in the sidebar.

The new legend bar height could also be adjusted in the bottom to open just after the legend button:
image

Also, I'd see the title style in the sidebar as it's a similar pattern.

Another weird style issue is that the width of the floating side panel is not uniform:
image

I guess it could look nicer if the width is fixed, it's noticeable when it changes.

jshaughn:
@lucasponce Thanks for the continued testing.  I'm planning to show to UX for feedback as well, here are some initial replies to your latest comments:

> I'm testing the latest changes (not yet entered in the code) and the performance issue seems gone.

Hooray, yes, this has been a battle.

> I see some style issues in the sidebar, as we may have 3 or 4 different styles and fonts for titles:

Not sure, maybe need to get together to make sure we have details here.  I did make the hover title consistent with the side-panel title, and made the positioning consistent.   I can see from the screenshot that the `Trace` header is not consistent.  The font is different (that has been the case for a while) but now also the positioning could change to be centered.   Not sure about what else you may be seeing.

> Another style issue is the badge in the labels ... It looks different from the badge in the sidebar.

The PF-badges in the box labels are different, they have less padding.  That is actually intentional and it has been that way for a while, but we can revisit.  The reason is practical, more narrow labels cause less overlap on graph elements.  If we want to be consistent then I would probably vote for the narrow badges everywhere in Kiali.  We can discuss with UX.  Also, I don't have an OS console at the moment, not sure if they are using more narrow badging that the default PF.

> The new legend bar height could also be adjusted in the bottom to open just after the legend button:

+1, I actually looked at this  but could not fine the correct place to make the adjustment :-)  I can revisit.

> Also, I'd see the title style in the sidebar as it's a similar pattern.

OK, I guess this PR could add that as well, since we are in there.

> Another weird style issue is that the width of the floating side panel is not uniform:

I actually like it better this way.  For me the uniform width was prone too too much whitespace.

leandroberetta:
@jshaughn I re tested this effort and looks great, the delay sensation is gone, and I guess the current delay is very cool and improves the experience using the graph.

The only thing I'm still thinking is that when you are far from any specific area, it's hard to target a namespace of interest:

image

It's true that you can go to the namespace directly using the filter, but I guess a lot of users will want to drill down direclty, and there is no clue on where to go.

The rest works super and is a huge improvement!

leandroberetta:
> The only thing I'm still thinking is that when you are far from any specific area, it's hard to target a namespace of interest:
> 
> image
> 

Dismiss this, I saw the side panel, that indicates in what namespace you are when hovering...

jshaughn:
> ...it's hard to target a namespace of interest:

True, when labeling is taken away it is not trivial to locate things by name.  You can always do a graph find like `ns=myNamespace`.  But going forward we may want to do some other things.  I had a proposal for a "namespace" graph type.  But also there are things like collapsing nodes, etc.

Thanks for the positive feedback, I'm still working on it, as UX (Andrew) was not thrilled with the fixed-position hover info.  Trying now a different approach...

jshaughn:
@lucasponce, @leandroberetta, I changed direction after the last UX meeting and abandoned the fixed-position hover info.  First, I tried positioning that same info/component closer to the node being hovered.  It looked OK but there were some positioning issues and it just looked sort of clunky to me.  So, I changed the approach and tried leveraging our existing contextMenu.  This already uses Tippy to get a nice popover style box with good positioning.  I sort of merged the look of my previous hover-info with the contextMenu, updating the contextMenu to be more consistent with our overall graph styling.

For me this is looking pretty good and behaving pretty well, I'd like your opinion.   The way it works is that a hover gives you contextInfo and a right-click gives you the traditional contextMenu (if that element actually has a contextMenu, some don't, like edges).

lucasponce:
My early feedback about the new UX change is pretty good, and I like the new style in both scenarios:
image
image

Perhaps the "Show" menu would need a minor style tweak as it looks as options are disabled, but that's a little comment.

I will perform more testing with large scenarios to check there are no side effects, but I think this work is pretty close.

lucasponce:
I'm not convinced about the style to center the titles; I like to see the same font (+1 on that), but I think that centering the headings gives the feeling that it provides a different style when navigating from overview or to the details.
So, I'd keep them on the left, and that would make feel the same style as the overview or details pages:
image
image

But now in the graph:
image
image

(Also, the "close" icon in the trace detail pushes the "centered" style a little bit on the left, and I guess it works better aligned on the left)

jshaughn:
@lucasponce Glad to hear this change of approach looks good to you.  Hopefully @leandroberetta will feel the same.

> Perhaps the "Show" menu would need a minor style tweak as it looks as options are disabled, but that's a little comment.

The context menu was actually using custom colors.  As part of the update I tried to convert to standard PF colors for Dropdown, but perhaps I got the wrong one.  I'll re-check.

> I'm not convinced about the style to center the titles...

I sort of like the centered titles and at least the graph was very consistent :-)  But I understand that it is now not consistent with Overview.  Also, looking at PF in general, it seems left-titles is more prevalent.  I will revert the centering.

jshaughn:
@lucasponce pushed changes.  The context menu options should look better now, and back to left-align.  As a note for CSS consistency, I only had to change one style to revert all of the titles from center to left.

lucasponce:
> @lucasponce pushed changes. The context menu options should look better now, and back to left-align. As a note for CSS consistency, I only had to change one style to revert all of the titles from center to left.

Oh, I think a push is missing, I fetch the PR and this is the latest commit I have:

```
commit 3de85cd42081ee8728469343d475045f26b4071a (HEAD -> pr2275)
Author: jshaughn <jshaughn@redhat.com>
Date:   Thu Jan 6 17:36:44 2022 -0500

    update snapshot

```

jshaughn:
@lucasponce Sorry about that, code pushed.  Also started some code cleanup for the PR, on the assumption that the latest approach will move forward.

lucasponce:
I hit a problem in the "minigraph", but it's also in master and reported previously 
https://github.com/kiali/kiali/issues/4589

I'm going to run another test in a clean environment, but it looks good to me.
Good job.

jshaughn:
@lucasponce I was unaware of the mini-graph bug, I'll assign to myself now, as I've been looking at the graph code.  Thanks for all the review, the code should now be in a reviewable state as well.

leandroberetta:
> @lucasponce Glad to hear this change of approach looks good to you. Hopefully @leandroberetta will feel the same.
> 

Same here, looks very nice @jshaughn.

jshaughn:
WIth respect to review, @lucasponce just asking you to complete/approve a functional review.  @leandroberetta requesting code review. Thanks!

lucasponce (approved):
All topics (performance, design, and style) discussed have been addressed.

nrfox:
Just leaving my 2c. The early performance issues are all gone and everything is very slick! I also like the new hover feature with the context manager. If you are new to the graph, it would even help you more quickly identify what the different shapes mean. Only a couple of things I'd mention.

I'm still not keen on the delay on hover although it's much less noticeable now. It just makes the graph feel less responsive to me and I didn't find the "flashing" a problem before. 

The only other thing I noticed was that clicking on a node disables the hover feature until a node is de-selected. I was expecting to still be able to hover over the other nodes and see the context menu even with a node selected.

Both minor things. All in all great feature that works very well!

jshaughn:
@nrfox thanks for the review.

> I'm still not keen on the delay on hover although it's much less noticeable now...

I keep playing with the delay.  I had it at 250ms bit currently it's upped to 300ms.  I will drop it down to 270 given your feedback.  Trying to find the sweet spot between unwanted hover and acceptably snappy.

> The only other thing I noticed was that clicking on a node disables the hover feature until a node is de-selected.

Hmmm, I'm not sure about that one.  Let me go back and play with that a little bit.  When a node is selected then unrelated nodes (those not involved in a request path involving the selected node) are dimmed.  I would think those should be ignored.  But maybe hover should still work for the non-dimmed nodes.  Not sure so probably will leave as-is for this round, and maybe revisit as more people play with it.

nrfox (approved):
+1 for functionality

leandroberetta (approved):
Although I'm not 100% familiar with the code, it looks good to me, I don't see any problems and seems clear. 

I tested the functionality several times during reviewing.

mattmahoneyrh (approved):

PR #2277, opened 2021-12-06
Enable cluster and namespace boxing by default, in the main graph.

jshaughn:
Part-of: https://github.com/kiali/kiali/issues/4547

Note that this change is in conjunction with a server-side change to omit boxing when unnecessary.

**Server PR:** https://github.com/kiali/kiali/pull/4548

### DIFF
--- a/src/pages/Graph/GraphToolbar/GraphSettings.tsx
+++ b/src/pages/Graph/GraphToolbar/GraphSettings.tsx
@@ -432,7 +432,12 @@ class GraphSettings extends React.PureComponent<GraphSettingsProps, GraphSetting
         labelText: 'Cluster Boxes',
         isChecked: boxByCluster,
         onChange: toggleBoxByCluster,
-        tooltip: <div style={{ textAlign: 'left' }}>When enabled the graph will box nodes in the same cluster.</div>
+        tooltip: (
+          <div style={{ textAlign: 'left' }}>
+            When enabled and there are multiple clusters, the graph will box nodes in the same cluster. The "unknown"
+            cluster is never boxed.
+          </div>
+        )
       },
       {
         id: 'boxByNamespace',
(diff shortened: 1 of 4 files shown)


israel-hdez (approved):
I can see the new defaults in the main graph, and I also see that the graph in detail pages is unaffected.

jshaughn:
Thanks for the review, @israel-hdez.  Adding Leandro and Lucas, as I think they both also participated on that Slack thread.  Want to make sure multiple folks think it's better.  The alternative is to keep the box reduction but not change the UI defaults.  Personally, I like the new approach.

lucasponce:
A side effect I see of this approach is that when the namespace boxes are enabled, there is always an overlap with the zoom/layouts/legend buttons:
image

And the user can't read the namespace label, and it should move the element manually.

lucasponce:
Another example that would need some work in the default layouts:

No need to go to a large scenario, with just 4 namespaces the boxes help, but it needs at least a better layout (as discussed in https://github.com/kiali/kiali/discussions/4518):

image

lucasponce (approved):
My comments are not blockers.
I think those are related to the discussion to study an "intelligent" layout to improve those visualizations.

jshaughn:
> there is always an overlap with the zoom/layouts/legend buttons:

This behavior has been there since we added the cytoscape toolbar.  And to be honest, I hate how often the legend obscures the graph (it's independent of boxing, layout, etc).  I think in another issue we should consider some alternatives.  For instance, maybe the cy toolbar should be vertical on the left side.  Or possibly have smaller buttons and have dedicated space.

jshaughn:
> Another example that would need some work in the default layouts:

The preference of dagre to stack things vertically often runs against what we'd like to see.  I wonder if we could contribute to that layout and add an option that better uses horizontal space.

PR #2280, opened 2021-12-22
Resolve tmpl 1.0.4 to 1.0.5, ws 5.2.2 to 5.2.3

lucasponce:
We have three places in the package.json to update dependencies ("dependencies", "resolutions", "devDependencies") and the possibility to update a dependency directly in the yarn.lock file (some of the bots try to do it in that way).

I used the "resolutions" for these couple of transitive dependencies,  but sometimes don't work well (on multiple nested dependencies, on this case it seems to work).


jshaughn:
I'm going to add assign this to QE as I think we can go straight to regression tests.

pbajjuri20:
Tested and Passed.

LGTM!

pbajjuri20 (approved):

PR #2281, opened 2022-01-07
Ensure workload is initialized with defaults expected in UI code.

jshaughn:
Fixes https://github.com/kiali/kiali/issues/4589

I wasn't able to reproduce the problem but it seems independent of the mini-graph and more about a workload being returned without any services attached.  Perhaps the workload detail was requested by navigating in the mini-graph (i.e. clicking on a different workload than the one currently selected), but the failure point was actually in rendering the Info section of the workload's Overview tab.

So, the theory here is that the server can return a workload with no services, but the UI expects an array (even if it has 0 length).  I suppose this could be fixed on the server, but since it's really a UI expectation, I thought the UI should ensure the workload is valid.  So, this PR makes sure the fetched workload is first filled with default values.

Note that first I thought maybe the API call was returning the wrong type, and I noticed that may of our client API functions didn't define a return type.  So I added that as well.  This resulted in a couple of tweaks unrelated to the core issue, the main change is in WorkloadDetailsPage.

@lucasponce Perhaps you can try and recreate the issue, or supply repro steps I can try.  Maybe it has to do with a specific demo's workload.

### DIFF
--- a/src/pages/WorkloadDetails/WorkloadDetailsPage.tsx
+++ b/src/pages/WorkloadDetails/WorkloadDetailsPage.tsx
@@ -3,7 +3,7 @@ import { connect } from 'react-redux';
 import { RouteComponentProps } from 'react-router-dom';
 import { EmptyState, EmptyStateBody, EmptyStateVariant, Tab, Title } from '@patternfly/react-core';
 import * as API from '../../services/Api';
-import { Workload, WorkloadId } from '../../types/Workload';
+import { emptyWorkload, Workload, WorkloadId } from '../../types/Workload';
 import WorkloadInfo from './WorkloadInfo';
 import * as AlertUtils from '../../utils/AlertUtils';
 import IstioMetricsContainer from '../../components/Metrics/IstioMetrics';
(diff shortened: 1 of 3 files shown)


lucasponce:
I'm reviewing this PR and checking the problem, I guess that the main root cause is a "promise" not well caught in the workload page in certain cases.
I'll update my review when I have more information, perhaps we don't need to refactor the API.ts file just for this issue.

jshaughn:
@lucasponce If this PR missed the root cause we can close it.  Or, feel free to modify it with a sub-pr.

lucasponce:
I think the issue is due to a non-expected null when a workload has no services.
I've created a one-liner PR https://github.com/kiali/kiali-ui/pull/2283 perhaps it's enough to fix this one.
I'll close it and I put you as reviewer @jshaughn, let me know if you would like to keep this one, but I guess that wouldn't be necessary at least for this issue.

jshaughn:
> I think the issue is due to a non-expected null when a workload has no services.

Right, that was the root cause I fixed as well, but I did it ui-side, by setting defaults, to ensure that any similar issue would be caught as well.  As I mentioned above, a server-side fix is also possible.  So your approach is fine, although a little less robust as it is a single-point fix.  Note, this PR used `emptyWorkload` from `types/Workload`, which I think was the only use of that structure,  It was previously (and so now again) unused.

PR #2282, opened 2022-01-10
Prevent fast-click in the Graph Display settings dropdown.

jshaughn:
The overapping graph processing can cause a problem.
- Add the same protection to the Traffic settings dropdown.
- Fix the actual button disabling as well.

Fixes https://github.com/kiali/kiali/issues/4638

I don't love the menus now get grayed out during graph refresh, but I guess it make sense to prevent fast-click and to finish one update before starting another.

cc https://github.com/Xunzhuo

### DIFF
--- a/src/components/ToolbarDropdown/ToolbarDropdown.tsx
+++ b/src/components/ToolbarDropdown/ToolbarDropdown.tsx
@@ -80,6 +80,7 @@ export class ToolbarDropdown extends React.Component<ToolbarDropdownProps, Toolb
           return (
             <SelectOption
               key={key}
+              isDisabled={this.props.disabled}
               isSelected={key === String(this.props.value || this.state.currentValue)}
               value={`${key}`}
             >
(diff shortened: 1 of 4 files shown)


nrfox:
> I don't love the menus now get grayed out during graph refresh, but I guess it make sense to prevent fast-click and to finish one update before starting another.

Agreed. Perhaps an alternative would be to queue updates if a graph refresh is in progress and apply them once the in progress refresh finishes. I have no idea how much work that would be to implement so this way seems fine for now.

nrfox (approved):

Xunzhuo:
Thanks @jshaughn 👍

jshaughn:
> Perhaps an alternative would be to queue updates if a graph refresh is in progress and apply them once the in progress refresh finishes. I have no idea how much work that would be to implement so this way seems fine for now.

I had the same thought and decided it was tricky and maybe not worth the bang-for-the-buck.  Another option could have been to just close the menu on any click, but that seemed a little draconian as I often find myself wanting to make more than one change in one go.  Yet another approach would be to batch the changes, a-la the Namespace selection, and apply all when the menu is closed.  In the end it seemed like the easy way was the the best way, at least as a first offering.

mattmahoneyrh (approved):

PR #2284, opened 2022-01-12
Remove Iter8 extension

lucasponce:
Fixes https://github.com/kiali/kiali/issues/4643 but it requires a backend PR.
Note that a new model for extensions will be defined in https://github.com/kiali/kiali/issues/4642.


src/components/Nav/RenderPage.tsx
```diff
@@ -19,16 +18,7 @@ class RenderPage extends React.Component<{ isGraph: boolean }> {
...
-        // Extensions are conditionally rendered
-        if (route.path.includes('iter8') && serverConfig.extensions!.iter8.enabled) {
-          return true;
-        }
-        return false;
-      })
-    );
-    return this.renderPaths(allPathRoutes);
+    return this.renderPaths(pathRoutes);
   }
```
jshaughn, line 22: Could just get rid of renderPathRoutes().

jshaughn (commented):
One trivial comment.  Looks good, need server PR to review and sanity check.

jshaughn (approved):
Built and sanity checked with operator, server and ui prs.  One note, with the underlying removal of the current extension support (as part of the iter8 removal, as it was the only extension), I think there may still be some removable extension-related code in BreadcrumbView.  Approving, but take a look...

lucasponce:
Thanks. Added to QE flow, it's good that they could review it for regressions-free.
There are too many changes that may break some scenarios.

pbajjuri20 (approved):

PR #2285, opened 2022-01-12
Clip long numbers in heatmaps, as opposed to wrapping.

jshaughn:
Fixes https://github.com/kiali/kiali/issues/4588

This issue title suggests a different font size, but a smaller font can become unreadable.  Clipping gives a uniform look while still allowing enough space for significant digits, and the tooltip can be used to see detail, if desired.

### DIFF
--- a/src/components/HeatMap/HeatMap.tsx
+++ b/src/components/HeatMap/HeatMap.tsx
@@ -22,12 +22,14 @@ type Props = {
 };
 
 const cellStyle = style({
-  overflow: 'hidden',
-  fontSize: '.7rem',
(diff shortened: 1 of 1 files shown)


lucasponce (approved):
LGTM

jshaughn:
@pbajjuri20 If you can't generate a large number I think it's OK if you just ensure the heat maps still look fine in a scenario you have (i.e. no regression).  @lucasponce and I both tested with large numbers and I don't have a simple scenario for you to recreate (I used fake data).

pbajjuri20 (approved):
LGTM!

PR #2288, opened 2022-01-19
Migrate cose and cola layouts to grid and concentric algorithms

lucasponce:
Part of https://github.com/kiali/kiali/issues/4601

~~This draft PR is used as a playground to test new layout algorithms and tweak the configuration for existing ones.~~

~~There is no intention to promote them to a formal PR, just to share some experiments and collect feedback before moving in any direction.~~

It includes a couple of minimal changes that I think will be consolidated:

- Move the BoxLayout for Namespaces to "dagre" algorithm. As it's defined per app, it's a minimal change that it has a good impact when managing large scenarios.
- In large scenarios, the "cola" algorithm has a performance penalty compared with others, so that one could be a candidate to be replaced.
- "Grid" layout works well on a high population of namespaces; results look promising.
- Still doing tests on scenarios that have a single namespace with a high population of workloads, "dagre" works well in most cases, but at some point, it seems there is space to change the layout.

### DIFF
--- a/src/components/CytoscapeGraph/CytoscapeGraphUtils.ts
+++ b/src/components/CytoscapeGraph/CytoscapeGraphUtils.ts
@@ -120,6 +120,7 @@ export const runLayout = (cy: Cy.Core, layout: Layout): Promise<any> => {
       ...layoutOptions,
       name: 'box-layout',
       appBoxLayout: 'dagre',
+      namespaceBoxLayout: 'dagre',
       defaultLayout: layout.name
     });
   } else {
--- a/src/components/CytoscapeGraph/graphs/ConcentricGraph.ts
+++ b/src/components/CytoscapeGraph/graphs/ConcentricGraph.ts
@@ -1,7 +1,7 @@
-export class CoseGraph {
+export class ConcentricGraph {
   static getLayout() {
     return {
-      name: 'cose-bilkent',
+      name: 'concentric',
       animate: false,
       fit: false,
       nodeDimensionsIncludeLabels: true
--- a/src/components/CytoscapeGraph/graphs/GridGraph.ts
+++ b/src/components/CytoscapeGraph/graphs/GridGraph.ts
@@ -1,12 +1,11 @@
-export class ColaGraph {
+export class GridGraph {
   static getLayout() {
     return {
-      name: 'cola',
+      name: 'grid',
       animate: false,
       fit: false,
-      flow: { axis: 'x' },
       nodeDimensionsIncludeLabels: true,
-      randomize: false
+      padding: 0
     };
   }
 }
--- a/package.json
+++ b/package.json
@@ -68,8 +68,6 @@
     "cy-node-html-label": "2.0.0",
     "cytoscape": "3.15.5",
     "cytoscape-canvas": "3.0.1",
-    "cytoscape-cola": "2.3.1",
-    "cytoscape-cose-bilkent": "4.1.0",
     "cytoscape-dagre": "2.2.2",
     "cytoscape-popper": "1.0.7",
     "deep-freeze": "0.0.1",
--- a/src/components/CytoscapeGraph/CytoscapeReactWrapper.tsx
+++ b/src/components/CytoscapeGraph/CytoscapeReactWrapper.tsx
@@ -4,17 +4,13 @@ import * as Cy from 'cytoscape';
 import { GraphStyles } from './graphs/GraphStyles';
 import canvas from 'cytoscape-canvas';
 import cytoscape from 'cytoscape';
-import cycola from 'cytoscape-cola';
 import dagre from 'cytoscape-dagre';
-import coseBilkent from 'cytoscape-cose-bilkent';
 import BoxLayout from './Layout/BoxLayout';
 import popper from 'cytoscape-popper';
 const nodeHtmlLabel = require('cy-node-html-label');
 
 cytoscape.use(canvas);
-cytoscape.use(cycola);
 cytoscape.use(dagre);
-cytoscape.use(coseBilkent);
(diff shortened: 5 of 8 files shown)


lucasponce:
Some screenshots about the new algorithms in a medium/high populated scenario.
- Dagre layout (default):
image

- Grid layout (new algorithm being evaluated):
image

On high populated namespaces, the grid helps to organize the namespaces better. The BoxLayout keeps the Dagre distribution inside the namespace.

On a high populated single namespace or multiple workloads but with no boxing, the Grid and Concentric may help visualize these cases better.
- Dagre layout (default):
image

- Grid layout:
image

- Concentric layout:
image

As discussed in the working sessions, these are just quick experiments with the default settings of the algorithms, but it seems that replacing the "cose-bikent" and "cola" with the "grid" and "concentric" looks a good direction.

Also, exploring combining different layouts, one for the elements and a different one inside the BoxLayout, looks interesting.

We have commented that we could provide some "heuristic function" to hint at which scenarios may look better with one of the three options (Dagre/Grid/Concentric) and make this recursive inside the box.

Combined this with the current optimization of the labeling increases a lot the UX experience:
image

And with a one-click zoom-in user can visualize all labeling:
image

There is still some work to do; with no refresh, there are some nondeterministic rendering when changing from one layout to another (the expectation would be that with the same data, the layout should keep consistent to some degree).

lucasponce:
I have started to investigate to auto-detect the best layout for given scenarios. Still, results are not yet noticeable and there are too many scenarios that would need to wait until we can have a good mockup to simulate larger clusters.

I'm going to promote this PR to consolidate three changes that I guess will benefit the work:
- Add "Dagre" as the default layout for namespaces.
- Migrate "Cola" / "Cose" to "Grid" / "Concentric" as they show better results in large scenarios.

There is an issue that I couldn't find the root cause; I think it's related to the work in the Cytoscape events (cc @jshaughn): sometimes changing the layout shows labels on/off in a no clear deterministic way to reproduce it. So I guess it's related to the async calls and the events to react to those.

I think it's unrelated to this PR, and these changes may consolidate to help progress in that area as well.

src/components/CytoscapeGraph/CytoscapeGraphUtils.ts
```diff
@@ -108,6 +108,7 @@ export const runLayout = (cy: Cy.Core, layout: Layout): Promise<any> => {
       ...layoutOptions,
       name: 'box-layout',
       appBoxLayout: 'dagre',
+      namespaceBoxLayout: 'dagre',
```
jshaughn, line 111: We may want to add logic here.  We know  'dagre' fails when there are a lot of orphans (or possibly when not showing edges, based on the stuff I'm looking into).  In that case we may want to set `namespaceBoxLayout: 'grid'`.   For now this is still an improvement.

src/components/CytoscapeGraph/graphs/ConcentricGraph.ts
```diff
@@ -1,8 +1,7 @@
-export class CoseGraph {
+export class ConcentricGraph {
   static getLayout() {
     return {
-      name: 'cose-bilkent',
-      animate: false,
```
jshaughn, line 5: I know 'animate' defaults to false, but I'm not sure it's a bad idea to make our intent explicit.   Not a blocker.

src/components/CytoscapeGraph/graphs/GridGraph.ts
```diff
@@ -0,0 +1,10 @@
+export class GridGraph {
+  static getLayout() {
+    return {
+      name: 'grid',
+      fit: false,
+      nodeDimensionsIncludeLabels: true,
+      padding: 0
```
jshaughn, line 7: maybe add `animate: false` to show explicit intent.  Not a blocker.

jshaughn (approved):
A couple of non-blocking comments.  I'd like to see this merged quickly, so that I can rebase on top of it as I look into #4666.  Cose-Bilkent seems to have some odd behavior, so I'd like to see it go away.

src/components/CytoscapeGraph/graphs/ConcentricGraph.ts
lucasponce: +1 I'll add it.

src/components/CytoscapeGraph/graphs/GridGraph.ts
lucasponce: +1 I'll add it.

jshaughn:
@mattmahoneyrh would like to see this one get priority if possible. Thanks!  Jay

lucasponce:
Side comment.
During this work, I was playing with a clone of the Grid algorithm, as I think that at some point that one could be enhanced for some specific Kiali needs (I'm still playing with it).

The code is under this gist https://gist.github.com/lucasponce/e4d4fc757135796e2d96f816343adc59.

This PR uses the algorithm with no modification, so there was no need to add a clone to it, but it's something to consider for the future.

I comment on it as the cloned algorithm helped to "debug" the async behavior and how "Cytoscape" handled the events.

lucasponce:
@jshaughn @prachiyadav I've rebased with the master branch and resolved a couple of minor conflicts.
Let me know if things look good.

Note that there is some follow-up (and experimental) work in https://github.com/kiali/kiali-ui/pull/2298 but this PR is good to be consolidated no matter that we refine these algorithms in the next steps.

jshaughn:
@lucasponce This still looks good to me.  With smaller meshes the layouts are at least as good, if not better than what we had before.  And with larger meshes things are definitely better.

prachiyadav (approved):
Regression test executed successfully

PR #2289, opened 2022-01-19
New edge hiding options and other changes related to graph at scale

jshaughn:
Relates to https://github.com/kiali/kiali/issues/4610

This converted draft PR introduces various changes to help users use the graph at scale.  In particular, how to let users de-clutter the graph, or more easily focus on elements of interest.

**Changes in this PR:**
- add two new options to the cytoscape toolbar:
  - show no edges
  - show unhealthy edges only
- remove the two zoom options from the cytoscape toolbar
  - these are likely never used by anyone and removing them lets us maintain the size of the cy toolbar
  - in general zoom is done via the mouse wheel, or drag-zoom
- do not show popup contextInfo when labels are active
  - this both reduces noise and stops it from covering the actual label.
- on hover and selection, increase the dimming of unhighlighted nodes and edges ((i.e. make them less visible)
  - the intent is to make the highlighted nodes and edges more prominent, and reduce the busy look of large graphs.
-  on hover, highlight the full E2E paths
  - the extended highlighting is now the same as selected-node-highlighting, offering a more consistent UX
- calculate edge health once, at fetch time, instead on every cy render
  - this is cleaner, faster, and lends itself better to server-side health assignment
  - this has no visible change to the user (unless they can notice a slightly snappier UX)

**QE Notes**
This doesn't yet have dev approval, but I expect it's coming soon, I added it to your project so that it's on your radar.


jshaughn:
@lucasponce, perhaps related to the layout investigation is the fact that certain options/settings may do better with different default layouts.  For example, in a local investigation around edge removal, the cose-bilkent looks pretty good but dagre is terrible.  Cose:

image

leandroberetta:
> @lucasponce, perhaps related to the layout investigation is the fact that certain options/settings may do better with different default layouts. For example, in a local investigation around edge removal, the cose-bilkent looks pretty good but dagre is terrible. Cose:
> 
> image

This is good for a kind of dashboard, to check the health of applications.

jshaughn:
@lucasponce, @leandroberetta, Here is a working example of being able to show edges (our usual default), no edges, or just unhealthy edges.  Note that this works best with cose-bilkent.  I have tried it against Lucas' layout PR, with grid and concentric, but neither does well with no edges, falling into the same stacked approach as dagre.  I expected grid to do well, but it may need some changes.  One option is to keep cose-bilkent, and force it for these options.  Note that in the PR, the decision to force dagre for laying out namespace-box content, is not going to work well with these options.  For the PoC I just have more options on the cytoscape toolbar, not sure how we would want to to potentially offer these options in general.

kiali#4610-1

lucasponce:
> Note that in the PR, the decision to force dagre for laying out namespace-box content, is not going to work well with these options. For the PoC I just have more options on the cytoscape toolbar, not sure how we would want to to potentially offer these options in general.

Nice progress. Yes, I think that we'll need more iterations on this, that's why I think that the work in https://github.com/kiali/kiali-ui/pull/2298 will be beneficial.

I have a playground with a custom "breadthfirst" algorithm, and also I see that the integration between the box + main algorithm needs some tweak, so I'm going to continue working on that PR as I guess it will give us a better position to adjust these scenarios.

jshaughn:
@lucasponce , Playing with this some more, I like having the ability to quickly change edge and layout options in the cytoscape toolbar.  I may just look for better icons.  I don't like that the cy toolbar is growing fairly large, we could consider removing the zoom-in and zoom-out options, in favor of just drag-zoom and mouse-wheel-zoom.

As for the layouts, I think we could start applying some basic logic to the layout decisions we make.  As mentioned above, we could keep cose-bilkent specifically for when the edge-removal options are active, or something like that.  I think that could give us the ability to add something like what's in this POC, in the near term.

lucasponce:
> Playing with this some more, I like having the ability to quickly change edge and layout options in the cytoscape toolbar. I may just look for better icons. I don't like that the cy toolbar is growing fairly large, we could consider removing the zoom-in and zoom-out options, in favor of just drag-zoom and mouse-wheel-zoom.

+1

> As for the layouts, I think we could start applying some basic logic to the layout decisions we make. As mentioned above, we could keep cose-bilkent specifically for when the edge-removal options are active, or something like that. I think that could give us the ability to add something like what's in this POC, in the near term.

Well, this is what I'm working on https://github.com/kiali/kiali-ui/pull/2298 it's still WIP, the BoxLayout doesn't work well with the grid and concentric, but it's the area that I'm actively working on. I'm not sure if I'd prefer to bring the "cose-bilkent" back. I'd prefer to wait if we can get a better control of the algorithms.

leandroberetta:
Good progress!.

> As for the layouts, I think we could start applying some basic logic to the layout decisions we make. As mentioned above, we could keep cose-bilkent specifically for when the edge-removal options are active, or something like that. I think that could give us the ability to add something like what's in this POC, in the near term.

I like the idea of a "dynamic layout" according on what the user is doing. Personally, I didn't use that much the other layouts.

jshaughn:
I'm going to convert this from Draft.  The results of the initial investigation look promising.  We will need decent layout support for the the new edge options.  That work is under way, so I'll also mark as DNM.

lucasponce:
Hi @jshaughn, I've started to take a look at the work of this PR.
I have a first comment idea that I guess will also affect the work related to the layouts.

I wonder if the strategy to hide the edges could be to apply a "transparent/hidden" style instead to remove the edges from the Cytoscape model.

Basically, the layouts take as an input the edges to calculate the distributions, then removing them will create scenarios as the "stack" case in the "concentric":

image

Perhaps exploring the "styles" on the edges may help to keep the original distribution, but just "hide" the connections.

Otherwise Cytoscape searches (like the ones used by the "dagre", "concentric" or the "breadth-first" algorithms won't work well).

Probably a "quasi-transparent style" on the edges when the option is selected can create the effect we are looking for (to reduce the noise of the connections but keep the relations provided by the edges).

jshaughn:
I think the value of removing the edges is the fact that you could get the compressed view of things, like in my screenshot above (using cose-bilkent), where you end up with a very dashboard-like view.    I know it's desirable to remove CB, and its dependencies,  but if we can't get a different layout to behave well with these new options then maybe we should keep it for this one option.

Having said that, it's probably not obvious, but you can get the behavior you want by disabling the "compress on hide" option in the Display menu.

lucasponce:
> Having said that, it's probably not obvious, but you can get the behavior you want by disabling the "compress on hide" option in the Display menu.

I'll keep reviewing this, it has some overlap with the PR https://github.com/kiali/kiali-ui/pull/2298 and probably we may need to adjust some behavior, like to select a specific layout algorithm when no edges and set some fixed parameters to look well.

I'm currently working on the 2298 but I'll be back to the PR review with some proposals.

nrfox:
Nice to play with some of these features.

The edge hiding on hover/selection does a good job of reducing visual noise especially for large graphs. I'm not sure I like it enabled by default though. It is a large change in behavior from the previous dimming behavior and imo people generally want their UIs to be familiar between versions. At least once I was wondering where one of my nodes went when I realized that my mouse was resting on a different node causing my other nodes to be hidden. Perhaps the edge hiding on hover/selection could be a display option but not enabled by default?

The additional edge hide buttons on the cyto toolbar seem to duplicate the functionality of the hide box and the display options. There's now three different places to change the display properties of nodes/edges: display options, hide box, cyto toolbar. It's definitely more convenient to just press a button on the cyto toolbar but the other cyto toolbar options are specific to the graph itself and do not deal with edges/nodes like "enable drag", or "zoom to fit". I think changing the edge behavior would fit more naturally in display options or as a predefined hide query.

nrfox:
50ish apps with grid layout and edges hidden:
Screenshot from 2022-02-16 08-48-52

nrfox:
> I'm reluctant to keep adding more options to our large number of current options.

Why is that? Shouldn't users be able to configure the graph in whatever way best fits their needs? I know adding any config option adds some overhead but it also gives more control for those who want it. It makes me wonder who the graph should be optimized for by default. For example, some layouts or options might make larger scale graphs look better but at the cost of making the smaller graphs look worse. Perhaps an "auto layout" feature could help choose good defaults for either in the future.

lucasponce:
The work is good. We are adding new usability features that will help to manage large scenarios.
In detail, I see that this PR's work is also close to the https://github.com/kiali/kiali-ui/pull/2298 and I guess we need to combine them.

When no edges, there are situations that don't look good and we'll need to play with those scenarios.

I'm fine with consolidating this first, and then I can rebase and apply the #2298, and when all together we can work on the adjustments.

Good thing is that the default layouts will remain intact and these can be seen as new features that will help users to navigate better.

jshaughn:
@nrfox , @jmazzitelli, you both had reservations about the change related to hover/selection, where I changed from "dimming" unhighlighted elements to hiding them completely.  So, I partially reverted the change, I returned to dimming, so that you don't completely lose sight of the whole graph structure, but I further reduced the opacity to make it even lighter, helping to further highlight the nodes and edges of interest.  I kept the change to have hover and selection provide the full e2e paths.

jshaughn:
@nrfox @lucasponce If you add an approval I think we can move this forward, because we need it to work well with https://github.com/kiali/kiali-ui/pull/2298 and it will be easier to incrementally change things if it's merged.

src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -513,7 +516,15 @@ export default class CytoscapeGraph extends React.Component<CytoscapeGraphProps>
       CytoscapeGraph.mouseInTimeout = setTimeout(() => {
         // timer expired without a mouseout so perform highlighting and show hover contextInfo
         this.handleMouseIn(cytoscapeEvent);
-        this.contextMenuRef!.current!.handleContextMenu(cytoscapeEvent.summaryTarget, true);
+
+        // if we are not showing labels (due to zoom level), show contextInfo
+        const zoom = cy.zoom();
+        const noLabels = this.zoomThresholds!.some(zoomThresh => {
```
nrfox, line 522: Is `this.zoomThresholds` always going to be set to something? This is defaulted on the backend even when not explicitly set?

lucasponce (approved):
After the discussion with @andrew-ronaldson, I don't remember if we commented to bring back the zoom buttons.

In any case, I see this PR work and #2298 related so consolidating in master and iterating on them sounds good to me.

I've tested it locally and it worked fine (with all the previous comments from this PR).

About the code, I haven't reviewed in detail as when rebasing we may need some adjustments to resolve conflicts.

nrfox (approved):
I really like the new hover behavior. It does a much better job of highlighting nodes/edges.

My only gripe is the new edge hide buttons but that's not a blocker.

jshaughn:
> after the discussion with @andrew-ronaldson, I don't remember if we commented to bring back the zoom buttons.

My impression was that they could be removed.

jshaughn:
> I really like the new hover behavior. It does a much better job of highlighting nodes/edges.

Glad to hear it, perhaps the more aggressive dimming was the best approach.  We'll see what others have to say, like @jmazzitelli.

> My only gripe is the new edge hide buttons but that's not a blocker.

There will still be time to revisit this before it's set in stone for SM 2.2.  My biggest concern at the moment is that we don't have a layout that works well with the new option, but that is under active development.    Again, we have a little time for people to use and comment, perhaps the convenience of the buttons will be seen as a confusing addition due to the overlap with the standard hide feature.

lucasponce:
Folks, this task is blocking some follow-up work as it needs to be rebased and adjusted.

How can we proceed with this?

jshaughn:
> Folks, this task is blocking some follow-up work as it needs to be rebased and adjusted.

@lucasponce This is one that is blocked by lack of QE resources and is one we will need to move forward with dev-as-qe, I think.  We may need to ask for a volunteer here.

nrfox (approved):
Tested with OSSM 2.1 with upstream kiali/kiali-operator and bookinfo deployed:
    - The two new edge graph buttons
        - show no edges
        - show unhealthy edges only
    - Confirmed the zoom buttons are removed.
    - The contextinfo popup does not show up when labels are active* (see note)
    - Confirmed the dimming is increased on hover and selection of nodes and edges
    - Confirmed the full E2E paths are highlighted on hover
    - Tested on each type of graph (workload, app, service, versioned app)
    - Visited each tab (workloads, services, etc.)
    - Created a routing object through kiali wizard
    - Viewed pod logs

The only minor issue I found was that the contextinfo popup sticks around until you exit the box when you go from zoomed out view without labels to a zoomed in view with labels. See behavior. Sorry for the weird flashing to my desktop background. Not sure why my screen recording is doing that:
popup_stays

Either way, I don't see that issue as a blocker here.

nrfox:
Note the issue above with the contextmenu sticking around goes away when you hover outside of the namespace box.

skondkar:
Polarion test case >> SWSQE-2547

PR #2290, opened 2022-01-20
Protect graph-hide from unwanted zoom events

jshaughn:
Fixes https://github.com/kiali/kiali/issues/4666

Graph-hide performs a layout after removing the graph elements.  Layouts
can generate intermediate zoom-level changes that should be ignored, and
so we must add that protection.

### DIFF
--- a/src/components/CytoscapeGraph/CytoscapeToolbar.tsx
+++ b/src/components/CytoscapeGraph/CytoscapeToolbar.tsx
@@ -38,6 +38,7 @@ type ReduxProps = {
 
 type CytoscapeToolbarProps = ReduxProps & {
   cytoscapeGraphRef: any;
+  disabled: boolean;
 };
 
 type CytoscapeToolbarState = {
@@ -143,11 +144,12 @@ export class CytoscapeToolbar extends React.PureComponent<CytoscapeToolbarProps,
               id="toolbar_layout_default"
               aria-label="Graph Layout Default Style"
               className={buttonStyle}
-              variant="plain"
-              onClick={() => {
-                this.props.setLayout(DagreGraph.getLayout());
-              }}
               isActive={this.props.layout.name === DagreGraph.getLayout().name}
+              isDisabled={this.props.disabled}
+              onClick={() => {
+                this.setLayout(DagreGraph.getLayout());
+              }}
+              variant="plain"
             >
               <TopologyIcon
                 className={this.props.layout.name === DagreGraph.getLayout().name ? activeButtonStyle : undefined}
@@ -163,11 +165,12 @@ export class CytoscapeToolbar extends React.PureComponent<CytoscapeToolbarProps,
                 id="toolbar_layout1"
                 aria-label="Graph Layout Style 1"
                 className={buttonStyle}
-                variant="plain"
-                onClick={() => {
-                  this.props.setLayout(CoseGraph.getLayout());
-                }}
                 isActive={this.props.layout.name === CoseGraph.getLayout().name}
+                isDisabled={this.props.disabled}
+                onClick={() => {
+                  this.setLayout(CoseGraph.getLayout());
+                }}
+                variant="plain"
               >
                 <TopologyIcon
                   className={this.props.layout.name === CoseGraph.getLayout().name ? activeButtonStyle : undefined}
@@ -183,11 +186,12 @@ export class CytoscapeToolbar extends React.PureComponent<CytoscapeToolbarProps,
               id="toolbar_layout2"
               aria-label="Graph Layout Style 2"
               className={buttonStyle}
-              variant="plain"
-              onClick={() => {
-                this.props.setLayout(ColaGraph.getLayout());
-              }}
               isActive={this.props.layout.name === ColaGraph.getLayout().name}
+              isDisabled={this.props.disabled}
+              onClick={() => {
+                this.setLayout(ColaGraph.getLayout());
+              }}
+              variant="plain"
             >
               <TopologyIcon
                 className={this.props.layout.name === ColaGraph.getLayout().name ? activeButtonStyle : undefined}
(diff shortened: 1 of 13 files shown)


lucasponce:
Interesting. In the context of the https://github.com/kiali/kiali-ui/pull/2288 work, I've also noticed some weird behavior with the zoom events when changing from one layout to another. 

I'm curious if those are also related to the same problem when merged I can rebase and comment as well.

israel-hdez (approved):
Since the cose layout is going to be removed, I think this is OK.
But worth to check if the new layout that is going to be introduced does not have the issue.

jshaughn:
@israel-hdez, @lucasponce:  Edgar had reviewed this before and had managed to still have a problem with Cose-Bilkent.  I'm not sure his recreation was limited to CB, because the more I looked at things the more I could cause problems.  Especially with graph hide enabled.   I have been playing with things for days, and have actually found a variety of issues, a couple of which have been around for a long time.  I have made a variety of changes and ask for a re-review from Edgar, as well as a review from Lucas.  I don't see any more browser hanging, but to be honest, I still see some odd label/zoom behavior at times.  I have run out of ideas for fixing it.  Issues tend to happen when changes in layout, or perhaps a graph hide, take place very near to one of our labeling thresholds.

So, try to break things, change layouts, zoom, drag-zoom, add a graphHide, remove a graphHide, resize, whatever; if you see something really bad we obviously need to fix it.  But if you see just a glitch here and there, I think maybe we need to consider leaving it alone for now.

I should note that in some cases I think I was able to make things better than before, especially when using graph hide.  Before, on each refresh you could see the graph layout slightly differently, even when nothing else seemingly changed.  That should no longer happen.  Also, I protected against a layout change during a refresh.  And there were also a couple of slight perf improvements resulting from some fixes.  Probably not noticeable in a small topology, but every bit helps.

I will be out for a few days, feel free to do what you like with this.  I think it's quite important for 1.45, and it should be able to backport directly.  For master, it will have conflicts with Lucas' layout change PR.

src/components/CytoscapeGraph/CytoscapeToolbar.tsx
```diff
@@ -183,11 +186,12 @@ export class CytoscapeToolbar extends React.PureComponent<CytoscapeToolbarProps,
               id="toolbar_layout2"
               aria-label="Graph Layout Style 2"
               className={buttonStyle}
-              variant="plain"
+              isActive={this.props.layout.name === ColaGraph.getLayout().name}
+              isDisabled={this.props.disabled}
               onClick={() => {
-                this.props.setLayout(ColaGraph.getLayout());
+                this.setLayout(ColaGraph.getLayout());
```
israel-hdez, line 192: Was calling directly `this.props.setLayout` doing a re-render?
If true, something may be weird with redux, because an unchanged redux state should not trigger the re-render :-/

israel-hdez:
@jshaughn I can no longer break it.

FWIW, may be it's my laptop, but I start to see some slowness on the graph responsiveness and it's only the bookinfo app. Not sure if it's really a concern.

I'll approve for now, and I'll keep doing more testing and see if it breaks for some reason.

israel-hdez (approved):

lucasponce (approved):
I ran some local tests and I didn't see anything "blocker", so ok.
About the code a couple of considerations, I'm not familiar with that code but working with the layouts I found it very handy to have the algorithms present [1]. That helped me understand all of these relations between the "events" handling and the "async" tasks, I have the "gut" feeling that something can be simplified on that side, and having the algorithms close to the code would help.

If QE doesn't find any blocker we can continue, BUT this PR would need to rebase with the master branch, otherwise, the backend may fail due to the previous "iter8" removal and that may create more noise.

[1] https://github.com/kiali/kiali-ui/pull/2288#issuecomment-1020916419

lucasponce:
In the context of https://github.com/kiali/kiali-ui/pull/2288, I'm trying to bring/clone the algorithms to be local to the graph code.
At least grid and concentric are possible, I'm going to give a try to the "dagre", I think it can help to debug these race conditions with the Cytoscape events.

src/components/CytoscapeGraph/CytoscapeToolbar.tsx
josejulio: Not really sure about the many pieces here, but [ColaGraph.getLayout](https://github.com/kiali/kiali-ui/blob/master/src/components/CytoscapeGraph/graphs/ColaGraph.ts#L3-L10) always returns a new object (even if has the same contents). Probably should put that in a variable and return it.

jshaughn: @israel-hdez It's because `ColaGraph.getLayout()` returns a new object every time.  I could have changed that behavior but instead just made this change.

jshaughn: Oh, right, @josejulio just said the same thing :-)

jshaughn:
> FWIW, may be it's my laptop, but I start to see some slowness on the graph responsiveness and it's only the bookinfo app. Not sure if it's really a concern.

@israel-hdez, I can't really understand any reason for a slowdown, if anything there should be some minor speedups.  I will attribute to your env.

>  That helped me understand all of these relations between the "events" handling and the "async" tasks, I have the "gut" feeling that something can be simplified on that side, and having the algorithms close to the code would help.

@lucasponce, it's possible.  One thing to know is that some layouts are discrete (dagre) and some not (cola, cose-bilkent).  So it's important to know that layout.run() can return before the layout is complete.  As such, the 'layoutstop' event is critical.  And this is why BoxLayout has some complicated `async` stuff.

src/components/CytoscapeGraph/CytoscapeToolbar.tsx
israel-hdez: OK, understood :-)

mattmahoneyrh (approved):

PR #2291, opened 2022-01-21
Protect trendlines from undefined values

lucasponce:
Fixes https://github.com/kiali/kiali/issues/4668


israel-hdez (approved):
Besides formatting, I only see one conditional changed.
I tried and seems to work properly.

mattmahoneyrh (approved):

PR #2292, opened 2022-01-21
Bump several third-party deps

lucasponce:
All of these are transitive dependencies.
I performed a basic smoke test on a local env, but it would require more eyes to check that no regression is introduced.


leandroberetta (approved):
Hi @lucasponce, everything seems ok.

mattmahoneyrh (approved):

PR #2293, opened 2022-01-21
[POC] Use cypress as proxy server for mock graph data

nrfox:
This change adds cypress.io to the project for the purpose of intercepting and mocking out the graph data response to return the pre-generated graph data created by the [generator tool](https://github.com/kiali/kiali/pull/4658).

To test this branch:
1. Update your deps to add cypress:  `yarn`
2. Update the `proxy` field in your `package.json` to point to your local kiali endpoint as per: https://github.com/kiali/kiali-ui#testing-with-kiali
3. ~Create namespaces `n[1-100]`. These namespaces need to exist for the graph page to work with the mock graph data.~ Run the [mock generator](https://github.com/kiali/kiali/pull/4658): `go run graph/generator/cmd/main.go`
4. Start your local front end server: `yarn start`
5. Start the cypress process: `yarn graph-proxy`
6. A cypress browser window should open. Click "Run 1 integration spec" in top right corner.
7. Navigate to the `graph` tab and select all the mock namespaces. You should see the mock graph data.

Something to note with using cypress as the proxy:

The request intercepting/proxy feature only works within the cypress browser context. If you navigate to the kiali url _outside_ of the cypress browser, you will not see the mock graph data but the real endpoint response. In other words, all testing of kiali with the mock data needs to be done within the context of the cypress browser.


ScriptingShrimp:
We need to add wildcards for cypress artifacts into `.gitignore` (screenshots/*.png, videos/*.mp4) - can send commit on that

> Create namespaces `n[1-100]` ...

This could be automated as a separate step (few test steps), i.e. `cy:prepare`, than you can go cy:open

> all testing of kiali with the mock data needs to be done within the context of the cypress browser

That's true. It's a bit clumsy for development, but I would say is's worthwhile. There is also possibility of adding specific dev tools (react, redux). 


If this approach will be considered as insufficient, I want to still find a way how to hook up cypress with mock service/endpoint. Testing on top this data is essential for quality.

lucasponce:
Thanks, @nrfox, and @ScriptingShrimp for this effort. I've started to review it (Nick feel free to tell me if it's too early, I wouldn't like to add noise with comments until you feel it's ready to receive feedback; feedback is always good, but I also like to wait until devs feel "hey, now it's a good moment to give it a try").

I have a basic question that's not 100% clear reading the steps. As I'm new with this framework, the "yarn cy:open" is similar to the "yarn start" or it's necessary to open a "yarn start" in the 3000 port and then open the "yarn cy:open"?

I ask because I can't make it work in my environment, so I guess I'm missing something.

nrfox:
> Thanks, @nrfox, and @ScriptingShrimp for this effort. I've started to review it (Nick feel free to tell me if it's too early, I wouldn't like to add noise with comments until you feel it's ready to receive feedback; feedback is always good, but I also like to wait until devs feel "hey, now it's a good moment to give it a try").

No, thank you for taking a look! I opened up the draft and assigned it to you to get feedback on using cypress for the proxy at this stage.

> I have a basic question that's not 100% clear reading the steps. As I'm new with this framework, the "yarn cy:open" is similar to the "yarn start" or it's necessary to open a "yarn start" in the 3000 port and then open the "yarn cy:open"?

It's similar to `yarn start` but instead of doing `yarn start` you would do `yarn cy:open`. This is because `yarn start` opens your normal browser (chrome, firefox, etc) whereas `yarn cy:open` will open the cypress browser which is requirement to using cypress for the proxy that all testing be done in the cypress browser because the requests are intercepted at the browser.

> I ask because I can't make it work in my environment, so I guess I'm missing something.

What's the issue?

nrfox:
> This could be automated as a separate step (few test steps), i.e. `cy:prepare`, than you can go cy:open

Yeah absolutely. My focus was on finishing the generator and deciding on the proxy solution before optimizing that but we that setup should be a part of the end result.

> If this approach will be considered as insufficient, I want to still find a way how to hook up cypress with mock service/endpoint. Testing on top this data is essential for quality.

That's a good point. Seems like there's value in automated tests for larger graph scenarios using the mock data even if the proxy solution proves to be too awkward for a development environment.

lucasponce:
I'm facing an aspect that may impact the Dev environments.
Hopefully is not a blocker but I guess it should require having a flexible config for this framework.

I'm having this error:
image

And I guess it means that "cypress" expects a single server for Kiali, both for backend and frontend.

BUT from a Dev perspective, I don't think this is a unified way to work; we typically have:

- A test cluster (minikube / openshift via crc).
- An external backend process (enabled to debug).
- An external frontend process (decoupled from the backend one and ready to debug).
- A browser with remote debug enabled to inspect and check issues.

So, from a user/dev perspective, I'd like the "cypress" framework to open the URL typically handled via "yarn start."

Otherwise, to use this framework, I'd need to prepare a less handy built because one of the aspects that I'd like to use is debugging my frontend with the mock data provided by the "cypress" server.

 I haven't investigated, but I have the feeling that it is something approachable via config.

We need this, as the developer setup is pretty typical. Yes, I know some ancient developers don't need to debug backend and frontend, but I'm not yet at that level, and having an IDE with debug mode is something that I need :)

nrfox:
Just to clarify, @lucasponce do you use the `proxy` setting in package.json to setup your dev env? The `cy:open` cmd uses this value to set `KIALI_URL` for cypress.

> So, from a user/dev perspective, I'd like the "cypress" framework to open the URL typically handled via "yarn start."

You need the dev server running that gets started with `yarn start` because that's what your IDE connects to for debugging? You might try running `yarn start` then passing that url directly in the `cy:open` [script in package.json](https://github.com/kiali/kiali-ui/pull/2293/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R42) and then try doing `cy:open`. Integration with an IDE for debugging was not something I kicked the tires on but it might work as expected or it might look different with cypress e.g. debugging within the cypress browser instead of within an IDE. 

> An external frontend process (decoupled from the backend one and ready to debug).
> A browser with remote debug enabled to inspect and check issues.

Cypress aims to combine/replace these two by executing both the frontend process and the browser you must test in.

lucasponce:
> You might try running yarn start then passing that url directly in the cy:open script in package.json and then try doing cy:open. Integration with an IDE for debugging was not something I kicked the tires on but it might work as expected or it might look different with cypress e.g. debugging within the cypress browser instead of within an IDE.

+1 I think this should work, if I understand this framework, populating here the URL should be enough.
I'll try it.

> Cypress aims to combine/replace these two by executing both the frontend process and the browser you must test in.

Well, we'd need to keep iterating on this, let's see if we can make it work with the yarn start, then the debug capabilities shouldn't be missed.

nrfox:
Updated the instructions to use the latest generator and run against the local front end server.

nrfox:
Closing this PR because it is specific to using cypress for the graph proxy. Going to open a different PR from the same branch that is just focused on adding cypress without the proxy.

PR #2297, opened 2022-01-27
Add Istio preview in wizards under servide details page

aljesusg:
Issue: https://github.com/kiali/kiali/issues/4681

Screenshot from 2022-01-27 12-25-24
Screenshot from 2022-01-27 12-25-41


lucasponce:
My initial feedback is positive, I think the feature works as expected.
I see a glitch with the Wizard when the doc is being generated, like if wizards are not properly closed or I think there is a problem in how the state is being handled.
I think that needs to be addressed:
kiali-services-preview-glitch

src/components/IstioConfigPreview/IstioConfigPreview.tsx
```diff
@@ -20,62 +27,67 @@ import { EditResources } from './EditResources';
 import { cloneDeep } from 'lodash';
 import _ from 'lodash';
 
-type PolicyItem = AuthorizationPolicy | Sidecar;
+export type PolicyItem =
```
lucasponce, line 30: "PolicyItem" is confusing. It's an Istio config resource, so "IstioConfigItem"/"IstioConfigResource" may help to not get lost on these layers.

src/components/IstioConfigPreview/IstioConfigPreview.tsx
```diff
@@ -20,62 +27,67 @@ import { EditResources } from './EditResources';
...
+  | AuthorizationPolicy
+  | Sidecar
+  | DestinationRule
+  | PeerAuthentication
+  | Gateway
+  | VirtualService;
+
+export interface ConfigPreviewObject {
```
lucasponce, line 38: ConfigPreviewObject -> "ConfigPreviewItem"?
Just to keep the terminology used?

(it's just naming, but that helped me to follow better the flow of the changes)

lucasponce:
Another issue, when copying/downloading, the types are missing:
```
metadata:
  namespace: travel-control
  name: control
  labels:
    kiali_wizard: request_routing
  annotations:
    kiali_wizard_related: PeerAuthentication/control
spec:
  host: control.travel-control.svc.cluster.local
  subsets:
    - name: v1
      labels:
        version: v1
  trafficPolicy:
    loadBalancer:
      simple: ~
      consistentHash:
        httpHeaderName: test
        httpCookie: ~
        useSourceIp: ~
    connectionPool:
      tcp:
        maxConnections: 1
      http:
        http1MaxPendingRequests: 1
    outlierDetection:
      consecutiveErrors: 1

---

metadata:
  namespace: travel-control
  name: control
  labels:
    kiali_wizard: request_routing
spec:
  selector:
    matchLabels:
      app: control
  mtls:
    mode: UNSET

---

metadata:
  namespace: travel-control
  name: control
  labels:
    kiali_wizard: request_routing
spec:
  http:
    - route:
        - destination:
            host: control.travel-control.svc.cluster.local
            subset: v1
          weight: 100
      match:
        - headers:
            test:
              regex: ^.*$
      fault:
        delay:
          percentage:
            value: 100
          fixedDelay: 5s
        abort:
          percentage:
            value: 100
          httpStatus: 503
    - route:
        - destination:
            host: control.travel-control.svc.cluster.local
            subset: v1
          weight: 100
  hosts:
    - control.travel-control.istio-cluster.org
  gateways:
    - travel-control/control-gateway-1
---

```

lucasponce:
I'm testing the PR with "view_only_mode: true" and I think that flow has not been properly managed in these changes:
image

If the user doesn't have rights, they should be able to create the configuration, check the preview, download/copy but not create in the cluster (similar flow as defined in the Overview page).

aljesusg:
Done !  I forgot to move the control. Thanks @lucasponce 

I think that we should show some kind of message like "You have not rights" and not only disable the button/option.

lucasponce:
> I think that we should show some kind of message like "You have not rights" and not only disable the button/option.

+1 

I have a minor comment here about style, I think that there is a similar "warning" used in the wizards locate close to the "Cancel" / "Create" button IIRC, but I haven't tested it yet, I put it on my TODO list to give a final review today.

@hhovsepy was the main reviewer in all the Wizards' work, so it would be nice if he can also give a review.

hhovsepy:
From UX perspective, I think the "Cancel" button on Preview mode should return back to Create page, not cancel the whole wizard, or add a "Back" button in Preview mode.

hhovsepy:
Page crashes when applying all advanced options and hit "Preview".
Screenshot from 2022-02-02 11-08-43
Screenshot from 2022-02-02 11-09-43
Screenshot from 2022-02-02 11-10-06
Screenshot from 2022-02-02 11-10-22

hhovsepy:
The generated VS empty "host:", so VS is not created.
Screenshot from 2022-02-02 12-03-51
Screenshot from 2022-02-02 12-04-06

hhovsepy:
Traffic shifting does not show weights:
Screenshot from 2022-02-02 14-02-06

aljesusg:
Let me check this because I didn't change anything from wizards, this is weird.

aljesusg:
> The generated VS has empty "route:", so VS is not created. "admission webhook "validation.istio.io" denied the request: configuration is invalid: HTTP route or redirect is required"
> 
> Screenshot from 2022-02-02 12-03-51 Screenshot from 2022-02-02 12-04-06

This is working for me @hhovsepy 
Screenshot from 2022-02-03 10-52-23
Can you provide me more information about how you got this error?

aljesusg:
> Traffic shifting does not show weights: Screenshot from 2022-02-02 14-02-06
 
Please @hhovsepy which example our service are you testing?
Screenshot from 2022-02-03 10-53-43

aljesusg:
> The generated VS has empty "route:", so VS is not created. "admission webhook "validation.istio.io" denied the request: configuration is invalid: HTTP route or redirect is required"
> 
> Screenshot from 2022-02-02 12-03-51 Screenshot from 2022-02-02 12-04-06

About this the preview button is showed if the configuration isValid like before so if this configuration cannot be applied so we have a bug in master. @lucasponce Can you verify this? we should backport this if the configuration is not valid.

lucasponce:
> About this the preview button is showed if the configuration isValid like before so if this configuration cannot be applied so we have a bug in master. @lucasponce Can you verify this? we should backport this if the configuration is not valid.

I can't see a regression in master.
It's probably that something in this PR is touching the state of the wizards, note that the main state is stored at the page level, and perhaps is in that area where something is not synced (I couldn't investigate in the code).

aljesusg:
> > About this the preview button is showed if the configuration isValid like before so if this configuration cannot be applied so we have a bug in master. @lucasponce Can you verify this? we should backport this if the configuration is not valid.
> 
> I can't see a regression in master. It's probably that something in this PR is touching the state of the wizards, note that the main state is stored at the page level, and perhaps is in that area where something is not synced (I couldn't investigate in the code).

Thanks @lucasponce 
The logic is the same I only changed the name and action, That's why I was asking, I'm going to take a look at it.

## before
```ts
<Button isDisabled={!this.isValid(this.state)} key="confirm" variant="primary" onClick={this.onCreateUpdate}>
            {this.props.update ? 'Update' : 'Create'}
          </Button>
```
## Now
```ts
<Button isDisabled={!this.isValid(this.state)} key="confirm" variant="primary" onClick={this.onPreview}>
              Preview
            </Button>
```

lucasponce:
In the master, you can create valid empty configurations.

This:
image

May create valid configuration:
image
image
image

aljesusg:
> The generated VS has empty "route:", so VS is not created. "admission webhook "validation.istio.io" denied the request: configuration is invalid: HTTP route or redirect is required"

So the behaviour is right. Why is not working for @hhovsepy ? I am try to replicate his environment

hhovsepy:
Sorry for the noise, deleted my screenshots, these were environment issues I cannot figure out.

jmazzitelli:
BTW, as a side note, this:
image
is not entirely correct English. It should say: "User does not have permissions"

aljesusg:
> BTW, as a side note, this: image is not entirely correct English. It should say: "User does not have permissions"

mmm thanks @jmazzitelli  this error sentence is in master, I am going to make the change here. Do we want to backport this?

jmazzitelli:
I just noticed that is not in this PR - so I created an issue about changing the english syntax of that error message. See: https://github.com/kiali/kiali/issues/4704

jmazzitelli:
> this error sentence is in master, I am going to make the change here.

It is in 4 different files. See my issue I just wrote. You can change it here in the PR if you want. No need to backport.

aljesusg:
> > this error sentence is in master, I am going to make the change here.
> 
> It is in 4 different files. See my issue I just wrote. You can change it here in the PR if you want. No need to backport.

Done ! 

Fix https://github.com/kiali/kiali/issues/4704

src/components/IstioActions/IstioActionsDropdown.tsx
```diff
@@ -75,7 +75,7 @@ class IstioActionDropdown extends React.Component<Props, State> {
       </DropdownItem>
     );
     const deleteActionWrapper = serverConfig.deployment.viewOnlyMode
-      ? this.renderTooltip('delete', TooltipPosition.left, 'User has not permissions', deleteAction)
+      ? this.renderTooltip('delete', TooltipPosition.left, 'User does not have permissions', deleteAction)
```
jmazzitelli:
```suggestion
      ? this.renderTooltip('delete', TooltipPosition.left, 'User does not have permission', deleteAction)
```

src/components/IstioActions/IstioActionsNamespaceDropdown.tsx
```diff
@@ -78,7 +78,7 @@ class IstioActionsNamespaceDropdown extends React.Component<Props, State> {
         className="kiali-group-menu"
         children={dropdownItemsRaw.map(r =>
           serverConfig.deployment.viewOnlyMode
-            ? this.renderTooltip(r.name, TooltipPosition.left, 'User has not permissions', r.action)
+            ? this.renderTooltip(r.name, TooltipPosition.left, 'User does not have permissions', r.action)
```
jmazzitelli:
```suggestion
            ? this.renderTooltip(r.name, TooltipPosition.left, 'User does not have permission', r.action)
```

src/components/IstioWizards/ServiceWizardDropdown.tsx
```diff
@@ -281,7 +281,7 @@ class ServiceWizardDropdown extends React.Component<Props, State> {
 
   getDropdownItemTooltipMessage = (): string => {
     if (serverConfig.deployment.viewOnlyMode) {
-      return 'User has not permissions';
+      return 'User does not have permissions';
```
jmazzitelli:
```suggestion
      return 'User does not have permission';
```

src/components/IstioWizards/WorkloadWizardDropdown.tsx
```diff
@@ -129,7 +129,12 @@ class WorkloadWizardDropdown extends React.Component<Props, State> {
         </DropdownItem>
       );
       const enableActionWrapper = serverConfig.deployment.viewOnlyMode
-        ? this.renderTooltip('enable_auto_injection', TooltipPosition.left, 'User has not permissions', enableAction)
+        ? this.renderTooltip(
+            'enable_auto_injection',
+            TooltipPosition.left,
+            'User does not have permissions',
```
jmazzitelli:
```suggestion
            'User does not have permission',
```

src/components/IstioWizards/WorkloadWizardDropdown.tsx
```diff
@@ -143,7 +148,12 @@ class WorkloadWizardDropdown extends React.Component<Props, State> {
         </DropdownItem>
       );
       const disableActionWrapper = serverConfig.deployment.viewOnlyMode
-        ? this.renderTooltip('disable_auto_injection', TooltipPosition.left, 'User has not permissions', disableAction)
+        ? this.renderTooltip(
+            'disable_auto_injection',
+            TooltipPosition.left,
+            'User does not have permissions',
```
jmazzitelli:
```suggestion
            'User does not have permission',
```

src/components/IstioWizards/WorkloadWizardDropdown.tsx
```diff
@@ -157,7 +167,12 @@ class WorkloadWizardDropdown extends React.Component<Props, State> {
         </DropdownItem>
       );
       const removeActionWrapper = serverConfig.deployment.viewOnlyMode
-        ? this.renderTooltip('remove_auto_injection', TooltipPosition.left, 'User has not permissions', removeAction)
+        ? this.renderTooltip(
+            'remove_auto_injection',
+            TooltipPosition.left,
+            'User does not have permissions',
```
jmazzitelli:
```suggestion
            'User does not have permission',
```

src/components/IstioWizards/WorkloadWizardDropdown.tsx
```diff
@@ -202,7 +217,7 @@ class WorkloadWizardDropdown extends React.Component<Props, State> {
           ? this.renderTooltip(
               'tooltip_wizard_actions',
               TooltipPosition.top,
-              'User has not permissions on this Workload',
+              'User does not have permissions on this Workload',
```
jmazzitelli:
```suggestion
              'User does not have permission on this Workload',
```

src/pages/IstioConfigNew/IstioConfigNewPage.tsx
```diff
@@ -159,7 +159,9 @@ class IstioConfigNewPage extends React.Component<Props, State> {
             () => {
               this.props.activeNamespaces.forEach(ns => {
                 if (!this.canCreate(ns.name)) {
-                  AlertUtils.addWarning('User has not permissions to create Istio Config on namespace: ' + ns.name);
+                  AlertUtils.addWarning(
+                    'User does not have permissions to create Istio Config on namespace: ' + ns.name
```
jmazzitelli:
```suggestion
                    'User does not have permission to create Istio Config on namespace: ' + ns.name
```

jmazzitelli (dismissed):
The English correction of the "User has not permissions" still needs a small tweak.

From the https://github.com/kiali/kiali/issues/4704 issue, `"User has not permissions" is not entirely correct English. Should be User does not have permission`

You want to change the plural form of `permissions` to the singular `permission`.

In other words, we want to say that the user does not have permission to do something. For example, "You do not have permission to create this resource" or "You do not have permission to go out and play" :) In this usage of the word, it is singular (i.e. no "s" at the end).

Note the difference in the two places where I left the plural form as-is in OverviewNamespaceActions.tsx: `User does not have enough permissions for this action`... here it is implied the user may need multiple "permissions" to do this action ("enough permissions"), hence plural.

lucasponce:
This PR has some noise with the grammar comments.
@hhovsepy then, can you confirm that the main flows from the Wizard work correctly?
I can give the last review after your confirmation.
Thanks

hhovsepy (approved):
Confirming that the main flows from wizard work correctly. Overview page looks good to me.

Will be good to have "cancel" button on overview page returning to main wizard.

lucasponce:
> Will be good to have "cancel" button on overview page returning to main wizard.

+1 I was to comment this as well.

aljesusg:
> > Will be good to have "cancel" button on overview page returning to main wizard.
> 
> +1 I was to comment this as well.

Done !

hhovsepy:
Cancelling the preview returns to wizard with reset state. I think this is fine.

aljesusg:
In my env the state was not reset maybe you pulled before my force push

lucasponce (changes requested):
Found a blocker/regression in the Overview, when trying to use the Traffic Policies
image

hhovsepy:
Clusteradmin user does not have permissions on Overview page with Traffic Policies.
Screenshot from 2022-02-04 16-59-39

aljesusg:
> Clusteradmin

can you query to /api/istio/permissions with the namespace that you are requesting and check permissions?

You must to have delete,update and create rights. We didn't change this, clusteradmin must have these rights but it makes no sense 

```ts
fetchPermission = () => {
    this.promises.register('namespacepermissions', API.getIstioPermissions([this.props.nsTarget])).then(result => {
      const permission = result.data[this.props.nsTarget][AUTHORIZATION_POLICIES];
      const disableOp = !(permission.create && permission.update && permission.delete);
      this.setState({
        confirmationModal: true,
        disableOp
      });
    });
  };

```
Curl example
```bash
curl -X GET \
  -H "Content-type: application/json" \
  -H "Accept: application/json" \
  -d "{'namespaces':'travel-agency,travel-portal'}" \
> "http://localhost:20001/api/istio/permissions" 
```

aljesusg:
> Found a blocker/regression in the Overview, when trying to use the Traffic Policies image

It's working for me :'( what namespace/example are you trying?

hhovsepy:
> can you query to /api/istio/permissions with the namespace that you are requesting and check permissions?

The permissions are correct on that cluster, this is QE cluster which was always used.
The issue seems like a glitch. Not happening constantly.
Steps to see it:
1. Refresh the Overview page.
2. Check Overview Actions - Update Traffic Policies.
3. You will see that Update button disabled and no permissions warning.
4. Try to execute any action, and cancel it, for instance "Delete Traffic Policies".
5. Try to Update Traffic Policies.
6. Issue is not available anymore. Create/Update buttons are active.
It seems some async load issue.

lucasponce:
I'm still facing this 
image

When the user should be able to create that config.

aljesusg:
> I'm still facing this image
> 
> When the user should be able to create that config.

o.o i found the problem, after rework with confirmation let me fix it

lucasponce:
kiali-overview-preview-PR-review-problem

More regressions.

lucasponce:
I still see a regression with the permissions logic.
If the user has no permissions, but there are existing configurations, the update flow works as expected but the user can delete.

This screenshot is correct:
image

But when clicking on the "Delete Traffic Policies" we have:
image

(And the user can perform the operation)

As we can see, the permissions query returns false:
```
{
"travel-agency":{
"authorizationpolicies":{"create":false,"update":false,"delete":false},
"gateways":{"create":false,"update":false,"delete":false},
"peerauthentications":{"create":false,"update":false,"delete":false},
"requestauthentications":{"create":false,"update":false,"delete":false},
"serviceentries":{"create":false,"update":false,"delete":false},
"sidecars":{"create":false,"update":false,"delete":false}}
}
```

Perhaps the same message used in the preview wizard can be used here and the "delete" button may show as disabled.

I've tested this using the "view_only_mode: false" flag, that feature will return all permissions queries as false.

skondkar:
If i click on 'Disable Auto Injection' 'Enable Auto Injection' or 'Remove Auto Injection' on overview page, I am not able to see the confirmation dialogs. I tested on master and the confirmation dialog appears on master.

skondkar (approved):
LGTM

Verified preview on overview page and in wizards under service details page. Verified Copy and Download resources on preview pages.
Also verified with user permission when view_only_.mode: true. 

Screenshot from 2022-02-10 17-34-44

Screenshot from 2022-02-10 17-34-13

Screenshot from 2022-02-10 17-33-56

lucasponce (approved):
@skondkar has tested the use cases I commented. So, +1.

PR #2298, opened 2022-02-01
BreadthFirst algorithm + Combine outer/inner layouts + Clone layouts

lucasponce:
This is a WIP/experimental branch.
It has several goals:
- To be able to debug the layout algorithms to better use the "layoutstop" event and async tasks (first for learning, but there is a suspect that layout can be optimized).
- The "dagre" algorithm is based on an algorithm deprecated and unmaintained since 2019, so I think it could be good to find an alternative that has the same functionality.
- The "breadthfirst" algorithm can be a substitute for the "dagre" but it requires a fix as it's not working as it should be designed.

Algorithms are easy to follow and can be handy for debugging or potential modifications (i.e. adjust size).


jshaughn:
It's an interesting idea to just clone the layouts we are interested in.   If it ever made sense, I suppose we could do the same with the html-node-label extension.

lucasponce:
> If it ever made sense, I suppose we could do the same with the html-node-label extension.

I can add it to the PR, then all critical pieces involved in a debug should be local and potentially it can give more flexibility in case some change is required, it can be local.

lucasponce:
The "breadth and first" algorithm looks promising for certain scalability scenarios, but it's not yet finished.
It works well for regular objects, but for boxes, it requires some "padding" adjustment in the position (to avoid a generic position non-optimized as other layouts).

Some examples of same cluster view:

a) Dagre
image

b) Grid
image

c) Concentric
image

d) Breadh and First
image

This last algorithm is a combination of a Grid and organizing the elements per level. Then the user can navigate better than a pure Grid, in which the elements are sorted by name.

But as in other situations, there is no magical layout, there are cases that this last algorithm performs really well, but in other cases that Dagre and concentric offers better results.

Still working to adjust the overlapping when boxes are activated:

a) Breadth and First + Dagre
image

b) Breadth and First in main elements + boxes:
image

lucasponce:
I've rebased & combined the work of this PR with the edges:

image

I see an aspect that I'd like to implement better: when edges are hidden, we have the "one line/row" of elements as the connective algorithms don't work (they don't have edges to evaluate levels).

In this case, the "overflow" in the "breadth-first" could distribute better the elements (i.e. in practice like a grid).

I'm going to prepare a change for this scenario.

I'm not sure if we will need the zoom buttons back, when the graph is highly populated I missed them.

jshaughn:
I've started by just trying a small layout, using travel demo, to try various options and make sure what used to look ok still looks ok.  In general things look good, I've tried a variety of combinations of graph types, layouts and namespace layouts.  One small anomoly was breadth-first, where it seemed the labels get crowded:

image

Actually, the label font seems much bigger, in general, for breadth-first, maybe that has something to with it.  This is easier to see when just looking at a small topology.

Note that, when removing edges, grid and breadth-first both look pretty good now, avoiding the horizontal look of dagre and concentric.  Both would possibly take advantage, if possible, of some sort of left margin to avoid having nodes occluded by the toolbar:

Either can look bad when showing just a few unhealthy edges, because the edges may just go horizontally through a row of nodes.  But I think that's just the nature of the layout.  As soon as you add some edges a different layout, dagre or concentric may be better.

In general, if there are a no edges I think dagre and concentric could possibly even be disabled.  If there are several orphan/disconnected nodes (including app, namespace, or cluster boxes), dagre could likely be disabled.

I then did some testing using the graph proxy, with 50 apps, and to be honest, I'm pretty happy with what I am seeing.   In general, given what I have available in the PR right now, I can always find something fairly reasonable both with a small topology and a medium topology.  I do have some separate concerns around our color scheme, as nodes and edges get hard to discern from the background, as the topology gets large and the nodes get small.

Looking good!

lucasponce:
>  One small anomaly was breadth-first, where it seemed the labels get crowded:

I see, I think we could add some padding for these cases.

> Actually, the label font seems much bigger, in general, for breadth-first, maybe that has something to with it. This is easier to see when just looking at a small topology.

This one is not related to the layout algorithm but a combination of how Cytoscape "scales" the elements. The algorithm is only adding positions (x, y) to all the elements, but trying to adjust the space to have a better picture when it's populated. One of the differences between this modified "breadth-first" and the others is how the positions are calculated, that is a little bit optimized (others algorithms take the assumption that all nodes have the same size, then when you enabled namespaces/boxes, you'll see all the big scale and bad aspect/ratio).

One of the benefits of having the algorithm code available is that now we could optimize this (not only in the "breadth-first" but also in others).

> In general, if there are a no edges I think dagre and concentric could possibly even be disabled. If there are several orphan/disconnected nodes (including app, namespace, or cluster boxes), dagre could likely be disabled.

If the idea of the conditional render of the buttons looks a good direction, I think I can try to adapt the logic of the toolbar and add some constraint here as it shows in the "namespace" boxes.

When no edges, probably as you comment, the user can only choose with two layouts (that we know will have a better result).

With boxes and edges, I'm still thinking letting the user choose the outer/inner layout is a good option, but I also see like an experimental, we can propose it in one version and optimize it in the following ones.

In general, I see that adapting the algorithms can be a direction to optimize these layouts for more cases; I also see in some cases the inconsistency between the state of Cytoscape, especially when the "box-layout" is involved, but I don't see how to solve it (but I don't see it as a blocker).

I'm going to keep iterating on the comments, but thanks for the feedback, and glad that the proposal looks good.

lucasponce:
@jshaughn I've updated the PR with better padding to fix the comments about the populated labels:
image

As discussed, the ideas to automatically identify the "best" layout can come in future iterations.

Let me know if you'd like to address more aspects of this PR or if it's good for consolidating it.

jshaughn (changes requested):
@lucasponce , only one minor change required for approval, I think you forgot to make node-drag active by default in the cyto-toolbar.

Otherwise, I think we should get this in so people can use it for a few days before the end of sprint.

I noticed one thing that can cause a small issue, we currently increase the font-size for a highlighted node.  For larger labels (with host info, for example), and especially with breadth-first, it can cause some overlap.  We could consider not increasing the font for highlighted nodes.  These larger labels do seem to still cause some other issues in breadth-first, but I'm inclined not to block on it:

image

lucasponce:
> only one minor change required for approval, I think you forgot to make node-drag active by default in the cyto-toolbar.

Oh, sure. I'll prepare a commit for this.

>  For larger labels (with host info, for example), and especially with breadth-first, it can cause some overlap. We could consider not increasing the font for highlighted nodes. 

I see what you mean. I'd say it's a more generic issue about how the height is calculated with labels, as for example, in pure Dagre layout:

image

So, I think is something we can iterate.

The good thing about this PR is that now we can make finer adjustments in all layouts to better calculate dimensions for these cases (the BoxLayout and external badges/labeling add additional logic that is not native in Cytoscape and those would require some tweaks).

lucasponce:
I wonder how it looks like a minor change in the label badge on ingress/VS, to hide the "N Host" wording and show all the hosts when hovering the badge in general.

I think that would increase the UX for these cases and when the user hovers a specific badge it will work as today.

jshaughn:
> I wonder how it looks like a minor change in the label badge on ingress/VS, to hide the "N Host"

It is something we could play.  The added height for these labels is currently hard to deal with and it may be useful to defer host info to contextInfo.

lucasponce:
> It is something we could play. The added height for these labels is currently hard to deal with and it may be useful to defer host info to contextInfo.

Thanks for the feedback, let me add a proposal to this PR in a separate commit, but I think at scale it's something we can give a try.

lucasponce:
I put a commit with a minor refactoring on the style of the labels:
image

I think it may offer a better result.
The host number is a little bit redundant; on Gateways and VirtualServices there will be always hosts.
The user can keep seeing them when "hovering" on them.

But thinking on scale, on a large populated graph, probably this minor change can improve the overall visualization.

jshaughn (approved):
Node-drag now active by default.   I like the label-hover for host display.  Non-blocking idea: add an entry for this the graph tour "Shortcuts".  Quite an effort on all of this, APPROVED!

lucasponce:
Thanks for the review @jshaughn.
I'm going to consolidate it on the master branch. 
I imagine that there could be some follow-up fixes as soon as more people give a try, I think we have time before the end of the sprint for additional testing/changes.

PR #2299, opened 2022-02-01
Side panel re-design for Istio config details page

leandroberetta:
Work in progress of the side panel re-design.

Resolves [#4553](https://github.com/kiali/kiali/issues/4553)

Backend PR: [#4740](https://github.com/kiali/kiali/pull/4740)

I tried to follow the same design language from the workload/services overview (type bade, name, status):

sp_vs_ok

Then, new sections like:

- Configuration Analysis (former analyzer messages)
- Configuration Assistant (here the idea is to use dynamic markers to highlight parts of the yaml and help the user, see https://github.com/ajaxorg/ace/issues/4626 to understand the blocker I guess I have)
- References (to Istio objects, workloads and services, i'm implementing the services from Hayk's draft)

This side panel will be present for all config resources.

The following is an example of a validation error, when the configuration status field is activated on Istio (https://istio.io/latest/docs/reference/config/config-status/)

sp_vs_fail


jshaughn:
I'm not sure what the Configuration Assistant section will contain.  Aren't the dynamic markers going to be in the yaml?  Or is there some sort of entry you click on to activate the dynamic marker?

Are the references going to be links to [service] detail pages?  What about the related config, like destination rules?

If a section doesn't have any entries it should probably have some value to indicate that it's not waiting to be filled in, like 'none found'.

leandroberetta:
@jshaughn thanks for the comments.

Sure, the sections are there to indicate a sort of layout, but there are parts missing.

Regarding the dynamic markers, the idea is to have them in the yaml, but when hovering, fill that "Configuration Assistant" section with the information or have the info in a popover.

jshaughn:
A popover would seem to make more sense, if it is possible.

lucasponce:
Hi @leandroberetta thanks for sharing this draft.

The main three sections have a lot of sense to me, regardless the final wording to cover "validations", "references", "help" have a lot sense. I'm sure that the final wording may change (i.e. "configuration assistance" sounds pretty AI, perhaps "Istio Reference" or something more clear may help, but keep it in your draft with your criteria until we can test it and have more discussion).

I know that the markers have some technical challenges, we faced some when we worked on the validations code.
If you find some challenges, you can explore other options but keep the same feature:
- Indicate to the user that clicking in some line will show additional info about the Istio reference of that part of the code (i.e. Istio Reference about what a route is in the VirtuaService).
- Personally, I think the right place is to add it in the Overview panel, but here I guess that we need to see what is technically doable.

In the validations we had to take some similar decisions, instead to highlight just a word, we decided to highlight the entire line to avoid some problems to implementing and maintaining the algorithm.

hhovsepy:
Hi @leandroberetta , can this UI PR be designed to include the References from backend PoC PR https://github.com/kiali/kiali/pull/4694 ?
It needs "references=true" param in istio config details request (alongside with validation=true param), and backend will return "references.serviceReferences" array including name and namespace for Services.
It is working for VS details for now.

leandroberetta:
> Hi @leandroberetta , can this UI PR be designed to include the References from backend PoC PR [kiali/kiali#4694](https://github.com/kiali/kiali/pull/4694) ? It needs "references=true" param in istio config details request (alongside with validation=true param), and backend will return "references.serviceReferences" array including name and namespace for Services. It is working for VS details for now.

Yes, in fact, yesterday I pulled your PR and I get the services references already!. I'm on it.

leandroberetta:
> Hi @leandroberetta thanks for sharing this draft.
> 
> The main three sections have a lot of sense to me, regardless the final wording to cover "validations", "references", "help" have a lot sense. I'm sure that the final wording may change (i.e. "configuration assistance" sounds pretty AI, perhaps "Istio Reference" or something more clear may help, but keep it in your draft with your criteria until we can test it and have more discussion).
> 

Ok, no problem, regarding the Assistant, yes, sounds like an AI feature, but well.. I was thinking that this could be the first step in a sort of help/assistance.

> I know that the markers have some technical challenges, we faced some when we worked on the validations code. If you find some challenges, you can explore other options but keep the same feature:
> 
> * Indicate to the user that clicking in some line will show additional info about the Istio reference of that part of the code (i.e. Istio Reference about what a route is in the VirtuaService).
> * Personally, I think the right place is to add it in the Overview panel, but here I guess that we need to see what is technically doable.
> 
> In the validations we had to take some similar decisions, instead to highlight just a word, we decided to highlight the entire line to avoid some problems to implementing and maintaining the algorithm.

I agree, we need to add more help to the users whatever be the technical implementation.

Thanks for the feedback.

leandroberetta:
@hhovsepy the following is an example of a references representing a Service:

sp_references

My idea is to instead of separate between configuration object references, services and workloads, is to list all together with the badges indicating the nature of each of them.

In the backend, we can do the same adding a "type" field, or continue like it is now, and I can gather them in the UI.

jshaughn:
> My idea is to instead of separate between configuration object references, services and workloads, is to list all together with the badges indicating the nature of each of them.

This is fairly consistent with the graph side panel, which also uses the PF badges (with tooltips) to identifie the related objects.  In the graph it's all in the side-panel header.  Below, the Service and Workload are related to the selected App:

image

leandroberetta:
Updated PR with the addition of the Gateways references:

sp-gw-ref

I'm still figuring out how to proceed with the dynamic markers feature without them.

leandroberetta:
@lucasponce @jshaughn 

I've been trying to use the dynamic markers from the Ace editor with no luck, I even tried to go a little bit deeper trying another way with a Renderer function but at least for now, I couldn't get it to work (problem in the description).

Si I thought a different approach to achieve (kinda) the same, or at least the same spirit, that is helping the user when configuring Istio resources.

So, for that reason, I'm sharing the following screenshots that describe the "Help" section:

sp_help

As you can see, each resource will have some field help, the data will be delivered by the backend, I'm thinking in adding an optional parameter like ?help=true to the existing Istio config detail endpoint.

Then, with a simple tooltip, more info is shown:

sp_help_tooltip

I know it would be nice to have this on the editor but maybe this is a mid term solution while we wait for help on the Ace side.

jshaughn:
I think I like this approach as all of the help is located is consolidated in one place, and the user can easily hover on things of interest.  Also, it may actually have an advantage over the ace/in-place help, as the popup doesn't occlude theyaml.   One tiny nit, as opposed to `/spec/hosts`, maybe `spec.hosts`?  I'mnot sure what is standard, I know in our CR docs we use the dot notation.

leandroberetta:
@jshaughn I agree with your comments, thanks for the feedback.

Regarding the notation, I guess that I thought in "/" because of my time using xpath queries, but dot notation looks good too, I'll try it.

leandroberetta:
New updates:

- Help is now coming from the backend
- New style help labels
- Moved from slashes to dots

sp_new

leandroberetta:
> What do you think about just linking to the istio reference docs for the particular object e.g. https://istio.io/latest/docs/reference/config/networking/virtual-service/?

We can add links but I think is fine to have a short description/explanation (although it will be a copy of the Istio's doc). I tried to reference to a particular field in a table and I couldn't, so for example for: "spec.http.route.destination.subset" I should link the parent. 

I guess I prefer to give a little bit of context inside Kiali rather than send the user to Istio site, but at least a link could be there for sure, thanks @nrfox.

lucasponce:
Hi @leandroberetta, I could play with the PoC/PR and I may have some suggestions:

- I see a little bit hard to use the current UX/UI design:
image

The idea is not bad itself, but I'd propose to explore if we could play with the cursor event:
image

So, I can detect the line where the user places the editor, and if it's in the "spec.hosts" area, then the side panel can show the content of the tooltip like a regular paragraph after the "spec.hosts" label.

If the tooltip would be within the editor, with a marker/icon, perhaps the current format may be useful, but need to move to the side panel, I'd prefer to see all the info in a single step.

The current validation markers use similar line-matching logic to place the validations, and I think you could use something similar to check when the cursor is under a section and then allow the specific help, then when a user is modifying the YAML, the side panel is automatically updating this help info.

Let me know how do you see this comment. I think with the current work, it would require some work in the UI to check if the events on the cursor can help on this.

leandroberetta:
@lucasponce thanks for the suggestion, it worked well, you can see the result for now:

simplescreenrecorder-2022-02-22_12 23 31 (1)

lucasponce:
> thanks for the suggestion, it worked well, you can see the result for now

Perhaps, the lines that have "help" text, can show an info marker (that would be likely removed on validations warnings/errors).
That would tell the user that a section is under the "help" scope.

lucasponce:
> I tried to put the full message in the tooltip that appears, but it's a lot if information, so instead the message indicates a little bit what the user should expect.

Yes, or perhaps with the "i" icon is just enough, the description can be shown only when the side panel is collapsed.

leandroberetta:
@jshaughn @lucasponce I turned this draft into a ready-to-review PR.

Features:

- Common header for all objects including the type of the object health and labels if any
- References for workloads, services and other configuration objects (in case of a validation failure, the references won't be shown)
- Analyzer messages if turned on when installing Istio (only for upstream Istio)
- Help section with hints on the editor (this is pretty much the same logic for every object, the help is shown if the backend returns information)

How to test?:

The only requirement is the following backend PR: https://github.com/kiali/kiali/pull/4740

By now, the server PR adds help information for VirtualServices, DestinationRules and Gateways

I tested this with our bookinfo demo that already provides a Gateway and a Virtual Service. For the destination rules I created the following: https://raw.githubusercontent.com/istio/istio/release-1.13/samples/bookinfo/networking/destination-rule-all.yaml

As an example playing a little bit with the VirtualService, the following example can be seen as follows:

sp_1

Also, for example, an Authorization Policy:

```yaml
apiVersion: security.istio.io/v1beta1
kind: AuthorizationPolicy
metadata:
 name: allow-nothing
 namespace: bookinfo
spec:
  selector:
    matchLabels:
      app: productpage

```

sp_2

lucasponce:
Adding @hhovsepy as he's added multiple references in the backend.

lucasponce (approved):
The backend code is still in Draft, but the UI PR looks ok.
I tested several scenarios:
image
image
image

Failed DR:
image

Sidecars:
image

AuthorizationPolicies:
image

leandroberetta:
@lucasponce thanks for the review. The backend PR is out of draft now.

hhovsepy:
One minor comment, to add info message "No References Available" or something  similar under References section, in case of empty:
Screenshot from 2022-03-15 14-13-29

leandroberetta:
> One minor comment, to add info message "No References Available" or something similar under References section, in case of empty: Screenshot from 2022-03-15 14-13-29

The message is in the code, but It seems that for any reason is not appearing. I will test with that object. Thanks @hhovsepy.

leandroberetta:
@hhovsepy that is correct, I was assuming that when no references, a null would be present like in the workloads case, but in some cases an empty string can be an option to according to:

`"references":{"objectReferences":[],"serviceReferences":null,"workloadReferences":[]}`

I will cover both cases. Thanks.

leandroberetta:
Now the message appears:

sp_pa

hhovsepy (approved):
Verified, all 3 cases: no reference, error in validation, all is green:
Screenshot from 2022-03-15 15-58-37
Screenshot from 2022-03-15 15-59-14
Screenshot from 2022-03-15 15-59-31

leandroberetta:
> Validation references are missing, these are not the object references but the validation ones whic hshould be shown in case of warning or error: Before: Screenshot from 2022-03-15 17-33-14 After: Screenshot from 2022-03-15 17-33-02

Oh, it's true, I guess I need to add the section, because today as you mentioned is not there, I guess back in the moment, I thought that with the new references would be enough, but I see that they are useful now.

Thanks.

leandroberetta:
@hhovsepy I added the section for the cases when there are validation references:

sp_sidecar_1_error

Thanks, good catch!.

PR #2303, opened 2022-02-08
[v1.36] bump axios

jshaughn:
@nrfox This should likely wait for the pending server-side changes.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -4552,14 +4552,7 @@ axios-mock-adapter@1.16.0:
   dependencies:
     deep-equal "^1.0.1"
 
-axios@0.21.2:
-  version "0.21.2"
-  resolved "https://registry.yarnpkg.com/axios/-/axios-0.21.2.tgz#21297d5084b2aeeb422f5d38e7be4fbb82239017"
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -4552,7 +4552,14 @@ axios-mock-adapter@1.16.0:
...
 
-axios@0.21.1, axios@^0.21.1:
+axios@0.21.2:
+  version "0.21.2"
+  resolved "https://registry.yarnpkg.com/axios/-/axios-0.21.2.tgz#21297d5084b2aeeb422f5d38e7be4fbb82239017"
+  integrity sha512-87otirqUw3e8CzHTMO+/9kh/FSgXt/eVDvipijwDtEuwbkySWZ9SBm6VEubmJ/kLKEoLQV/POhxXFb66bfekfg==
+  dependencies:
+    follow-redirects "^1.14.0"
+
+axios@^0.21.1:
```
nrfox, line 4562: It looks like one of the transitive dependencies is still resolving to `0.21.1`. Just to be safe can you also force the transitive deps to use a version > `0.21.1` by either adding "axios": "^0.21.2"` to the `resolutions` field or by removing this block from the `yarn.lock` file and re-running `yarn`?

jshaughn: updated, take a look now...  the section is still there but I think now it forces a more recent version.

jshaughn: updated again, requesting a new review...

nrfox (approved):

PR #2304, opened 2022-02-08
[v1.24] bump axios

jshaughn:
@nrfox This should likely wait for the pending server-side changes.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3149,10 +3149,10 @@ axios-mock-adapter@1.16.0:
   dependencies:
     deep-equal "^1.0.1"
 
-axios@0.21.2:
-  version "0.21.2"
-  resolved "https://registry.yarnpkg.com/axios/-/axios-0.21.2.tgz#21297d5084b2aeeb422f5d38e7be4fbb82239017"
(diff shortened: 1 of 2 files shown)


nrfox (approved):

yarn.lock
```diff
@@ -6435,18 +6435,16 @@ focus-trap@4.0.2:
...
     xtend "^4.0.1"
 
-follow-redirects@1.5.10:
-  version "1.5.10"
-  resolved "https://registry.yarnpkg.com/follow-redirects/-/follow-redirects-1.5.10.tgz#7b7a9f9aea2fdff36786a94ff643ed07f4ff5e2a"
-  integrity sha512-0V5l4Cizzvqt5D44aTXbFZz+FtyXV1vrDN6qrelxtfYQKW0KO0W2T/hkE8xvGa/540LkZlkaUjO4ailYTFtHVQ==
-  dependencies:
-    debug "=3.1.0"
-
 follow-redirects@^1.0.0:
```
leandroberetta, line 6438: Do we have 2 versions of follow-redirects?.

nrfox (approved):
Pkg bump LGTM

yarn.lock
jshaughn: This should be OK, I believe it means axios will use the newer version of follow-redirects, but some other transitive dep may use the older version.  The CVE is against axios, not follow-redirects, so I don't think we need to deal with that other dep.

PR #2307, opened 2022-02-15
[v1.24] due to the requirement of building with node 14, we need to bump node-sass to 5.0.0

jmazzitelli:
We need this so the production builds can work.

You must make sure the builds use Node 14.  Specifically, prod builds use 14.18.2, so use the same version to ensure compatibility.


israel-hdez (approved):
Works for me

PR #2309, opened 2022-02-22
Use dot notation for sourcing from pkg scripts

nrfox:
Some package scripts source env vars with the `source` keyword. Apparently this keyword is only available in `bash` but not `sh`. Some linux distros run bash for `/bin/sh` by default but some do not causing the package scripts to fail on the distros that do not. Using `.` instead of `source` works for all shell types apparently.


nrfox:
@jmazzitelli you mind taking a look at this when you get a chance?

jmazzitelli (approved):
. == source

lgtm

PR #2310, opened 2022-02-25
[WIP] Cypress initial dev tests

nrfox:
See [README](https://github.com/nrfox/kiali-ui/blob/7d3cb55d8d1e8a848c41f91674ff91d11c9e7aaf/cypress/README.md) for instructions on how to run tests.

A few basic tests for:

1. Graph - tests the graph loads for bookinfo and a large graph (200+ namespaces/workloads)
2. Istio Config - tests the editor loads
3. Workloads - tests the page loads


nrfox:
@lucasponce if you are kicking the tires on cypress you checkout this PR and run it for a starting point with Kiali.

cypress/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+```bash
+yarn cypress:run
+```
+
+## Testing Strategies
+
+### Graph
+
+The kiali graph is primarily a canvas element which makes it more difficult to test through query selectors with cypress. To get around this limitation, a simple way of testing visual elements is to do snapshot testing aka [visual testing](https://docs.cypress.io/guides/tooling/visual-testing). This boils down to comparing an existing image of the page to a new image. If any differences exist, the test will fail. There are other ways of testing canvas elements in Cypress but they seem to involve interacting with cypress through a custom api rather than visually which is not ideal.
```
lucasponce, line 27: Mmm, it's going to be a limitation if cypress can't interact with the canvas element.
I don't think we can't rely on snapshots comparison in the graph, for multiple reasons, like the same data can render different graphs easily due to the nature of the framework/technology we use.

lucasponce: But I'd guess it's an early comment from my side, worth giving a try to see how it works and which level of sensitivity of "snapshot-image-comparison" may offer here.

nrfox: It's definitely a limitation but more of a limitation of using html canvas than cypress itself I'd say. There's some settings to tune the diff sensitivity between two images. It'd be good to play around with those to make sure it'll work for two graphs that are rendered slightly differently but are mostly the same.

mattmahoneyrh:
Is there a reason not to use the BDD framework that @ScriptingShrimp is prototyping?

cypress/README.md
ScriptingShrimp: If we want to invest and research visual testing (i.e. that canvas renders), I highly recommend this talk (timestamp 8:30) https://youtu.be/Dl_XMd_1F6E?list=PLI1o8Ea2Lj01qMzZRg8F4tRpIxTklrnbV&t=511
 
^ that pretty much sums up the possible approaches, also he has a brief demo at the end

my opinion: We might want to try create our snapshots, see if that works. If not, some 3rd party commercial service with OSS plan could

aljesusg:
This PR must be submit to https://github.com/kiali/kiali

lucasponce:
I think this one may be outdated due to the current work.
I've also created a main epic for these efforts https://github.com/kiali/kiali/issues/4854.
I'm going to close it to avoid confusion, please feel free to re-arrange the PR for the unified repo.
Thanks!

PR #2311, opened 2022-02-25
Find/Hide support for Labels

jshaughn:
The UI-side to add label find/hide to the graph.

**SERVER PR: ** https://github.com/kiali/kiali/pull/4755


jmazzitelli (approved):
this is working with the server-side pr

src/services/GraphDataSource.ts
```diff
@@ -180,7 +182,7 @@ export default class GraphDataSource {
     }
 
     // Some appenders are expensive so only specify an appender if needed.
-    let appenders: AppenderString = 'deadNode,sidecarsCheck,serviceEntry,workloadEntry,istio,healthConfig';
+    let appenders: AppenderString = 'deadNode,istio,healthConfig,serviceEntry,sidecarsCheck,workloadEntry';
```
jmazzitelli, line 185: if you did this change to alphabetize the list, try again :p

jshaughn: rats

jshaughn:
Pushed a rebase to resolve conflicts with master, and also increased find/hide input box width to better accommodate the longer `label:label-name` operands.  No functional changes.

lucasponce:
I think with the labels feature, the find/hide width boxes would not be enough for the user experience.

I wonder if it could be a good idea to unify them in a single box as proposed in https://github.com/kiali/kiali/issues/4605#issuecomment-1029767625

jshaughn:
> I wonder if it could be a good idea to unify them in a single box...

As I mentioned in my original response, I'm not convinced it's necessary.  If nothing else it doesn't need to be part of this work and is something we can cansider in the future.

lucasponce:
It would need an update in the "Graph Fing/Hide" helper:
image

I can't find the syntax and some examples about how to use it.

lucasponce:
(I'm going to consolidate the feedback in the UI PR, but perhaps some aspects are related to the backend/frontend).

This scenario doesn't work in my environment:
image

I think the syntax is correct and I'd expect to highlight the "details" app.

lucasponce:
A first issue with the "negative" operator.

This scenario works as expected:
image

This other scenario is not working:
image

I'd expect the negative operator of this query:
image

jmazzitelli:
@lucasponce for "app" and "version" do NOT use the "label:" syntax. We stripped out the "app" and "version" labels from the generated graph to avoid duplicating data (so we don't send the data over the wire and so we don't unnecessarily bloat the graph). You already have the capabilities you want by simply using "app=<your app name>" or "version=<your version>".

Since we already had "app" and "version" support in the find/hide, we added the "label:" syntax for custom labels such as "datacenter=europe" or something of that nature.

jmazzitelli:
> But the "not" operator didn't work as well:

THAT sounds like a bug. @jshaughn ^^^

lucasponce:
> Since we already had "app" and "version" support in the find/hide, we added the "label:" syntax for custom labels such as "datacenter=europe" or something of that nature.

Yes, I saw the code, I deleted an older comment.

jshaughn:
@lucasponce , @jmazzitelli ,

For sure I missed the help additions, adding that now.

A note, I think it's understood but we strip the configured label names for `app` and `version`, which default to "app" and "version".

The situation you see with find `label=env` and hide `!label=env` is actually expected, and has nothing to do with the `label:` operand in particular.  You can see the same behavior in other hide scenarios.  When hiding nodes, any node that is left as an orphan is also removed.  In general this is the preferred behavior as otherwise you can end up with a bunch of disconnected nodes, especially when hiding on edge conditions.  We can discuss whether it's the best behavior but we haven't had complaints from actual users.

The issue with `not label:env=prod` resulting in an error is actually also expected behavior.  The reason is that only unary operands can accept a leading `not`.   I'll admit it is a little confusing because this new `label:` operand can be both unary and binary.  When checking for existence, it is unary and these are valid: `!label:env`, `! label:env`, `not label:env` .  But when used with an operator and value, it is binary, and the negation should be done via the operator, like `label:env != prod`.

I'll try and add that into the new help.

jshaughn:
> Just describe how a user should show a "find" node filtered from the label and hide the others.

It's not currently possible to do what you want if the "hide" would leave that node as an orphan.

lucasponce (approved):
The feature looks good and I'm +1 to consolidate it.

These are the tests performed on the work:

- Basic find by custom labels:
image

- Combined find/hide to spot only desired nodes:
image

- Combined find/hide with labels and values in the expression:
image

- Combined find/hide with a specific find but a wider hide expression:
image

- Regression on the find/hide with the tutorial, validating that previous flows work:
image

- Regression on the find/hide with the errors-rate demo, validating that previous flows work:
image

aljesusg (approved):
LGFM

PR #2308, opened 2022-02-16
Initial cypress integration

nrfox:
Adds cypress for integration testing. This PR is foundational for future cypress work for either QE or dev. It essentially just adds cypress to the project without yet adding any tests or making assumptions for how those tests should be structured.

There's a single test included to serve as an example of the integration working. This can be run when testing kiali locally by first running the UI dev server with `yarn start` and then running `yarn cypress`.

PR #2312, opened 2022-02-25
BDD extension for cypress testing

ScriptingShrimp:
This PR extends [Initial cypress integration](https://github.com/kiali/kiali-ui/pull/2308) with BDD plugin and simple `.feature` file for learning purposes. 

As QE's, we will extend this, refactor and make comprehensive structure over time. Right now, our QE's are familiar how to run BDD files and how to make small, incremental improvements. 

Already we spent ~4 hours learning and exercising on this approach and so far I have positive feedback.  This is also part of JIRA Ticket 

- to implement [Cypress plugin - Cucumber / BDD](https://issues.redhat.com/browse/OSSM-1039)
- have [Cypress PoC with working login](https://issues.redhat.com/browse/OSSM-1017)

image


cypress-bdd.json
```diff
@@ -0,0 +1,17 @@
+{
+  "baseUrl": "https://kiali-istio-system.apps.pmdown4.0216-eg3.fw.rhcloud.com",
```
nrfox, line 2: Can you change this to something else by default? Something like `http://localhost:3000`? The base url will likely need to be updated for whoever is running the tests through the cypress env var or through a command line flag.

cypress/integration/common/kiali_homepage.ts
```diff
@@ -0,0 +1,13 @@
+import { Given } from "cypress-cucumber-preprocessor/steps";
+import { Then } from "cypress-cucumber-preprocessor/steps";
+
+const url = "https://kiali-istio-system.apps.pmdown4.0216-eg3.fw.rhcloud.com"
```
nrfox, line 4: If you set `baseUrl` through an env var you shouldn't need to set this explicitly: https://docs.cypress.io/guides/references/best-practices#Setting-a-global-baseUrl. `cy.visit(url)` then becomes `cy.visit("/")`.

cypress/integration/check_graph.spec.ts
```diff
@@ -0,0 +1,6 @@
+import { Given } from "cypress-cucumber-preprocessor/steps";
+
+const url = 'https://google.com'
+Given('user is logged in as administrator', () => {
+  cy.visit(url)
+})
```
nrfox, line 6: This looks like just an example. Can we remove it?

nrfox (changes requested):
Just a couple comments about the hardcoded urls otherwise LGTM

cypress-bdd.json
jshaughn: +1 to the suggested URL, it works well when already using `yarn start`.

cypress/integration/featureFiles/kiali_homepage.feature
```diff
@@ -0,0 +1,8 @@
+Feature: Kilai login
```
jshaughn, line 1:
```suggestion
Feature: Kiali login
```

cypress-bdd.json
ScriptingShrimp: changed to suggested value

nrfox (approved):
Works great! Excited to begin using this.

PR #2314, opened 2022-02-28
Reduce health req lists

aljesusg:
Require https://github.com/kiali/kiali/pull/4763

Issue: https://github.com/kiali/kiali/issues/4748

This Reduce the requests, If we have N namespaces and each namespace has W workloads.

We move from **N*W requests** to **N requests**

Screenshot from 2022-02-28 10-08-38

# Test 
Open network tab and check that there are only one health request instead of N request.

The behaviour must be the same that before. Try to change a health of a service/workload/app and check that in list is updated after refresh interval.

### DIFF
--- a/src/pages/AppList/AppListPage.tsx
+++ b/src/pages/AppList/AppListPage.tsx
@@ -68,12 +68,10 @@ class AppListPageComponent extends FilterComponent.Component<AppListPageProps, A
     this.promises.cancelAll();
   }
 
-  sortItemList(items: AppListItem[], sortField: SortField<AppListItem>, isAscending: boolean): Promise<AppListItem[]> {
+  sortItemList(items: AppListItem[], sortField: SortField<AppListItem>, isAscending: boolean): AppListItem[] {
     // Chain promises, as there may be an ongoing fetch/refresh and sort can be called after UI interaction
     // This ensures that the list will display the new data with the right sorting
-    return this.promises.registerChained('sort', items, unsorted =>
-      AppListFilters.sortAppsItems(unsorted, sortField, isAscending)
-    );
+    return AppListFilters.sortAppsItems(items, sortField, isAscending);
   }
 
   updateListItems() {
@@ -88,7 +86,9 @@ class AppListPageComponent extends FilterComponent.Component<AppListPageProps, A
   }
 
   fetchApps(namespaces: string[], filters: ActiveFiltersInfo, rateInterval: number) {
-    const appsPromises = namespaces.map(namespace => API.getApps(namespace));
+    const appsPromises = namespaces.map(namespace =>
+      API.getApps(namespace, { health: 'true', rateInterval: String(rateInterval) + 's' })
+    );
     this.promises
       .registerAll('apps', appsPromises)
       .then(responses => {
@@ -99,18 +99,9 @@ class AppListPageComponent extends FilterComponent.Component<AppListPageProps, A
         return AppListFilters.filterBy(appListItems, filters);
       })
       .then(appListItems => {
-        this.promises.cancel('sort');
-        this.sortItemList(appListItems, this.state.currentSortField, this.state.isSortAscending)
-          .then(sorted => {
-            this.setState({
-              listItems: sorted
-            });
-          })
-          .catch(err => {
-            if (!err.isCanceled) {
-              console.debug(err);
-            }
-          });
+        this.setState({
+          listItems: this.sortItemList(appListItems, this.state.currentSortField, this.state.isSortAscending)
+        });
       })
       .catch(err => {
         if (!err.isCanceled) {
--- a/src/pages/AppList/FiltersAndSorts.ts
+++ b/src/pages/AppList/FiltersAndSorts.ts
@@ -2,7 +2,7 @@ import { ActiveFiltersInfo, FILTER_ACTION_APPEND, FilterType } from '../../types
 import { calculateErrorRate } from '../../types/ErrorRate';
 import { AppListItem } from '../../types/AppList';
 import { SortField } from '../../types/SortFilters';
-import { WithAppHealth, hasHealth } from '../../types/Health';
+import { hasHealth } from '../../types/Health';
 import {
   istioSidecarFilter,
   healthFilter,
@@ -173,19 +173,8 @@ export const sortAppsItems = (
   unsorted: AppListItem[],
   sortField: SortField<AppListItem>,
   isAscending: boolean
-): Promise<AppListItem[]> => {
-  if (sortField.title === 'Health') {
-    // In the case of health sorting, we may not have all health promises ready yet
-    // So we need to get them all before actually sorting
-    const allHealthPromises: Promise<WithAppHealth<AppListItem>>[] = unsorted.map(item => {
-      return item.healthPromise.then((health): WithAppHealth<AppListItem> => ({ ...item, health }));
-    });
-    return Promise.all(allHealthPromises).then(arr => {
-      return arr.sort(isAscending ? sortField.compare : (a, b) => sortField.compare(b, a));
-    });
-  }
-  const sorted = unsorted.sort(isAscending ? sortField.compare : (a, b) => sortField.compare(b, a));
-  return Promise.resolve(sorted);
+): AppListItem[] => {
+  return unsorted.sort(isAscending ? sortField.compare : (a, b) => sortField.compare(b, a));
 };
 
 export const compareObjectReference = (a: ObjectReference, b: ObjectReference): number => {
--- a/src/pages/ServiceList/FiltersAndSorts.ts
+++ b/src/pages/ServiceList/FiltersAndSorts.ts
@@ -1,5 +1,5 @@
 import { ActiveFiltersInfo, FilterType, FILTER_ACTION_APPEND, FilterTypes } from '../../types/Filters';
-import { WithServiceHealth, hasHealth } from '../../types/Health';
+import { hasHealth } from '../../types/Health';
 import { ServiceListItem } from '../../types/ServiceList';
 import { SortField } from '../../types/SortFilters';
 import {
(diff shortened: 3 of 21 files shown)


src/pages/ServiceList/ServiceListPage.tsx
```diff
@@ -126,19 +130,12 @@ class ServiceListPageComponent extends FilterComponent.Component<
...
-            if (!err.isCanceled) {
-              console.debug(err);
-            }
-          });
+        this.setState({
+          listItems: this.sortItemList(serviceListItems, this.state.currentSortField, this.state.isSortAscending)
+        });
       })
       .catch(err => {
+        console.log(err);
```
nrfox, line 138: Was this for testing? Can it be removed?

src/services/Api.ts
```diff
@@ -202,8 +202,8 @@ export const createIstioConfigDetail = (
   return newRequest(HTTP_VERBS.POST, urls.istioConfigCreate(namespace, objectType), {}, json);
 };
 
-export const getServices = (namespace: string) => {
-  return newRequest<ServiceList>(HTTP_VERBS.GET, urls.services(namespace), {}, {});
+export const getServices = (namespace: string, params: any = {}) => {
```
nrfox, line 205: Is there a more specific type other than `any` that can be used here?

nrfox (commented):
Couple minor comments. Looks like this also makes things much simpler w/o having a promise for each item :+1:

src/services/Api.ts
aljesusg: We can create a type woth queryparams that we let in the backend

src/pages/ServiceList/ServiceListPage.tsx
aljesusg: Totally :)

lucasponce:
I've started to take a look, but the services page gives errors in the first try:
image

The backend query looked ok.

src/services/Api.ts
nrfox: Yeah this might be a larger API so it's fine if we don't want to include it here.

> We can create a type woth queryparams that we let in the backend

Yes exactly. Something like:
```
getServicesParams interface {
  health?: bool
  rateInterval?: string
}
```
This way we can catch at compile time passing bogus query params.

Perhaps we could also update `newRequest` to accept query params as a second type parameter so that it would be typed as well:
```
const newRequest = <P, QP>(method: HTTP_VERBS, url: string, queryParams: QP, data: any) => ...
...
newRequest<ServiceList, getServicesParams>(HTTP_VERBS.GET, urls.services(namespace), params, {});
```

aljesusg: Opened https://github.com/kiali/kiali/issues/4783 to track this

israel-hdez: Perhaps tools like this one may help: https://github.com/acacode/swagger-typescript-api

israel-hdez:
@aljesusg I'm seeing the same problem as Lucas. I think it may be related to my comment in the back-end side: https://github.com/kiali/kiali/pull/4763#discussion_r818964529

aljesusg:
I did some changes maybe I forgot to push them sorry 😞 let me resume this tomorrow. Thanks @israel-hdez

src/pages/AppList/AppListPage.tsx
```diff
@@ -100,17 +100,9 @@ class AppListPageComponent extends FilterComponent.Component<AppListPageProps, A
...
-            this.setState({
-              listItems: sorted
-            });
-          })
-          .catch(err => {
-            if (!err.isCanceled) {
-              console.debug(err);
-            }
-          });
+        this.setState({
```
israel-hdez, line 103: I think you can remove the previous line `this.promises.cancel('sort');`.

src/services/Api.ts
nrfox: +1 ideally we'd be able to generate stuff like this from the swagger spec. Last time I investigated this there were some problems with Kiali's swagger spec: https://github.com/kiali/kiali-ui/pull/2248#issuecomment-1027974421. Probably those will need to be resolved in the upstream go swagger generator or else manually patched before being able to generate models from the spec.

src/pages/ServiceList/FiltersAndSorts.ts
```diff
@@ -255,18 +255,6 @@ export const sortServices = (
...
-    });
-    return Promise.all(allHealthPromises).then(arr => {
-      return arr.sort(isAscending ? sortField.compare : (a, b) => sortField.compare(b, a));
-    });
-  }
-  // Default case: sorting is done synchronously
-  const sorted = services.sort(isAscending ? sortField.compare : (a, b) => sortField.compare(b, a));
-  return Promise.resolve(sorted);
+): ServiceListItem[] => {
+  return services.sort(isAscending ? sortField.compare : (a, b) => sortField.compare(b, a));
```
israel-hdez, line 259: Not related to this line....

Perhaps change the declaration of the `filterBy` method to not return `Promise<ServiceListItem[]>`? Because I see it can no longer be a Promise.

src/pages/ServiceList/__tests__/ServiceListComponent.test.ts
```diff
@@ -78,20 +78,18 @@ describe('ServiceListContainer#sortServices', () => {
...
     const services = [makeService('A', {}, {}), makeService('B', {}, {})];
-    ServiceListFilters.sortServices(services, sortField, true).then(sorted => {
-      expect(sorted[0].name).toBe('A');
-      expect(sorted[1].name).toBe('B');
-      done();
-    });
+    const sorted = ServiceListFilters.sortServices(services, sortField, true);
+    expect(sorted[0].name).toBe('A');
+    expect(sorted[1].name).toBe('B');
+    done();
```
israel-hdez, line 84: Since these tests no longer deal with Promises, I think you can remove the call to `done()` AND also remove it as an argument to the spec. So, line 78 would be `it('should sort ascending', () => {`

src/pages/ServiceList/__tests__/ServiceListComponent.test.ts
```diff
@@ -78,20 +78,18 @@ describe('ServiceListContainer#sortServices', () => {
...
     const services = [makeService('A', {}, {}), makeService('B', {}, {})];
-    ServiceListFilters.sortServices(services, sortField, false).then(sorted => {
-      expect(sorted[0].name).toBe('B');
-      expect(sorted[1].name).toBe('A');
-      done();
-    });
+    const sorted = ServiceListFilters.sortServices(services, sortField, false);
+    expect(sorted[0].name).toBe('B');
+    expect(sorted[1].name).toBe('A');
+    done();
```
israel-hdez, line 93: ...and same here.

src/pages/WorkloadList/FiltersAndSorts.ts
```diff
@@ -353,17 +353,6 @@ export const sortWorkloadsItems = (
...
-    const allHealthPromises: Promise<WithWorkloadHealth<WorkloadListItem>>[] = unsorted.map(item => {
-      return item.healthPromise.then((health): WithWorkloadHealth<WorkloadListItem> => ({ ...item, health }));
-    });
-    return Promise.all(allHealthPromises).then(arr => {
-      return arr.sort(isAscending ? sortField.compare : (a, b) => sortField.compare(b, a));
-    });
-  }
-  const sorted = unsorted.sort(isAscending ? sortField.compare : (a, b) => sortField.compare(b, a));
-  return Promise.resolve(sorted);
+): WorkloadListItem[] => {
```
israel-hdez, line 356: Same comment on this file. I think in line 322 the `Promise<WorkloadListItem[]>` return type can be removed, since the `filterBy` function will no longer return that type.

src/types/ErrorRate/ErrorRate.ts
```diff
@@ -52,7 +52,11 @@ export const calculateErrorRate = (
 ): { errorRatio: ErrorRatio; config: ToleranceConfig[] } => {
   // Get the first configuration that match with the case
   const rateAnnotation = new RateHealth(requests.healthAnnotations);
-  const conf = rateAnnotation.toleranceConfig || getRateHealthConfig(ns, name, kind).tolerance;
+  let conf = rateAnnotation.toleranceConfig;
+  if (!conf) {
+    const healthRateConf = getRateHealthConfig(ns, name, kind);
+    conf = healthRateConf ? healthRateConf.tolerance : [];
+  }
```
israel-hdez, line 59: This is changing the behavior a little bit, no?
Just to understand it, what was the motivation?

src/pages/AppList/FiltersAndSorts.ts
```diff
@@ -173,19 +173,8 @@ export const sortAppsItems = (
...
-      return item.healthPromise.then((health): WithAppHealth<AppListItem> => ({ ...item, health }));
-    });
-    return Promise.all(allHealthPromises).then(arr => {
-      return arr.sort(isAscending ? sortField.compare : (a, b) => sortField.compare(b, a));
-    });
-  }
-  const sorted = unsorted.sort(isAscending ? sortField.compare : (a, b) => sortField.compare(b, a));
-  return Promise.resolve(sorted);
+): AppListItem[] => {
+  return unsorted.sort(isAscending ? sortField.compare : (a, b) => sortField.compare(b, a));
```
israel-hdez, line 177: ...an same comment as the other `FilterAndSort.ts` files. I think the declaration of the `filterBy` function can be fixed to no longer return `Promise<AppListItem[]>`

israel-hdez (commented):
Mostly minor comments here...
Comments on the back-end side are more important, I think.

aljesusg:
> I've started to take a look, but the services page gives errors in the first try: image
> 
> The backend query looked ok.

Sorry i forgot init the health in services, now should work.

src/types/ErrorRate/ErrorRate.ts
aljesusg: Because I saw a corner case but it was caused by the backend so I did the change int he backend let my change the code like before

israel-hdez (approved):
This part is fine IMO.
It's a matter of fixing the CI failure.

lucasponce (changes requested):
I do still see problems with the service's health:
image

image

Expected (from master branch):
image

israel-hdez:
I'm seeing the same problem on list pages.
It looks like the problem is on the back-end due to a pass-by-value issue while you may want to pass by reference.

aljesusg:
> I'm seeing the same problem on list pages. It looks like the problem is on the back-end due to a pass-by-value issue while you may want to pass by reference.

Fixed

israel-hdez (approved):
It works as expected in my env.

prachiyadav:
Working for me.

Screenshot from 2022-03-14 16-45-18


Screenshot from 2022-03-14 16-44-51

prachiyadav:
Test case needs to be created.

lucasponce (approved):
Tested locally.
image

As discussed, more work-related changes on health from graph and details are on the way.
We are going to consolidate to reduce the risk of conflicts.

PR #2315, opened 2022-03-01
Support 'name.namespace' key for Istio config validation result instead of 'name' only, to support multiple namespaces

hhovsepy:
Fix for https://github.com/kiali/kiali/issues/4766

Core changes are required: https://github.com/kiali/kiali/pull/4768

### DIFF
--- a/src/types/IstioConfigList.ts
+++ b/src/types/IstioConfigList.ts
@@ -15,6 +15,7 @@ import {
   WorkloadGroup
 } from './IstioObjects';
 import { ResourcePermissions } from './Permissions';
+import _round from 'lodash/round';
 
 export interface IstioConfigItem {
   namespace: string;
@@ -90,6 +91,14 @@ export const dicIstioType = {
   workloadgroup: 'WorkloadGroup',
   envoyfilter: 'EnvoyFilter'
 };
+
+export function validationKey(name: string, namespace?: string): string {
+  if (namespace != undefined) {
+    return name + '.' + namespace;
+  } else {
+    return name;
+  }
+}
 
 const includeName = (name: string, names: string[]) => {
   for (let i = 0; i < names.length; i++) {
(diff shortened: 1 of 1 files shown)


aljesusg (approved):
LGFM

src/types/IstioConfigList.ts
```diff
@@ -178,14 +178,15 @@ export const toIstioItems = (istioConfigList: IstioConfigList): IstioConfigItem[
     }
 
     entries.forEach(entry => {
+      const fqdn = entry.metadata.name + '.' + entry.metadata.namespace;
```
nrfox, line 181: Looks like combining the name + namespace is done in a few different places to form the validation key. Can this be a separate func so that we don't forget to add the `.` later?

To me, `fqdn` is related to dns which isn't what this var is for. Can this be renamed to something like `validationKey` or perhaps `validationKey(name: string, namespace: string)` if it's a separate func?

nrfox (commented):
Minor comment but tested the fix and everything works correctly.

Tested by:
deploying bookinfo, bookinfo2, bookinfo3
breaking the istio config in bookinfo causing an error to appear
refreshing the istio config list page and confirming that results are consistent

src/types/IstioConfigList.ts
```diff
@@ -156,8 +164,8 @@ export const filterByConfigValidation = (unfiltered: IstioConfigItem[], configFi
 export const toIstioItems = (istioConfigList: IstioConfigList): IstioConfigItem[] => {
   const istioItems: IstioConfigItem[] = [];
 
-  const hasValidations = (type: string, name: string) =>
-    istioConfigList.validations[type] && istioConfigList.validations[type][name];
+  const hasValidations = (type: string, name: string, namespace: string) =>
+    istioConfigList.validations[type] && istioConfigList.validations[type][name + '.' + namespace];
```
nrfox, line 168:
```suggestion
    istioConfigList.validations[type] && istioConfigList.validations[type][validationKey(name, namespace)];
```

pbajjuri20 (approved):

nrfox (approved):

PR #2317, opened 2022-03-03
Don't show Envoy metrics tab if those are not present

lucasponce:
Fixes https://github.com/kiali/kiali/issues/4769

There is a bug when a workload may have an Envoy sidecar but Envoy metrics are not available (for some reason, like a misconfiguration in the dashboard).

In this case, the UI will not crash and the "Metrics" tab will be omitted.


jmazzitelli (approved):
I ran the test procedures specified in the [description](https://github.com/kiali/kiali/pull/4778#issue-1157406761) (for both Istio 1.12 and 1.13) and the behavior is as expected.

PR #2320, opened 2022-03-04
Fix time selection issue in replay custom startTime picker

jshaughn:
- in graph replay's custom startTime, remove minTime and maxTime properties, they did not act as desired.  Instead of being used to establish a date-time range, and enabling all times between min and max (potentially across multiple days), the date portion was ignored and the time range was applied to each day.  As the time range was tied to the prom retention period,  this created seemingly random valid time ranges, or even a single, fixed valid time.  **With the fix the time-range should allow any time to be chosen, and only the days should be restricted based on retention time.**
- update react-datepicker from 2.14 to 3.8.  This fixes a variety of component glitches and updates some dependencies.  Note that this updates popper but not to a new major version.  V4 of react-datepicker had negative impact on the UI, so the update moves to the highest 3.x release.
- this also includes a fix to the graph contextMenu, which had excessive padding when built for prod, but looked fine with yarn start.  Now it is consistent.  This was discovered when testing popper update impact, which includes things like context menu and graph tour.

relates-to https://github.com/kiali/kiali/issues/4784
fixes https://issues.redhat.com/browse/OSSM-999

This PR requires a server PR, which fixes an issue with the TSDB retention config setting, which in turn affected the datepicker config.

**SERVER PR:** https://github.com/kiali/kiali/pull/4785

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
     "react": "16.13.1",
     "react-ace": "9.1.3",
     "react-copy-to-clipboard": "5.0.1",
-    "react-datepicker": "2.14.0",
+    "react-datepicker": "3.8.0",
     "react-dom": "16.13.1",
     "react-flexview": "4.0.3",
     "react-redux": "7.2.0",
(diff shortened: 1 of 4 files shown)


nrfox (approved):
LGTM

israel-hdez (approved):
It's been a long time since we have been having such CSS issues where `yarn start` looks right, but the result of `yarn build` generates something different.

We may be impacted by this: https://github.com/facebook/create-react-app/issues/7190. Apparently there is no definitive solution.

Anyway, I tried and this LG!

PR #2321, opened 2022-03-04
Improve protection against graph numbers that are actually string variables

jshaughn:
javascript is super flexible in that it actually handles things like
numeric comparisons that may involve a string (like '10' > 0 => true). But
you can't call a numeric function on a string var.

Our 'num' variables are *supposed* to be numbers but we have an undiscovered
issue where very infrequently the number variable is a string.  Until we
fix that with confidence, we need to protect ourselves.

This extends some previous protection to be more robust.

https://issues.redhat.com/browse/OSSM-1099

### DIFF
--- a/src/components/CytoscapeGraph/graphs/GraphStyles.ts
+++ b/src/components/CytoscapeGraph/graphs/GraphStyles.ts
@@ -666,7 +666,7 @@ export class GraphStyles {
       return `${trimFixed((num / 1000.0).toFixed(2))}s`;
     };
 
-    // This due to us never having figured out why a tiny fraction of what-we-expect-to-be-numbers
+    // This is due to us never having figured out why a tiny fraction of what-we-expect-to-be-numbers
     // are in fact strings.  We don't know if our conversion in GraphData.ts has a flaw, or whether
     // something else happens post-conversion.
     const safeNum = (num: any): number => {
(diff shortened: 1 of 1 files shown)


src/components/CytoscapeGraph/graphs/GraphStyles.ts
```diff
@@ -657,16 +654,32 @@ export class GraphStyles {
...
 
     const toFixedDuration = (num: number): string => {
+      num = safeNum(num);
       if (num < 1000) {
         return `${num.toFixed(0)}ms`;
       }
       return `${trimFixed((num / 1000.0).toFixed(2))}s`;
     };
 
+    // This due to us never having figured out why a tiny fraction of what-we-expect-to-be-numbers
```
jmazzitelli, line 669: This ..is... due
```suggestion
    // This is due to us never having figured out why a tiny fraction of what-we-expect-to-be-numbers
```

src/components/CytoscapeGraph/graphs/GraphStyles.ts
```diff
@@ -657,16 +654,32 @@ export class GraphStyles {
...
+    // are in fact strings.  We don't know if our conversion in GraphData.ts has a flaw, or whether
+    // something else happens post-conversion.
+    const safeNum = (num: any): number => {
+      if (Number.isFinite(num)) {
+        return num;
+      }
+      if (typeof num === 'string' || num instanceof String) {
+        console.log(`Expected number but received string: |${num}|`);
+      }
+      // this will return NaN is the string is 'NaN' or any other non-number
```
jmazzitelli, line 679: will return NaN ...if...the string
```suggestion
      // this will return NaN if the string is 'NaN' or any other non-number
```

jmazzitelli (commented):
trivial comment typos. non-blocker

nrfox (approved):

jshaughn:
Merging this, it's a pure code change and we have no recreation steps for the reported issue.  Best testing is to get it into the code base.

PR #2322, opened 2022-03-08
Make Service List Item validation to be undefined to support ServiceEntries

hhovsepy:
Fix for https://github.com/kiali/kiali/issues/4793

When service list contains only ServiceEntries, which does not have validations, error it thrown in console.
Now the validation existence is checked before using.

For QE:
Create namespace.
Create only ServiceEntry in that namespace.
List Services from this namespace only in Kiali.
Error should not be shown. 
Service entry is shown.
Verify also listing all Services is not broken.


lucasponce (approved):
Before:
image

After:
image
image

PR #2323, opened 2022-03-08
Cypress improvements

ScriptingShrimp:
Adding new smoke test to test login page. This first smoke test should work for both environments, local and stage/QA. 

Also introducing pre-test check, that is hitting Kiali API at `/api/auth/info` for determining authentication method. Test suite will throw error and will not start if authentication endpoint is not reachable. 

This PR is also part of Cypress integration plan
https://issues.redhat.com/browse/OSSM-1017


cypress/integration/common/kiali_login.ts
```diff
@@ -0,0 +1,52 @@
...
+const KUBEADMIN_IDP = 'my_htpasswd_provider';
+const auth_strategy = JSON.parse(Cypress.env('auth_strategy'))
+
+Given('user opens base url', () => {
+    cy.visit('/'); // visits baseUrl which is set in plugins/index.ts or bdd.ts
+    cy.window().then((win: any) => {
+        if (win = true) {
+            cy.log(win.localStorage + 'skipping login, console is running with auth disabled');
+            return;
+        }
```
nrfox, line 14: Not really sure what this is doing but it doesn't seem like it is necessary here. If we want to check if auth is disabled, can we just do `if auth_stragegy === 'anonymous'`?

cypress/integration/common/kiali_login.ts
```diff
@@ -0,0 +1,52 @@
...
+        // Make sure we clear the cookie in case a previous test failed to logout.
+        cy.clearCookie('openshift-session-token');
+    });
+});
+
+And('user clicks Log In With OpenShift', () => {
+    if (auth_strategy.strategy == 'openshift') {
```
nrfox, line 21: Generally doing strict equality with three equals `===` leads to less weirdness. Otherwise javascript will do type conversions for you like `['openshift'] == 'openshift'`.

nrfox (commented):
A few comments but the dev tests are still working as expected. I will try to test in CRC as well.

cypress-bdd.json
```diff
@@ -2,7 +2,6 @@
   "baseUrl": "http://localhost:3000",
   "viewportWidth": 1920,
   "viewportHeight": 1080,
-  "watchForFileChanges": true,
```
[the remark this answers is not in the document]
ScriptingShrimp: Its still working, it defaults to `true`. Don't know why, but this key/value was throwing warning with each start, so I removed it.

cypress/plugins/bdd.ts
```diff
@@ -1,12 +1,37 @@
...
     typescript: path.join(path.resolve('..'), 'kiali-ui/node_modules/typescript'),
   };
 
   on('file:preprocessor', cucumber(options));
+
+  async function exportConfig() {
+    const getAuthStrategy = async () => {
+      try {
+        const resp = await axios.get(config.env.base_URL)
+        return JSON.stringify(resp.data)
```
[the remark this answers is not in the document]
ScriptingShrimp: fixed

cypress/plugins/bdd.ts
```diff
@@ -1,12 +1,37 @@
...
+      } catch (err) {
+        console.error(err);
+        throw new Error(`Kiali API is not reachable at ${JSON.stringify(err.config.url)}`)
+      }
+    }
+
+    config.env.cypress_config = default_config
+    config.env.base_URL = `${default_config.baseUrl}/api/auth/info`
```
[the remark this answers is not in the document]
ScriptingShrimp: fixed, I reduced variables and add option for env variable

cypress/integration/common/kiali_login.ts
ScriptingShrimp: yes, you are right, it was form example. changed condition, fixed

cypress/integration/common/kiali_login.ts
ScriptingShrimp: fixed, added strict equals

nrfox:
@ScriptingShrimp it looks like a lot of unrelated recent changes got pulled into this PR when master got merged into this branch. Can you get rid of those commits so that this PR includes just the cypress changes?

cypress/plugins/bdd.ts
```diff
@@ -1,12 +1,35 @@
 const browserify = require('@cypress/browserify-preprocessor');
 const cucumber = require('cypress-cucumber-preprocessor').default;
 const path = require('path');
+const axios = require('axios');
+const default_config = require('../../cypress-bdd.json')
+
 
 module.exports = (on, config) => {
+  config.env.baseUrl = process.env.CYPRESS_BASE_URL || default_config.baseUrl
```
nrfox, line 9: By default cypress has `config.baseUrl` option. You can override this by env var or command line option: `cypress --config baseUrl=http://mycluster.crc/kiali`. IMO providing the URL via command line is the preferred way to run cypress against different environments. Importing the bdd json directly does not allow you to do this. If you remove this line and the import on line 5 then on line 29 use `config.baseUrl` you'll get the same functionality plus keep the ability to override via cmd line.

Example running cypress via cmd line:
```
yarn cypress:bdd --config baseUrl="https://localhost:20001" --env username=testuser,password=xxxx
```

cypress/plugins/bdd.ts
```diff
@@ -1,12 +1,35 @@
 const browserify = require('@cypress/browserify-preprocessor');
 const cucumber = require('cypress-cucumber-preprocessor').default;
 const path = require('path');
+const axios = require('axios');
+const default_config = require('../../cypress-bdd.json')
```
nrfox, line 5: As mentioned below, you can remove this import.

cypress/integration/common/kiali_login.ts
```diff
@@ -0,0 +1,52 @@
...
+        cy.clearCookie('openshift-session-token');
+    });
+});
+
+And('user clicks Log In With OpenShift', () => {
+    if (auth_strategy === 'openshift') {
+        const idp = KUBEADMIN_IDP;
+        cy.log(`Logging in as ${USERNAME}`);
+        cy.get('button[type="submit"]').should('be.visible');
+        cy.get('button[type="submit"]').click();
```
nrfox, line 25: We talked offline about this but cypress hangs here if the auth redirect domain does not match the same domain as the app. We don't need to decide how we test third party auth integrations in this PR but something to keep in mind for the future.

nrfox (changes requested):

mattmahoneyrh:
Environment var CYPRESS_PASSWD apparently is not yet supported. Workaround is to hard code password.

nrfox:
> Environment var CYPRESS_PASSWD apparently is not yet supported. Workaround is to hard code password.

It might be `CYPRESS_passwd` or `cypress_passwd` with a lowercase `passwd`: https://github.com/kiali/kiali-ui/pull/2323/files#diff-1bf72959fc9321e5a985ca544c185c72e70e965eed524d18c681c8ea001fb4e4R4

cypress/integration/common/kiali_login.ts
```diff
@@ -0,0 +1,52 @@
+import { Given, When, And, Then } from "cypress-cucumber-preprocessor/steps";
+
+const USERNAME = 'jenkins';
```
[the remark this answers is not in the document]
nrfox: @ScriptingShrimp can this be an env var as well as opposed to being hardcoded to jenkins? You can add a "env" section to the `cypress-bdd.json` config and set the default value there to `jenkins`.

cypress-bdd.json
```
{
...
  "env": {
    "USERNAME": "jenkins"
  }
...
}
```

There's a bunch of different ways of setting env vars with cypress but being able to specify a different username for the tests will be good for running in CRC: https://docs.cypress.io/guides/guides/environment-variables#Setting

cypress/plugins/bdd.ts
```diff
@@ -1,12 +1,35 @@
...
+      try {
+        const resp = await axios.get(url+'/api/auth/info')
+        return resp.data.strategy
+      } catch (err) {
+        console.error(`ERROR: Kiali API is not reachable at ${JSON.stringify(err.config.url)}`);
+        throw new Error(`Kiali API is not reachable at ${JSON.stringify(err.config.url)}`)
+      }
+    }
+
+    config.baseUrl = config.env.BASEURL || config.baseUrl
```
nrfox, line 27: This isn't a blocker but I don't see why you need `config.env.BASEURL` when setting the env var `CYPRESS_BASE_URL` will populate `config.baseUrl`.

cypress/plugins/bdd.ts
ScriptingShrimp: funny, this works with YARN OK as you mentioned, but not with NPM
`yarn cypress:bdd --config baseUrl="https://localhost:20001" --env username=testuser,password=xxxx`

nrfox (approved):

PR #2325, opened 2022-03-11
List Gateways for All Namespaces in Service Wizzard, instead of query per namespace

hhovsepy:
Fix for https://github.com/kiali/kiali/issues/4692

It used to list Gateways per namespace in Service Details page.
Screenshot from 2022-03-11 09-44-37

Now it sends only one query and lists all Gateways, which is used in Wizard Gateways section.
Screenshot from 2022-03-11 10-14-11


lucasponce (approved):
Tested together with backend work:
image

Instead, to iterate per namespace to collect all gateways, it fetches the list from a single result.

PR #2326, opened 2022-03-11
improve graph  contrast 

jshaughn:
minor changes to graph background color, box fill colors, and node edge color, to try and improve contrast and clarity, especially when zoomed out.   Not trying to make drastic changes, and the current look is UX-approved.

@lucasponce You may not even notice on a casual look, but if you do some side-by-side comparison you should see a difference.  Let me know if you see a difference, and whether you think it's better.

Related to https://github.com/kiali/kiali/issues/4610

### DIFF
--- a/src/components/CytoscapeGraph/graphs/GraphStyles.ts
+++ b/src/components/CytoscapeGraph/graphs/GraphStyles.ts
@@ -172,7 +172,7 @@ export class GraphStyles {
     EdgeColorFailure = PFColorVals.Danger;
     EdgeColorTCPWithTraffic = PFColorVals.Blue600;
     EdgeTextOutlineColor = PFColorVals.White;
-    NodeColorBorder = PFColorVals.Black400;
+    NodeColorBorder = PFColorVals.Black500;
     NodeColorBorderBox = PFColorVals.Black600;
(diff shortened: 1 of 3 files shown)


lucasponce (approved):
It looks good to me:
image

Perhaps it's time to review the global style for a major version and we can be more open to changes.

jshaughn:
> Perhaps it's time to review the global style for a major version and we can be more open to changes.

Yes, this was aimed to be a very minor change.  In another effort we may want to take a big swing and see what we sort of ideas we may want to consider.

jshaughn:
Merged directly, only color changes.

lucasponce:
Hi @jshaughn, I've noticed some cases where the node is selected and the contrast is a little bit different:
image

Just wondering if you wanted to have this effect or if that is due to some unexpected bug.

jshaughn:
I think this is mostly expected and is the result of the more aggressive dimming we introduced earlier in this cycle, not in this PR.  Having said that, I think maybe the dimming should be slightly less aggressive, as in this scenario, when there are no edges connecting the namespaces, it does seem a little drastic.  Also, I think the label dimming seems actually not aggressibe enough.  Let me look at that...

PR #2327, opened 2022-03-14
Remove 'null' array values in Istio Config update JSON

hhovsepy:
Fix for https://github.com/kiali/kiali/issues/4753

### DIFF
--- a/src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
+++ b/src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
@@ -284,7 +284,9 @@ class IstioConfigDetailsPage extends React.Component<RouteComponentProps<IstioCo
 
   onUpdate = () => {
     jsYaml.safeLoadAll(this.state.yamlModified, (objectModified: object) => {
-      const jsonPatch = JSON.stringify(mergeJsonPatch(objectModified, getIstioObject(this.state.istioObjectDetails)));
+      const jsonPatch = JSON.stringify(
+        mergeJsonPatch(objectModified, getIstioObject(this.state.istioObjectDetails))
+      ).replace(new RegExp(',null', 'g'), '');
       API.updateIstioConfigDetail(
         this.props.match.params.namespace,
         this.props.match.params.objectType,
--- a/src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
+++ b/src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
@@ -286,7 +286,7 @@ class IstioConfigDetailsPage extends React.Component<RouteComponentProps<IstioCo
     jsYaml.safeLoadAll(this.state.yamlModified, (objectModified: object) => {
       const jsonPatch = JSON.stringify(
         mergeJsonPatch(objectModified, getIstioObject(this.state.istioObjectDetails))
-      ).replace(new RegExp(',null', 'g'), '');
+      ).replace(new RegExp('"(,null)+]', 'g'), '"]');
       API.updateIstioConfigDetail(
         this.props.match.params.namespace,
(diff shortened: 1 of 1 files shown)


src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
```diff
@@ -272,7 +272,9 @@ class IstioConfigDetailsPage extends React.Component<RouteComponentProps<IstioCo
 
   onUpdate = () => {
     jsYaml.safeLoadAll(this.state.yamlModified, (objectModified: object) => {
-      const jsonPatch = JSON.stringify(mergeJsonPatch(objectModified, getIstioObject(this.state.istioObjectDetails)));
+      const jsonPatch = JSON.stringify(
+        mergeJsonPatch(objectModified, getIstioObject(this.state.istioObjectDetails))
+      ).replace(new RegExp(',null', 'g'), '');
```
lucasponce, line 277: I don't think this will work in all cases.
https://datatracker.ietf.org/doc/html/rfc7386

"null" values are allowed in the merge/patch syntax and differentiate between an explicit field with "null" or removing a field from an object.

But it seems the problem happens with the "arrays" values, so this global regex will produce side effects when you want to remove a field in other editions.

hhovsepy: yes, it should apply only arrays, as in json, null  values in array are moved on the end, ',null/g' will replace them all, and it seems the original issue existed always

lucasponce:
@hhovsepy I couldn't reproduce the problem in master with the VirtualService object in the gateways host which is a similar scenario, and I spot this difference that may affect to "lodash" helper function and the yaml editor:

image
image
image

I wonder if you'd like to give a try to adjust the type and see if the problem is gone rather than apply an overall regex to the JSON.

I was curious why we didn't see that problem in other similar cases and perhaps that's the case.

(I know is minor, but the global regex looked suspicious that perhaps the root cause was in a different place).

hhovsepy:
It seems "hosts: string[];" does not resolve the problem.

lucasponce:
And ?
```
hosts?: string[] | null;
```

hhovsepy:
It seems with all combinations for "hosts", including "hosts?: string[] | null;", issue still exists. I will try to come with more specific regexp instead of general one.

src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx
hhovsepy: changed the regexp to '"(,null)+]' to make sure only null in arrays are considered

lucasponce:
Hi @hhovsepy I'm back to this PR, can you rebase with master? It can't work with the backend master branch.
Thanks!

hhovsepy:
thanks @lucasponce rebased

lucasponce:
@hhovsepy in master I can't reproduce the problem

I wonder if the root cause of this is the "poll/ping" mechanism that we have in the IstioConfigDetails to check newer versions.
Could you test one thing?

Try to disable it locally and see if you can update/remove the hosts correctly.

P.S.: I'm using Istio 1.13 (are you in 1.13 or perhaps this is something related to Istio 1.12?).

hhovsepy:
I can reproduce the issue after disabling refresh/version check mechanism.
This issue exists even in Kiali v1.36.

It's coming from "return null" of IstioConfigUtils customizer function.
Screenshot from 2022-03-18 12-00-53

lucasponce:
Thanks @hhovsepy for the investigation, I'm going to give a final review and finish it.

lucasponce (approved):
I've tested the specific Sidecar objects and done regression tests in other string arrays from other types of objects (VirtualServices, Gateways).

PR #2328, opened 2022-03-14
Don't allow a graph type change for node graphs

jshaughn:
Don't allow a graph type change for node graphs, they should be fixed as not every graph type makes sense for the selected node type.   I think it's reasonable to disable the entire dropdown, and not worry about potentially valid combinations.  Users can always return to the main graph and change the graph type, if necessary.

Fixes https://github.com/kiali/kiali/issues/4797

cc @Xunzhuo


israel-hdez (approved):
I tried it. Easy fix! :smiley:

Xunzhuo (approved):
Thanks! LGTM

PR #2332, opened 2022-03-16
Fix for BreadthFirstLayout on races conditions

lucasponce:
There is a race condition when the Cytoscape model is not populated with the position information.

### DIFF
--- a/src/components/CytoscapeGraph/Layout/KialiBreadthFirstLayout.ts
+++ b/src/components/CytoscapeGraph/Layout/KialiBreadthFirstLayout.ts
@@ -278,8 +278,16 @@ export default class KialiBreadFirstLayout {
     }
 
     var getPosition = function (ele) {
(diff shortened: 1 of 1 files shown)


jmazzitelli:
I thought this fixed it - but after doing a bunch of clicking and zooming, I ended up breaking it:
image

jmazzitelli:
BTW: When I do break it, it always is on the "namespace layout 2"

jmazzitelli:
Peek 2022-03-16 11-43

lucasponce:
Can you see the browser console when the problem happens (check if there is a new issue or is it clean)?

I see you have the "namespace box" selected and one namespace, do you see the problem with more namespaces (namespace box enabled with a single namespace is a different case)?

lucasponce:
@jmazzitelli updated the patch.
I guess that is the "offending" method that under a race scenario it may trigger the problem.

Let me know if you could test it again.

PR #2335, opened 2022-03-17
Remove unused Application/Service/Workload health endpoints

lucasponce:
Follow-up task to clean unused health endpoints from the UI.

I'm preparing a task in the backend as I've noticed a hidden bug in the Workload logic (it was before that the health refactoring effort, but thanks to this work, that issue is more evident).

Please, I did a smoke test and everything worked as expected, let me know if this PR is good.


aljesusg (approved):
LGFM, we need to remove this in the backend too

jshaughn (approved):
Proves that we did in fact move this to the backend :-)

PR #2336, opened 2022-03-17
Resolve critical, high and dev-dep CVEs

jshaughn:
- resolve ansi-html, follow-redirects, prismjs, url-parse to newer (non-major) versions
- bump node-sass dev-dep

This is based on dependabot security issues.  I created a clean node_modules, ran yarn install, sanity checked various kiali pages against the build and also yarn start, and ran the tests.    No regressions found.


lucasponce (approved):
Tested locally.
Perform several tests on all pages and nothing is broken from a UI perspective.

PR #2340, opened 2022-03-21
[backport v1.36] Improve protection against graph numbers that are actually string variables

jshaughn:
javascript is super flexible in that it actually handles things like
numeric comparisons that may involve a string (like '10' > 0 => true). But
you can't call a numeric function on a string var.

Our 'num' variables are *supposed* to be numbers but we have an undiscovered
issue where very infrequently the number variable is a string.  Until we
fix that with confidence, we need to protect ourselves.

This extends some previous protection to be more robust.

Backport of https://github.com/kiali/kiali-ui/pull/2321
https://issues.redhat.com/browse/OSSM-1099


jmazzitelli (approved):
Is there a git commit this cherry picked ? If so, put a link in here to document what master branch commit was cherry picked.

PR #2351, opened 2022-05-23
Resolve OSSM CVEs

jshaughn:
Resolves:
- OSSM-1522
- OSSM-1358

@aljesusg These fixes would get picked up in the next 2.1 cpaas build (for 2.1.3).  We need to wait on the build until 1.1 and 2.0 builds are complete, so that the 2.1 operator can pick up the new hash values.


aljesusg (approved):
LGFM

